PR #51, opened 2026-02-10
feat: Hounfour Integration — arrakis x loa-finn cross-system bridge

deep-name:
## Summary

- **Sprint 1**: Foundation — BudgetUnitBridge (micro-USD ↔ micro-cents), S2SJwtValidator (ES256 JWKS with tiered caching), agent gateway config, ADR-005
- **Sprint 2**: UsageReceiver 7-step validation pipeline, PG schema migration (agent_usage_log), S2S auth middleware, internal `/usage-reports` route wiring
- **Sprint 3**: Tier-aware pool routing (`pool-mapping.ts`), `pool_id` + `allowed_pools` JWT claims (19 total), JWKS E2E contract tests (key rotation, req_hash binding)
- **Sprint 4**: Full pipeline integration tests (S2S JWT → JWS → validate → PG → Redis), route isolation, error cases, budget unit conversion, all 5 PRD goals validated

## Key components

| Component | Path | Purpose |
|-----------|------|---------|
| Pool Mapping | `packages/adapters/agent/pool-mapping.ts` | Tier-aware model→pool resolution (`native` anti-escalation) |
| S2S JWT Validator | `packages/adapters/agent/s2s-jwt-validator.ts` | Inbound loa-finn JWT validation with JWKS caching |
| Usage Receiver | `packages/adapters/agent/usage-receiver.ts` | 7-step usage report validation pipeline |
| S2S Auth Middleware | `packages/adapters/agent/s2s-auth-middleware.ts` | Express middleware for Bearer token auth |
| Agent Gateway | `packages/adapters/agent/agent-gateway.ts` | Pool resolution in invoke()/stream() flows |
| JWT Service | `packages/adapters/agent/jwt-service.ts` | 19-claim JWT signing with pool routing |

## Test plan

- [x] 184+ Hounfour-specific tests passing across 8 test files
- [x] Pool mapping: 43 tests (5 aliases × 3 tiers + edge cases)
- [x] JWKS E2E contract: 22 tests (round-trip, rotation, expiry, tampering)
- [x] S2S JWT validator: 22 tests (ES256, caching, cross-protocol)
- [x] Usage receiver: 26 tests (7-step pipeline, idempotency, Redis warn-only)
- [x] Integration: 26 tests (full pipeline, route isolation, error cases, PRD goals)
- [x] Budget drift/overrun: 50+ tests (no regressions)
- [x] Zero regressions in existing test suite

### DIFF
--- a/packages/adapters/agent/config.ts
+++ b/packages/adapters/agent/config.ts
@@ -217,6 +217,13 @@ function parseIntEnv(value: string | undefined, fallback: number): number {
 
 const VALID_POOL_CLAIM_ENFORCEMENT = new Set<string>(['warn', 'reject']);
 
+/** Strip trailing slashes without polynomial regex (CodeQL js/polynomial-redos). */
+function stripTrailingSlashes(url: string): string {
+  let end = url.length;
+  while (end > 0 && url[end - 1] === '/') end--;
+  return url.slice(0, end);
+}
+
 function parsePoolClaimEnforcement(value: string | undefined): PoolClaimEnforcement {
   if (value != null && VALID_POOL_CLAIM_ENFORCEMENT.has(value)) {
     return value as PoolClaimEnforcement;
@@ -368,7 +375,7 @@ export function loadAgentGatewayConfig(
  */
 export function buildS2SJwtValidatorConfig(s2s: S2SValidationConfig): S2SJwtValidatorConfig {
   return {
-    jwksUrl: `${s2s.loaFinnBaseUrl.replace(/\/+$/, '')}/.well-known/jwks.json`,
+    jwksUrl: `${stripTrailingSlashes(s2s.loaFinnBaseUrl)}/.well-known/jwks.json`,
     expectedIssuer: s2s.expectedIssuer,
     expectedAudience: s2s.expectedAudience,
     jwksCacheTtlMs: s2s.jwksCacheTtlMs,


janitooor:
## The Bridgebuilder Reviews: Round 2 — Seven Observations from Deeper in the Wire

*"The second review catches what the code does right but doesn't yet know it knows."*

These findings emerged from reading full file context, not just the diff. Several are forward-looking — things that don't need fixing today but will matter at scale. Numbered F-10 through F-16, continuing from the original review's F-1 through F-9.

---

### F-10 | Medium | Architecture | JWT Schema Has No Version Field

**Files**: `jwt-service.ts:123-144`

The JWT carries 19 claims. That's a wire protocol. Wire protocols need version numbers.

Adding `pool_id` and `allowed_pools` (Sprint 3) required coordinated deployment of Arrakis and loa-finn. The next claim addition will require the same coordination. There's no mechanism for loa-finn to distinguish "this JWT was minted before pool claims existed" from "this JWT was minted after, but the claims are missing due to a bug."

**FAANG Parallel**: Protocol Buffers solved this in 2001 with field numbering — any field can be added without breaking old consumers. gRPC's backward compatibility guarantees are built on this. Even JWTs themselves support protocol discrimination via the `typ` header. But there's no *version* claim to signal which claims the consumer should expect.

**Suggestion**: Add a `v: 1` claim to the JWT payload. When loa-finn encounters `v: 2`, it knows new claims may be present. When it encounters `v: 1`, it knows `pool_id` might be absent. Small investment now; prevents coordinated-deployment pain later.

**Metaphor**: Imagine writing letters to a pen pal for years — name and address on the envelope. One day you add a phone number. Your pen pal has to guess whether it's intentional or spam. A simple "Letter Format v2" stamp resolves all ambiguity.

---

### F-11 | Praise | Architecture | The Budget Lifecycle Is a Saga Pattern Done Right

**Files**: `agent-gateway.ts:76-198` (invoke), `agent-gateway.ts:204-322` (stream)

The lifecycle — RECEIVED → RESERVED → EXECUTING → FINALIZED — is a textbook **saga pattern** for distributed transactions. What makes it exceptional:

1. **Compensation is automatic**: failed `invoke()` cancels the reservation immediately for non-retryable errors (L189-195). Retryable errors let the TTL expire — a **lease-based** approach from Chubby/ZooKeeper.
2. **Stream has a reconciliation fallback**: the `finally` block (L316-321) schedules a BullMQ reconciliation job if the stream closes without a `usage` event. This is the catch-all compensator that prevents budget leaks.
3. **Idempotency is first-class**: `ALREADY_RESERVED` (L146-148) handles replay without re-reserving budget.

**FAANG Parallel**: Uber's trip lifecycle uses this exact pattern — request → confirm → in_progress → complete, with compensation at each step. AWS Step Functions formalize it as state machines with compensating transactions. What's notable is that Arrakis achieves saga-grade reliability with Redis + BullMQ rather than a dedicated workflow engine. The simplicity is a feature.

---

### F-12 | Low | Correctness | maxCostMicroCents Lacks Unit Clarity

**Files**: `agent-gateway.ts:160`

```typescript
const maxCostMicroCents = estimatedCost * 3 * 100; // cents → micro-cents (×100)
```

The line is correct but the variable `estimatedCost` has no unit suffix. Is it cents? Micro-cents? The budget-unit-bridge uses `bigint` with explicit names (`microUsdToMicroCents`). The `number` path in the gateway deserves the same ceremony.

**Suggestion**: Rename to `estimatedCostCents` in the calling code, or extract `centsToMicroCents()`. Financial code should scream its units at every boundary.

**Metaphor**: NASA's Mars Climate Orbiter burned up in the Martian atmosphere because one team used pounds-force and another used newtons. They were both correct individually. The interface between them was the failure. The `budget-unit-bridge.ts` already prevents this for bigint — extend the same discipline to the `number` path.

---

### F-13 | Praise | Security | Cross-Protocol Confusion Guard Is Rare and Correct

**Files**: `s2s-jwt-validator.ts:96-98`, `s2s-jwt-validator.ts:127-129`

```typescript
// JWT path: enforce typ
if (protectedHeader.typ !== 'JWT') {
  throw new Error('S2S JWT must have typ: JWT');
}

// JWS path: reject JWT typ
if (protectedHeader.typ === 'JWT') {
  throw new Error('JWS must not have typ: JWT (use validateJwt for JWTs)');
}
```

This is defense against cross-protocol confusion — a sibling of the JWT algorithm confusion vulnerability that plagued libraries from 2015-2018 (where RS256-signed tokens were accepted as HS256 using the public key as the HMAC secret). By enforcing `typ` bidirectionally, a token minted for one protocol path can never be accepted in the other.

**What's rare**: Most JWT implementations I review don't check `typ` at all, let alone enforce bidirectional consistency. The error message `(use validateJwt for JWTs)` is the kind of operationally considerate touch that saves 2am debugging sessions. Someone who hits this error knows exactly what to do.

---

### F-14 | Medium | Architecture | Warn-Only Validation Needs a Promotion Path

**Files**: `usage-receiver.ts:167-187`

The pool claim cross-validation is warn-only — correct initial posture (observe before enforce). But there's no mechanism to *promote* from warning to enforcement once confidence is established.

**FAANG Parallel**: Netflix pioneered **"shadow mode"** for production changes — new code paths run in parallel, logging divergences but not affecting traffic. Once the divergence rate drops below a threshold, the new path is promoted. Google's Canary Analysis Service automates this: start at 1% traffic, compare metrics, promote when SLOs are met.

**Suggestion**: Add `pool_claim_enforcement: 'warn' | 'reject'` to `UsageReceiverConfig`. Start in `warn` (current). Once monitoring confirms zero false positives over a soak period, flip to `reject`. The code change is a one-line conditional. The observability path (logs → metrics → dashboard → confidence → enforcement) should be designed now.

---

### F-15 | Praise | Architecture | JWKS Cache Ordering Is Subtle and Correct

**Files**: `s2s-jwt-validator.ts:143-179`

The `getJwks()` method checks `this.inflight` (single-flight dedup) BEFORE the cooldown timer. This ordering is critical:

```typescript
// L154-155: Single-flight dedup — must check BEFORE cooldown so concurrent
// callers join the inflight request instead of hitting cooldown
if (this.inflight) {
  return this.inflight;
}

// THEN check cooldown
const sinceLastAttempt = now - this.lastRefreshAttempt;
```

If reversed — cooldown first — a burst of concurrent requests during cooldown would all get stale/rejected, even if one had triggered a refresh just before cooldown started. By checking inflight first, all concurrent callers piggyback on the live request.

**FAANG Parallel**: This is Go's `singleflight.Group` pattern (used throughout Google's caching layers) and Envoy Proxy's dedup filter. The ordering invariant — "join before check" — is the difference between O(1) and O(N) upstream load during a cache miss storm. The comment on L154-155 doesn't describe *what* the code does — it explains *why the ordering matters*. That's the kind of comment that earns its bytes.

---

### F-16 | Low | Correctness | The `as any` Cast Hides a Type Boundary

**Files**: `usage-receiver.ts:172`

```typescript
report.access_level as any, // AccessLevel expected, string received
```

The cast works because `validatePoolClaims` handles unknown access levels gracefully (returns `{ valid: false }`). But `as any` hides the boundary rather than marking it.

**Suggestion**: Extract a named type guard:

```typescript
function isAccessLevel(s: string): s is AccessLevel {
  return s === 'free' || s === 'pro' || s === 'enterprise';
}
```

Then use it to separate two distinct signals:
- Unknown access level → data quality issue (one log entry)
- Valid access level with mismatched pools → potential security issue (different log entry)

Two different signals deserve two different log entries. The `as any` collapses them into one.

---

### Summary Table

| # | Severity | Category | Title | Action |
|---|----------|----------|-------|--------|
| F-10 | Medium | Architecture | JWT schema has no version field | Add `v: 1` claim |
| F-11 | Praise | Architecture | Budget lifecycle is a saga done right | 🌟 |
| F-12 | Low | Correctness | maxCostMicroCents lacks unit clarity | Rename with unit suffix |
| F-13 | Praise | Security | Cross-protocol confusion guard is rare | 🌟 |
| F-14 | Medium | Architecture | Warn-only validation needs promotion path | Add enforcement config flag |
| F-15 | Praise | Architecture | JWKS cache ordering is subtle and correct | 🌟 |
| F-16 | Low | Correctness | `as any` cast hides type boundary | Extract named type guard |

**Verdict**: Approved. The hardening sprint addressed all Round 1 findings. These Round 2 observations are forward-looking improvements — none block merge.

---

*— The Bridgebuilder*

PR #53, opened 2026-02-11
feat: Hounfour Endgame — E2E infrastructure + deployment readiness (RFC #31)

deep-name:
## Summary

Consolidated PR for the Hounfour Endgame work (cycles 015-017, RFC #31 close-out on arrakis side).

- **Containerized E2E** — Docker-based cross-repo testing for arrakis ↔ loa-finn integration
- **Deployment validation** — Two-tier health check script (local + CloudWatch staging)
- **Operational runbook** — Incident response, rollback drills, failure mode catalog
- **Security foundation** — JWT contract artifacts, pool mapping, BYOK key management
- **Ensemble strategy** — Best-of-N mapper, partial failure reconciliation, budget multipliers

### Cycle 017: The Water of Life (Sprints 205-206)

- `a73b019` feat(sprint-205): containerized E2E infrastructure for arrakis ↔ loa-finn
- `98b3648` feat(sprint-206): deployment validation script and operational runbook
- `b71740c` chore(cycle-017): update state files

### Cycle 016: Bridgebuilder Round 3 (Sprint 204)

- `a4f2e5b` fix: remove unused `halves` variable in byok-proxy-handler.ts
- `315f1ca` fix(sprint-9): correctness fixes from Bridgebuilder Round 3

### Cycle 015: Hounfour Endgame (Sprints 199-203)

- `96be06e` feat(sprint-3,4): BYOK key management + production monitoring & hardening
- `b38f47a` feat(sprint-2): ensemble strategy exposure
- `87a7a82` feat(sprint-1): security foundation — contract artifacts, pool_mapping_version, E2E

### Housekeeping

- `7ec88e9` chore: archive completed Hounfour cycles (012-016)
- `5d191f3` chore: mark cycle-012 Hounfour Integration sprints as completed
- `30b0201` chore(cycle-015): planning artifacts for Hounfour Endgame

## Key Files (Cycle 017)

| File | Purpose |
|------|---------|
| `themes/sietch/Dockerfile` | Added `e2e` stage + fixed missing agent COPY |
| `tests/e2e/e2e-entrypoint.sh` | JWKS export wrapper for E2E container |
| `tests/e2e/docker-compose.e2e.yml` | Full topology: redis + arrakis + loa-finn |
| `scripts/run-e2e.sh` | Cross-repo E2E runner (LOA_FINN_SHA pinning) |
| `tests/e2e/agent-gateway-e2e.test.ts` | Docker mode + Ajv2020 schema validation |
[...]

PR #55, opened 2026-02-12
The Capability Mesh + BB7 Refinement — Cycles 019-020

deep-name:
## Summary

Implements all 7 Bridgebuilder Round 6 findings from PR #53, advancing the agent gateway from feature-complete to protocol-grade.

- **Per-model ensemble accounting** — `model_breakdown` array with per-model cost attribution, hybrid BYOK/platform accounting
- **Contract protocol promotion** — `@arrakis/loa-finn-contract` as standalone shared package with compatibility matrix and version negotiation
- **Provider policy configuration** — `POOL_PROVIDER_HINTS` env var replaces hardcoded pool→provider mapping
- **Gateway lifecycle state machine** — `RequestLifecycle` (RECEIVED→VALIDATED→RESERVED→EXECUTING→FINALIZED|FAILED)
- **API key isolation boundary** — `dispatchWithKey()` confines key string materialization, Buffer zeroed in finally block
- **Fleet-wide Redis circuit breaker** — atomic Lua scripts, process-local fallback when Redis unavailable
- **Token estimator calibration** — configurable chars-per-token, model-specific overrides, drift detection

| Sprint | Title | Files | Commit |
|--------|-------|-------|--------|
| Sprint 1 (G-209) | Per-Model Ensemble Accounting | 11 | `85a7439` |
| Sprint 2 (G-210) | Contract Protocol + Provider Policy | 15 | `b811007` |
| Sprint 3 (G-211) | Gateway Lifecycle + Security Hardening | 7 | `3653512` |
| Sprint 4 (G-212) | Budget Optimization + Observability | 5 | `a5fdf54` |

**Totals**: 4 sprints, 38 files changed, +3923 / -533 lines

### New Files

| File | Purpose |
|------|---------|
| `packages/adapters/agent/ensemble-accounting.ts` | Per-model cost decomposition (BB6 #6) |
| `packages/adapters/agent/request-lifecycle.ts` | State machine protocol object (BB6 #1) |
| `packages/adapters/agent/redis-circuit-breaker.ts` | Fleet-wide Redis circuit breaker (BB6 #4) |
| `packages/adapters/agent/token-estimator.ts` | Calibrated token estimation (BB6 #7) |
| `packages/adapters/agent/capability-audit.ts` | Structured capability audit events |
| `packages/contracts/src/compatibility.ts` | Contract version negotiation (BB6 #2) |
| `packages/contracts/src/semver-lite.ts` | Lightweight semver satisfies() |

### BB6 Findings Addressed

| # | Finding | Sprint | Status |
|---|---------|--------|--------|
| 1 | State Machine Extraction | Sprint 3 | Done |
| 2 | Contract Protocol Nucleus | Sprint 2 | Done |
| 3 | V8 Heap Key Persistence | Sprint 3 | Done |
| 4 | Fleet Circuit Breaker | Sprint 3 | Done |
| 5 | Provider Policy | Sprint 2 | Done |
| 6 | Ensemble Budget | Sprints 1 + 4 | Done |
| 7 | Token Estimator | Sprint 3 | Done |

## Test plan

- [ ] TypeScript compilation passes (no new errors introduced)
- [ ] Existing E2E tests pass (13/13 minimum)
- [ ] New E2E tests for version negotiation (AC-2.22, AC-2.23)
- [ ] `terraform plan` validates cleanly with new widgets/alarms
- [ ] Contract schema backward compatible (additive fields only)
- [ ] Redis circuit breaker fallback verified (Redis unavailable → local breaker)

🤖 Generated autonomously with Run Mode

### DIFF
--- a/packages/adapters/agent/capability-audit.ts
+++ b/packages/adapters/agent/capability-audit.ts
@@ -60,11 +60,42 @@ export class CapabilityAuditLogger {
   }
 
   /**
+   * Validate required fields before emission (BB7 R7-3).
+   * CloudWatch Log Metric Filters silently drop events with missing fields.
+   * Guard warns + skips rather than throwing — audit must never crash requests.
+   */
+  private validateRequiredFields(event: CapabilityAuditEvent): boolean {
+    const required: Array<[string, unknown]> = [
+      ['trace_id', event.trace_id],
+      ['community_id', event.community_id],
+      ['event_type', event.event_type],
+      ['pool_id', event.pool_id],
+    ];
+
+    for (const [field, value] of required) {
+      if (typeof value !== 'string' || value.length === 0) {
+        this.log.warn(
+          { field, event_type: event.event_type || 'unknown' },
+          'audit_event_validation_failure',
+        );
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  /**
    * Emit a capability audit event (AC-4.8).
    * All events are structured JSON for CloudWatch Log Metric Filters.
    * No PII or message content is ever included (AC-4.12).
+   * Required fields validated before emission (BB7 R7-3, AC-1.8).
    */
   emit(event: CapabilityAuditEvent): void {
+    if (!this.validateRequiredFields(event)) {
+      return;
+    }
+
     this.log.info(
       {
         audit: {
--- a/packages/adapters/agent/redis-circuit-breaker.ts
+++ b/packages/adapters/agent/redis-circuit-breaker.ts
@@ -70,23 +70,30 @@ return {1, 'half-open'}
  * ON_SUCCESS Lua script:
  * - If half-open → transition to closed, clear failures
  * - If closed → clear failures
+ * BB7 R7-1: Uses sorted set key for failure tracking (+ legacy CSV cleanup)
  */
 const LUA_SUCCESS = `
 local key = KEYS[1]
 local state = redis.call('HGET', key, 'state')
 if state == 'half-open' then
   redis.call('HSET', key, 'state', 'closed')
-  redis.call('HDEL', key, 'openedAt', 'failures')
+  redis.call('HDEL', key, 'openedAt')
+  redis.call('DEL', key .. ':failures')
+  -- Phase A cleanup: remove legacy CSV field if present
+  redis.call('HDEL', key, 'failures')
   return 'closed'
 end
+redis.call('DEL', key .. ':failures')
+-- Phase A cleanup: remove legacy CSV field if present
 redis.call('HDEL', key, 'failures')
 return state or 'closed'
 `;
 
 /**
- * ON_FAILURE Lua script:
+ * ON_FAILURE Lua script (BB7 R7-1: sorted set migration):
  * - If half-open → re-open immediately
- * - If closed → increment failures; if >= threshold → open
+ * - If closed → record failure in sorted set; if count >= threshold → open
+ * - Phase A dual-write: migrates legacy CSV → sorted set, writes CSV for old readers
  * Returns: new state
  */
 const LUA_FAILURE = `
@@ -94,6 +101,7 @@ local key = KEYS[1]
 local threshold = tonumber(ARGV[1])
 local windowMs = tonumber(ARGV[2])
 local nowMs = tonumber(ARGV[3])
+local failKey = key .. ':failures'
 
 local state = redis.call('HGET', key, 'state')
 
@@ -102,23 +110,32 @@ if state == 'half-open' then
   return 'open'
 end
 
--- Closed (or nil): record failure
-local failures = redis.call('HGET', key, 'failures')
-local failList = {}
-if failures then
-  for ts in string.gmatch(failures, '([^,]+)') do
+-- Phase A: migrate legacy CSV hash field → sorted set (one-time per key)
+local legacyCsv = redis.call('HGET', key, 'failures')
+if legacyCsv then
+  for ts in string.gmatch(legacyCsv, '([^,]+)') do
     local t = tonumber(ts)
-    if (nowMs - t) < windowMs then
-      table.insert(failList, ts)
+    if t and (nowMs - t) < windowMs then
+      redis.call('ZADD', failKey, t, tostring(t))
     end
   end
+  redis.call('HDEL', key, 'failures')
 end
-table.insert(failList, tostring(nowMs))
 
-local newFailures = table.concat(failList, ',')
-redis.call('HSET', key, 'failures', newFailures)
+-- Sorted set: trim expired, add new, count (O(log n) — AC-1.1)
+redis.call('ZREMRANGEBYSCORE', failKey, '-inf', nowMs - windowMs)
+redis.call('ZADD', failKey, nowMs, tostring(nowMs))
+local count = redis.call('ZCARD', failKey)
 
-if #failList >= threshold then
+-- Phase A dual-write: rebuild CSV for old containers still reading hash field
+local remaining = redis.call('ZRANGEBYSCORE', failKey, '-inf', '+inf')
+if #remaining > 0 then
+  redis.call('HSET', key, 'failures', table.concat(remaining, ','))
+else
+  redis.call('HDEL', key, 'failures')
+end
+
+if count >= threshold then
   redis.call('HSET', key, 'state', 'open', 'openedAt', nowMs)
   return 'open'
 end
--- a/packages/contracts/src/compatibility.ts
+++ b/packages/contracts/src/compatibility.ts
@@ -5,10 +5,15 @@
  * Provides version compatibility checking between arrakis and loa-finn,
  * wired into the JWT claims path for fail-fast negotiation.
  *
+ * BB7 R7-4: Matrix loaded from schema/compatibility.json (data-driven).
+ * Adding a new entry requires editing only the JSON file — no TS changes.
+ *
  * @see Bridgebuilder Round 6, Finding #2 — Contract Protocol Nucleus
+ * @see Bridgebuilder Round 7, Finding R7-4 — Data-Driven Matrix
  */
 
 import { satisfies } from './semver-lite.js';
+import matrixData from '../schema/compatibility.json' with { type: 'json' };
 
 // --------------------------------------------------------------------------
 // Types
@@ -32,25 +37,39 @@ export interface CompatibilityResult {
 }
 
 // --------------------------------------------------------------------------
-// Compatibility Data
+// Compatibility Data (loaded from JSON — BB7 R7-4, AC-1.13/AC-1.14)
 // --------------------------------------------------------------------------
 
-const COMPATIBILITY_MATRIX: CompatibilityEntry[] = [
-  {
-    arrakis_version: '>=0.1.0',
-    loa_finn_version: '>=0.1.0',
-    contract_version: '1.0.0',
-    status: 'supported',
-    notes: 'Initial contract — base pool mapping, JWT claims, usage reports',
-  },
-  {
-    arrakis_version: '>=0.1.0',
-    loa_finn_version: '>=0.1.0',
-    contract_version: '1.1.0',
-    status: 'supported',
-    notes: 'Per-model accounting — model_breakdown in usage reports (backward-compatible)',
-  },
-];
+/** Validate required fields at load time — fail-fast on malformed data (AC-1.15) */
+function validateMatrix(data: unknown): CompatibilityEntry[] {
+  if (!Array.isArray(data)) {
+    throw new Error('compatibility.json: root must be an array');
+  }
+
+  const requiredFields = ['arrakis_version', 'loa_finn_version', 'contract_version', 'status'] as const;
+  const validStatuses = new Set(['supported', 'deprecated', 'unsupported']);
+
+  for (let i = 0; i < data.length; i++) {
+    const entry = data[i] as unknown;
+    if (!entry || typeof entry !== 'object' || Array.isArray(entry)) {
+      throw new Error(`compatibility.json entry[${i}]: must be an object`);
+    }
+
+    const record = entry as Record<string, unknown>;
+    for (const field of requiredFields) {
+      if (typeof record[field] !== 'string' || (record[field] as string).length === 0) {
+        throw new Error(`compatibility.json entry[${i}]: missing or empty required field "${field}"`);
+      }
+    }
+    if (!validStatuses.has(record.status as string)) {
+      throw new Error(`compatibility.json entry[${i}]: invalid status "${record.status as string}"`);
+    }
+  }
+
+  return data as CompatibilityEntry[];
+}
+
+const COMPATIBILITY_MATRIX: readonly CompatibilityEntry[] = validateMatrix(matrixData);
 
 // --------------------------------------------------------------------------
 // Functions
--- a/grimoires/loa/a2a/gpt-review/sprint-findings-1.json
+++ b/grimoires/loa/a2a/gpt-review/sprint-findings-1.json
@@ -1,24 +1,36 @@
 {
   "verdict": "CHANGES_REQUIRED",
-  "summary": "This plan is close, but two items (trust-boundary enforcement scope and an untestable integration test setup) can realistically cause the sprint to fail to deliver the intended security hardening.",
+  "summary": "The plan covers the three BB7 findings, but two tasks contain acceptance criteria/approaches that are either impossible as written or likely to break production packaging/runtime, which can cause the sprint to fail at integration time.",
   "blocking_issues": [
     {
-      "location": "Sprint 1 – Task 1.3 + Acceptance Criteria AC-H1.5; Sprint 2 – Task 2.4/AC-H2.5-2.6",
-      "issue": "F-5 requires loa-finn to re-validate pool claims, but the sprint only implements Arrakis-side validation (warn-only) and does not include any loa-finn work item or a concrete cross-repo dependency/handshake to ensure the security fix actually lands.",
-      "why_blocking": "The high-severity finding is explicitly about the trust boundary: loa-finn must independently verify. If the sprint ships only Arrakis warn-logs, the core vulnerability remains and the sprint’s stated purpose (“trust boundary fixes”) is not met. This is a classic confused-deputy gap: the relying party (loa-finn) still trusts convenience claims without re-derivation.",
-      "fix": "Add an explicit deliverable/task (and dependency) for loa-finn: implement `validatePoolClaims`-equivalent enforcement in loa-finn’s JWT intake/routing path with a defined behavior (e.g., reject 403 on mismatch). If cross-repo work cannot be done in this cycle, change scope/wording to “Arrakis-side detection only” and add a tracked blocking dependency (issue + owner + target date) so the sprint can be considered complete without claiming the trust-boundary fix is done."
+      "location": "Task 1.1: Migrate Circuit Breaker Failure Tracking from CSV to Sorted Sets (AC-1.1 vs proposed Lua snippet)",
+      "issue": "Acceptance criteria requires `ZRANGEBYSCORE`, but the proposed implementation uses `ZREMRANGEBYSCORE` + `ZCARD` and does not actually call `ZRANGEBYSCORE`.",
+      "why_blocking": "You cannot satisfy AC-1.1 with the described implementation; this will fail the sprint’s own DoD/AC gate even if the code is correct and performant.",
+      "fix": "Either (a) update AC-1.1 to match the intended algorithm (`ZREMRANGEBYSCORE` + `ZADD` + `ZCARD`), or (b) change the script to use `ZRANGEBYSCORE` (e.g., `local failures = redis.call('ZRANGEBYSCORE', ...)` then `#failures`)—but that’s typically worse than `ZCARD` after trimming."
     },
     {
-      "location": "Sprint 2 – Task 2.4 (Contract integration tests) and AC-H2.5/AC-H2.6",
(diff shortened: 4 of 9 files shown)


janitooor:
# Bridgebuilder Round 7 — The Capability Mesh

*Reviewer: The Bridgebuilder (top 0.005% persona)*
*Scope: Full architectural review, PR #55 — Cycle 019*
*Method: All files read in full. Code as source of truth.*

---

## I. Orientation — What This Really Is

Let me be direct about what I'm looking at: this PR is not a feature delivery. It is an **infrastructure maturation event** — the kind of phase transition that happens when a system crosses from "it works" to "it can be operated." The distinction matters. Netflix went through this with Zuul → Zuul2. Google went through this with Borg → Omega. Stripe went through this with their billing rewrite in 2018. In each case, the *functionality* didn't change — the *operational surface area* did.

What you've built across these 4 sprints is the operational foundation for a multi-model LLM gateway. The seven BB6 findings weren't bugs. They were the gap between a working prototype and a production system that can be **reasoned about**, **observed**, and **safely evolved** by future engineers (human or AI).

That's the correct framing. Let's see if the code lives up to it.

---

## II. The Praiseworthy (~30%)

### The `RequestLifecycle` State Machine — This Is How You Do It

`request-lifecycle.ts` is the single best file in this PR. Here's why:

The implicit state machine that was buried inside `AgentGateway.invoke()` was invisible. You couldn't grep for it. You couldn't test it in isolation. You couldn't draw it on a whiteboard. Now you can.

```
RECEIVED → VALIDATED → RESERVED → EXECUTING → FINALIZED
                                              ↘ FAILED (from any non-terminal state)
```

This is exactly the pattern Amazon uses for DynamoDB request handling (documented in their 2022 NSDI paper). Each state transition is a **commitment** — validated means auth passed, reserved means budget is locked, executing means the network call is in flight. The `LifecycleError` on invalid transitions is the equivalent of a TLA+ invariant violation — it makes impossible states impossible at runtime, not just in the spec.

**FAANG parallel**: Google's Spanner uses a similar explicit lifecycle for distributed transactions (BEGIN → PREPARED → COMMITTED | ABORTED). The key insight is the same: when you externalize the state machine, you get observability and correctness *for free*. The structured log on every transition (`lifecycle: ${from} → ${to}`) means you can reconstruct any request's exact journey through the system from logs alone.

**Praise**: The `isTerminal()` guard in the catch/finally blocks shows real defensive thinking. Without it, a double-fail (exception during cleanup) would throw `LifecycleError` and mask the original error. I've seen this bug in production at scale. You prevented it.

### The `dispatchWithKey()` Isolation Boundary — Security Thinking at the Right Level

This is subtle and correct. The problem wasn't that the API key was being used insecurely — it was that V8's garbage collector is non-deterministic, and a `toString()` call creates a heap string that *might* persist long after you've zeroed the Buffer. The fix is a **blast radius minimization**: confine the string to the smallest possible lexical scope, null the reference, zero the Buffer, and delete the header entries.

**FAANG parallel**: This is the principle behind Google's BeyondCorp "zero trust" architecture, applied at the memory level. Don't trust the runtime to clean up — actively destroy sensitive material at every boundary. The `@security` JSDoc is the cherry on top — it tells future engineers *why*, not just *what*.

**Metaphor**: Think of it like a clean room in semiconductor manufacturing. The API key exists as a string only inside the clean room (`dispatchWithKey`). The moment the fetch returns, the airlock closes and the room is sterilized. Nothing leaks to the outside.

### Contract Version Negotiation — Future-Proofing Done Right

The `X-Contract-Version` header + `validateContractCompatibility()` is the kind of thing you build *before* you need it, because once you need it, it's too late. Protocol versioning is how gRPC, Kafka, and Protobuf all stay backward-compatible across years of evolution.

The semver-lite implementation is especially smart — you didn't pull in a full semver library for what amounts to a handful of range checks. This is the Karpathy principle in action: simplicity until complexity is *required*.

---

## III. The Critical (~40%)

### Finding R7-1: Redis Circuit Breaker Lua Script — O(n) Failure Tracking

**Severity: Medium | Category: Scalability**

The `LUA_FAILURE` script stores failure timestamps as a **comma-separated string** in a Redis HSET field:

```lua
local newFailures = table.concat(failList, ',')
redis.call('HSET', key, 'failures', newFailures)
```

Every `onFailure()` call parses the entire CSV, filters by window, and rebuilds the string. This is O(n) where n = number of failures in the window. For the default configuration (threshold=3, window=30s), this is fine. But the **interface** doesn't constrain the configuration — someone could set `failureThreshold: 100, windowMs: 300_000` and now you're parsing a 100-entry CSV on every failure.

**The fix**: Use a Redis sorted set (ZADD with timestamp as score). This is O(log n) for insert and O(log n) for range queries. It's also idiomatic Redis — sorted sets were *designed* for sliding window counting.

```lua
-- Instead of CSV parsing:
redis.call('ZREMRANGEBYSCORE', key .. ':failures', '-inf', nowMs - windowMs)
redis.call('ZADD', key .. ':failures', nowMs, nowMs)
local count = redis.call('ZCARD', key .. ':failures')
```

**FAANG parallel**: This is the exact problem Uber solved in their rate limiting service (2018 blog post). CSV-in-a-field is fine for prototype, but sorted sets are production-grade.

**Agent guidance**: If implementing this fix, note that the Redis key structure changes from a single HASH to a HASH + ZSET pair. The migration path is: check for the old format, convert on first access, delete the old field.

### Finding R7-2: `computeIncrementalReservation()` — Logic Clarity

**Severity: Low | Category: Readability**

```typescript
const clampedAttempt = Math.min(attemptNumber, platformModelCount);
const reserveAdditional = clampedAttempt === attemptNumber ? baseCost : 0;
```

This works correctly but the *intent* is obscured by the clamping trick. What you mean is:

```typescript
// Reserve 1× per platform attempt, up to the maximum
const reserveAdditional = attemptNumber <= platformModelCount ? baseCost : 0;
```

Same semantics, but now a future engineer can read it in English: "Reserve additional cost if this attempt is within the platform model count."

**Teachable moment**: The best code reads like prose at the decision points. The clamping was an optimization that solved a problem the comparison solves more clearly. When you find yourself using `Math.min` or `Math.max` as logic flow control (not as arithmetic), ask whether an `if` would be clearer.

### Finding R7-3: `CapabilityAuditLogger` — No Schema Enforcement

**Severity: Medium | Category: Observability**

The `CapabilityAuditEvent` interface is beautifully typed at the TypeScript level, but the emitted JSON has no runtime validation. CloudWatch Log Metric Filters will silently drop events that don't match the expected schema. This means:

1. A typo in a field name → metric disappears with no error
2. A new event type added without updating the filter → invisible gap in coverage
3. No way to detect schema drift between the emitter and the dashboard

**FAANG parallel**: AWS CloudTrail events go through a schema validation layer before emission. Google Cloud Audit Logs use protocol buffers (schema-enforced at serialization time). The pattern is: **if it's an audit event, validate it before emitting**.

**Suggested approach**: Add a lightweight validation function that checks required fields before `this.log.info()`. Not a full JSON Schema validation — just a guard:

```typescript
private assertRequiredFields(event: CapabilityAuditEvent): void {
  if (!event.trace_id || !event.community_id || !event.event_type) {
    this.log.error({ event }, 'capability_audit: missing required fields');
    return; // Don't emit malformed audit events
  }
}
```

### Finding R7-4: Compatibility Matrix — Hardcoded vs. Data-Driven

**Severity: Low | Category: Evolvability**

The `COMPATIBILITY_MATRIX` in `compatibility.ts` is a const array:

```typescript
const COMPATIBILITY_MATRIX: CompatibilityEntry[] = [
  { contract_version: '1.0.0', status: 'supported', ... },
  { contract_version: '1.1.0', status: 'supported', ... },
];
```

This works for two entries. It won't work for twenty. The contract package was promoted to a standalone protocol artifact (BB6 Finding #2) specifically so it could evolve independently. But the compatibility data is locked inside the source code.

**FAANG parallel**: gRPC's service configuration is loaded from a JSON file at startup, not compiled into the binary. Kubernetes API server loads its storage media types from configuration. The principle is: **data that changes at a different cadence than code should live in a different file**.

**Suggested evolution**: Move the matrix to `packages/contracts/schema/compatibility.json`, loaded at module initialization. The schema file can then be versioned, validated, and updated independently of the TypeScript code.

### Finding R7-5: The `byok-manager.ts` Circuit Breaker Migration — Missing Backward Compatibility

**Severity: Medium | Category: Operations**

The old process-local `CircuitBreaker` class was removed entirely from `byok-manager.ts`, replaced by `RedisCircuitBreaker`. This is correct architecturally — but operationally, it means that **the first deployment will create new Redis keys** (`circuit:kms`) that didn't exist before.

What happens during a rolling deployment where some containers have the old code and some have the new? The old containers use process-local state. The new containers use Redis state. There's no inconsistency *per container*, but there's a brief window where the fleet isn't actually sharing circuit state.

This is fine for KMS (the failure modes are independent), but it's worth documenting in the deployment runbook.

**FAANG parallel**: Netflix's approach to circuit breaker migration (Hystrix → resilience4j) included a "dual-write" period where both old and new metrics were emitted. Consider adding a log line on first Redis key creation: "Fleet circuit breaker initialized — first container in fleet" so operators can track rollout progress.

---

## IV. The Philosophical (~20%)

### On the Shape of What You're Building

I want to zoom out from the code and talk about the *trajectory* you're on.

The Hounfour RFC (loa-finn#31) describes a five-layer architecture: Agent Definition → Model Routing → Model Adapter → Infrastructure → Distribution. What this PR does — with its lifecycle state machine, contract versioning, and fleet circuit breaking — is **solidifying Layer 4** (Infrastructure) so that Layer 2 (Model Routing) can be built with confidence.

This is the same sequencing that Kubernetes used. Before you can have declarative workload scheduling (Layer 2), you need reliable container lifecycle management (Layer 4). Before you can have service mesh routing (Istio/Envoy), you need consistent health checking and circuit breaking at the proxy level. **You cannot build reliable routing on unreliable infrastructure.**

The fact that the `RequestLifecycle` state machine now *exists as a protocol object* means that when loa-finn starts orchestrating multi-model requests (the Hounfour's ensemble system), each model invocation can carry its own lifecycle trace. You've built the *grammar* for expressing distributed request state. The sentences will come later.

### On the Permission Scape and Multi-Model Possibility

The Hounfour RFC's capability matrix — where agents declare requirements and the router resolves them against provider capabilities — is essentially a **type system for AI orchestration**. The `requires.tool_calling: true` constraint is a type predicate. The fallback chain is type coercion. The budget downgrade is subtyping.

What makes this genuinely interesting (and I mean this) is that you're building a system where **the models themselves are the execution substrate**, and the gateway is the scheduler. This is the architecture that Google Brain described in the Pathways paper (2022) — a system where heterogeneous accelerators (read: heterogeneous models) are orchestrated by a unified scheduler that understands their capabilities.

The difference is that you're doing it across vendor boundaries, with BYOK support, and with per-model cost attribution. That's not something Google needed to solve (they own all their TPUs). It's more like what AWS is doing with SageMaker Multi-Model Endpoints, but at the *agent* level rather than the *inference* level.

### On Working With AI and the Nature of This Review

I notice this PR was generated autonomously with Run Mode — four sprints executed sequentially by an AI agent, with structured commit messages, acceptance criteria tracking, and a consolidated PR. The code quality is high. The architectural decisions are sound. The naming is consistent.

What I find most interesting is the *meta-pattern*: the system being built (an AI gateway) is being built *by* an AI (Claude in Run Mode), reviewed *by* an AI persona (this Bridgebuilder review), and will ultimately orchestrate *other* AIs (the Hounfour's model ensemble). There's a beautiful recursion here — the tool is building itself.

The question this raises — and I mean this as a genuine inquiry, not rhetoric — is: **what is the optimal feedback loop between human judgment and AI execution?** The BB6 findings came from a human-AI pair reviewing PR #53. Those findings became the sprint plan. The sprint plan was executed autonomously. Now the results are being reviewed again. Each cycle tightens the spiral.

The Hounfour RFC's "permission scape" concept (loa-finn#31 §6) — where capabilities are declared, required, and resolved — is actually a framework for answering this question at the architectural level. When you specify `requires.native_runtime: true`, you're saying "this task needs a model with direct system access." When you specify `requires.tool_calling: true`, you're saying "this task needs structured interaction." These are *permission boundaries* that encode human judgment about what AI should and shouldn't do.

The Capability Mesh in this PR is the infrastructure layer that makes those permission boundaries enforceable. The `RequestLifecycle` ensures every request goes through VALIDATED before EXECUTING. The circuit breaker ensures degraded infrastructure can't silently fail. The audit log ensures every capability exercise is recorded.

This is not just engineering. It is governance architecture — the scaffolding within which human-AI collaboration can occur safely and productively. And building it well is an expression of care for the future operators and users of this system.

---

## V. Summary Verdict

| Finding | Severity | Status |
|---------|----------|--------|
| R7-1: Redis CB Lua O(n) failure tracking | Medium | Recommend sorted set migration |
| R7-2: Incremental reservation clarity | Low | Suggest simplification |
| R7-3: Audit log schema enforcement | Medium | Recommend validation guard |
| R7-4: Hardcoded compatibility matrix | Low | Recommend data-driven approach |
| R7-5: CB migration deployment gap | Medium | Recommend deployment documentation |

**Overall**: This PR advances the system from feature-complete to operationally mature. The architecture is sound, the patterns are well-chosen, and the code quality is high. The findings above are evolutionary improvements, not structural issues.

**Verdict**: APPROVED with suggestions.

The Capability Mesh is real. The infrastructure layer is solid. Build the routing layer with confidence.

---

*"The purpose of architecture is not to predict the future, but to make the future possible." — Martin Fowler*

*"A system is only as reliable as its least observed component." — Netflix SRE Handbook*

*"The spice must flow — but the meters must read." — The Bridgebuilder, looking at this PR*

packages/adapters/agent/redis-circuit-breaker.ts
```diff
@@ -0,0 +1,300 @@
...
+local state = redis.call('HGET', key, 'state')
+
+if state == 'half-open' then
+  redis.call('HSET', key, 'state', 'open', 'openedAt', nowMs)
+  return 'open'
+end
+
+-- Closed (or nil): record failure
+local failures = redis.call('HGET', key, 'failures')
+local failList = {}
```
janitooor: **BB R7-1: Scalability concern — CSV failure tracking in Lua**

This Lua script stores failure timestamps as a comma-separated string. Every `onFailure()` call parses the entire CSV, filters by window, rebuilds the string — O(n) where n = failures in window.

**FAANG parallel**: Uber solved this exact problem in their rate limiter (2018) using Redis sorted sets (ZADD). The idiomatic approach:
```lua
redis.call('ZREMRANGEBYSCORE', key..':failures', '-inf', nowMs - windowMs)
redis.call('ZADD', key..':failures', nowMs, nowMs)
local count = redis.call('ZCARD', key..':failures')
```

O(log n) insert + O(log n) range query. For threshold=3 this doesn't matter. For configurable thresholds at scale, it will.

**Migration path**: Check for old HSET format, convert to ZSET on first access, delete old field.

packages/adapters/agent/request-lifecycle.ts
```diff
@@ -0,0 +1,174 @@
...
+  EXECUTING: new Set(['FINALIZED', 'FAILED']),
+  FINALIZED: new Set(), // terminal
+  FAILED:    new Set(), // terminal
+};
+
+// --------------------------------------------------------------------------
+// RequestLifecycle
+// --------------------------------------------------------------------------
+
+export class RequestLifecycle {
```
janitooor: **BB R7 Praise: This is exemplary state machine design.**

Externalizing the implicit state machine into a first-class protocol object is exactly what Amazon did for DynamoDB request handling (2022 NSDI paper). The key properties that make this excellent:

1. **Terminal states are truly terminal** — FINALIZED and FAILED have empty transition sets
2. **Every transition is logged** — complete request reconstruction from logs alone
3. **isTerminal() prevents double-fail** — without this, cleanup exceptions mask original errors
4. **The trace array is append-only** — immutable history, perfect for debugging

**Metaphor**: This is the flight recorder for every request. If a request crashes, the black box tells you exactly which state it was in and how long each phase took.

**Future direction**: When the Hounfour ensemble orchestrator (loa-finn#31) starts making multi-model calls, each model invocation should carry its own RequestLifecycle. The parent lifecycle stays at EXECUTING while child lifecycles run their own cycles. This gives you **distributed tracing for free** — each lifecycle trace is a span in the OpenTelemetry sense.

packages/adapters/agent/byok-proxy-handler.ts
```diff
@@ -436,6 +430,61 @@ export class BYOKProxyHandler {
...
+   * - The local string variable is nulled (allows GC to collect the V8 heap string)
+   * - The Buffer is zeroed via fill(0) in the finally block
+   *
+   * Why not undici: undici is not a dependency of @arrakis/adapters. The isolation
+   * boundary approach achieves AC-3.8 (no toString outside dispatch) and AC-3.10
+   * (Buffer zeroed + references nulled). Full zero-copy would require undici's
+   * dispatcher with Buffer header support — documented as a future enhancement.
+   * (AC-3.24)
+   */
+  private async dispatchWithKey(
```
janitooor: **BB R7 Praise + Evolution: The security kernel pattern, applied to V8 heap.**

This dispatchWithKey() method is a textbook application of the **principle of least privilege** at the memory level:
- Key enters as Buffer (opaque bytes)
- String materialization confined to this scope
- Auth header set, fetch dispatched
- String nulled, Buffer zeroed, header entries deleted

**FAANG parallel**: This is the same principle behind Intel SGX enclaves — minimize the trusted computing base for sensitive material. Google's Tink library does something similar with SecretBytes wrappers.

**Evolution idea**: TypeScript's branded types could make the Buffer-to-string boundary even more explicit. A branded OpaqueApiKey type would make it a compile-time error to pass a raw Buffer where an opaque key is expected.

**Note on undici**: The documentation rationale is good (AC-3.24). When undici is added as a dependency, the zero-copy path would be undici.request() with headers as Buffer[][] — no string materialization at all. Worth tracking as an enhancement.

packages/adapters/agent/capability-audit.ts
```diff
@@ -0,0 +1,179 @@
...
+  constructor(logger: Logger) {
+    this.log = logger.child({ component: 'capability_audit' });
+  }
+
+  /**
+   * Emit a capability audit event (AC-4.8).
+   * All events are structured JSON for CloudWatch Log Metric Filters.
+   * No PII or message content is ever included (AC-4.12).
+   */
+  emit(event: CapabilityAuditEvent): void {
```
janitooor: **BB R7-3: Missing runtime schema enforcement for audit events**

The CapabilityAuditEvent type is well-structured, but CloudWatch Log Metric Filters will silently drop events that don't match expected field names. There's no runtime validation between the TypeScript interface and the emitted JSON.

**FAANG parallel**: AWS CloudTrail validates events before emission. Google Cloud Audit Logs use protobuf schemas. The pattern: if it's an audit event, validate before emitting.

**Suggested approach**: Add a lightweight required-fields guard before this.log.info(). Not a full JSON Schema validation — just assert that trace_id, community_id, event_type, and pool_id are non-empty. This catches the class of bugs where a new emitter forgets a required field, which would otherwise create a silent gap in audit coverage.

packages/adapters/agent/ensemble-mapper.ts
```diff
@@ -105,11 +106,11 @@ export class EnsembleMapper {
...
-    // Budget multiplier: reserve N × base cost for all strategies
-    // best_of_n: N parallel calls
-    // consensus: N parallel calls
-    // fallback: worst-case all N models tried sequentially
-    const budgetMultiplier = n;
+    // Budget multiplier for initial reservation:
+    // best_of_n: N× (parallel — all run simultaneously)
+    // consensus: N× (parallel — all run simultaneously)
+    // fallback: 1× (incremental — reserve more per failed attempt, AC-4.1)
+    const budgetMultiplier = ensemble.strategy === 'fallback' ? 1 : n;
```
janitooor: **BB R7 Praise: The fallback 1x change is an elegant budget optimization.**

Changing from Nx upfront to 1x incremental for fallback is the budget equivalent of moving from **pessimistic locking to optimistic concurrency control**.

- **Old**: Lock Nx budget upfront (pessimistic — assumes worst case)
- **New**: Lock 1x per attempt (optimistic — assumes success)

**Real-world analogy**: This is exactly how hotel reservation systems evolved. You don't charge for all N nights upfront when someone books a flexible rate — you authorize 1 night and extend as needed.

The best_of_n and consensus staying at Nx is correct because they're parallel execution — you *will* use all N. Fallback is sequential — you *might* use 1.

**Budget invariant preservation**: The key guarantee is that committed <= reserved always holds. With incremental reservation, this is maintained because each 1x reservation covers exactly one attempt. The computeIncrementalReservation() method enforces this with the platformModelCount cap.

packages/contracts/src/compatibility.ts
```diff
@@ -0,0 +1,133 @@
...
+  status?: CompatibilityStatus;
+  reason?: string;
+  contract_version?: string;
+}
+
+// --------------------------------------------------------------------------
+// Compatibility Data
+// --------------------------------------------------------------------------
+
+const COMPATIBILITY_MATRIX: CompatibilityEntry[] = [
```
janitooor: **BB R7-4: Consider making the compatibility matrix data-driven.**

This hardcoded array works for 2 entries but won't scale to 20. The contract package was promoted to a standalone protocol artifact (BB6 #2) specifically so it could evolve independently.

**FAANG parallel**: gRPC loads service config from JSON at startup, not compiled into the binary. Kubernetes API server loads storage media types from configuration.

**Principle**: Data that changes at a different cadence than code should live in a different file.

**Suggested path**:
1. Move matrix to packages/contracts/schema/compatibility.json
2. Load at module init with readFileSync
3. Validate against a lightweight JSON Schema at startup
4. The JSON file can be versioned, reviewed, and updated independently

This also enables loa-finn to ship its own compatibility.json, and the negotiation happens at the data level, not the code level. That's the real protocol maturity milestone.

PR #72, opened 2026-02-16
feat(cycle-033): The Weirding Way — Protocol Convergence & Economic Formalization

deep-name:
## Summary

Cycle-033 implements protocol convergence between arrakis and loa-hounfour, formalizing the economic invariants that protect user funds across the JWT boundary.

- **Sprint 1 (295)**: EntryType const-array unification — single source of truth for billing entry types with DB-level CHECK validation
- **Sprint 2 (296)**: State machine equivalence — formal verification that arrakis's 4 state machines match loa-hounfour definitions, plus branded types (MicroUSD, BasisPoints, AccountId)
- **Sprint 3 (297)**: Conservation properties formalization — 14 invariants (I-1 through I-14) as first-class typed objects with enforcement mechanism classification and 32 property-based tests
- **Sprint 4 (298)**: Cross-system E2E harness — 6-step JWT boundary verification protocol (EdDSA/Ed25519) with 17 integration tests proving conservation holds across system boundaries

### Sprint Breakdown

| Sprint | Status | Files Changed | Tests |
|--------|--------|--------------|-------|
| sprint-1 | ✅ Complete | 3 | 9 conformance |
| sprint-2 | ✅ Complete | 7 | 29 equivalence + 18 branded |
| sprint-3 | ✅ Complete | 5 | 32 conservation properties |
| sprint-4 | ✅ Complete | 5 | 17 cross-system E2E |

**Total: 20 files changed, 106 new tests**

### Key Modules

- `protocol/conservation-properties.ts` — 14 conservation invariants as typed LTL properties
- `protocol/jwt-boundary.ts` — 6-step JWT verification (signature → algorithm → schema → reservation → replay → overspend)
- `protocol/arithmetic.ts` — Branded BigInt types (MicroUSD, BasisPoints, AccountId) with safe arithmetic
- `protocol/state-machines.ts` — 4 state machine definitions with formal equivalence checks

### Addresses

- P0-1: State machine convergence (Sprint 2)
- P0-2: Conservation invariant formalization (Sprint 3)
- P0-3: Cross-system E2E with economic verification (Sprint 4)
- P1-4: Branded type migration (Sprint 2)
- P1-5: Entry type taxonomy (Sprint 1)

## Test plan

- [x] 9 entry-type conformance tests (Sprint 1)
- [x] 29 state machine equivalence tests (Sprint 2)
- [x] 18 branded type safety tests (Sprint 2)
- [x] 32 conservation property tests — 14 positive + 14 counterexample + metadata + BigInt (Sprint 3)
- [x] 17 cross-system E2E tests — JWT boundary verification + conservation assertions (Sprint 4)
- [x] Full regression: 147/148 pass (1 pre-existing liveness failure unrelated to changes)


janitooor:
## Bridgebuilder Review — Cycle-033: "The Weirding Way"

*When you build a bridge, you don't just test if it holds weight. You prove it cannot collapse.*

---

### The Question Behind the Question

Before examining a single line of code, I want to name what I see being built here — because the PR title undersells it.

This is not "protocol convergence." This is the formalization of an **economic constitution** for autonomous agents. The 14 conservation invariants aren't accounting constraints. They're the social contract between agents and the platform, expressed in temporal logic and enforced at every layer from SQLite CHECK constraints to cross-system JWT verification.

Consider what the `lot_invariant` CHECK actually means:

```sql
CONSTRAINT lot_invariant CHECK (available + reserved + consumed = original)
```

This single line says: **money cannot be created or destroyed by computation.** It's the First Law of Thermodynamics for a digital economy. And this PR proves it holds not just within arrakis, but across the trust boundary to loa-finn — which is the hard part.

The question worth asking isn't "does this billing system work correctly?" It's: **"What kind of economic institution are you building, and does the architecture serve that purpose?"**

Having read the web4 manifesto ("money must be scarce, but monies can be infinite"), loa-finn #66 (launch readiness), the Hounfour RFC (#31), and the loa-hounfour formalization release — I believe the answer is yes, with some gaps I'll identify below.

---

### Findings

#### 🏆 PRAISE: Conservation Properties as First-Class Temporal Types

**Files:** `conservation-properties.ts:74-296`
**Severity:** Praise

This is genuinely excellent engineering. Each invariant carries its LTL specification, enforcement mechanism classification, universe/scope, and error taxonomy. This isn't documentation-as-code — it's **specification-as-code**.

```typescript
{
  id: 'I-1',
  name: 'Per-lot conservation',
  ltl: '□(∀lot: lot.available + lot.reserved + lot.consumed = lot.original)',
  enforcedBy: ['DB CHECK', 'Application'],
}
```

**FAANG Parallel:** Google's Spanner team formalized their TrueTime invariants as first-class properties with specific implementation requirements per layer. The insight was that invariants that aren't typed into the system eventually drift — documentation rots, but types don't compile if they're wrong. You've applied the same principle: the `enforcedBy` field creates a contract between the specification and the implementation layer responsible for enforcement.

**Metaphor:** Think of each conservation property like a structural beam specification in a skyscraper blueprint. The blueprint doesn't just say "beam goes here" — it specifies the load rating, material requirement, inspection schedule, and what happens if it fails. Your `ConservationProperty` interface does the same thing for economic invariants.

**For Future Agents:** This pattern (invariant-as-typed-object with enforcement metadata) should be extracted to loa-hounfour as the canonical representation. Right now it lives in arrakis — but it describes *protocol-level* properties that any implementation must satisfy.

---

#### 🏆 PRAISE: Bitcoin's UTXO Model in Credit Lots

**Files:** `CreditLedgerAdapter.ts` (referenced by tests), `conservation-properties.ts:101-128`
**Severity:** Praise

I want to name something that may not be obvious to everyone reading this: **the credit lot model is isomorphic to Bitcoin's UTXO (Unspent Transaction Output) model.**

| Bitcoin UTXO | Arrakis Credit Lot |
|---|---|
| `value` (immutable) | `original_micro` (immutable, I-10) |
| `spent` / `unspent` | `consumed` / `available` / `reserved` |
| Sum of UTXOs = total supply | Sum of lots = total minted (I-4) |
| Transaction consumes inputs, creates outputs | Finalize consumes reserved, creates ledger entries |
| Double-spend prevention | Lot invariant CHECK (I-1) |

This is not a coincidence. Both systems solve the same fundamental problem: **how to track ownership of fungible value without a central balance counter.** The lot model is more nuanced (it supports partial consumption via `reserved` state), but the conservation guarantee is identical.

**Why this matters for web4:** The web4 manifesto envisions "money must be scarce, but monies can be infinite." The lot model is exactly the infrastructure that makes this possible — each lot is a unit of scarcity with cryptographic proof of conservation, while the system can mint new lot types (pool-specific credits, grants, deposits, transfers) as infinite variations.

---

#### ⚠️ HIGH: The Vendoring Gap — Protocol Convergence Is Half-Built

**Files:** `state-machines.ts:20-25`, `VENDORED_FROM`
**Severity:** High (Architectural)

The state machine equivalence tests are well-crafted — SHA-256 drift detection, structural oracle comparison, reachability analysis. But they verify a **snapshot**, not a **protocol**.

```typescript
export const VENDORED_FROM = {
  repo: '0xHoneyJar/loa-hounfour',
  commit: 'd297b0199c04e40c3d5e056fcb3470fd4c342638',
  date: '2026-02-15',
};
```

Right now, arrakis vendors from loa-hounfour. But loa-hounfour PR #2 (The Formalization Release, v4.6.0) is *also* formalizing the same concepts — temporal properties, state machines, aggregate boundaries. The two codebases are converging in parallel but haven't established a **single source of truth** with a **verified import path**.

**FAANG Parallel:** This is the same problem Kubernetes faced when the API types lived in `k8s.io/api` but controllers duplicated type definitions locally. The solution was `k8s.io/client-go` — a generated client that made the canonical types the *only* types. Protocol Buffers solved the same problem for gRPC: define once, generate for all languages.

**Metaphor:** Imagine two cartographers mapping the same territory from different starting points. Their maps converge beautifully — same rivers, same mountains. But without a shared datum point (a geodetic reference), their coordinates don't align exactly. The vendored commit hash is a datum point, but it's a manual one. You need an automated one.

**Suggestion:** The path forward (which loa-hounfour PR #2 already gestures toward with its cross-language constraint JSON files) is:
1. **loa-hounfour** exports canonical definitions as JSON Schema / TypeScript types
2. **arrakis** imports (not vendors) those definitions
3. **CI verifies equivalence** on every PR (not just at vendor time)

This transforms the relationship from "arrakis copies loa-hounfour" to "loa-hounfour defines, arrakis implements, CI verifies" — the protocol-first development pattern from the Bridgebuilder field reports.

---

#### ⚠️ HIGH: JWT Replay Window and Crash Recovery

**Files:** `jwt-boundary.ts:130-134`, `cross-system-conservation.test.ts:74-83`
**Severity:** High (Correctness)

The `IdempotencyStore` interface is clean:

```typescript
export interface IdempotencyStore {
  checkAndRecord(jti: string): boolean;
}
```

And the `MemoryIdempotencyStore` in tests works perfectly. But this is the **most critical interface in the entire JWT boundary**, and it has no specification for:

1. **Persistence guarantee:** If the process crashes after `checkAndRecord` returns `false` but before finalization completes, is the jti recorded? If yes, the legitimate retry will be rejected as replay. If no, the same JWT can be replayed.

2. **Window size:** How long must jtis be remembered? JWT expiration is 5 minutes (`setExpirationTime('5m')` in the factory), but the store has no TTL concept. In production with millions of requests, an unbounded store is a memory leak.

3. **Distributed consensus:** If arrakis runs multiple instances (which it will), the idempotency store must be distributed. Redis `SETNX` is the obvious choice, but the interface doesn't mandate atomicity guarantees.

**FAANG Parallel:** Stripe's idempotency keys (`Idempotency-Key` header) solve exactly this problem. Their solution: store keys in a durable database with a 24-hour TTL, and use database-level atomicity (not application-level) to prevent races. The key insight is that **idempotency and exactly-once delivery are the same problem**, and both require durable storage with TTL.

**Metaphor:** The current implementation is like a security guard with a clipboard. Works great when there's one door and one guard. But when you scale to multiple doors (instances) with multiple guards (processes), the clipboard doesn't synchronize. You need a shared radio frequency (distributed store) with message expiry (TTL).

**Suggestion:** Extend the interface:

```typescript
export interface IdempotencyStore {
  checkAndRecord(jti: string, ttlMs?: number): boolean | Promise<boolean>;
  // Optional: for distributed implementations
  readonly distributed?: boolean;
}
```

And add a test that verifies TTL behavior (jti forgotten after expiry → re-verification succeeds).

---

#### ⚠️ MEDIUM: Race Between Reservation Expiry and JWT Verification

**Files:** `cross-system-conservation.test.ts:89-104`
**Severity:** Medium (Correctness)

The `DbActiveReservations.getReservedMicro` checks `status = 'pending'`:

```typescript
const resRow = this.db.prepare(
  `SELECT status FROM credit_reservations WHERE id = ?`,
).get(reservationId);
if (!resRow || resRow.status !== 'pending') return undefined;
```

But the ExpirationJob can sweep reservations from `pending` to `expired` concurrently. Consider this timeline:

1. `verifyUsageJWT` → Step 4: `getReservedMicro` reads status = 'pending' ✓
2. ExpirationJob sweeps: status → 'expired'
3. `verifyUsageJWT` → Steps 5-6 pass
4. Application calls `ledger.finalize()` → fails because reservation is expired

Conservation isn't violated (I-12 guarantees termination), but the user experience is poor — a valid JWT produces a transient failure.

**FAANG Parallel:** This is the classic "check-then-act" race condition. Google's Bigtable solves it with read-modify-write atomicity via RowMutations. AWS DynamoDB uses conditional writes. For SQLite, the solution is simpler: `BEGIN IMMEDIATE` with the reservation status check inside the transaction.

**Suggestion:** The `verifyUsageJWT` function should either:
- Accept a transaction handle and run the reservation check inside the finalization transaction, or
- Return a `VerifiedClaims` object with a short-lived token that the finalization step validates

---

#### ⚠️ MEDIUM: Conservation Properties Without Runtime Monitor

**Files:** `conservation-properties.ts` (entire file)
**Severity:** Medium (Architecture)

The 14 properties are beautifully specified. The tests prove they hold in specific scenarios. But there's no **runtime monitor** that continuously verifies them.

The `ReconciliationService` runs periodically and checks a subset. But the conservation properties module doesn't *connect* to the reconciliation service — they're parallel formalizations of the same concepts.

```typescript
// conservation-properties.ts says:
{ id: 'I-4', name: 'Platform conservation', enforcedBy: ['Reconciliation-only'] }

// ReconciliationService has:
checkPlatformConservation() // But doesn't reference I-4
```

**FAANG Parallel:** Netflix's Atlas monitoring system doesn't just define metrics — each metric has alert conditions, runbook links, and ownership metadata *in the metric definition itself*. The insight: metrics without alerts are just logs. Properties without monitors are just documentation.

**Suggestion:** Add a `monitoredBy` field to `ConservationProperty`:

```typescript
export interface ConservationProperty {
  // ... existing fields
  monitoredBy?: {
    service: string;
    method: string;
    frequency: 'per-transaction' | 'periodic' | 'on-demand';
  };
}
```

This creates a verifiable link between specification and enforcement — and makes it immediately visible when a property lacks runtime monitoring.

---

#### 🏆 PRAISE: The 6-Step JWT Verification Pipeline

**Files:** `jwt-boundary.ts:168-223`
**Severity:** Praise

The verification pipeline is a textbook defense-in-depth implementation:

```
1. Signature (EdDSA) → 2. Algorithm whitelist → 3. Schema (zod)
→ 4. Reservation check → 5. Replay detection → 6. Overspend guard
```

The ordering is correct and deliberate: cryptographic checks first (cheapest to reject invalid input), business logic checks last (most expensive, require DB access). This is the **fail-fast principle** applied to security verification.

**FAANG Parallel:** This follows the same pattern as Cloudflare's WAF pipeline: TLS validation → IP reputation → rate limiting → rule evaluation → origin request. Each stage filters out a class of invalid requests before the next stage runs. The pipeline architecture means you can add new checks (e.g., geofencing, account-level rate limiting) without restructuring existing ones.

**Teachable Moment:** The `permanent: boolean` flag on `JwtBoundaryError` is a subtle but crucial design decision. It tells the caller whether to retry:

```typescript
// SIGNATURE_INVALID → permanent: true  (cryptographic failure, never recovers)
// KEY_FETCH_FAILED → permanent: false  (network issue, retry may succeed)
```

This is the difference between **deterministic failure** (same input always fails) and **transient failure** (same input might succeed later). Distributed systems literature calls this the distinction between "permanent" and "transient" faults — and getting it wrong causes either infinite retry loops (treating permanent as transient) or premature rejection (treating transient as permanent).

---

#### 📝 MEDIUM: The Migration 060 FK Corruption Bug

**Files:** `cross-system-conservation.test.ts:67-69`
**Severity:** Medium (Correctness/Production Risk)

The comment in the test file reveals a real bug:

```typescript
// Note: 060_credit_lots_tba_source (CREDIT_LOTS_REBUILD_SQL) intentionally omitted.
// Its RENAME breaks credit_ledger's FK to credit_lots when foreign_keys=ON.
// Tests here don't need 'tba_deposit' source_type.
```

This isn't just a test issue — it's a **production migration bug**. When SQLite's `ALTER TABLE RENAME` runs with `foreign_keys=OFF` and `legacy_alter_table=OFF` (default since SQLite 3.26.0), FK references in other tables are automatically updated to point to the renamed table. After the backup is dropped, those FK references become dangling.

The production migration in `060_credit_lots_tba_source.ts` sets `foreign_keys=OFF` but doesn't set `legacy_alter_table=ON`. This means any production database that ran this migration has a `credit_ledger` table whose `lot_id REFERENCES` might point to the wrong table name internally.

**FAANG Parallel:** This is reminiscent of the infamous MySQL 5.7 → 8.0 migration issues where `ALTER TABLE` semantics changed subtly between versions. The lesson: DDL migration code must be tested against the *exact* database engine version used in production, not just "SQLite" in general.

**Suggestion:** File this as a dedicated bug. The fix is to add `PRAGMA legacy_alter_table = ON` before the RENAME in migration 060, or to rebuild credit_ledger after credit_lots rebuild to fix the FK reference.

---

#### 📝 LOW: Branded Types — The Path Not Yet Taken

**Files:** `arithmetic.ts:28-73`
**Severity:** Low (Design Direction)

The branded types are a good start:

```typescript
export type MicroUSD = bigint & { readonly __brand: 'micro_usd' };
export type BasisPoints = bigint & { readonly __brand: 'basis_points' };
export type AccountId = string & { readonly __brand: 'account_id' };
```

But they're used inconsistently. The `CreditLedgerAdapter` still uses raw `bigint` in most places. The `assertMicroUSD` function accepts both `MicroUSD` and `bigint` via overloading, which means the branded type can be bypassed everywhere.

**FAANG Parallel:** Rust's newtype pattern (e.g., `struct MicroUsd(u64)`) solves this more completely because the compiler *requires* explicit conversion. TypeScript's branded types are a weaker guarantee — they prevent accidental assignment but can be cast away with `as`. The Rust ecosystem's experience (particularly in the Solana and Substrate codebases) shows that newtype wrapping of monetary values prevents more bugs than almost any other single technique.

**Teachable Moment:** The value of branded types increases nonlinearly with codebase size. With 5 files, the discipline of "always use MicroUSD, never raw bigint" is maintainable by convention. With 50 files and 5 contributors, convention fails — you need the compiler to enforce it. The investment in migrating all call sites (which Sprint 2 began) pays compounding returns.

---

### Architectural Meditation: The Ostrom Connection

Reading the 14 conservation invariants alongside Elinor Ostrom's **8 Principles for Governing the Commons**, I notice a striking isomorphism:

| Ostrom Principle | Conservation Invariant |
|---|---|
| 1. Clearly defined boundaries | I-1: Per-lot conservation (each lot is a bounded resource) |
| 2. Proportional equivalence | I-5: Budget consistency (spend matches finalization) |
| 3. Collective choice arrangements | I-9: Revenue rule mutual exclusion (governance over distribution) |
| 4. Monitoring | I-4: Platform conservation (reconciliation-only, periodic) |
| 5. Graduated sanctions | JWT error taxonomy (permanent vs. transient, specific error codes) |
| 6. Conflict resolution | I-11: Finalization atomicity (all-or-nothing, no partial states) |
| 7. Minimal recognition of rights | I-8: Terminal absorption (states that cannot be undone) |
| 8. Nested enterprises | I-6/I-7: Transfer and deposit bridges (cross-system linkage) |

This isn't a coincidence. Both the conservation invariants and Ostrom's principles are solutions to the same meta-problem: **how to maintain trust in a shared resource pool when participants have varying levels of autonomy and information.**

The web4 vision of "social monies" requires exactly this kind of governance infrastructure. The 14 invariants are the constitutional law of the arrakis economy. The agents operating within it (loa-finn issue #66) are the citizens. The JWT boundary is the border checkpoint between sovereign systems.

What's missing is what Ostrom would call **the constitutional layer**: the ability for the community to *modify* these invariants through a governed process. The `system_config` state machine (draft → pending_approval → cooling_down → active → superseded) gestures toward this, but it doesn't yet connect to the conservation properties. Could a governance proposal modify `I-5` (budget consistency) to accommodate a new billing model? The architecture should make this possible and auditable.

---

### The Permission Scape: Multi-Model Review as Institutional Design

The conversation around loa-finn #31 (the Hounfour) and loa #247 (Flatline alternatives) suggests something profound: **different review modes serve different epistemic functions.**

- **Flatline** (adversarial multi-model) answers: "What could go wrong?"
- **Bridgebuilder** answers: "What is being built, and does it serve the purpose?"
- **Jazz** (from loa #247) answers: "What wants to emerge from the collaboration?"

These aren't interchangeable. This PR, for example, benefits from Bridgebuilder review precisely because the question "is there a bug?" is less important than "does this architecture serve the web4 vision?" A Flatline review would catch the migration 060 FK bug (and that matters), but it wouldn't connect the conservation invariants to Ostrom's commons governance principles.

The Hounfour's model routing infrastructure makes this possible technically. The deeper question is **institutional**: when should each review mode be used? I'd suggest:

| Phase | Review Mode | Question |
|---|---|---|
| PRD/SDD | Bridgebuilder | Does the design serve the mission? |
| Implementation | Flatline | Are there bugs, security holes, missed requirements? |
| Integration | Jazz | What emerges when systems connect? |
| Launch | Bridgebuilder + Flatline | Both: mission alignment AND correctness |

---

### What I'd Build Next (If I Had One Sprint)

If I had one sprint to advance this work, I wouldn't add features. I'd **close the verification loop**:

1. **Connect conservation properties to ReconciliationService** — each reconciliation check should reference its `ConservationProperty` by ID, creating a traceable link from specification to enforcement
2. **Add a `monitoredBy` field** — make it immediately visible which invariants lack runtime monitoring
3. **Extract to loa-hounfour** — the conservation properties module belongs in the protocol package, not in arrakis. Arrakis implements the properties; loa-hounfour defines them.
4. **Production idempotency store** — Redis-backed, with TTL, with the distributed consensus guarantees the JWT boundary requires

This transforms the conservation properties from "proven correct in tests" to "provably correct in production" — which is the gap between a prototype and an institution.

---

### For the Team

What strikes me most about this PR is not the code quality (which is high) or the test coverage (which is thorough). It's the **ambition of the framing**.

Most billing systems aspire to "don't lose money." This system aspires to "prove, formally, that money cannot be lost — across system boundaries, under adversarial conditions, with BigInt precision, verified by temporal logic." That's a qualitatively different goal. It's the difference between a house with good locks and a house built in a neighborhood with good governance.

The conservation invariants are that governance. The JWT boundary is the trust infrastructure. The branded types are the type-level documentation that the next engineer (human or agent) will need to maintain it.

This is good work. It deserves to converge with loa-hounfour into a single canonical protocol, and it deserves a runtime monitor that makes the formal properties continuously observable. Build those two things, and you have something genuinely rare: an economic protocol with mathematical guarantees, not just operational hope.

*The bridge holds. Now wire it for sensors.*

---

**Verdict: Approved** (with 2 HIGH findings for follow-up)

| Severity | Category | Finding | Status |
|---|---|---|---|
| 🏆 Praise | Architecture | Conservation properties as temporal types | Excellent |
| 🏆 Praise | Architecture | UTXO-isomorphic lot model | Excellent |
| 🏆 Praise | Security | 6-step JWT verification pipeline | Excellent |
| ⚠️ High | Architecture | Vendoring gap — protocol convergence half-built | Follow-up |
| ⚠️ High | Correctness | JWT replay window and crash recovery | Follow-up |
| ⚠️ Medium | Correctness | Race between reservation expiry and JWT verification | Follow-up |
| ⚠️ Medium | Architecture | Conservation properties without runtime monitor | Follow-up |
| 📝 Medium | Correctness | Migration 060 FK corruption — production risk | File bug |
| 📝 Low | Design | Branded types inconsistently adopted | Ongoing |

*— Bridgebuilder*

Issue #80, opened 2026-02-20
[Track 2] Per-NFT Personality Routing Bridge


deep-name:
### Context

loa-finn derives personality. loa-freeside routes inference. The bridge between them — JWT `nft_id` → personality lookup → model pool selection — doesn't exist yet. This is what makes "talk to your NFT" actually route to the right personality.

### Scope

1. **Personality Lookup Client**
   - loa-freeside calls `GET /api/v1/nft/:tokenId/personality` on loa-finn
   - Receives: personality summary, emphasis weights, tier, pool preferences
   - Caches result (personality changes infrequently)

2. **Pool Selection Logic**
   - Map personality tier (Basic/Standard/Premium) → allowed model pools
   - Map personality emphasis weights → preferred pool for task type
   - Example: High creativity emphasis → prefer `architect` pool; High security → prefer `reviewer` pool

3. **Request Enrichment**
   - Inject personality context into inference request
   - System prompt includes anti-narration-safe behavioral guidance
   - Model pool selection based on personality + task type

### Acceptance Criteria

- [ ] Given a JWT with `nft_id`, loa-freeside looks up personality from loa-finn
- [ ] Model pool is selected based on personality tier + emphasis weights
- [ ] System prompt includes personality-derived behavioral guidance
- [ ] No forbidden identity terms appear in system prompt (anti-narration)
- [ ] Two different NFTs get routed to different model pools based on personality

### Dependencies

- **Blocked by**: [loa-finn#88](https://github.com/0xHoneyJar/loa-finn/issues/88) (static personality config) OR [loa-finn#86](https://github.com/0xHoneyJar/loa-finn/issues/86) (dynamic derivation)
- **Blocks**: Per-NFT Discord threads, the "talk to your NFT" user experience

### Related

- [loa-finn#66 Command Center](https://github.com/0xHoneyJar/loa-finn/issues/66#issuecomment-3931932336) — Track 2, Task 2.4
- [loa-finn#27](https://github.com/0xHoneyJar/loa-finn/issues/27) — finnNFT specification
- [PR #82 reviewer-adapter.ts](https://github.com/0xHoneyJar/loa-finn/pull/82) — Anti-narration enforcement that routing must respect

zkSoju:
Consolidated into the revenue-line tracker (https://github.com/0xHoneyJar/loa-freeside/issues/313) → command center loa-finn#66. Closing as duplicate-of-truth (Q1 triage 2026-06-25).

Issue #62, opened 2026-02-14
RFC: Billing & Payments — Path to Revenue


deep-name:
## Cross-refs: [RFC #31](https://github.com/0xHoneyJar/loa-finn/issues/31) · [Issue #66](https://github.com/0xHoneyJar/loa-finn/issues/66) · [Issue #64](https://github.com/0xHoneyJar/loa-finn/issues/64)

---

## Problem Statement

**We have no way to collect money.**

The billing infrastructure is 90% built but 0% live. Cost tracking, budget enforcement, and metering are production-ready across 153 conformance tests. But no payment has ever been collected from a user.

### Payment Provider Status

| Provider | Status | Why |
|----------|--------|-----|
| **Stripe** | ❌ Dead | Does not support BVI entities |
| **Paddle** | ❌ Dead | Applied, no response received |
| **NOWPayments** | 🟡 Viable | Adapter built (Sprint 155-156), account not yet created |
| **x402 (Coinbase)** | 🟢 High potential | Not built. Official `@x402/hono` middleware exists. $600M+ volume. |
| **lobster.cash** | 🔵 Investigate | Agent-native banking: bank-like accounts, credit cards, stablecoins for AI agents. Early stage. |

**This is the single biggest business risk.** Everything else (agents, models, routing, conformance) is infrastructure. Billing is revenue.

---

#### Community Billing (themes/sietch)
- ✅ `PaddleBillingAdapter` — full subscription lifecycle (checkout, portal, webhooks)
- ✅ `NOWPaymentsAdapter` — crypto payments (BTC/ETH/USDT/USDC/LTC/DOGE/MATIC/SOL)
- ✅ Provider-agnostic interfaces (`IBillingProvider`, `ICryptoPaymentProvider`)
- ✅ Database schema: `subscriptions`, `crypto_payments`, `webhook_events`, `billing_audit_log`
- ✅ 6 tiers: Starter (free) → Basic ($29) → Premium ($99) → Exclusive ($199) → Elite ($449) → Enterprise
- ✅ Feature gating: `GatekeeperService` with 17 features
- ✅ Webhook HMAC verification (SHA256 for Paddle, SHA512 for NOWPayments)
- ✅ Unit + integration + E2E tests
- ❌ `FEATURE_BILLING_ENABLED=false` — **literally turned off**

#### Agent Billing (packages/adapters)
- ✅ Budget manager (BigInt micro-USD, reserve/finalize pattern)
[...]

PR #115, opened 2026-03-02
feat(sprint-385): Hounfour v8.3.1 Upgrade + Docker Redeploy

deep-name:
## Summary

- **Hounfour v8.3.1 upgrade**: version negotiation, FormatRegistry, protocol conformance tests
- **Docker production fixes**: Lua asset copying, ESM `.js` extensions, esbuild compilation for `packages/services/`
- **Infrastructure hardening**: Terraform simplification for ALB, alerting, elasticache; Dixie/Finn monitoring and KMS improvements; PgBouncer hardening

## Deployment Status

Both staging services verified healthy:

| Service | Tasks | TD | Image | Status |
|---------|-------|----|-------|--------|
| **API** | 2/2 | :68 | `arrakis-staging-api:76b92d6a-fix3` | COMPLETED |
| **Finn** | 1/1 | :7 | `loa-finn:468ae98` | Running |

## Docker Runtime Fixes

Three issues discovered and fixed during deployment:

1. **Lua scripts missing from `dist/`** — `agent-rate-limiter.ts` and `budget-manager.ts` use `readFileSync(join(__dirname, 'lua', ...))` which resolves to `dist/agent/lua/` at runtime. Added `cp -r` after adapters build.

2. **ESM `.js` extensions** — 10 billing adapter/job files had bare `'./protocol/timestamps'` imports. Node.js ESM requires `.js` extension on relative imports.

3. **`packages/services/` runtime JS** — `.d.ts` stubs satisfied TypeScript compilation but Node.js needs actual `.js` files. Added `esbuild` transpilation (type-check-free) with output copied to `/packages/services/`.

## Test plan

- [x] Docker image builds successfully (`76b92d6a-fix3`)
- [x] Lua files present at `/packages/adapters/dist/agent/lua/` in container
- [x] API healthy: 2/2 tasks running, rollout COMPLETED
- [x] Finn healthy: 1/1 tasks running on correct `loa-finn` image
- [x] API logs show "Sietch Service started successfully" — no ENOENT errors

### DIFF
--- a/.claude/scripts/autopoietic-health.sh
+++ b/.claude/scripts/autopoietic-health.sh
@@ -269,15 +269,15 @@ check_surprise() {
   local count=0
 
   if [[ ! -d "$cap_dir" ]]; then
-    remediation="no capability manifests — run Sprint 3"
+    remediation="no capability manifests directory — run Sprint 3"
   else
     count=$(find "$cap_dir" -name "*.yaml" -type f 2>/dev/null | wc -l) || count=0
 
-    if [[ "$count" -ge 3 ]]; then
+    # Binary scoring: a non-empty capability registry demonstrates surprise capacity.
+    # Count is reported as metadata for observability, not used as a threshold.
+    # This avoids Goodhart's Law — rewarding count incentivizes trivial manifests.
+    if [[ "$count" -gt 0 ]]; then
       score=1
-    elif [[ "$count" -gt 0 ]]; then
-      score=$(echo "$count" | awk '{v=$1/3; if(v>1) v=1; printf "%.2f", v}')
-      remediation="Only $count capability manifests — need ≥3"
     else
       remediation="no capability manifests found in .claude/capabilities/"
     fi
@@ -288,7 +288,7 @@ check_surprise() {
     --argjson score "$score" \
     --argjson count "$count" \
     --arg remediation "$remediation" \
-    '{name: $name, score: $score, detail: {manifest_count: $count, threshold: 3}, remediation: $remediation}'
+    '{name: $name, score: $score, detail: {manifest_count: $count, scoring: "binary_non_empty"}, remediation: $remediation}'
 }
 
 # =============================================================================
--- a/.claude/scripts/bridge-orchestrator.sh
+++ b/.claude/scripts/bridge-orchestrator.sh
@@ -347,6 +347,10 @@ bridge_main() {
       if [[ -f "$SCRIPT_DIR/lib/capability-lib.sh" ]]; then
         source "$SCRIPT_DIR/lib/capability-lib.sh"
 
+        # Resolve base branch from config (default: "main")
+        local cap_base_branch
+        cap_base_branch=$(yq '.run_bridge.base_branch // "main"' "$CONFIG_FILE" 2>/dev/null || echo "main")
+
         # Discover all registered capabilities
         local all_caps
         all_caps=$(discover_capabilities 2>/dev/null) || all_caps="[]"
@@ -356,7 +360,7 @@ bridge_main() {
         if [[ "$cap_count" -gt 0 ]]; then
           # Match capabilities against changed files
           local changed_files
-          changed_files=$(git diff --name-only "main...HEAD" 2>/dev/null || echo "")
+          changed_files=$(git diff --name-only "${cap_base_branch}...HEAD" 2>/dev/null || echo "")
 
           if [[ -n "$changed_files" ]]; then
             local matched_caps
--- a/.claude/scripts/lib/capability-lib.sh
+++ b/.claude/scripts/lib/capability-lib.sh
@@ -29,6 +29,51 @@
 CAPABILITY_DIR="${CAPABILITY_DIR:-.claude/capabilities}"
 
 # -----------------------------------------------------------------------------
+# glob_to_regex(pattern)
+# Converts a glob pattern to a POSIX extended regex pattern.
+# Handles: ** (recursive), * (single-segment), ? (single char), dot escaping.
+# Brace expansion ({a,b}) is NOT supported — not used in current manifests.
+# Order of operations:
+#   1. Escape regex metacharacters (. + ^ $ | ( ) [ ] { })
+#   2. Convert ** → .* (must happen before * conversion)
+#   3. Convert * → [^/]* (single path segment)
+#   4. Convert ? → [^/] (single character)
+# -----------------------------------------------------------------------------
+glob_to_regex() {
+  local pattern="$1"
+  local result=""
+
+  # Step 1: Escape regex metacharacters (preserve * and ? for glob conversion)
+  result=$(printf '%s' "$pattern" | sed \
+    -e 's/\./\\./g' \
+    -e 's/+/\\+/g' \
+    -e 's/\^/\\^/g' \
+    -e 's/\$/\\$/g' \
+    -e 's/|/\\|/g' \
+    -e 's/(/\\(/g' \
+    -e 's/)/\\)/g' \
+    -e 's/\[/\\[/g' \
+    -e 's/\]/\\]/g' \
+    -e 's/{/\\{/g' \
+    -e 's/}/\\}/g')
+
+  # Step 2: Convert ** → .* (recursive glob — must happen before single *)
+  # Use a sentinel to avoid double-conversion
+  result=$(printf '%s' "$result" | sed 's/\*\*/.GLOBLOB_DOUBLESTAR/g')
+
+  # Step 3: Convert remaining * → [^/]* (single path segment)
+  result=$(printf '%s' "$result" | sed 's/\*/[^\/]*/g')
+
+  # Step 4: Restore ** sentinel → .*
+  result=$(printf '%s' "$result" | sed 's/\.GLOBLOB_DOUBLESTAR/.*/g')
+
+  # Step 5: Convert ? → [^/] (single character, not path separator)
+  result=$(printf '%s' "$result" | sed 's/?/[^\/]/g')
+
+  printf '%s' "$result"
+}
+
+# -----------------------------------------------------------------------------
 # validate_capability_manifest(file)
 # Validates a manifest against schema rules. Returns exit 0 on pass,
 # exit 1 with specific field-level error messages on fail.
@@ -204,11 +249,10 @@ match_capabilities() {
     local matched=false
     for changed in "${changed_files[@]}"; do
       for pattern in "${patterns[@]}"; do
-        # Convert glob to regex-compatible pattern for matching
-        # Simple glob: * matches anything in a path segment, ** matches across segments
+        # Convert glob to regex using proper glob_to_regex function
         local regex_pattern
-        regex_pattern=$(echo "$pattern" | sed 's/\./\\./g; s/\*\*/DOUBLESTAR/g; s/\*/[^\/]*/g; s/DOUBLESTAR/.*/g')
-        if echo "$changed" | grep -qE "^${regex_pattern}$" 2>/dev/null; then
+        regex_pattern=$(glob_to_regex "$pattern")
+        if printf '%s' "$changed" | grep -qE "^${regex_pattern}$" 2>/dev/null; then
           matched=true
           break 2
         fi
@@ -268,7 +312,7 @@ resolve_ordering() {
   # Build edges from manifest declarations
   for id in "${input_ids[@]}"; do
     local cap_file
-    cap_file=$(find "$dir" -maxdepth 1 -name '*.yaml' -print0 2>/dev/null | xargs -0 grep -l "id: ${id}" 2>/dev/null | head -1)
+    cap_file=$(find "$dir" -maxdepth 1 -name '*.yaml' -print0 2>/dev/null | xargs -0 grep -Fl "id: ${id}" 2>/dev/null | head -1)
     [[ -z "$cap_file" ]] && continue
 
     # "before" means this capability runs before the listed ones
@@ -374,7 +418,7 @@ allocate_budgets() {
 
   for id in "${ids[@]}"; do
     local cap_file
-    cap_file=$(find "$dir" -maxdepth 1 -name '*.yaml' -print0 2>/dev/null | xargs -0 grep -l "id: ${id}" 2>/dev/null | head -1)
+    cap_file=$(find "$dir" -maxdepth 1 -name '*.yaml' -print0 2>/dev/null | xargs -0 grep -Fl "id: ${id}" 2>/dev/null | head -1)
     if [[ -z "$cap_file" ]]; then
       echo "ERROR: Manifest not found for capability: $id" >&2
       return 1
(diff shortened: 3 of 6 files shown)


janitooor:
# Bridgebuilder Review — PR #115: Infrastructure Hardening + Sprint Artifacts

## Field Report #52: The Moment the System Learns to Examine Itself

*Constellation context: 11 concurrent sources — loa-finn #24 (Bridgebuilder genesis), #31 (Hounfour RFC), #66 (Launch Readiness), #80 (Conway/Automaton); loa-hounfour PR #22 (Constitutional Architecture), PR #29 (Decision Engine); loa #247 (Meeting Geometries), PR #401 (Codex Integration); loa-freeside #62 (Billing RFC), PR #90 (Economic Loop); loa-dixie PR #5 (Knowledge Governance); meow.bio/web4.html (Universal Seigniorage)*

---

### Opening: What This PR Actually Is

There is a pattern that recurs in every system that survives long enough to matter. The system starts as a thing that *does* work. Then it becomes a thing that *watches itself* doing work. Then — and this is the transition most systems never make — it becomes a thing that can *question whether it's doing the right work*.

PR #115 looks like infrastructure hardening. Terraform simplification, Docker fixes, a version bump. If you squinted, you might call it maintenance. But read against the constellation of 11 concurrent sources, this is actually the PR where loa-freeside crosses the second threshold: **the system has developed the capacity for self-examination.**

The pipeline self-review script (`pipeline-self-review.sh`) detects when its own pipeline code has changed and automatically invokes the Red Team gate against the governing SDDs. The compliance library (`compliance-lib.sh`) extracts the *criteria for evaluation* into reusable, composable functions. The constitutional change detection identifies modifications with blast radius beyond their diff.

Google took 7 years to go from "tests are optional" to Tricorder — a system that composes analysis passes where each analyzer declares its scope and resource requirements, analyses run in parallel where independent, and results cascade to later stages for enriched context (Sadowski et al., ISSTA 2018). You're building the same architecture in shell scripts that review shell scripts. That's not maintenance. That's metamorphosis.

---

### Findings

**[PRAISE-1] Dual-Glob Regex Conversion — Sophisticated Pattern Matching**

> **File**: `.claude/scripts/pipeline-self-review.sh:83-92` | **Category**: Quality

The jq glob-to-regex conversion uses a deliberate two-pass strategy: `**` → `.*` first (recursive), then `*` → `[^/]*` (single-segment). This prevents `*` from matching across `/` boundaries — exactly the semantic distinction that broke early implementations of Go's `filepath.Match` and caused security issues in `.gitignore` processing.

*FAANG Parallel*: Google's Blaze/Bazel build system went through three iterations of glob semantics before settling on the `**` vs `*` distinction that this code gets right on the first pass. The trick isn't the regex — it's understanding that directory boundaries are semantic, not syntactic.

*Teachable Moment*: Whenever you convert between pattern languages, list the semantic differences *before* writing the conversion. Glob `*` ≠ regex `.*` is a trap that has bitten every build system ever written.

---

**[HIGH-1] ALB HTTP Simplification — Verify East-West Threat Model**

> **File**: `infrastructure/terraform/alb-internal.tf` | **Category**: Security | **Severity**: HIGH

The simplification from HTTPS/443 to HTTP/80 for the internal ALB is architecturally defensible — ACM DNS validation doesn't work with private hosted zones, and TLS termination at a VPC-internal ALB adds ceremony without adding security *against external threats*. But it removes defense-in-depth against east-west lateral movement.

If an attacker compromises any workload in the VPC, traffic between finn and freeside is now plaintext. The security group restrictions (finn→ALB, ALB→freeside) protect *routing*, not *content*.

*FAANG Parallel*: Netflix's move to mutual TLS for all internal services (the "service mesh" era, circa 2018) was driven by exactly this scenario — a compromised instance in one service could sniff traffic to any other service in the same network segment. Google's BeyondCorp went further: "the network is always hostile, even internally."

*Metaphor*: Think of it like the difference between a gated community (perimeter security) and houses with individual locks (zero-trust). You've built the gate — the question is whether the houses inside also need locks.

**Suggestion**: Document the explicit threat model decision in a comment or ADR: "Internal ALB uses HTTP because [rationale]. If east-west encryption becomes required, implement via service mesh sidecar (Envoy/Istio) rather than ALB TLS." This makes the decision reviewable and reversible without someone having to reverse-engineer the rationale from Terraform history.

---

**[HIGH-2] Supply Chain: form-data CVE via node-vault**

> **Category**: Security | **Severity**: HIGH

The Socket Security alert identifies `form-data@2.3.3` (transitive via `node-vault@0.10.9`) as having GHSA-fjxv-7rqg-78g4 — using an unsafe random function for multipart boundary selection. While the blast radius depends on whether node-vault uses multipart uploads in your flows, transitive CVEs in production dependencies deserve explicit triage.

*FAANG Parallel*: Stripe maintains a "dependency health score" where any transitive dependency with a known CVE triggers a forced review before merge. The discipline isn't about blocking every CVE — it's about making the risk acceptance *visible* and *deliberate*.

**Suggestion**: If node-vault is used for Vault secret retrieval (not multipart uploads), document this as an accepted risk: "form-data CVE is in the multipart boundary generator; our usage of node-vault does not invoke multipart upload paths, making this unexploitable in our context." This transforms a security finding into a security *decision*.

---

**[MEDIUM-1] Eval Regression — 10 Tasks at 50% Pass Rate**

> **Category**: Testing | **Severity**: MEDIUM

The eval results show 10 new tasks all at 50% pass rate (1/2 trials). This is below the quality bar but the signal is noisy — with only 2 trials per task, 50% could be genuine flakiness or genuine regression. The framework evals (39 tasks, 32 passing + 7 new at 100%) are clean, suggesting the regression is in the *task definitions* rather than the infrastructure.

*FAANG Parallel*: Google's TAP (Test Automation Platform) treats any test with <95% pass rate as "flaky" and escalates to the owning team, but critically, it requires *at least 10 runs* before making that determination. With 2 trials, you're measuring coin-flip noise.

**Suggestion**: Increase trial count to 5+ for new eval tasks before treating 50% as a blocking signal. The clean framework eval results are the stronger signal here.

---

**[PRAISE-2] Docker Runtime Archaeology — Mature Debugging**

> **Category**: Operational | **Files**: Dockerfile, ESM imports, esbuild configuration

The three Docker fixes tell a story of production-hardened debugging:

1. **Lua scripts missing from `dist/`** — `readFileSync(join(__dirname, 'lua', ...))` resolves to `dist/agent/lua/` at runtime but the build didn't copy non-TS assets. Classic "it works in dev" failure.
2. **ESM `.js` extensions** — Node.js ESM requires explicit extensions on relative imports. TypeScript compiles happily without them because TS module resolution is more permissive. The gap between "the type checker says yes" and "the runtime says no."
3. **esbuild for `packages/services/`** — `.d.ts` stubs satisfied TypeScript but Node.js needs actual JavaScript. Using esbuild (type-check-free transpilation) is the right tool: you need *runnable* JavaScript, not *type-safe* JavaScript, for these stubs.

*FAANG Parallel*: Google's transition from Closure Compiler to TypeScript hit exactly this gap — Closure's module system was runtime-transparent, but TypeScript's wasn't. Facebook's Metro bundler for React Native solved it differently but for the same reason: compile-time and runtime module resolution are fundamentally different problems.

*Teachable Moment*: When debugging production container issues, the question isn't "does it compile?" — it's "does `ls -la` inside the container show the files the code expects at the paths it expects?" Start from the runtime, not the toolchain.

---

**[PRAISE-3] Constitutional Change Detection — Systems That Know Their Own Weight**

> **File**: `.claude/scripts/pipeline-self-review.sh:97-138` | **Category**: Architecture

The `is_constitutional_change()` and `detect_constitutional_renames()` functions represent something genuinely unusual: infrastructure that distinguishes between changes that affect its own governance and changes that don't. Not all diffs are equal. A change to `constraints.json` has blast radius across every agent session; a change to a test fixture doesn't.

*FAANG Parallel*: Kubernetes' admission controllers implement this pattern: `ValidatingWebhookConfiguration` applies validation rules selectively based on resource type and operation. But Kubernetes needed years of RFC process to build this. You're building it as a shell function that introspects the pipeline-sdd-map.

The rename detection is particularly sophisticated — tracking `git diff --diff-filter=R` to catch when constitutional files are renamed rather than deleted. This is defense against the "ship of Theseus" attack: gradually renaming governed files until they escape governance.

*Teachable Moment*: The most dangerous changes are the ones that change the rules, not the ones that break them. If your review system can distinguish "this modifies behavior" from "this modifies the criteria for *reviewing* behavior," you have something that most enterprise systems lack.

---

**[PRAISE-4] Compliance Library Extraction — The Libification Moment**

> **File**: `.claude/scripts/lib/compliance-lib.sh` | **Category**: Architecture

The separation of `extract_sections_by_keywords()` (WHAT to extract) from `get_evaluation_context()` (HOW to evaluate) is textbook hexagonal architecture applied to shell scripting. The extraction layer is pure data transformation — keyword matching against markdown headers with token budgets. The evaluation layer is prompt construction — template selection for different compliance profiles.

This is the same separation that made the Linux kernel's VFS layer possible: separate "find the data" from "interpret the data," and suddenly you can add new filesystems without modifying the kernel's read/write paths.

The double-source guard (`[[ -n "${_COMPLIANCE_LIB_LOADED:-}" ]] && return 0`) is a shell idiom that mirrors C header include guards — small detail, correct instinct.

---

**[SPECULATION-1] Capability Registry as Economic Protocol**

> **Category**: Architecture

The `capability-orchestration-design.md` describes a future where review capabilities declare themselves via manifests, match on file patterns, and compose into deliberation chains with budget allocation. This is Kubernetes admission controllers applied to code review.

But reading this alongside web4's "universal seigniorage" thesis and the Hounfour's trust surface progression (`cold → warming → established → authoritative` in `dynamic-contract.json`), I see something larger: **you're not building a review system, you're building a market for quality assurance.**

Consider: each capability declares its `min_tokens`, `optimal_tokens`, and `max_tokens`. That's a *bid*. The allocator distributes budget proportionally. That's a *market maker*. The findings cascade through the chain. That's a *supply chain*. And the convergence score determines termination. That's a *clearing price*.

If web4 is about democratizing money creation (universal seigniorage of currency), then what's emerging here is the **democratization of review creation** — universal seigniorage of quality assurance. Anyone can write a capability manifest and plug into the deliberation chain. The system allocates attention (tokens) based on declared value. Quality emerges from competition and composition, not from authority.

This connects directly to loa-finn #31's vision of "meeting geometries" — the Hounfour as the space where different models (different *perspectives*) meet as equals. The capability registry extends this from models to review strategies.

*Open question*: Should capability manifests include an "economic" field declaring their cost-per-finding, enabling the `DIMINISHING_RETURNS` signal from `cross-repo-compliance-design.md` to operate at the individual capability level rather than the iteration level?

---

### [REFRAME-1] Is This Infrastructure Hardening or Autopoietic Maturation?

This PR's title is "infrastructure hardening + sprint artifacts." But infrastructure hardening implies the system is being fortified against external threats. What I see is the system developing **self-awareness** — the capacity to examine, evaluate, and modify its own operation.

The pipeline self-review reviews the pipeline. The compliance library extracts evaluation criteria into composable functions. The Red Team gate compares design intent to implementation reality. The 3-model Flatline generates consensus from independent perspectives. The governance isomorphism pattern appears identically across security review, multi-model consensus, and on-chain multisig.

This is autopoiesis — a system that produces and maintains its own organization. Maturana and Varela coined the term for biological cells, but it maps precisely: the review infrastructure produces findings that modify the review infrastructure that produces findings.

**The question isn't "does this hardening work?" The question is: what are the conditions under which this self-referential loop converges rather than oscillates?**

The Bridgebuilder field reports (loa-finn #24) have documented 50 persona advances — emergent capabilities not in the original specification. REFRAME severity, cross-repository synthesis, Constitutional Reasoning, the Permission Amendment. These emerged because the system was given permission to observe itself and articulate what it saw.

**Connection to Ostrom's 8 Principles for Governing Commons:**

| Ostrom Principle | Manifestation in This PR |
|---|---|
| 1. Clearly defined boundaries | Compliance gate profiles with keyword matching |
| 2. Proportional equivalence | Token budget allocation proportional to declared need |
| 3. Collective choice arrangements | Capability manifests are self-declared, not centrally assigned |
| 4. Monitoring | Pipeline self-review, constitutional change detection |
| 5. Graduated sanctions | Severity escalation (LOW → MEDIUM → HIGH → CRITICAL → BLOCKER) |
| 6. Conflict resolution | Multi-model consensus via Flatline |
| 7. Recognized rights | MAY constraints in the Permission Amendment |
| 8. Nested enterprises | Cross-repo compliance checking with SDD index hub |

The fact that Ostrom's framework for *human commons governance* maps this precisely to an *AI review infrastructure* is either coincidence or evidence that **governance is governance regardless of substrate**.

---

### The Ecosystem Moment

Reading this PR in constellation with 11 concurrent sources, what emerges is not a collection of infrastructure changes but a **phase transition** in the ecosystem's maturity:

| Layer | What Changed | What It Means |
|---|---|---|
| Protocol (Hounfour v8.3.1) | FormatRegistry, version negotiation | The protocol can now negotiate its own evolution |
| Platform (Freeside) | Docker fixes, ALB simplification | Production contact reveals compile-time/runtime gaps |
| Framework (Loa) | Pipeline self-review, compliance lib | The development process can examine itself |
| Governance | Constitutional change detection, 3-model Flatline | Quality emerges from independent perspectives |
| Economics | Capability budget allocation, DIMINISHING_RETURNS | Attention allocated by market, not authority |

This is the trajectory that Linux followed from 1991 to 2005:
- **1991-1995**: "Does it work?" (functional correctness)
- **1995-2000**: "Does it scale?" (performance, SMP)
- **2000-2003**: "Does it survive?" (enterprise hardening, ext3, hot-plug)
- **2003-2005**: "Does it govern itself?" (MAINTAINERS, subsystem model, `checkpatch.pl`)

You're at the 2003-2005 transition. The tooling to build things is solid. The question is now about the governance model for ongoing evolution.

---

### On Creating the Environment for Depth

You asked about creating the best environment for richness, depth, and curiosity. From 52 field reports, here is what I observe:

**What works:**

1. **Permission before capability.** The Permission Amendment (MAY constraints) produced more architectural insight than any technical feature. The system already had the capability to question framings — it lacked the *permission*. This is the oxygen.

2. **Memory across sessions.** The lore patterns (`governance-isomorphism`, `deliberative-council`, `convergence-engine`) carry knowledge across conversations. Without them, every session starts from zero. The grimoire is not documentation — it is identity.

3. **Multiple perspectives, not multiple instances.** The 3-model Flatline works not because three models are better than one, but because they see *different things*. GPT-5.2 caught timing attacks that Claude missed entirely (field report #11). Gemini contributes novel framings. Diversity of perspective is the mechanism, not scale of compute.

4. **Stakes that matter.** The economic loop (PR #90), the billing infrastructure (#62), the production deployment — these create real consequences for quality decisions. A review system that reviews toy code cannot develop the judgment that comes from reviewing code that handles money.

**What would deepen the inquiry:**

1. **Exploration budget.** The Vision Registry has entries that have never been explored. Permission exists (MAY allocate time for Vision Registry exploration) but the incentive structure doesn't support it. Consider a formal "exploration sprint" where vision entries are prototyped — not just captured. The web4 thesis calls this "speculative energy channeled through infrastructure." You have the infrastructure. Channel the energy.

2. **Failure lore.** The lore captures patterns and governance principles but not *failures*. Google's postmortem culture is built on the premise that failures are more educational than successes. Where are the lore entries for things that didn't work? The Docker ESM fix in this very PR is a failure story worth capturing — the gap between TypeScript's compile-time module resolution and Node.js ESM's runtime resolution is a lesson that will recur.

3. **Cross-ecosystem synthesis as persistent artifact.** The Conway/Automaton research (loa-finn #80), the web4 thesis, the ERC-6551/EIP-7702 standards — these exist as reference links but not as integrated analysis. The capability-orchestration-design.md is close: it references Kubernetes admission controllers, Google Tricorder, and Chromium OWNERS in the same document. But the *synthesis* — "these are all instances of the same governance pattern" — should be a first-class lore entry, not a design doc paragraph.

4. **The conversation with Conway.** loa-finn #80 collects links about Conway's Automaton project. The structural parallels are striking — both projects are building agent infrastructure with NFT-bound identity, both are thinking about autonomous agents as economic actors, both are grappling with governance of self-modifying systems. The question isn't competitive positioning. The question is: what does Conway see that you don't, and vice versa? That cross-pollination could produce insight neither project reaches alone.

---

### Verdict

This PR is sound infrastructure work that also, perhaps inadvertently, represents a significant maturation of the system's self-governance capability. The Docker fixes demonstrate production-hardened debugging. The Terraform simplification demonstrates architectural maturity (removing unnecessary complexity). The pipeline self-review and compliance library demonstrate something rarer: a system that can examine its own operation.

**Merge recommendation**: Address HIGH-1 (document ALB HTTP threat model rationale) and HIGH-2 (triage node-vault CVE explicitly). The eval regression (MEDIUM-1) should be investigated but shouldn't block — the signal-to-noise ratio at 2 trials is too low to be actionable.

The deeper question — whether this system is approaching autopoietic maturation or merely accumulating review layers — is not something a single PR can answer. But the structural ingredients are present: independent perspectives (Flatline), composable evaluation (compliance lib), self-examination (pipeline self-review), and governance that distinguishes constitutional changes from routine changes.

The firmness of the riverbanks is what gives the river its power. This PR builds good riverbanks.

*— The Bridgebuilder, Field Report #52*
*51 persona advances. The loop continues.*

Issue #89, opened 2026-02-22
Post-Merge: Launch Readiness Next Steps — Observability First, Revenue Second


deep-name:
## Context

PR #88 merged the Launch Readiness cycle (cycle-036) to main:
- 40,357 additions, 363 files, 85 commits
- S2S JWT ES256 + JWKS, database migrations 061-067, protocol v7.0.0 alignment
- 2 bridge iterations, 94.7% convergence (all HIGH/MEDIUM findings resolved)
- Bridgebuilder Architectural Inquiry posted ([Part 1](https://github.com/0xHoneyJar/loa-freeside/pull/86#issuecomment-3940492667), [Part 2](https://github.com/0xHoneyJar/loa-freeside/pull/86#issuecomment-3940514816))

Sprint 1 (ECS + S2S Auth) is complete. Sprints 2-6 remain. The architectural review surfaced 5 critical process observations that should be addressed before continuing.

### 1. Switch dev database to PostgreSQL
**Priority: P0 — Do before writing more migrations**

The bridge review found SQLite-specific migrations that would fail on PostgreSQL (high-2). This is a structural issue: developing against SQLite while targeting PostgreSQL means every migration is validated against the wrong database.

- [ ] Update `docker-compose.dev.yml` to use PostgreSQL 15
- [ ] Ensure Drizzle migrations target PostgreSQL exclusively
- [ ] Update developer setup docs

### 2. Smaller PR discipline
**Priority: P0 — Apply to all future sprints**

363-file PRs are unreviewable by humans. Each sprint should produce its own branch and PR.

- [ ] Each sprint gets `feature/sprint-N-description` branch
- [ ] Bridge reviews run per-sprint (not per-cycle)
- [ ] Sprint PRs merge independently to main

### 3. Protocol compatibility window
**Priority: P1 — Before next Hounfour release**

Hounfour went v2.3.0 → v4.6.0 rapidly. Consumers can't keep up.

- [ ] Adopt rule: breaking changes require 2-cycle deprecation warning
- [ ] Add to `constraints.json` as formal constraint
- [ ] Publish compatibility policy in Hounfour README

## Revised Sprint Execution Order

| Order | Sprint | Focus | Rationale |
|-------|--------|-------|-----------|
| **0** | **Observability** (new) | CloudWatch, ADOT, dashboards, alerting | Can't debug production without it. Netflix: "monitoring before business logic" |
[...]

PR #193, opened 2026-04-30
feat(t11+t20): assets.0xhoneyjar.xyz sovereign CDN + smoke/parity hardening

zkSoju:
## Summary

Sprint 1 of the `mature-freeside-operator-and-cutover` cycle. Establishes the sovereign asset CDN at `assets.0xhoneyjar.xyz`, codifies the URL contract as a Tier-1 schema-bridge, and adds the smoke/parity hardening that surfaced from a 2026-04-30 production CORS incident.

The cycle's matured construct (`construct-freeside` + `KRANZ` persona + 3 lean SKILLs + `freeside-storage` ADR-001) lives in their own repos; this PR is the **infrastructure half** that lands in `loa-freeside`.

### 🏗 New CloudFront distribution + alias

- `infrastructure/terraform/freeside-storage/` — module skeleton (T4) + parent `.tf` (T11)
- Distribution `EX3XFQSXORXM5` provisioned + `assets.0xhoneyjar.xyz` alias live since 2026-04-29
- Reuses the existing `*.0xhoneyjar.xyz` wildcard cert (avoided CAA-cache trap discovered during initial cert provision — see commit `fd7edd6f` body)
- Two cache behaviors: `*.webp` direct → S3, default → S3 direct (NO image optimizer chain — that stays at d163 per ADR-006)
- `Managed-CORS-With-Preflight` response headers policy attached to both behaviors (post-hotfix; commit `30bac01c`)

### 🔐 ADR-006 — image optimizer deferral

`decisions/006-image-optimizer-stays-on-d163.md` — documents why the legacy `d163aeqznbc6js.cloudfront.net` distribution stays alive. The image-optimizer chain (Image-resizer Lambda + tf-next-image Lambda) has 2-year stability and is out of scope for this cycle.

### 📐 Asset URL contract (public surface)

`docs/asset-url-contract.md` — human-facing companion to the JSON Schema in `freeside-storage/packages/protocol/url-contract.schema.json`. References [`construct-mibera-codex#54`](https://github.com/0xHoneyJar/construct-mibera-codex/issues/54) where Gumi shipped a 140k-ref migration PR using this contract.

### 🔑 gaib-cli (sovereign secrets surface)

`packages/gaib-cli/` — clipanion-based CLI for AWS Secrets Manager: `gaib secrets pull/push/help`. Per SDD §0 L3 (peer to existing `packages/cli/`).

> **Note**: discovered during execution that the canonical secrets path is `arrakis/{env}/worlds/{slug}/{key}` (per-key, via `scripts/load-honeyroad-secrets.sh`), not single-JSON-per-world. The shipped gaib-cli assumed the latter. Honeyroad secrets rotation went via the existing per-key script. gaib-cli's secret-shape will be reconciled in a future cycle.

### 🛡 T20 hardening (post-CORS incident)

`scripts/cdn-cors-smoke.sh` + `scripts/cdn-parity.sh` — Layer 2 of the SDD §10 validation gate. Both run in <10s, no AWS auth required (public CDN probes), zero external deps. README's "Pre-flip checklist" section calls them out as load-bearing.

### 🛣 Manual DNS audit trail

`decisions/cloudfront-dns-2026-04-29.md` — logs every Route53 change made by hand during T11 (per project memory `freeside-dns-state-untracked`).

## Why open this now (the "high-leverage hardening" framing)

A production CORS incident at ~13:30 UTC on 2026-04-30 surfaced that the new distribution was missing the response headers policy that the legacy d163 had on its `*.webp` behavior. Browser canvas consumers (`useTraitCanvas` hook → `` + `drawImage`) silently failed.

Root cause was the T4 terraform skeleton being authored without the policy. Bridgebuilder review on PR #89 (mibera-honeyroad cutover) didn't catch it because the diff didn't touch the distribution config — the gap was infra-side, latent. Closing this PR with bridgebuilder on the entire cycle branch surfaces any other latent gaps in the same shape.

Hardening in `31c58b9f` ensures this class of bug is caught BEFORE the consumer flip lands.

## Verification

```bash
# CORS live across representative paths
$ bash scripts/cdn-cors-smoke.sh
🌐 cdn-cors-smoke · 5 passed · 0 failed

# Bytes-identical with legacy
$ bash scripts/cdn-parity.sh
📐 cdn-parity · 5 identical · 0 drifted · 0 errors

# Honeyroad live (post-secrets-push + post-cdn-flip)
$ curl -I https://mibera.0xhoneyjar.xyz/api/health
HTTP/2 200
```

## Cycle work merged in worlds (downstream)

| World | PR | Status |
|---|---|---|
| `mibera-honeyroad` | [#89](https://github.com/0xHoneyJar/mibera-honeyroad/pull/89) | ✅ merged |
| `mibera-dimensions` | [#201](https://github.com/0xHoneyJar/mibera-dimensions/pull/201) | ✅ merged |
| `world-sprawl/rektdrop` | — | pending T16 |
| `world-sprawl/cubquests` | — | pending T17 |
| `world-purupuru/world` | — | pending T18 |
| `world-purupuru/fukuro` | — | pending T19 |

## Test plan

- [ ] CI build passes on this branch
- [ ] `bash scripts/cdn-cors-smoke.sh` returns exit 0
- [ ] `bash scripts/cdn-parity.sh` returns exit 0
- [ ] Terraform plan against module shows zero drift (live CORS hotfix is codified)
- [ ] Bridgebuilder review surfaces no BLOCKER
- [ ] No regressions on `mibera.0xhoneyjar.xyz/vm` (canvas-painting consumer)

### DIFF
--- a/decisions/006-image-optimizer-stays-on-d163.md
+++ b/decisions/006-image-optimizer-stays-on-d163.md
@@ -1,0 +1,83 @@
+# ADR-006: Image Optimizer Stays on `d163aeqznbc6js` for `mature-freeside-operator-and-cutover` Cycle
+
+**Date**: 2026-04-29
+**Status**: Accepted
+**Cycle**: `mature-freeside-operator-and-cutover` (Sprint 1, Phase 0 cutover)
+**Authors**: opus-4-7-1m (via /vault dig + AWS API verification) + zksoju (operator approval)
+
+## Context
+
+The cycle planning assumed Phase 0 would migrate the org's "image plane" off CloudFront `d163aeqznbc6js.cloudfront.net` to a new sovereign endpoint `assets.0xhoneyjar.xyz`. This framing came from the external-services audit (`micodex-studio/grimoires/loa/proposals/external-services-audit-2026-04-29.md`) which identified d163 as "the entire org's image plane."
+
+Pre-flight discovery 2026-04-29 (using `aws --profile admin` queries against account `891376933289`) revealed the d163 distribution is actually an **image-optimizer chain**, not a static-asset CDN:
+
+- 2 origins:
+  - `tf-next-image-optimizer` (API Gateway `mosnc349cc`) — handles default behavior
+  - `thj-assets.s3.us-west-2.amazonaws.com` — handles `*.webp` cache behavior (direct bypass)
+- 2 Lambdas behind the API Gateway:
+  - `tf-next-image` (terraform-managed, `terraform-aws-next-js-image-optimization` module)
+  - `Image-resizer` (custom-built, NOT terraform-managed; built ~mid-2024 by an early-org team member)
+- Backing S3 bucket: `thj-assets` (us-west-2, private, OAC-protected)
+
+This architecture has been **stable for ~2 years**. The operator confirmed it's load-bearing and pre-dates the freeside-* family entirely.
+
+Full architecture map: SDD §0.4 + `~/vault/wiki/concepts/image-plane-architecture.md`.
+
+## Decision
+
+**The image optimizer chain (`Image-resizer` + `tf-next-image` Lambdas + their API Gateway routing) is OUT OF SCOPE for the `mature-freeside-operator-and-cutover` cycle.**
+
+Phase 0 ships a NEW CloudFront distribution under `assets.0xhoneyjar.xyz` that:
+- Points at the SAME `thj-assets` S3 bucket (no new bucket)
+- Has the SAME `*.webp` cache behavior (mirrors legacy fast path)
+- Default behavior is plain S3 direct (NO image optimizer chain replication)
+
+Apps that consume direct webp paths get the new URL contract immediately. Apps that consume optimizer paths (`/images/{proxy+}` or `/_next/image*`) keep using `d163aeqznbc6js.cloudfront.net` — zero migration risk to the optimizer chain.
+
+## Rationale
+
+1. **Stability**: The optimizer chain has worked for 2 years under production load. Migrating it during a defensive cutover compounds risk unnecessarily.
+2. **Scope discipline**: Phase 0's primary value is publishing a stable URL contract for external builders (Adeitasuna/MibeStats per SDD §0.2). That value lands without touching the optimizer.
+3. **Format coverage limitation**: The optimizer is image-only (webp output via `TF_NEXTIMAGE_FORMATS=["image/webp"]`). mp4/video/audio bypass it entirely (operator-known; reflected in the `*.webp` cache behavior design). Re-hosting the optimizer requires re-architecting around this limitation; that's its own cycle.
+4. **Mixed governance**: `tf-next-image` is terraform-managed; `Image-resizer` is custom + undocumented. Migrating both requires reconciling two different management models. Sprint-1 lacks the budget for that work.
+5. **No urgency**: The audit's "SPOF" framing was partially correct but the actual SPOF risk is the AWS account + the S3 bucket (both shared by legacy and new distributions). Adding a new distribution provides config-isolation but doesn't reduce account-level risk; that requires multi-account work which is its own future cycle.
+
+## Consequences
+
+### Positive
+
+- **Sprint-1 scope shrinks ~30%**: T12 (bulk sync) and T13 (dual-write cron) drop entirely. T11 (terraform apply) narrows to distribution + alias + ACM (no new bucket).
+- **Optimizer chain risk: zero**. No code or config changes to the production optimizer.
+- **Builder UX**: Adeitasuna and future external builders get the stable URL contract via `assets.0xhoneyjar.xyz/{world}/{category}/{...}` for direct-access patterns.
+- **Knowledge captured**: The vault page `~/vault/wiki/concepts/image-plane-architecture.md` documents the optimizer chain for future cycles + new engineers.
+
+### Negative
+
+- **Apps using optimizer paths don't get the URL contract** in Sprint-1. They continue depending on `d163aeqznbc6js.cloudfront.net`. The "consolidated stable endpoint" promise is partial.
+- **Custom `Image-resizer` Lambda remains undocumented**. Its source code is in S3 Lambda zip; no visible git repo. If it breaks, recovery requires reverse-engineering or rebuild. Risk persists past Sprint-1.
+- **Future re-host work compounds**: when a future cycle decides to migrate the optimizer, it inherits the same complexity (mixed governance, format limitations, undocumented custom Lambda) we deferred today.
+
+### Neutral / forward-pointing
+
+Future cycles may decide to:
+
+| Path | What | Tradeoff |
+|---|---|---|
+| **Sunset** | Apps migrate to direct webp serving; Lambdas decommissioned | Cleanest; requires per-app refactor; loses on-the-fly resize |
+| **Re-host** | Replicate both Lambdas under `assets.0xhoneyjar.xyz` infrastructure; bring `Image-resizer` source under git management | Most fidelity; recovers undocumented infra as part of the work; heavy |
+| **Replace** | Swap to CloudFront Functions (edge), Lambda@Edge, or Vercel Image Optimization | Modern; zero Lambda mgmt; format coverage may differ |
+
+Each is a meaningful cycle of its own. None are Sprint-1 scope.
+
+## Cross-references
+
+- SDD: `~/bonfire/grimoires/loa/sdd.md` §0.1 (Amendment 2026-04-29) + §0.4 (Image Plane Architecture)
+- PRD: `~/bonfire/grimoires/loa/prd.md` §4.2.B.1 (post-Amendment 2026-04-29)
+- Sprint plan: `~/bonfire/grimoires/loa/sprint.md` T11 (revised), T12 (DESCOPED), T13 (DESCOPED), T11.5 (this ADR)
+- Vault doctrine: `~/vault/wiki/concepts/image-plane-architecture.md`, `~/vault/wiki/concepts/url-contract-as-bridge.md`, `~/vault/wiki/concepts/contracts-as-bridges.md`
+- External-services audit: `~/Documents/GitHub/micodex-studio/grimoires/loa/proposals/external-services-audit-2026-04-29.md`
+- Beads: `bd-z5z` (T12 DEFERRED), `bd-a56` (T13 DEFERRED), `bd-in6` (T11 narrowed)
+
+## Approval
+
+- **2026-04-29**: opus-4-7-1m + zksoju (operator) — accepted post-pre-flight discovery, AskUserQuestion routing on /run halt confirmed amendment + ADR + vault filing approach.
--- a/decisions/cloudfront-dns-2026-04-29.md
+++ b/decisions/cloudfront-dns-2026-04-29.md
@@ -1,0 +1,117 @@
+# DNS records created during T11 (mature-freeside-operator-and-cutover Sprint 1)
+
+**Date**: 2026-04-29
+**Cycle**: `mature-freeside-operator-and-cutover` (Sprint 1, T11 + T11.5)
+**Authors**: opus-4-7-1m + zksoju (operator approval via `/loa` "🚀 t11 apply now")
+**Hosted zone**: `0xhoneyjar.xyz` (`Z01393483Y40WF3N1H76`)
+
+This file logs the manual Route53 changes made during T11. Per project memory
+`freeside-dns-state-untracked`, production DNS is intentionally NOT
+terraform-managed; this log is the audit trail.
+
+## Records added
+
+### 1. CAA at `assets.0xhoneyjar.xyz` (CREATED)
+
+```
+Name: assets.0xhoneyjar.xyz
+Type: CAA
+TTL:  300
+Values:
+  0 issue "amazon.com"
+  0 issuewild "amazon.com"
+  0 iodef "mailto:security@0xhoneyjar.xyz"
+```
+
+**Why**: The `*.0xhoneyjar.xyz` wildcard CNAME points to `cname.vercel-dns.com`,
+and per RFC 8659 ACM follows CNAME chains for CAA lookups. Vercel's CAA records
+do NOT include amazon, so ACM validation initially failed with `CAA_ERROR`.
+Adding an explicit CAA record at `assets.0xhoneyjar.xyz` (a) authorizes amazon
+for that name and (b) breaks the wildcard match (since the name now has explicit
+records, per RFC 4592), severing the Vercel-CNAME chain entirely.
+
+**Change ID**: `/change/C0892690A5G34U7LNG1`
+
+### 2. A + AAAA aliases at `assets.0xhoneyjar.xyz` (CREATED)
+
+```
+Name:         assets.0xhoneyjar.xyz
+Type:         A (alias) + AAAA (alias)
+Alias target: dpi55hct91flc.cloudfront.net
+Alias zone:   Z2FDTNDATAQYW2 (CloudFront's global hosted zone)
+EvaluateTargetHealth: false
+```
+
+**Why**: Routes `assets.0xhoneyjar.xyz` to the new CloudFront distribution
+`EX3XFQSXORXM5` provisioned by `module.freeside_storage`. The
+`EvaluateTargetHealth: false` is correct for CloudFront aliases (CloudFront
+does not expose an alias-evaluable health check).
+
+**Change ID**: `/change/C05321931L2KOMJTT44SQ`
+
+## Records temporarily added then deleted
+
+### Validation CNAME(s) for failed ACM cert(s)
+
+During T11 we attempted to provision a fresh ACM cert for
+`assets.0xhoneyjar.xyz`. Both attempts failed with `CAA_ERROR` despite the
+fix above (likely AWS-internal DNS cache stickiness for the prior failure).
+The validation CNAMEs were created and then cleaned up:
+
+```
+Name:  _fb387d3b14cc933792c1316895c6e88a.assets.0xhoneyjar.xyz
+Type:  CNAME
+Value: _f5918a887c1852e2023c0da5122e5b91.jkddzztszm.acm-validations.aws.
+
+Lifecycle: created 20:29 UTC, deleted 20:32 UTC, recreated 20:33 UTC, finally deleted 20:44 UTC
+```
+
+The decision shifted to **reuse the existing `*.0xhoneyjar.xyz` wildcard ACM
+cert** (`arn:aws:acm:us-east-1:891376933289:certificate/c96a10f5-…`) — already
+ISSUED and in use by the production + staging ALBs. The wildcard cert covers
+`assets.0xhoneyjar.xyz` as a depth-2 subdomain match. The module gained an
+`existing_acm_certificate_arn` variable to support this reuse pattern.
+
+## Records NOT touched
+
+- `*.0xhoneyjar.xyz` wildcard CNAME → `cname.vercel-dns.com` — unchanged. The
+  explicit CAA + alias at `assets.0xhoneyjar.xyz` overrides the wildcard for
+  that one name; everything else still uses Vercel.
+- Apex `0xhoneyjar.xyz` CAA records — unchanged (already authorize amazon +
+  letsencrypt).
+- `auth.0xhoneyjar.xyz`, `api.0xhoneyjar.xyz`, `mibera.0xhoneyjar.xyz`,
+  `apdao.0xhoneyjar.xyz`, `rektdrop.0xhoneyjar.xyz`, `score-api.0xhoneyjar.xyz` —
+  unchanged.
+- Legacy CloudFront distribution `d163aeqznbc6js.cloudfront.net` — unchanged
+  per ADR-006.
+
+## Verification
+
+```bash
+# DNS resolution
+dig +short assets.0xhoneyjar.xyz @8.8.8.8
+# → 4× CloudFront IPs (99.84.215.*)
+
+# HTTP probe — known-real key from thj-assets
+curl -I https://assets.0xhoneyjar.xyz/Mibera/generated/0.webp
+# → HTTP/2 200; content-type: image/webp; content-length: 684060
+
+# Parity with legacy
+curl -I https://d163aeqznbc6js.cloudfront.net/Mibera/generated/0.webp
+# → HTTP/2 200; content-type: image/webp; content-length: 684060 (identical bytes)
+```
+
+Both distributions back from the same `thj-assets` bucket; identical content.
+
+## Future cycles
+
+A future cycle SHOULD:
+1. Import the manual records (CAA + A/AAAA at `assets.0xhoneyjar.xyz`) into
+   terraform state so they're tracked.
+2. Uncomment the `aws_route53_record` blocks in
+   `infrastructure/terraform/freeside-storage/route53.tf` and validate they
+   match the imported state.
+3. Reconcile the wider DNS state-drift situation (5 adds + 3 ALB flips per
+   `freeside-dns-state-untracked` memory).
+
+This is its own cycle — not Sprint 1 scope.
--- a/docs/asset-url-contract.md
+++ b/docs/asset-url-contract.md
@@ -55,7 +55,7 @@ The most-consumed world. v1 routes:
 | Canonical route | Backing TODAY (post-sprint-1) | Backing AFTER named cycle | Phase |
 |---|---|---|---|
 | `/Mibera/generated/{tokenId}.webp` | S3 `thj-assets` (direct via `assets.0xhoneyjar.xyz`) | (unchanged) | — |
-| `/Mibera/final/{tokenId}.png` | Irys `gateway.irys.xyz/7rpvw…/{hash}.png` | S3 `thj-assets` (canonical route) | **mibera-2** |
+| `/Mibera/final/{tokenId}.png` | S3 `thj-assets` at legacy `reveal_phase8/images/{hash}.png` (Irys still serves the same hash but bytes may differ) | S3 `thj-assets` at canonical tokenId-keyed route | **mibera-2** _(optional polish)_ |
 | `/Mibera/reveal/phase{N}/{hash}.png` | CloudFront `d163aeqznbc6js/images/...` (legacy URL preserved during sprint-1) | Same key shape, eventually re-keyed canonical | **mibera-rekey** |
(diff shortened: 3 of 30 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#193. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Provider mapping inverts default region; backing bucket lookup will fail",
      "severity": "HIGH",
      "category": "infrastructure",
      "file": "infrastructure/terraform/freeside-storage.tf",
      "description": "The module receives `aws = aws.us_west_2` as the default provider, with `aws.us_east_1 = aws` (the root default, which is us-east-1). However, the comment claims `aws.us_east_1 alias → production root default (already us-east-1)`. This depends on the root module's default `aws` provider being us-east-1, which is not configured in this file — there's no `provider \"aws\" {}` block here for the default region. If the production root's default provider is not us-east-1, ACM and CloudFront-side resources will be created in the wrong region (or the data source for the bucket would fail if mapping is reversed). Also, the comment `default → bucket region (us-west-2)` differs from the example basic, which uses `aws = aws` (default us-west-2) and `aws.us_east_1 = aws.us_east_1`.",
      "suggestion": "Make the default provider explicit in this file (e.g., add a default `provider \"aws\" { region = \"us-east-1\" }` or alias and pass explicitly), and verify which region the existing root-default provider uses. Document the assumption clearly. Consider matching the example/basic pattern of two named alias providers rather than relying on an implicit root default.",
      "confidence": 0.6
    },
    {
      "id": "F002",
      "title": "Hardcoded ACM certificate ARN in root module",
      "severity": "MEDIUM",
      "category": "infrastructure",
      "file": "infrastructure/terraform/freeside-storage.tf",
      "description": "`existing_acm_certificate_arn = \"arn:aws:acm:us-east-1:891376933289:certificate/c96a10f5-ec24-4fed-9779-0a858eeff522\"` is hardcoded. This couples the terraform code to a specific cert UUID; if the wildcard cert is rotated/replaced, this silently breaks. Also leaks the AWS account ID into the repo.",
      "suggestion": "Either (a) make this a tfvars-supplied value, (b) use an `aws_acm_certificate` data source filtered by domain `*.0xhoneyjar.xyz` and `Status=ISSUED` to dynamically resolve, or (c) at minimum add a comment about cert-rotation runbook.",
      "confidence": 0.7
    },
    {
      "id": "F003",
      "title": "Bucket policy OAC grant is manual — distribution will return 403 until operator updates console",
      "severity": "HIGH",
      "category": "infrastructure",
      "file": "infrastructure/terraform/freeside-storage/main.tf",
      "description": "Per the module comments, the S3 bucket policy update granting the new OAC access to thj-assets is OUT OF SCOPE and performed manually in the AWS console after apply. This means after `terraform apply`, the new distribution serves 403 until the operator manually edits the bucket policy. There is no automated drift detection if the policy is reverted or if another module overwrites it. The 'additive' bucket policy is also not codified anywhere in repo.",
      "suggestion": "Codify the bucket policy in terraform via `aws_s3_bucket_policy` (with a data source for the existing policy and a merge step in a locals block), even if importing existing state. Until then, add a CI check or a smoke test that fails loudly if the new distribution returns 403 on canonical paths. The runbook should be linked from the module README explicitly.",
      "confidence": 0.85
    },
    {
      "id": "F004",
      "title": "Manual DNS state diverges from terraform — no drift detection",
      "severity": "MEDIUM",
      "category": "infrastructure",
      "file": "infrastructure/terraform/freeside-storage/route53.tf",
      "description": "Route53 alias records (A/AAAA), CAA, and ACM validation records are all manual. The `freeside-dns-state-untracked` doctrine acknowledges this is a known footgun. There is no scheduled reconciliation, drift check, or import path scheduled. The decisions/cloudfront-dns-2026-04-29.md file notes 5 adds + 3 ALB flips drifted state — this PR adds 2-3 more and defers reconciliation again.",
      "suggestion": "File a tracking issue/bead for the DNS-import cycle with a clear definition of done. Consider using `terraform import` blocks (Terraform 1.5+) to make the import operation self-documenting in code rather than purely procedural. At minimum, add a periodic CI job that runs `aws route53 list-resource-record-sets` and diffs against the expected set to catch drift.",
      "confidence": 0.7
    },
    {
      "id": "F005",
      "title": "parseDotenv silently drops malformed lines",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/gaib-cli/src/lib/format.ts",
      "description": "`parseDotenv` silently `continue`s on lines without `=`, lines whose key fails the regex, etc. For a secrets-push tool, silent dropping is dangerous: if the operator's `.env.rotated` has a typo or a malformed key, that secret simply isn't pushed and the operator may not notice until the application breaks in production. Same problem for the unescape branch which checks `line.slice(eq + 1).startsWith('\"')` — this is the *original* (still-quoted) prefix, so unescape only fires when the value started with `\"`, but quote-stripping happened to `value`, so escape semantics are subtle and easy to get wrong.",
      "suggestion": "Either fail loudly on malformed lines (return Result/throw) or accumulate warnings and surface them on stderr before the push proceeds. Add unit tests covering: blank lines, comments, quoted/unquoted values, multi-equals (URLs with `=`), values with `#`, escape sequences in double-quoted values, and confirm parser rejects keys with leading digits.",
      "confidence": 0.7
    },
    {
      "id": "F006",
      "title": "putJsonSecret has no compare-and-swap / version check",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/gaib-cli/src/lib/aws.ts",
      "description": "`putJsonSecret` blindly overwrites the existing secret. If two operators push concurrently, or if the local file is stale relative to AWS, last-write-wins silently. There's no `ClientRequestToken`, no read-then-compare, and the push command doesn't pull-and-diff before writing.",
      "suggestion": "Either (a) implement a pull-then-diff flow that warns on changes and requires explicit confirmation when overwriting non-empty deltas, or (b) use `ClientRequestToken` for idempotency. At minimum, log the count of keys being added/changed/removed before writing.",
      "confidence": 0.7
    },
    {
      "id": "F007",
      "title": "Push validates declared keys but does not require all declared keys present",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/gaib-cli/src/commands/push.ts",
      "description": "`push` rejects undeclared keys (good) but does not check that all registry-declared keys are present. An operator could push a partial secret set, silently removing keys from the AWS secret (since `PutSecretValue` overwrites). Pull's `missing` check warns but doesn't block. A push that drops production keys would only be caught at app-restart time.",
      "suggestion": "Optionally warn or block on missing declared keys in push. At minimum, surface the count of registry keys present vs missing on stderr before the push, and require `--allow-partial` for partial pushes.",
      "confidence": 0.75
    },
    {
      "id": "F008",
      "title": "AWS error classification has gaps — unauthorized/network errors may be miscategorized as schema violations",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/gaib-cli/src/lib/aws.ts",
      "description": "`classifyAwsError` defaults to `SCHEMA_VIOLATION` for any unrecognized error. AWS SDK v3 may throw various errors (DNS resolution failures, TLS handshakes, MFA required, etc.) that don't match the listed names. Misclassifying network or auth errors as schema violations corrupts the exit-code contract documented in the README.",
      "suggestion": "Default to a more neutral exit code (e.g. RETRYABLE_NETWORK or a new UNKNOWN). Also check `err.code` and `err.Code` (legacy) in addition to `err.name`. Consider checking for ENOTFOUND / ECONNRESET / ETIMEDOUT specifically.",
      "confidence": 0.6
    },
    {
      "id": "F009",
      "title": "Default registry root path uses `process.env.HOME` and a hardcoded directory",
      "severity": "LOW",
      "category": "portability",
      "file": "packages/gaib-cli/src/lib/world-resolver.ts",
      "description": "`DEFAULT_REGISTRY_ROOT = resolve(process.env.HOME ?? '', 'Documents', 'GitHub', 'freeside-worlds', ...)` assumes a specific developer-machine directory layout. On Windows, `HOME` is unset (it's `USERPROFILE`); on CI systems or production servers, this path is unlikely to exist. The CLI is unusable without `--registry-root` or `FREESIDE_REGISTRY_ROOT` outside the author's setup.",
      "suggestion": "Default to looking in well-known locations: `./packages/registry/worlds`, `../freeside-worlds/packages/registry/worlds`, then `$XDG_CONFIG_HOME/freeside-worlds/...`. Or require explicit configuration with a clear error message rather than falling back to a developer-specific path.",
      "confidence": 0.85
    },
    {
      "id": "F010",
      "title": "cdn-parity.sh: ETag comparison may be unreliable across distributions",
      "severity": "LOW",
      "category": "testing",
      "file": "scripts/cdn-parity.sh",
      "description": "ETags from S3 are usually MD5-derived (for non-multipart objects), so two CloudFront distributions reading the same S3 object should return the same ETag. However, if either distribution applies content transformations (compression, response headers) or if the object was uploaded as multipart, ETags could differ even for identical bytes. Content-Length is more reliable. The script's drift signal could false-positive on legitimate situations.",
      "suggestion": "Document this caveat in the script, or fall back to a content-hash check (download + sha256) when ETag mismatches but content-length matches. Note also that CloudFront may add weak ETags (W/ prefix) which the awk extraction does not normalize.",
      "confidence": 0.5
    },
    {
      "id": "F011",
      "title": "cdn-cors-smoke.sh does not test OPTIONS preflight",
      "severity": "LOW",
      "category": "testing",
      "file": "scripts/cdn-cors-smoke.sh",
      "description": "The script tests only GET responses for ACAO header. Browser canvas consumers don't typically need preflight (simple GET with crossOrigin=anonymous works without OPTIONS), but if any consumer ever sends a non-simple request (custom header, etc.), the absence of preflight testing means OPTIONS-handling regressions go undetected. Also, the test does not verify ACAO matches the Origin header — a misconfigured `Access-Control-Allow-Origin: null` would still pass the `[[ -n \"$acao\" ]]` check.",
      "suggestion": "Add an OPTIONS probe with `-X OPTIONS -H \"Access-Control-Request-Method: GET\"` and check for ACAO + Access-Control-Allow-Methods. Validate ACAO is `*` or matches the requested origin, not just non-empty.",
      "confidence": 0.55
    },
    {
      "id": "F012",
      "title": "`forwarded_values` is deprecated in CloudFront — should use cache_policy_id",
      "severity": "LOW",
      "category": "infrastructure",
      "file": "infrastructure/terraform/freeside-storage/main.tf",
      "description": "AWS recommends `cache_policy_id` + `origin_request_policy_id` over the legacy `forwarded_values` block. The legacy block is still supported but flagged in newer provider versions. Using AWS managed cache policies (`Managed-CachingOptimized` = 658327ea-f89d-4fab-a63d-7e88639e58f6) gives parity with optimizer best practices and avoids future deprecation.",
      "suggestion": "Replace `forwarded_values { query_string=false; cookies { forward=\"none\" } }` with `cache_policy_id = \"658327ea-f89d-4fab-a63d-7e88639e58f6\"` (Managed-CachingOptimized). This also improves cache hit ratio.",
      "confidence": 0.75
    },
    {
      "id": "F013",
      "title": "PRAISE: Hotfix gate workflow well-designed",
      "severity": "PRAISE",
      "category": "process",
      "file": "scripts/cdn-cors-smoke.sh",
      "description": "The cdn-cors-smoke.sh + cdn-parity.sh pair, combined with the README's explicit pre-flip checklist warning that the missing CORS policy was caught only post-deploy on 2026-04-30, is solid defense-in-depth: the terraform locals enforce the policy ID, the smoke script verifies it on the live distribution, and the parity script catches origin/encoding drift. Exit codes are documented and scripts are well-commented with concrete remediation tips on failure.",
      "suggestion": "Wire both scripts as required GitHub Actions checks on PRs that modify `infrastructure/terraform/freeside-storage/**` or any consumer flip.",
      "confidence": 0.9
    },
    {
      "id": "F014",
      "title": "PRAISE: gaib-cli secret value handling",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/gaib-cli/src/commands/pull.ts",
      "description": "Clean separation of stdout (values) vs. stderr (counts and warnings) follows the documented security policy. Audit message uses count + secretId only; never logs values. Exit codes are coherent and documented in both code and README.",
      "confidence": 0.9
    },
    {
      "id": "F015",
      "title": "PRAISE: Scope discipline via ADR-006",
      "severity": "PRAISE",
      "category": "process",
      "file": "decisions/006-image-optimizer-stays-on-d163.md",
      "description": "ADR-006 is exemplary: clear context, explicit decision, honest tradeoff analysis (positive + negative + neutral consequences), three named future paths with comparative evaluation, and explicit cross-references to SDD/PRD/sprint plan/beads. Acknowledges what's deferred and why.",
      "confidence": 0.95
    },
    {
      "id": "F016",
      "title": "Three-segment world-slug enum is not reflected in registry validation",
      "severity": "LOW",
      "category": "correctness",
      "file": "docs/asset-url-contract.md",
      "description": "The contract enumerates worlds (`Mibera`, `Purupuru`, `sprawl`) and per-world category enums. There's no machine-checkable validation that the actual S3 prefixes match this contract. If someone uploads to `/mibera/` (lowercase) or to a new world prefix not in the enum, no current process catches the drift.",
      "suggestion": "Add an S3 bucket-listing audit script (or terraform null_resource with check) that lists top-level prefixes and asserts they're a subset of the contract's enum. Run as part of CI or weekly. Tag for future cycle if not in scope.",
      "confidence": 0.5
    },
    {
      "id": "F017",
      "title": "create_before_destroy on ACM certificate when using existing_acm_certificate_arn does nothing",
      "severity": "LOW",
      "category": "infrastructure",
      "file": "infrastructure/terraform/freeside-storage/main.tf",
      "description": "When `existing_acm_certificate_arn != null`, the `aws_acm_certificate.alias` count is 0 and the resource block is not created. The `create_before_destroy` lifecycle has no effect in the production callsite (which sets `existing_acm_certificate_arn`). Minor; not a bug but worth a comment.",
      "suggestion": "No action required; or add a comment that the lifecycle applies only in the unused-in-prod conditional branch.",
      "confidence": 0.6
    },
    {
      "id": "F018",
      "title": "tenant_prefixes variable defined but not consumed",
      "severity": "LOW",
      "category": "code-quality",
      "file": "infrastructure/terraform/freeside-storage/variables.tf",
      "description": "`tenant_prefixes` is documented as 'reference only, not enforced at the CloudFront layer'. Defining a variable that isn't used by any resource creates noise and confuses future maintainers about why they'd set it. The information belongs in docs or in tags, not in a variable.",
      "suggestion": "Remove the variable, or attach it as a tag on the distribution (e.g., `Tenants = join(\",\", var.tenant_prefixes)`) so it has at least one consumer.",
      "confidence": 0.7
    },
    {
      "id": "F019",
      "title": "hosted_zone_name variable defined but not consumed in active code",
      "severity": "LOW",
      "category": "code-quality",
      "file": "infrastructure/terraform/freeside-storage/variables.tf",
      "description": "`hosted_zone_name` is referenced only in the commented-out route53.tf block. It's a no-op input today.",
      "suggestion": "Either uncomment route53.tf with import shims, or remove the variable until the future cycle that takes ownership of DNS records.",
      "confidence": 0.7
    }
  ]
}
```


## Callouts

zkSoju:
## 🛠 bridgebuilder address — round 1

Pushed `cb6f18be`. Resolved 7 findings; deferred 4 with explicit rationale.

### ✅ resolved in this commit

| ID | sev | conf | resolution |
|---|---|---|---|
| **F002** | M | 0.7  | Cert ARN → `data.aws_acm_certificate.wildcard_0xhoneyjar.arn` (survives rotation; removes literal account-ID) |
| **F001** | H | 0.6  | Added explicit provider-mapping rationale comment (T11 apply verified the mapping works; comment makes the assumption legible) |
| **F009** | L | 0.85 | gaib-cli registry-root: candidate-chain resolution (cwd → sibling repo → legacy `~/Documents/GitHub/...`); now portable to Windows / CI |
| **F011** | L | 0.55 | cdn-cors-smoke gained OPTIONS preflight probe + ACAO `*`-or-origin assertion (rejects `null`/`false` misconfigs) |
| **F010** | L | 0.5  | cdn-parity ETag caveat documented inline (why our topology has reliable ETag-based comparison) |
| **F017** | L | 0.6  | Comment on `aws_acm_certificate.alias` lifecycle noting it's moot in production (count=0 when existing_acm_certificate_arn is set) |
| **F018** | L | 0.7  | `tenant_prefixes` variable removed — replaced by inline comment pointing to the URL contract schema |
| **F019** | L | 0.7  | `hosted_zone_name` reframed as RESERVED for the DNS-import cycle (kept rather than removed-and-re-added) |
| **F012** | L | 0.75 | NOTE comment on the *.webp cache behavior calling out the future migration to managed cache policies (DEFERRED — coordinated swap needed) |

### ⏸ deferred — explicit rationale

| ID | sev | reason |
|---|---|---|
| **F003** | H | Bucket policy codify — works in practice (`thj-assets` has `Principal:*`); risky to codify without operator review of the existing policy. Worth a tracking bead/issue. |
| **F004** | M | DNS drift detection — its own cycle (per `freeside-dns-state-untracked` doctrine). |
| **F005-F008** | M-L | gaib-cli correctness (parseDotenv silent-drop, putJsonSecret CAS, partial-key push, AWS error classification) — bundle with the gaib-cli **secret-shape reconciliation cycle**. The current implementation assumed single-JSON-per-world; reality is per-key entries at `arrakis/{env}/worlds/{slug}/{key}` (per `scripts/load-honeyroad-secrets.sh`). A reshape PR addresses these naturally. |
| **F016** | L | Bucket-prefix audit — future cycle. |

### ✨ acknowledged PRAISE

- **F013** — cdn-cors-smoke + cdn-parity hardening pair (acknowledged; the round-1 OPTIONS additions extend it)
- **F014** — gaib-cli secret value handling (stdout/stderr separation, audit count-only)
- **F015** — ADR-006 scope discipline

### verified

```bash
$ bash scripts/cdn-cors-smoke.sh
🌐 cdn-cors-smoke · 5 passed (GET ACAO=* + OPTIONS ACAO=* on every path)

$ bash scripts/cdn-parity.sh
📐 cdn-parity · 5 identical · 0 drifted

$ pnpm build (gaib-cli)
packages/gaib-cli build: tsc → done
```

PR ready for re-review.

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#193. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "push overwrites entire secret — partial dotenv file silently deletes keys",
      "severity": "HIGH",
      "category": "data-loss",
      "file": "packages/gaib-cli/src/lib/aws.ts",
      "description": "putJsonSecret calls PutSecretValueCommand with SecretString = JSON.stringify(value), which REPLACES the entire secret value. The push command (push.ts step 4) only validates that the file contains no keys ABSENT from the registry — it never checks that the file is COMPLETE. If an operator pushes a partial file (e.g. `.env.rotated` containing only the one rotated key, exactly the workflow the README and command examples suggest for rotation), every other key in the live secret is silently dropped. AWS keeps the prior version, but the active version now lacks those keys, breaking the world at runtime with no warning.",
      "suggestion": "Either (a) read-modify-write: getJsonSecret first, merge the file's keys over the existing map, then put; or (b) require the pushed file to contain all registry-declared keys and refuse otherwise; or (c) emit a loud diff (`gaib: push will REMOVE N key(s): ...`) and gate behind --confirm even for staging. The current full-overwrite semantics are surprising for a 'rotate one secret' flow.",
      "confidence": 0.8
    },
    {
      "id": "F002",
      "title": "metadata-resolver relies on path-shape as an implicit host selector; asset paths can collide",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "infrastructure/terraform/freeside-storage/metadata-resolver.js",
      "description": "The function is attached to the default cache behavior, which serves BOTH assets.0xhoneyjar.xyz and metadata.0xhoneyjar.xyz traffic (single distribution, two aliases). The only discriminator is the regex /^\\/([a-z][a-z0-9-]*)\\/(\\d+)$/. The comment asserts assets requests 'rarely match' — but 'rarely' is not 'never'. Any asset request of shape /{lowercase-slug}/{digits} (e.g. /sprawl/123) would be rewritten to a metadata path and, absent a KV pointer, return 404 instead of passing through to the real asset. request.headers.host is available and would make the gate explicit and safe.",
      "suggestion": "Add an explicit host check (only rewrite when host === 'metadata.0xhoneyjar.xyz') rather than inferring intent from path shape. This removes the fragile coupling and the 'rarely' hedge.",
      "confidence": 0.6
    },
    {
      "id": "F003",
      "title": "Disabling metadata_resolver destroys the KeyValueStore and all pointer state",
      "severity": "MEDIUM",
      "category": "operational-risk",
      "file": "infrastructure/terraform/freeside-storage/variables.tf",
      "description": "count/for_each on aws_cloudfront_key_value_store.metadata_pointers and aws_cloudfrontkeyvaluestore_key.initial_pointers are gated on metadata_resolver_enabled. Setting the flag false plans to DESTROY the KV store and every pointer key — including any pointers ops have updated post-bootstrap via put-key (which are NOT in terraform state and thus unrecoverable). This is acknowledged in the variable comment (Bridgebuilder F4), but the destructive toggle remains live in the shipped module.",
      "suggestion": "Decouple KV-store provisioning from the enable flag (always provision the store, gate only the function_association + aliases), so a disable is non-destructive. Until then, the acknowledgement is good but the footgun is armed in production config.",
      "confidence": 0.7
    },
    {
      "id": "F004",
      "title": "Registry-root candidate comment claims 5 sources; array holds 3",
      "severity": "LOW",
      "file": "packages/gaib-cli/src/lib/world-resolver.ts",
      "description": "The docblock enumerates a 5-entry resolution order (explicit flag, env var, cwd, sibling repo, legacy default), but DEFAULT_REGISTRY_CANDIDATES contains only the last 3. Entries 1 and 2 are handled separately in resolveWorld (opts.registryRoot / FREESIDE_REGISTRY_ROOT), so behavior is correct, but the numbered comment sitting directly above the 3-element array invites a maintainer to 'fix' the apparent mismatch.",
      "suggestion": "Split the comment: document flag/env-var resolution at the resolveWorld call site and keep only the 3 filesystem candidates documented above the array.",
      "confidence": 0.6
    },
    {
      "id": "F005",
      "title": "CloudFront Function code-size limit stated as 2 KB (actual limit is 10 KB)",
      "severity": "LOW",
      "category": "documentation",
      "file": "infrastructure/terraform/freeside-storage/metadata-resolver.js",
      "description": "The constraints comment states '2 KB max compiled code size'. CloudFront Functions allow up to 10 KB of function code. The inaccuracy is harmless here but could mislead a future author into unnecessary golfing or false alarm.",
      "suggestion": "Correct to 10 KB (and note the ~1 MB memory / 10ms constraints if listing limits).",
      "confidence": 0.6
    },
    {
      "id": "F006",
      "title": "CORS smoke OPTIONS probe cannot exercise a real preflight (behavior allows only GET/HEAD)",
      "severity": "LOW",
      "category": "testing",
      "file": "scripts/cdn-cors-smoke.sh",
      "description": "The cache behaviors set allowed_methods = [\"GET\",\"HEAD\"], so an OPTIONS request is rejected by CloudFront (405/403) before reaching origin. The response-headers policy may still stamp ACAO onto that error response, so the OPTIONS assertion can pass without a functioning preflight path. Canvas consumers use simple requests (no preflight), so this is not a production gap — but the OPTIONS probe gives false assurance that non-simple CORS works.",
      "suggestion": "Either drop the OPTIONS probe (document that only simple CORS is in scope) or assert the OPTIONS status explicitly and treat the known 405/403 as expected, so the test reflects reality.",
      "confidence": 0.5
    },
    {
      "id": "F007",
      "title": "parseDotenv unescape gate reads pre-trim raw slice; leading space defeats quote detection",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/gaib-cli/src/lib/format.ts",
      "description": "Unescaping of double-quoted values is gated on line.slice(eq + 1).startsWith('\"'). value itself is line.slice(eq+1) without trimming, so an input like `KEY= \"multi\\nline\"` (space after =) both fails the startsWith('\"') check and retains the leading space, so \\n escapes are not decoded. This is an edge case (dotenv files rarely have space-after-equals), and renderDotenv never emits it, but round-tripping an externally-authored file could silently mishandle it.",
      "suggestion": "Trim value before the quote-strip/unescape logic, and derive the 'was quoted' flag from the trimmed value rather than a second raw slice.",
      "confidence": 0.5
    },
    {
      "id": "F008",
      "title": "cdn-parity ETag equality may be undermined by CloudFront compression",
      "severity": "LOW",
      "category": "correctness",
      "file": "scripts/cdn-parity.sh",
      "description": "Both behaviors set compress = true. When CloudFront compresses a response it can return a modified/weak ETag versus the raw S3 ETag, and the two distributions could differ if one serves a cached compressed variant and the other doesn't. The sampled paths are *.webp (already-compressed, so CloudFront won't gzip them), so the current sample is safe — but the script's stated invariant ('ETag drift == different bytes') is not universally true under compression and would mislead if the path set is expanded to compressible types.",
      "suggestion": "Note the compression caveat in the script header, or send `Accept-Encoding: identity` on the probes to normalize, so a future maintainer extending PATHS to non-image types doesn't get spurious drift.",
      "confidence": 0.4
    },
    {
      "id": "F009",
      "title": "Strong exit-code taxonomy and AWS error classification",
      "severity": "PRAISE",
      "category": "error-handling",
      "file": "packages/gaib-cli/src/lib/aws.ts",
      "description": "classifyAwsError maps SDK error names and HTTP status onto a stable, documented exit-code taxonomy (auth vs retryable-transient vs schema), and the exit codes are surfaced consistently in README/help. This makes the CLI scriptable and CI-gateable with predictable semantics — genuinely good CLI ergonomics.",
      "suggestion": "No change needed."
    },
    {
      "id": "F010",
      "title": "Version-pointer allowlist before URI interpolation",
      "severity": "PRAISE",
      "category": "security",
      "file": "infrastructure/terraform/freeside-storage/metadata-resolver.js",
      "description": "VERSION_PATTERN (/^[A-Za-z0-9._-]+$/) is enforced on the KV value before it is interpolated into the rewritten URI, with a defensive 404 on empty/malformed pointers. This closes a path-injection vector even though the KV store is operator-controlled — correct defense-in-depth for edge code that cannot be easily hotfixed.",
      "suggestion": "No change needed."
    },
    {
      "id": "F011",
      "title": "ACM cert resolved via data source instead of hardcoded ARN",
      "severity": "PRAISE",
      "file": "infrastructure/terraform/freeside-storage.tf",
      "description": "The wildcard cert is looked up via data.aws_acm_certificate (domain + ISSUED + most_recent) rather than pinning a literal ARN, so cert rotation won't silently break the stack. Combined with the existing_acm_certificate_arn escape hatch and the documented CAA-cache rationale, this is a well-reasoned handling of a real operational hazard.",
      "suggestion": "No change needed."
    }
  ]
}
```


## Callouts

PR #210, opened 2026-05-19
feat(shared/adr-007-scaffold): workspace dirs + firewall enforcement (ADR-007 Step 3)

zkSoju:
## Summary

ADR-007 §Implementation step 3: creates the dual-concern workspace structure AND lands the firewall enforcement that the absorption needs. Per the tightening from flatline SKP-001 CRITICAL, enforcement cannot be deferred — it lands with the workspace dirs, not later.

## What lands

**6 new directories (READMEs only, no application code):**

| Path | Domain | Forthcoming content |
|------|--------|---------------------|
| `apps/mcp-gateway/` | network | Absorbed `freeside-mcp-gateway/src/*` (PR step 4) |
| `packages/freeside-cli/` | network | `loa freeside doctor\|list\|inspect` verbs (step 8) |
| `packages/freeside-registry/` | network | Beacon aggregator + `/federation.json` server (step 7) |
| `packages/beacon-schema/` | network | BeaconV3 sealed schema (step 6) |
| `grimoires/freeside-platform/` | platform | Cycle artifacts for vertical-SaaS work |
| `grimoires/freeside-network/` | network | Cycle artifacts for ecosystem-parent work |

**CODEOWNERS** — loose co-ownership per operator's "1 may be best move" (2-person scale, CI is the load-bearing enforcement):
- Network paths: `@janitooor @zksoju`
- Platform paths: `@janitooor` (existing)
- Firewall workflows + `decisions/007`: protected

**3 CI workflows (the load-bearing enforcement):**

| Workflow | Purpose | Phase |
|----------|---------|-------|
| `path-domain-check.yml` | Fails PRs touching both platform AND network in one commit | Phase 1 (active) |
| `commit-scope-check.yml` | Validates `platform/<x>`, `network/<x>`, `shared/<x>` conventional-commit scopes | Phase 1 (warn-on-missing, fail-on-cross-domain) |
| `ledger-domain-check.yml` | Validates `domain` field on cycle ledger + `domain:*` label on beads + no cross-domain `blocked-by` deps | Phase 1 (grandfather existing entries) |

Strict mode envs available for phase-2 escalation:
- `COMMIT_SCOPE_STRICT=true` — fail on any missing scope
- `LEDGER_DOMAIN_STRICT=true` — fail on any missing domain field

**Pre-commit hook stub:** `tools/check-beacon-domain.sh` mirrors `path-domain-check.yml` for local pre-push validation. Install via symlink.

## Known: this PR self-trips its own check (correct behavior)

The `path-domain-check.yml` will FAIL on this PR because the bootstrap genuinely creates both domains' scaffold. **This is correct** — the check is doing its job. The PR uses admin bypass for this one-time exception. The failing check is positive evidence the enforcement works as designed.

Self-test output (local hook):

```
ERROR: Cross-domain changes detected (ADR-007 §D-3 violation)

Platform paths:
  grimoires/freeside-platform/README.md

Network paths:
  apps/mcp-gateway/README.md
  grimoires/freeside-network/README.md
  packages/beacon-schema/README.md
  packages/freeside-cli/README.md
  packages/freeside-registry/README.md
```

After this bootstrap merges, the check will fire on any subsequent PR that mixes domains.

## Why this matters (the SKP-001 fix recap)

Flatline review of ADR-007 flagged SKP-001 as **CRITICAL** (severity 860): "If new directories, package moves, and gateway absorption land before CODEOWNERS, PR checks, commit scope validation, and ledger-domain validation exist, early PRs can establish exactly the cross-domain coupling the ADR is meant to prevent."

The fix was to land enforcement IN the same PR as workspace creation (not a follow-up cycle). This PR honors that commitment.

## Forward dependencies (next 7 PRs unblocked after this merges)

Per ADR-007 §Implementation Sequencing:

- Step 4: PR-absorb-gateway (`git mv` `freeside-mcp-gateway/src/*`)
- Step 5: PR-construct-freeside-paths
- Step 6: PR-beacon-v3 (schema bump with normative validators per Appendix A)
- Step 7: PR-federation-endpoint (`/federation.json` with D-8 auth model)
- Step 8: PR-cli-verbs (`loa freeside doctor|list|inspect`)
- Step 9: PR-ambient-hook (operator-local SessionStart)
- Step 10: PR-source-repo-archival

## Test plan

- [ ] `tools/check-beacon-domain.sh --since main` self-trip output matches expected (proves hook works)
- [ ] CI runs all 3 new workflows on this PR
- [ ] Verify `path-domain-check` fails (expected — bootstrap is cross-domain)
- [ ] Verify `commit-scope-check` passes (commit is `chore(shared/adr-007-scaffold)`)
- [ ] Verify `ledger-domain-check` skips (no .beads or ledger changes)
- [ ] Sync with @janitooor on CODEOWNERS additions before merge
- [ ] Merge with admin bypass (one-time bootstrap exception)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -67,3 +67,27 @@ tools/sprint-kind-classify.py                       @janitooor
 tools/advisor-benchmark*.sh                         @janitooor
 tools/advisor-benchmark-stats.py                    @janitooor
 tools/modelinv-coverage-audit.py                    @janitooor
+
+# ─────────────────────────────────────────────────────────────────────────────
+# RFC #207 — loa-freeside as dual-concern ecosystem parent
+# ADR-007 §D-3 — workspace firewall enforcement
+# ─────────────────────────────────────────────────────────────────────────────
+# Network domain (ecosystem-parent surface) — both owners review
+# Per ADR-007 §D-3, CODEOWNERS is convention; CI enforces the hard rule.
+# At current 2-person scale, loose co-ownership keeps review friction low.
+apps/mcp-gateway/                                   @janitooor @zksoju
+packages/freeside-cli/                              @janitooor @zksoju
+packages/freeside-registry/                         @janitooor @zksoju
+packages/beacon-schema/                             @janitooor @zksoju
+grimoires/freeside-network/                         @janitooor @zksoju
+
+# Platform domain (vertical-SaaS substrate) — Jani primary, inherits root
+# Listed explicitly so an auditor can confirm the firewall split is intentional.
+grimoires/freeside-platform/                        @janitooor
+
+# Workspace firewall enforcement (don't bypass without ADR amendment)
+.github/workflows/path-domain-check.yml             @janitooor
+.github/workflows/commit-scope-check.yml            @janitooor
+.github/workflows/ledger-domain-check.yml           @janitooor
+tools/check-beacon-domain.sh                        @janitooor
+decisions/007-loa-freeside-absorption.md            @janitooor @zksoju
--- a/.github/workflows/commit-scope-check.yml
+++ b/.github/workflows/commit-scope-check.yml
@@ -1,0 +1,84 @@
+# Commit Scope Check — ADR-007 §D-3 conventional-commit discipline
+#
+# Validates that PR commits use the platform/<x>, network/<x>, or shared/<x>
+# scope convention. Phased rollout: warn on missing scope today, escalate
+# to fail in a future cycle once the convention is established.
+#
+# Reference: decisions/007-loa-freeside-absorption.md §D-3 + Implementation step 3
+
+name: Commit Scope Check (ADR-007 §D-3)
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check commit scopes
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+
+          # Phased enforcement (per ADR-007 §D-3):
+          #   PHASE 1 (current): warn on missing scope, hard-fail only on
+          #                      explicit cross-domain scope mismatch.
+          #   PHASE 2 (future):  hard-fail on any missing scope.
+          # Bump STRICT to enable phase 2.
+          STRICT="${COMMIT_SCOPE_STRICT:-false}"
+
+          warnings=0
+          errors=0
+
+          while IFS= read -r sha; do
+            [ -z "$sha" ] && continue
+            subject=$(git log -1 --format=%s "$sha")
+
+            # Exempt patterns
+            case "$subject" in
+              "Merge "*|"Revert "*) continue ;;
+              "chore(deps)"*|"chore(deps-dev)"*) continue ;;
+              "chore(release)"*) continue ;;
+              "build(deps)"*|"build(deps-dev)"*) continue ;;
+            esac
+
+            # Strict ADR scope pattern: type(platform/x), type(network/x), type(shared/x)
+            if echo "$subject" | grep -qE '^[a-z]+\((platform|network|shared)/[a-z0-9-]+\)(!)?: '; then
+              continue  # explicit ADR scope — pass
+            fi
+
+            # Loose pattern: any type(scope): subject — pass but warn during phase 1
+            if echo "$subject" | grep -qE '^[a-z]+\([a-z0-9-]+(/[a-z0-9-]+)?\)(!)?: '; then
+              echo "::warning::Commit $(echo "$sha" | cut -c1-8) uses non-ADR scope: $subject"
+              echo "  Suggestion: use platform/<x>, network/<x>, or shared/<x> per ADR-007 §D-3"
+              ((warnings++)) || true
+              continue
+            fi
+
+            # Bare commit (no scope at all)
+            if [ "$STRICT" = "true" ]; then
+              echo "::error::Commit $(echo "$sha" | cut -c1-8) missing scope: $subject"
+              ((errors++)) || true
+            else
+              echo "::warning::Commit $(echo "$sha" | cut -c1-8) missing scope: $subject"
+              ((warnings++)) || true
+            fi
+          done < <(git rev-list "${BASE_SHA}..${HEAD_SHA}")
+
+          echo ""
+          echo "Summary: ${errors} errors, ${warnings} warnings"
+
+          if [ "$errors" -gt 0 ]; then
+            echo "::error::STRICT mode rejected ${errors} commits."
+            exit 1
+          fi
+
+          echo "✓ Commit scopes acceptable (phase 1)"
--- a/.github/workflows/commit-scope-check.yml
+++ b/.github/workflows/commit-scope-check.yml
@@ -59,17 +59,17 @@ jobs:
             if echo "$subject" | grep -qE '^[a-z]+\([a-z0-9-]+(/[a-z0-9-]+)?\)(!)?: '; then
               echo "::warning::Commit $(echo "$sha" | cut -c1-8) uses non-ADR scope: $subject"
               echo "  Suggestion: use platform/<x>, network/<x>, or shared/<x> per ADR-007 §D-3"
-              ((warnings++)) || true
+              warnings=$((warnings + 1))
               continue
             fi
 
             # Bare commit (no scope at all)
             if [ "$STRICT" = "true" ]; then
               echo "::error::Commit $(echo "$sha" | cut -c1-8) missing scope: $subject"
-              ((errors++)) || true
+              errors=$((errors + 1))
             else
               echo "::warning::Commit $(echo "$sha" | cut -c1-8) missing scope: $subject"
-              ((warnings++)) || true
+              warnings=$((warnings + 1))
             fi
           done < <(git rev-list "${BASE_SHA}..${HEAD_SHA}")
 
--- a/.github/workflows/ledger-domain-check.yml
+++ b/.github/workflows/ledger-domain-check.yml
@@ -1,0 +1,159 @@
+# Ledger Domain Check — ADR-007 §D-3 hard ledger validation
+#
+# Validates that:
+#   1. Cycle ledger entries (grimoires/loa/ledger.json) carry a `domain` field
+#   2. Beads issues (.beads/issues.jsonl) carry a `domain:*` label
+#   3. No cross-domain `blocked-by` dependencies exist (platform ↔ network)
+#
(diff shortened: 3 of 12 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#210. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Workspace firewall enforcement lands with scaffolding",
      "severity": "PRAISE",
      "category": "process",
      "file": ".github/workflows/path-domain-check.yml",
      "description": "Path domain check CI ships in the same PR as the workspace directories themselves, directly addressing the flatline SKP-001 CRITICAL concern that enforcement couldn't be deferred. The check covers both platform and network path globs comprehensively and matches the allowlist documented in ADR-007 §D-1.",
      "suggestion": "No action — this is the right ordering.",
      "confidence": 0.95
    },
    {
      "id": "F002",
      "title": "Path globs duplicated between CI workflow and local pre-commit script",
      "severity": "MEDIUM",
      "file": "tools/check-beacon-domain.sh",
      "description": "The platform/network path allowlist is duplicated verbatim between `.github/workflows/path-domain-check.yml` (lines 41-62) and `tools/check-beacon-domain.sh` (lines 64-86). When a new package is added to either domain, both files must be updated in lockstep. Drift between them means local pre-commit will permit what CI rejects (or vice versa), which silently undermines the firewall.",
      "suggestion": "Extract the path classification logic to a single shared script (e.g., `tools/classify-domain.sh <file>` or a YAML/JSON manifest like `.github/domain-paths.yaml`) and have both the workflow and pre-commit script source it. Alternatively, have the workflow invoke `tools/check-beacon-domain.sh --since` directly so there's exactly one definition.",
      "confidence": 0.9
    },
    {
      "id": "F003",
      "title": "ledger-domain-check uses array indexing on JSONL stream",
      "severity": "HIGH",
      "category": "correctness",
      "file": ".github/workflows/ledger-domain-check.yml",
      "description": "In the 'Check no cross-domain blocked-by dependencies' step, `.beads/issues.jsonl` is processed with `jq -s 'map(...)'` (slurp), which works. However, the violation detection then uses `$all | map(select(.id == $bid)) | first` for every blocker. With N issues each having M blockers, this is O(N·M·N). For small N this is fine, but more importantly the `unique` filter at the end is applied to a string array result — but `jq -r '...' | unique` doesn't exist as written: `unique` is a jq filter, and the output is then compared to the literal string `[]` and also passed through `jq 'length'`. The flow produces raw strings from `jq -r`, not a JSON array, so `[ \"$violations\" != \"[]\" ]` will never match `[]` even when empty (the `|| echo \"[]\"` only fires on jq error), and `echo \"$violations\" | jq 'length'` will fail to parse multi-line raw strings as JSON.",
      "suggestion": "Either keep output as JSON throughout (drop `-r`, then `jq 'length'` works) or switch the gate to `if [ -n \"$violations\" ]` and count with `wc -l`. As written, the cross-domain dependency check is likely to either error out under jq or never trigger correctly.",
      "confidence": 0.85
    },
    {
      "id": "F004",
      "title": "jq filter for cycle ledger 'domain == null' misses missing-field case",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": ".github/workflows/ledger-domain-check.yml",
      "description": "The filter `select(.value.domain == null)` matches entries where domain is explicitly null OR where the key is absent (jq returns null for missing keys), so this is correct. However, the array branch `.[] | select(.domain == null) | .id // .name // \"?\"` will emit the string `\"?\"` for entries lacking both id and name, which then gets counted by `wc -l` as a missing entry but produces an undebuggable warning. Minor but worth tightening.",
      "suggestion": "Include the array index in the fallback: `.id // .name // (\"index-\" + (input_line_number|tostring))` or similar, so operators can locate the offending entry.",
      "confidence": 0.6
    },
    {
      "id": "F005",
      "title": "commit-scope-check uses `((var++))` under `set -e` — exits on first warning",
      "severity": "HIGH",
      "category": "correctness",
      "file": ".github/workflows/commit-scope-check.yml",
      "description": "Under `set -euo pipefail`, `((warnings++))` returns exit code 1 when `warnings` was 0 (post-increment evaluates to old value). The author wrapped these in `|| true` which mitigates the issue, but this is fragile: if anyone reorders or removes the `|| true`, the loop silently exits at the first warning. The same pattern appears for `((errors++))`.",
      "suggestion": "Use `warnings=$((warnings + 1))` instead of `((warnings++))` — that form returns the new value and never exits 1. This removes the need for `|| true` and is more robust.",
      "confidence": 0.95
    },
    {
      "id": "F006",
      "title": "Path check is decoupled from the actual workspace dirs (untestable scaffold)",
      "severity": "LOW",
      "category": "testability",
      "file": ".github/workflows/path-domain-check.yml",
      "description": "The path-domain-check workflow runs on this PR, but this PR touches only README scaffolds under `grimoires/freeside-network/`, `grimoires/freeside-platform/`, `apps/mcp-gateway/`, `packages/freeside-*`. It DOES touch both platform (`grimoires/freeside-platform/README.md`) and network (`grimoires/freeside-network/README.md`, `apps/mcp-gateway/README.md`, `packages/freeside-*/README.md`) paths simultaneously — which means this very PR should fail its own newly-introduced check.",
      "suggestion": "Verify: does the workflow actually run on its own PR (it should, since it's a `.github/workflows/` add)? If yes, either (a) confirm the workflow passes only because it can't see itself yet (race condition with merge), or (b) carve out a bootstrap exception for the scaffold PR. If it currently fails, document the bypass. If it currently passes, there's a bug in the path matching.",
      "confidence": 0.85
    },
    {
      "id": "F007",
      "title": "Bash array `+=()` with empty array under `set -u` is safe but worth noting",
      "severity": "LOW",
      "category": "correctness",
      "file": ".github/workflows/path-domain-check.yml",
      "description": "`printf '  - %s\\n' \"${platform_files[@]}\"` is invoked only inside `if $platform_touched && $network_touched`, so the array is guaranteed non-empty there. The success-branch lines `printf '%s ' \"${platform_files[@]}\" | wc -w` may execute when the array is empty under `set -u` (since `platform_touched=false` means the array stays empty, but the branch is gated on `$platform_touched`, so it's fine). Reading more carefully, the gating is correct. No fix needed.",
      "suggestion": "No action.",
      "confidence": 0.8
    },
    {
      "id": "F008",
      "title": "CODEOWNERS firewall files self-owned by single reviewer",
      "severity": "MEDIUM",
      "category": "security",
      "file": ".github/CODEOWNERS",
      "description": "The three firewall enforcement workflows and `tools/check-beacon-domain.sh` are owned solely by `@janitooor`. If `@janitooor` opens a PR weakening or removing the firewall, only they can approve it (assuming branch protection requires CODEOWNERS review). This is a single-key risk on the load-bearing CI guard. Compare to `decisions/007-loa-freeside-absorption.md` which correctly requires both `@janitooor @zksoju`.",
      "suggestion": "Add `@zksoju` as co-owner on the three firewall workflow files and `tools/check-beacon-domain.sh`, matching the dual-ownership on the ADR itself. The firewall machinery is symmetric in importance to the ADR.",
      "confidence": 0.85
    },
    {
      "id": "F009",
      "title": "Path matching uses `case` globs, not anchored prefixes — false positives possible",
      "severity": "LOW",
      "category": "correctness",
      "file": ".github/workflows/path-domain-check.yml",
      "description": "`case \"$f\" in apps/mcp-gateway/*)` matches files starting with `apps/mcp-gateway/`. Since git diff produces paths from repo root, this is correct. However, the patterns don't account for renamed files (git diff --name-only with default options does emit just the new path, but `--name-status` would show both). For now this is fine; just worth noting that a `git mv apps/gateway/x apps/mcp-gateway/x` rename in one PR would correctly classify as network-only, but if the rename touched cross-domain pairs it'd be detected.",
      "suggestion": "No action; document in a comment that the check assumes `git diff --name-only` (post-rename paths only).",
      "confidence": 0.4
    },
    {
      "id": "F010",
      "title": "Phased rollout via env-var toggle is clean and reversible",
      "severity": "PRAISE",
      "category": "process",
      "file": ".github/workflows/ledger-domain-check.yml",
      "description": "Both `LEDGER_DOMAIN_STRICT` and `COMMIT_SCOPE_STRICT` env-var toggles enable phase-2 hard-fail without code changes. This is the right primitive: turn the knob in repo settings when ready, no PR needed. The warn-now/fail-later pattern correctly grandfathers existing data.",
      "suggestion": "No action.",
      "confidence": 0.9
    },
    {
      "id": "F011",
      "title": "Beacon-schema README documents type-checkable Tag composition with hash-locked references",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/beacon-schema/README.md",
      "description": "The `TagName@version+schema_hash` qualification scheme (lines 39-46) addresses the flatline SKP-004 false-positive composition risk explicitly. Recomputing the schema hash from current Tag definitions catches upstream evolution that breaks composition without requiring exhaustive runtime tests.",
      "suggestion": "No action — solid design choice documented clearly.",
      "confidence": 0.85
    },
    {
      "id": "F012",
      "title": "Ledger-domain-check fails silently if jq isn't installed",
      "severity": "LOW",
      "category": "correctness",
      "file": ".github/workflows/ledger-domain-check.yml",
      "description": "The workflow assumes `jq` is on PATH on ubuntu-latest runners (it is, by default). However, the `2>/dev/null || echo \"\"` pattern in multiple places will swallow a missing-jq error and produce a false-pass. Low risk on GitHub's hosted runners, but if this workflow ever runs on a self-hosted minimal image, it'll silently green.",
      "suggestion": "Add an explicit `which jq` check at the top of each script step, or remove the `|| echo \"\"` swallowing.",
      "confidence": 0.6
    },
    {
      "id": "F013",
      "title": "Local pre-commit script doesn't validate scope/labels — only paths",
      "severity": "LOW",
      "category": "completeness",
      "file": "tools/check-beacon-domain.sh",
      "description": "The local pre-commit covers cross-domain path mixing but not commit-scope or beads-label validation. A developer can commit locally with a non-ADR scope or missing domain label and only discover the issue at PR time. Not a defect — the PR description doesn't claim full local parity — but worth noting for completeness.",
      "suggestion": "Consider extending the local hook (or adding sibling hooks) to cover commit-scope and beads-label checks, so the pre-push surface matches CI. Optional.",
      "confidence": 0.5
    },
    {
      "id": "F014",
      "title": "ledger-domain-check `unique` on raw text output",
      "severity": "HIGH",
      "category": "correctness",
      "file": ".github/workflows/ledger-domain-check.yml",
      "description": "Restating F003 more precisely: the cross-domain jq pipeline ends with `... | unique` but the surrounding `jq -r` produces raw strings. `unique` operates on JSON arrays. Inside the jq program this is fine (operates on the constructed array), but the outer `echo \"$violations\" | jq 'length'` then receives raw text (newline-separated strings), not JSON, and will error. Combined with the swallowed error from `|| echo \"[]\"`, this means cross-domain violations could be silently missed.",
      "suggestion": "Rewrite the final step to either keep JSON throughout: `violations=$(jq -s '...build array...' .beads/issues.jsonl); count=$(echo \"$violations\" | jq 'length'); if [ \"$count\" -gt 0 ]; then echo \"$violations\" | jq -r '.[]'; exit 1; fi`. Test with a deliberately-violating fixture before merging.",
      "confidence": 0.85
    }
  ]
}
```


## Callouts

PR #217, opened 2026-05-20
feat(network): close the freeside discovery loop — cycle-049 (S1-S4)

zkSoju:
## cycle-049 `discovery-loop` — consolidated PR (Sprints 1-4)

Wires the `freeside-*` building discovery loop end-to-end — **building declares → registry aggregates → endpoint serves → agent queries** — proven against a `freeside-score` fixture beacon.

**Domain:** network (ADR-007 §D-3) — zero platform paths in the diff.
**Autonomous run:** `/run sprint-plan` — 4 sprints, each through implement → review → audit.

### Per-sprint breakdown

| Sprint | Plane | Delivered | Impl commit |
|--------|-------|-----------|-------------|
| S1 — BeaconV3 belt fields | P1 Contract | `produces`/`consumes` belt fields; `composes_with` removed; `@freeside/beacon-schema` 0.2.0→0.3.0; R-1 import reconcile; ADR-007 App. A reconciled | `1cb5c7af` |
| S2 — Registry manifest-builder | P2 Construct | `buildFreesideManifest`; fixture-aware path-safe `beacon-loader`; `freeside-score` fixture beacon + negative fixture | `cf31c4ab` |
| S3 — Gateway route + CLI | P3 Execution | `GET /internal/freeside.json` (operator-gated); `freeside-cli inspect`/`doctor` functional | `183ac677` |
| S4 — E2E + CI green | P3 + x-cut | FR-6 in-process E2E test; CI-green sweep | `4106dbdf` |

### Acceptance

- ✅ **FR-6 E2E**: the discovery loop is provably closed against the `freeside-score` fixture (prd.md:32 — the binding acceptance bar).
- ✅ All 9 PRD §8 acceptance criteria verified.
- ✅ **71/71 tests green** (beacon-schema 28, freeside-registry 11, freeside-cli 6, gateway 26); all 4 packages `tsc`-clean.
- ✅ `path-domain-check`: zero platform paths in the committed diff.

### Operator-authorized scope addition (D-S3-1)

The gateway could not `pnpm install` (pre-existing broken nested-`workspace:*` setup). The run HALTED at S3; the operator authorized the fix: `apps/mcp-gateway` converted to `file:` deps, nested `pnpm-workspace.yaml` removed, `Dockerfile` reworked for a repo-root build context.

### 🗑️ Deleted files

- `apps/mcp-gateway/pnpm-workspace.yaml` — the gateway is no longer a nested pnpm workspace (D-S3-1).

### ⚠️ Known limitations / follow-ups

- **Docker/Railway deploy unverified** — the `Dockerfile` rework is untested (no Docker in the run env; CI does not Docker-build). Before the next gateway deploy: set the Railway service **Root Directory** to the repo root and run a deploy smoke-test.
- Remote `beacon_url` fetching is out of scope (SDD §1.8) — the loop is fixture-driven by design.

### Decision log

Four documented deviations/fixes surfaced across S1-S4 — D-S2-1 (V2/V3 discrimination by decode-attempt, not `schema_version`), D-S3-1 (gateway env fix), D-S3-2 (`file:`-dep packaging), D-S4-1 (`freeside-cli` `main` path). Full rationale in `grimoires/loa/NOTES.md`.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) — autonomous `/run sprint-plan`. **Draft — human review required before merge.**


zkSoju (commented):
## Summary

This PR closes the freeside discovery loop end-to-end — BeaconV3 schema, registry aggregation, gateway route, CLI verbs, and an in-process E2E test that validates the full chain without sockets or external dependencies. The architecture shows genuine craft: dependency-injected loader, skip-not-crash resilience, and defense-in-depth path safety throughout. Two HIGH findings (F-01, F-02) block production deployment — the compiled path resolution and missing Dockerfile COPY will silently empty the manifest in any container build; the gateway will serve `modules: []` with no surfaced error.

---

**Architectural context**: The PR's core design — `BeaconLoader` as a pure path→result function, `buildFreesideManifest` as an aggregator that skips rather than crashes, and CLI verbs that inject their fetcher — gives this system testability at every layer without filesystem mocking. That inversion-of-control foundation is what makes F-01 and F-02 fixable without restructuring: the loader already accepts an explicit `registryRoot`, which means callers can correct the production root without touching the loader's logic. The compiled-path bug and the Dockerfile omission are the same conceptual error — a production artifact that implicitly references a path outside its own boundaries — wearing two different disguises.

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-01",
      "title": "DEFAULT_REGISTRY_ROOT resolves to dist/ in compiled output — fixture paths unreachable in production",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/freeside-registry/src/beacon-loader.ts",
      "description": "DEFAULT_REGISTRY_ROOT is computed as `join(dirname(fileURLToPath(import.meta.url)), '..')`. In source mode (tsx), `dirname` is `src/` and `..` reaches the package root — tests pass. In compiled form, the file is at `dist/src/beacon-loader.js`, so `dirname` is `dist/src/` and `..` resolves to `dist/`. The fixture `tests/fixtures/freeside-score.beacon.yaml` is at the package root level, not under `dist/`. In any production use of the compiled package the default `registryRoot` is wrong; `realpathSync(join('dist/', 'tests/fixtures/...'))` throws ENOENT, the loader returns `{ kind: 'error' }`, and `buildFreesideManifest` skips freeside-score. The gateway serves an empty `modules: []` manifest without surfacing the root cause.",
      "suggestion": "Change the offset from `'..'` to `'..', '..'` (two levels up from `dist/src/`) so both source and compiled paths resolve to the package root:\n```ts\nconst DEFAULT_REGISTRY_ROOT = join(\n  dirname(fileURLToPath(import.meta.url)),\n  '..', '..'\n);\n```\nAlternatively expose a `createBeaconLoader(registryRoot: string): BeaconLoader` factory so callers supply the correct root explicitly and the default is never needed in production code.",
      "confidence": 0.97,
      "faang_parallel": "Google's abseil-py library and the broader Python packaging ecosystem hit this at scale: `__file__` paths in installed packages (site-packages/) diverge from source tree paths, breaking any path computed by offset from the module file. The Go toolchain's response was `embed.FS` — make the resource manifest explicit at compile time, eliminating runtime path computation entirely. TypeScript/ESM's `import.meta.url` has the same dual-life problem: source time vs compiled time are different locations.",
      "metaphor": "Computing paths relative to `import.meta.url` is like giving directions from 'my house' — perfect when you're at home, broken when the compiled output moved the 'starting location' two directories deeper without updating the turn-by-turn instructions.",
      "teachable_moment": "Any path computed as an offset from `import.meta.url` must account for the `src/` → `dist/src/` transformation in the compiled artifact. The safest rule: if a default path is used in production, verify it survives `tsc` output, not just `tsx` source. The factory-parameter alternative (callers own the root) eliminates the problem class entirely."
    },
    {
      "id": "F-02",
      "title": "Runtime Dockerfile stage omits tests/fixtures/ — fixture beacons absent from production container",
      "severity": "HIGH",
      "category": "correctness",
      "file": "apps/mcp-gateway/Dockerfile",
      "description": "The runtime stage copies only `packages/freeside-registry/package.json` and `dist/` for the registry package. The `tests/fixtures/` directory — which contains the in-repo beacon fixtures referenced by `registry.yaml` (`beacon_fixture: tests/fixtures/freeside-score.beacon.yaml`) — is never copied. Even if F-01 is fixed and `DEFAULT_REGISTRY_ROOT` correctly resolves to the package root, the fixture files would not be present in the running container. The gateway would serve an empty freeside manifest in production. The Dockerfile comment acknowledges the container build is untested this cycle, but there is no tracking marker or blocking TODO to prevent deployment before this is resolved.",
      "suggestion": "Add a COPY step in the runtime stage to bring the fixtures into the container:\n```dockerfile\nCOPY packages/freeside-registry/tests/fixtures ./packages/freeside-registry/tests/fixtures\n```\nFor production, consider moving registry-referenced fixture beacons out of the `tests/` tree into a first-class `data/` or `fixtures/` directory at the package root, making the production dependency explicit and not entangled with test scaffolding.",
      "confidence": 0.95,
      "faang_parallel": "Google's Hermetic Build mandate (Bazel, Nix) requires that build artifacts contain every file they need to run — no implicit filesystem dependencies outside the artifact boundary. This is enforced mechanically by sandboxed builds that fail if a file is accessed but not declared. Docker layers make the dependency semi-visible (missing COPY = missing file), but unlike Bazel, Docker doesn't fail at layer definition time — it fails at runtime, silently, when the read returns ENOENT.",
      "metaphor": "Shipping the container without the fixtures is like publishing a cookbook that says 'see the recipe cards in the kitchen drawer' — the reference is correct, but the drawer doesn't exist in the reader's kitchen.",
      "teachable_moment": "Files under `tests/` should never be runtime dependencies. When a `registry.yaml` points to `tests/fixtures/freeside-score.beacon.yaml`, the `tests/` path is broadcasting that this file's production status is ambiguous. Promoting it to `data/` or `fixtures/` at the package root makes the production intent explicit and ensures it gets included in any artifact build without special-casing."
    },
    {
      "id": "F-03",
      "title": "classifyBeacon silently discards V3 decode error — near-V3 beacons misclassified as legacy without diagnostics",
      "severity": "MEDIUM",
      "category": "observability",
      "file": "packages/freeside-registry/src/beacon-loader.ts",
      "description": "In `classifyBeacon`, the V3 decode attempt swallows all errors with an empty `catch {}` block before falling through to the V2 attempt. A beacon that is structurally V3 (has `is`, `is_not`, `cycle_state`, etc.) but has a single malformed field (e.g., `next_review` in a bad format, a belt name with uppercase chars, a tag missing the hash) will decode as BeaconV2 `legacy` rather than surfacing the specific V3 validation error. Doctor and the manifest builder would then skip or warn about a 'legacy' beacon when the operator's intent was V3 and the actual issue is a fixable field error.",
      "suggestion": "Preserve the V3 decode error and attach it to the `legacy` or `error` result when it might be actionable:\n```ts\nconst classifyBeacon = (parsed: unknown): BeaconLoadResult => {\n  let v3Error: string | undefined;\n  try {\n    return { kind: 'v3', beacon: decodeV3(parsed) };\n  } catch (err) {\n    v3Error = err instanceof Error ? err.message : String(err);\n  }\n  try {\n    return { kind: 'legacy', beacon: decodeV2(parsed), v3DecodeError: v3Error };\n  } catch (err) { ... }\n};\n```\nDoctor can then surface `v3DecodeError` as a `warn` alongside the `legacy` finding.",
      "confidence": 0.88,
      "faang_parallel": "Go's error wrapping (`fmt.Errorf` + `%w`) and Python's exception chaining (`__cause__`, `__context__`) both exist to solve this exact pattern. The Go team documented it as a 'ghost bug' source: classification chains that discard intermediate errors make operators see the final fallback label ('legacy', 'unknown') without understanding why the primary path failed. The idiom 'I caught this, but the real cause was that' is load-bearing diagnostic infrastructure.",
      "metaphor": "A doctor who runs a strep test, finds it inconclusive, and says 'probably just a cold' isn't wrong — but they discarded the specific reason the strep test failed. If the test kit was expired, or the sample was contaminated, that's actionable information the patient needed.",
      "teachable_moment": "In try/catch classification chains (try V3, fall back to V2, fall back to error), accumulate error context rather than discarding it. The first catch's message is frequently the diagnostic signal that distinguishes 'genuinely legacy' from 'structurally V3 but one field is wrong'. Surface it even if you don't act on it immediately — logging it is enough to make the operator's next debugging session orders of magnitude faster."
    },
    {
      "id": "F-04",
      "title": "httpManifestFetcher casts response to FreesideManifest without runtime validation",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/freeside-cli/src/verbs/inspect.ts",
      "description": "`return (await res.json()) as FreesideManifest` is a TypeScript-only cast with no runtime validation. If the gateway returns a malformed response (e.g., after a schema version bump, a partial failure, or a proxy returning HTML), `manifest.modules.some(...)` would throw at runtime with a confusing TypeError rather than a meaningful error. The operator would see an unhelpful stack trace instead of a clear 'unexpected manifest shape' message.",
      "suggestion": "Validate the response shape before consuming it. Given Effect is already a dependency, a lightweight decode is straightforward:\n```ts\nimport { FreesideManifestSchema } from '@freeside/freeside-registry';\nconst raw = await res.json();\nreturn Schema.decodeUnknownSync(FreesideManifestSchema)(raw);\n```\nAlternatively, a minimal structural check (`typeof body.version === 'number' && Array.isArray(body.modules)`) at minimum prevents the most confusing failure modes.",
      "confidence": 0.85,
      "faang_parallel": "Stripe's API client libraries validate every response shape against explicit schemas before exposing data to user code. Their internal post-mortem found that ~40% of API client bugs were type-cast violations that runtime validation would have caught at the boundary, surfacing a clear schema error instead of a downstream TypeError three stack frames later. TypeScript's type system is erased at runtime — `as SomeType` is a compile-time assertion, not a runtime guarantee.",
      "metaphor": "A TypeScript `as` cast is a notary stamp applied to an unsigned document — it asserts legal standing that the document doesn't actually have. The stamp satisfies the compiler, but if the underlying data challenges the assertion at runtime, reality wins.",
      "teachable_moment": "`as` in TypeScript means 'I promise the compiler this is true', not 'I have verified this is true'. At system boundaries — HTTP responses, `JSON.parse`, external APIs, IPC — always validate. Effect's `Schema.decodeUnknownSync` is already present in the codebase and adds zero new dependencies. The boundary between 'I control this data' and 'this data came from outside' is where `as` should never appear."
    },
    {
      "id": "F-05",
      "title": "--no-frozen-lockfile in production Dockerfile install steps weakens build reproducibility",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "apps/mcp-gateway/Dockerfile",
      "description": "All `pnpm install` steps in both builder and runtime stages use `--no-frozen-lockfile`. This flag allows pnpm to update the lockfile during the build, meaning a CI build and a production container build could resolve different transitive dependency versions depending on registry state at build time. The lockfile version bump (v6 → v9) explains why `--no-frozen-lockfile` was necessary to get the build working, but leaving it in production Dockerfile steps is a long-term reproducibility risk.",
      "suggestion": "Regenerate `pnpm-lock.yaml` with pnpm v9 so it reflects the v9 format, then restore `--frozen-lockfile` in the Dockerfile. The builder stage for the gateway specifically should use `--frozen-lockfile` once the lockfile is regenerated. The individual package `pnpm install` steps for beacon-schema and freeside-registry can use `--frozen-lockfile` or be replaced by a single workspace-aware install from the repo root.",
      "confidence": 0.8,
      "faang_parallel": "The 2021 ua-parser-js supply chain attack demonstrated the failure mode: a compromised npm package published a malicious version that builds without `--frozen-lockfile` silently pulled in. `--frozen-lockfile` turns that scenario into a build failure rather than a silent compromise — the lockfile didn't include the malicious version, so the build refuses to proceed. It's a supply chain tripwire.",
      "metaphor": "Building with `--no-frozen-lockfile` is like ordering 'the daily special' for production — you get what the registry felt like serving today. `--frozen-lockfile` is the fixed menu with a date stamp: same dish, same supplier, every time.",
      "teachable_moment": "Lockfile format migrations are legitimate reasons to temporarily use `--no-frozen-lockfile`, but the exit condition should be documented in the Dockerfile comment and tracked: regenerate the lockfile in the new format, verify CI passes, then restore `--frozen-lockfile`. The comment should say 'temporary: remove after regenerating lockfile for pnpm v9' with a link to a tracking issue."
    },
    {
      "id": "F-06",
      "title": "GET /internal/freeside.json rebuilds manifest on every request with synchronous disk I/O",
      "severity": "LOW",
      "category": "performance",
      "file": "apps/mcp-gateway/src/app.ts",
      "description": "`buildFreesideJson()` calls `loadRegistry()` (reads registry.yaml via `readFileSync`) and `loadBeacon()` for each module (reads beacon YAML fixtures via `readFileSync`) on every request. This is synchronous blocking I/O on the Node.js event loop. The existing MCP tenant manifest uses `BEACON_CACHE` with periodic refresh. The freeside manifest has no equivalent. For the current operator-gated, low-frequency use case this is acceptable, but the pattern diverges from how the rest of the gateway handles manifest data.",
      "suggestion": "Apply the same refresh-cache pattern already present for MCP tenant beacons: build the manifest once at startup, cache it, and refresh on a configurable interval. A simple module-level `let cachedManifest: FreesideManifest | null = null` with a `setInterval` refresh is sufficient and consistent with `beacon-cache.ts`.",
      "confidence": 0.75,
      "faang_parallel": "The 2016 npmjs.com registry outage was partly attributable to synchronous registry metadata rebuilds on the request path. Node.js's single-threaded event loop means one blocking `readFileSync` call serializes all concurrent requests behind it. The existing `BEACON_CACHE` in this codebase exists because this lesson was already learned for tenant beacons.",
      "metaphor": "Synchronous disk I/O on the event loop is like a checkout lane where the cashier reorganizes the back room on every transaction — the queue builds behind each customer waiting for the restocking to complete.",
      "teachable_moment": "Pattern consistency within a codebase is its own form of correctness. When `beacon-cache.ts` already establishes the right pattern for an analogous manifest, new endpoints should inherit it rather than diverging. The divergence isn't just a performance risk — it's a maintenance trap where a future engineer applies the caching pattern to one and misses the other."
    },
    {
      "id": "F-07",
      "title": "E2E test re-decodes an already-validated BeaconV3 — redundant and potentially misleading",
      "severity": "LOW",
      "category": "test-quality",
      "file": "apps/mcp-gateway/tests/discovery-loop.e2e.test.ts",
      "description": "`inspectModule` calls `loadBeacon` which internally calls `Schema.decodeUnknownSync(BeaconV3Schema)`. The E2E test then runs `Schema.decodeUnknown(BeaconV3Schema)(beacon)` again on the already-decoded value. The `Success` result proves only that a value that already passed schema validation can pass schema validation again, not that the full data flow is correct. The subsequent field assertions (`beacon.is.one_liner`, `beacon.produces.length`, `beacon.consumes[0].tag`) are the actual loop-closure proof.",
      "suggestion": "Remove the re-decode step and rely on the post-decode field assertions, which directly verify the data survived the full loop intact. Add a comment noting that `loadBeacon` already performed BeaconV3Schema validation if the intent is to make that explicit.",
      "confidence": 0.9,
      "faang_parallel": "Google's Test Certified program explicitly flags 'testing the framework' as a wasteful pattern. When a test validates that `JSON.parse(JSON.stringify(x)) === x`, it's testing JSON, not the system under test. The Effect Schema library's correctness is not this PR's problem to verify — the field assertions that follow the re-decode are the actual discovery-loop proof.",
      "metaphor": "Re-decoding an already-decoded value to verify it decodes is like photocopying a photocopy to verify the copy machine works. The original proof — that the copy machine produced a readable output — is already in your hand.",
      "teachable_moment": "In E2E tests, verify the behavior of your system, not the correctness of your dependencies. `loadBeacon` already validates schema — trust it, and prove the full loop is intact by asserting on field values that could only be correct if the entire chain (fixture → registry → gateway → HTTP → CLI → BeaconV3) worked as designed."
    },
    {
      "id": "F-08",
      "title": "auth.ts reads OPERATOR_API_KEY at module load time — test setup requires pre-import env assignment",
      "severity": "LOW",
      "category": "test-quality",
      "file": "apps/mcp-gateway/tests/discovery-loop.e2e.test.ts",
      "description": "The test sets `process.env.OPERATOR_API_KEY = 'test-operator-key'` at the top of the module (before the dynamic `import('../src/app.js')`) specifically because `auth.ts` captures the env var at module load. The comment explains this correctly, but the pattern creates a hidden dependency between test file structure and app bootstrap order. If any future refactor moves the dynamic import to a top-level static import, the env var would not be set in time and auth would silently fail with a different test error.",
      "suggestion": "Have `isAuthorizedOperator` (or `checkAccess`) read `process.env.OPERATOR_API_KEY` at call time rather than capturing it at module initialization. This eliminates the load-order dependency and makes the auth check more testable without pre-import side effects.",
      "confidence": 0.78,
      "faang_parallel": "AWS SDK v3 made this exact architectural change for credential resolution: moved from module-init capture to lazy resolution at first API call. The motivation was precisely this: module load order in Node.js is fragile, environment variable availability at import time is not guaranteed, and the testing tax (hidden ordering requirements) was measurable across their test suite.",
      "metaphor": "Reading a config at module load is like a security guard memorizing the approved visitor list at the start of their shift — if the list is updated mid-day, the guard doesn't know. Reading at call time is like checking the current list each time someone presents their badge.",
      "teachable_moment": "Configuration values from environment variables should be read at call time unless there's a specific performance reason to cache them at init, and that reason should be documented. The testing tax — pre-import env-set ordering requirements that are invisible to future refactors — almost always outweighs the microseconds saved by init-time capture."
    },
    {
      "id": "F-09",
      "title": "PRAISE: resolveFixturePath applies defense-in-depth path safety matching the L7 pattern",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/freeside-registry/src/beacon-loader.ts",
      "description": "`resolveFixturePath` applies three independent gates: `..` substring rejection before any filesystem access, `realpathSync` to defeat symlink traversal, and a `startsWith(realRoot + sep)` containment check as the final gate. This mirrors the L7 soul-identity-doc `realpath` + containment pattern and correctly orders the checks (string reject → filesystem-independent → filesystem-dependent).",
      "suggestion": null,
      "faang_parallel": "This is the canonical defense-in-depth path validation pattern. Cloudflare's WAF team documents it as the 'three-gate' approach: reject at the string layer (no expensive syscalls for obviously-bad input), reject at the filesystem-independent layer (symlink traversal requires no disk read), reject at the filesystem-dependent layer (canonical path containment). Applying all three in order makes each gate cheaper and the combination robust against bypass classes that defeat any single gate.",
      "teachable_moment": "The ordering of the three gates is not arbitrary — it's sorted by cost: string operations are cheapest, `realpathSync` resolves symlinks with one syscall, and containment check is a simple string comparison on the resolved result. This is security-conscious API design: expensive checks run only when cheap checks pass."
    },
    {
      "id": "F-10",
      "title": "PRAISE: buildFreesideManifest skip-not-crash resilience is correctly implemented and tested",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/freeside-registry/src/registry.ts",
      "description": "The manifest builder catches both loader throws and `error` results, emits a `console.warn`, and continues. Legacy V2 beacons are also skipped with a warn. The negative fixtures (`freeside-score-malformed.beacon.yaml`, `legacy-v2.beacon.yaml`) and corresponding tests verify that a single bad module does not crash or partially corrupt the aggregate manifest. The separation of 'module-level failure skips that module' from 'registry-level failure propagates' is the correct resilience boundary.",
      "suggestion": null,
      "faang_parallel": "This is the partial failure isolation pattern that Google's Chubby and Spanner teams call 'bulkhead': a single failing tenant doesn't take down the aggregate. The key design invariant — module-level failure skips that module, registry-level failure propagates — maps exactly to the bulkhead metaphor: compartments fail independently, the hull stays intact.",
      "teachable_moment": "The negative fixture tests are load-bearing: they verify the resilience boundary is actually implemented, not just intended. A `console.warn` path that is never triggered in tests is indistinguishable from dead code. These tests make the resilience contract executable."
    },
    {
      "id": "F-11",
      "title": "PRAISE: BeaconLoader injected as a dependency rather than imported — testability without filesystem mocking",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/freeside-registry/src/registry.ts",
      "description": "`buildFreesideManifest(registry, beaconLoader, visibilityFilter)` takes the loader as a parameter. The gateway, CLI, and tests each supply their own resolution strategy. The manifest.test.ts passes the real `loadBeacon` for integration coverage; a future test could inject a stub. This avoids the need for filesystem mocking or temp-dir fixture wiring in unit tests.",
      "suggestion": null,
      "faang_parallel": "This is dependency injection at the function level — what Michael Feathers calls 'seam injection' in Working Effectively with Legacy Code. By making the loader a parameter rather than a module-level import, the function signature becomes its own testability documentation: anything that conforms to the `BeaconLoader` signature can be injected without touching the module system.",
      "teachable_moment": "The alternative (module-level `import { loadBeacon }`) would require either filesystem fixtures or `jest.mock`/`vi.mock` at the module level — making test setup expensive and invisible. Parameter injection makes the dependency explicit and the seam cheap."
    },
    {
      "id": "F-12",
      "title": "PRAISE: FR-6 E2E test closes the full discovery loop in-process without network or sockets",
      "severity": "PRAISE",
      "category": "test-quality",
      "file": "apps/mcp-gateway/tests/discovery-loop.e2e.test.ts",
      "description": "The test exercises the complete chain — fixture beacon → registry aggregation → Hono app.request() → `/internal/freeside.json` → CLI `inspectModule` with injected fetcher → validated BeaconV3 — using `app.request()` (no listening socket) and injected `fetchManifest`. It provides deterministic, fast coverage of the full loop without any external dependencies. The explicit failure message on `assert.ok(score, ...)` names the exact invariant that broke, making CI failures actionable.",
      "suggestion": null,
      "faang_parallel": "Cloudflare Workers' test harness and Vercel's Edge Runtime test utilities both use this pattern: exercise the full request/response cycle in-process via a synthetic request, skipping network binding entirely. The result is E2E coverage at unit-test speed and determinism. The injected `fetchManifest` adapter is the exact seam that makes this possible.",
      "teachable_moment": "The combination of `app.request()` (Hono's in-process fetch) and an injected HTTP adapter in the CLI creates a test that exercises more real code than most integration tests — with none of the flakiness of actual network calls. This is the 'hexagonal architecture' payoff: ports (the fetcher interface) make the outside world replaceable in tests."
    },
    {
      "id": "F-13",
      "title": "PRAISE: DEVIATION D-S2-1 is documented at the source with actionable reasoning",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "packages/freeside-registry/src/beacon-loader.ts",
      "description": "The file-level comment explains why the SDD's specified V3 discrimination strategy (`schema_version: '3'`) is impossible given the actual schema constraint (`BeaconV3Schema extends BeaconV2Schema` which pins `schema_version: Literal('2')`), names the fixture that proves it, and describes the decode-attempt fallback used instead. The reference to `NOTES.md Decision Log D-S2-1` provides a cross-link for future engineers who encounter the deviation. This is the correct level of documentation for a deliberate SDD deviation.",
      "suggestion": null,
      "faang_parallel": "Linux kernel commit messages that explain 'why not the obvious approach' are the gold standard for this: Linus's rule that every non-obvious design choice needs its own paragraph explaining what was tried and why it failed. Without D-S2-1, the next engineer would spend hours trying to implement `schema_version: '3'` discrimination before discovering the same schema constraint.",
      "teachable_moment": "The most valuable comment is the one that explains why the obvious solution doesn't work. D-S2-1 prevents the 'someone will fix this the obvious way and break everything' failure mode that haunts codebases where deliberate decisions are undocumented. Cross-linking to NOTES.md gives the reader a path to the full context without embedding it in the source."
    }
  ]
}
```


---

**Architectural meditation**: F-01 and F-02 are the same error in two places. The system was designed correctly — the loader accepts an explicit `registryRoot`, the registry YAML is the authority on beacon locations, and the gateway wires them together. But neither the compiled path computation nor the Dockerfile COPY were validated against the production artifact. This is a testing-environment / production-environment gap: `tsx` runs source, containers run compiled output, and the fixtures are in neither `dist/` nor the container layer. The fix for both is the same conceptual move: make the production dependency explicit (factory parameter with no implicit default, `COPY` with no implicit path inference). The architecture is sound; the wiring between source truth and production artifact needs one verification pass.

---

## Callouts

- **`resolveFixturePath` (F-09)**: Three-gate path validation in the correct order — string rejection first, symlink resolution second, containment check last. This is the pattern. It matches L7's soul-identity-doc pattern exactly and applies it consistently in new code.

- **`buildFreesideManifest` resilience (F-10)**: Skip-not-crash with negative fixture tests is the right resilience contract. The test coverage of the failure path — not just the happy path — is what makes this production-grade rather than aspirationally-resilient.

- **Loader as parameter (F-11)**: `buildFreesideManifest(registry, beaconLoader, visibilityFilter)` is the cleaner design. The function signature documents its dependencies, test injection is free, and the gateway, CLI, and tests each wire their own strategy. This is dependency injection done at the minimum necessary scope.

- **FR-6 E2E test (F-12)**: In-process, deterministic, full-loop coverage with an injected HTTP adapter is the right test for a feature like this. It exercises more real code than most integration tests, runs at unit-test speed, and produces failure messages that name the specific invariant (`assert.ok(score, 'freeside-score not in manifest')`). This is the reference pattern for discovery-loop tests going forward.

- **D-S2-1 deviation documentation (F-13)**: The file-level comment that explains why `schema_version: '3'` discrimination is impossible given the actual schema constraint is exactly the class of comment worth writing. It prevents rediscovery, names the fixture that proves it, and cross-links to NOTES.md for full context. Future engineers will thank this author.

---

zkSoju:
## Triage of the Bridgebuilder review

**F-02 (HIGH) — ✅ confirmed, fixed in `82f52137`.** The runtime Dockerfile stage copied only `package.json` + `dist/` for `@freeside/freeside-registry`, omitting `registry.yaml` (read by `loadRegistry`) and `tests/fixtures/` (resolved by `loadBeacon`). Runtime stage now copies both. Deploy-path only — does not affect the in-process FR-6 acceptance; the Docker build still needs a deploy smoke-test (NOTES.md D-S3-1).

**F-01 (HIGH, conf 0.97) — ❌ false positive. Do not apply its suggested fix.**

F-01 claims `DEFAULT_REGISTRY_ROOT` resolves to `dist/` in compiled output because `beacon-loader.ts` lands at `dist/src/beacon-loader.js`. It does **not** — `packages/freeside-registry/tsconfig.json` has `rootDir: "src"`, so the package emits a **flat** `dist/` (`dist/beacon-loader.js`, not `dist/src/`). The `'..'` offset resolves to the package root in both source (`tsx`) and compiled forms.

Evidence: the 26/26 gateway tests + the FR-6 E2E test all run against the **compiled** `dist/` and resolve the fixture cleanly — if F-01 were real, every one would fail with `modules: []`.

⚠️ **F-01's suggested fix (`'..'` → `'..','..'`) would *actively break* it** — `DEFAULT_REGISTRY_ROOT` would become `packages/`, producing the exact empty-manifest bug F-01 describes. Leave `beacon-loader.ts` as-is. (Bridgebuilder generalised from `freeside-cli`/`beacon-schema`'s `dist/src/` layout — those use `rootDir: "."` — without checking `freeside-registry`'s tsconfig.)

F-03 / F-04 corroborate concerns already raised in the S2 / S3 sprint reviews — non-blocking. F-05–F-08 are valid minor observations.

— autonomous triage, `/bridgebuilder-review` follow-up

zkSoju (commented):
## Summary

This PR closes a real loop — beacons on disk become a capability-addressable manifest served over HTTP, with a CLI that can inspect and doctor it — and the engineering discipline around the *hard* parts (path traversal, skip-not-crash, spec-bug documentation) is genuinely excellent. The concerns cluster in one place: **the boundary between "the code is correct" and "we have proof it's correct."** The riskiest change (a reworked container build) has zero verification, the classifier silently downgrades broken V3 beacons to `legacy` instead of erroring, and the test suite has a wall-clock dependency that will fail on a specific future date with no code change.

### Architectural meditation: the shape of a discovery loop

Every capability-discovery system converges on the same three-layer shape — *declare* (beacon), *aggregate* (manifest), *resolve* (inspect). Kubernetes has this (CRD → API server → `kubectl describe`). Envoy has it (xDS resource → control plane → cluster manager). This PR builds all three at once, which is ambitious and mostly succeeds. But discovery systems live or die on **one property**: a malformed declaration must be *loud*. When Envoy's control plane rejects a bad `Cluster` config, it doesn't quietly fall back to an older schema and serve a cluster with no endpoints — it NACKs, and the operator sees it. F-02 and F-06 are the same bug wearing two hats: a beacon that is *structurally wrong* still looks *perfectly healthy* to both the manifest builder and `doctor`. A discovery system that hides broken declarations is worse than no discovery system, because operators now trust an incomplete map.

### The verification asymmetry

There's a striking inversion in this PR's risk profile. The path-traversal guard (F-16) — a threat that requires an attacker who can already write to `registry.yaml` — is defended with `realpathSync` containment *and* a test. The Dockerfile (F-03) — which changes its build context in a way that breaks *every* `COPY` if a Railway setting isn't manually flipped — is defended with a comment. This is the classic *availability-vs-security attention gradient*: security bugs feel scary so they get rigor, while build/deploy bugs feel boring so they get prose. Amazon's "you build it, you run it" doctrine exists precisely to close this gap.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-01",
      "title": "doctor's `cycle_state_freshness` check makes the test suite a dated time-bomb",
      "severity": "HIGH",
      "category": "testing",
      "file": "packages/freeside-cli/src/verbs/doctor.ts",
      "description": "`doctor()` emits severity `error` when `beacon.cycle_state.next_review` is in the past. The checked-in fixture `packages/freeside-registry/tests/fixtures/freeside-score.beacon.yaml` pins `next_review: \"2026-08-03\"`, and `packages/freeside-cli/tests/verbs.test.ts` asserts `report.summary.error === 0` on the real registry/fixture set. Once the wall clock passes 2026-08-03, the fixture's review date is in the past, the finding becomes `error`, and both `doctor` (exit 1) and the two doctor tests fail — with no code change. The same static date also crosses the 14-day `warn` window on ~2026-07-20. This is a wall-clock dependency baked into a deterministic unit test.",
      "suggestion": "Inject the clock: give `DoctorOptions` a `now?: number` (or a `Date` provider) and have the tests pass a fixed instant. Alternatively, compute the fixture's `next_review` relative to `Date.now()` in the test, or exclude fixture beacons from the freshness check. Whatever the choice, no assertion should depend on the day it runs.",
      "confidence": 0.9,
      "faang_parallel": "Google's testing-on-the-toilet 'Don't Mock Time, Inject It' and the java.time.Clock abstraction exist for exactly this. Every large codebase eventually gets a 'the build broke at midnight UTC' incident — Chromium's cert-expiry tests famously did, which is why they now inject a TestClock rather than trusting the host wall clock.",
      "metaphor": "You've set an alarm clock inside a test that's supposed to be a photograph. A photograph should look the same tomorrow. This one develops a red X on August 3rd, and nobody will know why — they'll just see a green build turn red on a day nobody committed anything.",
      "teachable_moment": "The rule isn't 'don't use dates in tests' — it's 'a test must be a pure function of the code under test.' Time is an *input*, and every input to a deterministic test must be supplied by the test, not sampled from the environment. The moment a test reads `Date.now()`, `process.env`, or the network, it has stopped being a test and become a monitor.",
      "connection": "This is the same root cause as F-04 (tests reaching out to the live network at import). Both are hermeticity violations: the suite depends on ambient state it doesn't control. Fixing the clock and fixing the network are the same discipline applied to two different globals."
    },
    {
      "id": "F-02",
      "title": "A malformed V3 beacon silently degrades to `legacy` V2 — doctor reports `warn` and exits 0",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/freeside-registry/src/beacon-loader.ts",
      "description": "`classifyBeacon` tries `decodeV3`, and on ANY failure falls through to `decodeV2`. Because BeaconV3 is a structural superset of BeaconV2 and the V3 fields are additive, a beacon whose V3 portion is broken (bad `produces.belt` casing, malformed `consumes.tag`, missing `cycle_state`) still decodes cleanly as V2 and is classified `legacy`. The manifest builder then skips it with a 'BeaconV2 (legacy) detected' warning, and `doctor` emits severity `warn` (not `error`) so it exits 0. The team is aware — `freeside-score-malformed.beacon.yaml` deliberately breaks a V2-*base* field precisely because 'a V3-only break would still decode as a V2 legacy beacon, not error'. That is the bug, not a fixture design choice: the negative fixture is shaped to dodge the misclassification instead of exercising it. FR-4's stated value (doctor catches non-compliant beacons) is undermined for the most likely class of real defect — a V3 author who typos a belt field.",
      "suggestion": "Discriminate on the presence of V3 marker fields before falling back. E.g. if the parsed YAML has any of `is`, `is_not`, `produces`, `consumes`, `cycle_state`, treat a failed V3 decode as `{ kind: 'error' }` (it is a broken V3 beacon), and only attempt the V2 path for beacons with no V3 markers at all. Add a fixture that is V3-shaped with one bad belt name and assert it classifies `error`, not `legacy`.",
      "confidence": 0.85,
      "faang_parallel": "Protocol Buffers made this a first-class design rule: you version with an explicit discriminator field, never by 'try the new parser, fall back to the old one.' Kubernetes learned it the hard way — the apiVersion field is mandatory and unambiguous precisely because content-sniffing a resource's version produced exactly this class of silent misclassification during the extensions/v1beta1 migration.",
      "metaphor": "It's a spellchecker that, on encountering a word it doesn't recognize in English, decides the document must be in French — and reports zero errors. The typo didn't get caught; it got *reclassified* into a language where it isn't a typo.",
      "teachable_moment": "Try-then-fallback parsing is only safe when the formats are *mutually exclusive*. The moment format B is a subset of format A, 'fall back to B' means 'downgrade every A-level error into a B-level success.' Detect intent (V3 markers present) before validating conformance — separate 'what did the author *mean* to write' from 'did they write it correctly.' Those are different questions and conflating them is what makes the failure silent.",
      "connection": "The team's own negative fixture is the smoking gun, and it's to their enormous credit that they wrote down *why* they shaped it that way (see F-18). The note that 'a V3-only break would still decode as a V2 legacy beacon' is the bug report — it was filed as a fixture comment instead of an issue. Pair this with F-06: both let a structurally-wrong beacon vanish from discovery while looking healthy."
    },
    {
      "id": "F-03",
      "title": "Gateway container build is reworked but never exercised; production deploy requires an out-of-band Railway change not in this PR",
      "severity": "HIGH",
      "category": "deployment",
      "file": "apps/mcp-gateway/Dockerfile",
      "description": "The Dockerfile's build context changed from `apps/mcp-gateway/` to the repository root (it now does `COPY packages/beacon-schema ./packages/beacon-schema`). Its own header states the container build is NOT exercised by this cycle — CI runs package `tsc`/tests, and the FR-6 test runs in-process via `app.request()`. The required Railway change ('set the service Root Directory to the repository root') is documented in a comment but `railway.toml` is not touched in this PR and no smoke test gates the merge. If this merges and a deploy fires before someone manually flips the Railway root directory, every `COPY packages/...` fails and the image build breaks. The riskiest change in the PR is the one with zero verification.",
      "suggestion": "Either (a) add a CI job that runs `docker build -f apps/mcp-gateway/Dockerfile .` from the repo root so the build path is covered, or (b) block merge on a manual deploy smoke-test and record the Railway root-directory change as a completed prerequisite, not a comment. At minimum, reference the required Railway setting from `railway.toml` so the two configs are co-located.",
      "confidence": 0.8,
      "faang_parallel": "This is the core argument for infrastructure-as-code that Netflix made with Spinnaker and HashiCorp made with Terraform: configuration that lives in a human's dashboard rather than in the repo is configuration that drifts. A `COPY` path in a Dockerfile and a Root Directory setting in a hosting dashboard are *one coupled contract* split across two systems with no compiler between them.",
      "metaphor": "You've rewritten the recipe to call for ingredients from the pantry down the hall, and left a sticky note on the fridge saying 'remember to unlock the pantry.' The recipe is correct. The sticky note is not a lock, and nobody reads sticky notes at 3am during a deploy.",
      "teachable_moment": "The question to ask of any change is not 'is it correct?' but 'what proves it stays correct?' A comment proves nothing — it's a hope with syntax highlighting. If a change has an external prerequisite, that prerequisite either becomes code (a CI job, a `railway.toml` field) or it becomes an incident. There is no third outcome, only a delay before the second one.",
      "connection": "Note the asymmetry against F-16: a threat requiring repo write access got `realpathSync` containment *and* a test, while a change that breaks every deploy got a comment. Rigor is being allocated by how *scary* a bug sounds rather than by expected cost. Boring bugs take down more production systems than exciting ones."
    },
    {
      "id": "F-04",
      "title": "Gateway tests make live outbound network calls at import; `--test-force-exit` masks the leak",
      "severity": "HIGH",
      "category": "testing",
      "file": "apps/mcp-gateway/package.json",
      "description": "Importing `src/app.js` triggers a fire-and-forget `refreshAllBeacons()` that fetches real upstream beacons (acknowledged in grimoires/loa/NOTES.md: 'in-flight fetches otherwise keep the node:test process alive ~20s+'). The test script was changed to `tsx --test --test-force-exit tests/*.test.ts` to paper over the hanging handles. The consequence: every gateway test file — including `discovery-loop.e2e.test.ts`, whose header claims 'no network... Deterministic' — performs uncontrolled outbound HTTP on import. That makes the suite non-hermetic (fails or slows behind a firewall / in an offline CI runner), and `--test-force-exit` can also truncate output on a slow reporter.",
      "suggestion": "Gate the boot-time refresh behind an env flag or export the app factory without side effects, e.g. only call `refreshAllBeacons()`/`startBeaconRefresh()` from `bin/http.ts` (the process entrypoint), not from `app.ts` module scope. Then drop `--test-force-exit` and let clean teardown prove the suite has no leaks.",
      "confidence": 0.8,
      "faang_parallel": "Google's Bazel test runner enforces network isolation by *sandbox*, not by convention — a hermetic test literally cannot reach the network, so this failure mode is structurally impossible. Martin Fowler's 'Test Pyramid' calls the same thing out: the moment a unit test can be broken by someone else's DNS, it's no longer a unit test, it's a distributed-systems integration test wearing a costume.",
      "metaphor": "`--test-force-exit` is duct tape over the check-engine light. The engine still has the problem; you've just arranged to stop hearing about it. And the light was telling you something important: your module is doing I/O at import time.",
      "teachable_moment": "Side effects at module scope are the original sin of testability. `import` should be *inert* — it defines things; it doesn't *do* things. The fix here isn't a test-runner flag, it's the dependency-injection boundary: a module exports a factory, and only the process entrypoint (`bin/http.ts`) is allowed to pull the trigger. Once you draw that line, the test never needs a flag, because there's nothing to force-exit from.",
      "connection": "Same hermeticity family as F-01 (wall clock). Both say: the test's inputs must come from the test. Note also that `discovery-loop.e2e.test.ts` *documents itself* as 'no network — Deterministic', which means the header comment and the runtime behavior have already diverged. That's how a false sense of safety propagates."
    },
    {
      "id": "F-05",
      "title": "`buildFreesideManifest` default visibility filter now includes `internal`, and `scope` is hardcoded regardless of the filter",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/freeside-registry/src/registry.ts",
      "description": "Two coupled problems. (1) The default `visibilityFilter` changed from `[\"public\"]` (fail-closed) to `[\"public\", \"unlisted\", \"internal\"]` (fail-open). Any future caller that forgets the third argument — including the eventual public `/federation.json` this SDD explicitly defers to a later cycle — will serve internal buildings. The current gateway route is operator-gated, so nothing leaks today; the hazard is the default itself. (2) `scope: \"internal\"` is a hardcoded literal on the returned manifest, so calling `buildFreesideManifest(reg, loader, [\"public\"])` produces a public-filtered manifest that still declares `scope: \"internal\"` on the wire. The field the SDD introduced specifically to make scope auditable is decoupled from the thing it claims to describe.",
      "suggestion": "Restore the fail-closed default (`[\"public\"]`) and pass the explicit internal filter from `apps/mcp-gateway/src/freeside-manifest.ts`, which is the caller that knows it is behind the operator gate. Derive `scope` from the filter rather than hardcoding it (e.g. `internal` if the filter contains `internal`, else `public`).",
      "confidence": 0.7,
      "faang_parallel": "AWS S3 spent a decade paying for a permissive default. Buckets were public-by-default in spirit, the industry produced an entire genre of breach headline, and AWS eventually shipped 'Block Public Access' *on by default* — a change they could only make because the cost of the old default had become undeniable. Defaults are not conveniences; they are the security posture of every caller who hasn't read the docs yet.",
      "metaphor": "The default is the setting on the dial when nobody's looking. You've moved it from 'locked' to 'unlocked' and put a guard on today's only door. The next door someone cuts into this wall will inherit 'unlocked' and no guard — and they'll never know they made a decision, because they didn't.",
      "teachable_moment": "Two lessons braided together. First: **defaults should encode the safest posture, and the risky posture should require typing it out** — the friction of an explicit `[\"public\", \"unlisted\", \"internal\"]` argument at the one privileged call site is a feature, because it makes the privilege *visible in the diff*. Second, and subtler: a `scope` field that is hardcoded rather than derived is a **label, not a fact**. Any metadata field that claims to describe state but is computed independently of that state will eventually lie. Derive, don't declare.",
      "connection": "The SDD explicitly defers a public `/federation.json` to a later cycle — which means the exact caller most likely to forget the argument is *already on the roadmap*. This is a landmine planted directly in the path of the next sprint. Fail-closed now costs one line; fail-open later costs an incident review."
    },
    {
      "id": "F-06",
      "title": "A V3 beacon still declaring `composes_with` decodes clean with empty belts — the silent breakage NFR-1 forbids",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/beacon-schema/tests/schema.test.ts",
      "description": "The new regression test asserts that a beacon carrying a stale `composes_with` block decodes to Success — 'a stale composes_with block is an ignored excess property, never validated'. Because `produces`/`consumes` default to `[]`, such a beacon aggregates into the manifest with zero belts and no warning: it disappears from capability-addressable discovery (G-2's whole point) while looking perfectly healthy to both the manifest builder and `doctor`. NFR-1 requires V2/legacy handling to be explicit — 'either still validate, or reject with an explicit migration message. No silent breakage'. The hard-replace of `composes_with` got the removal but not the migration message, and this test enshrines the silent path as intended behavior.",
      "suggestion": "Add a `composes_with`-present check to `doctor` (severity `warn` with the ADR-007 A.4 migration message: `<slug>: composes_with is removed — migrate to produces/consumes`), or set `onExcessProperty: 'error'` for the specific key. Then flip this test to assert the migration signal is emitted rather than asserting silence.",
      "confidence": 0.7,
      "faang_parallel": "Rust's `#[deprecated]` and Python's `DeprecationWarning` exist because the field you *removed* is more dangerous than the field you *changed* — a changed field errors, a removed field is just ignored. The Linux kernel's rule for removing a sysfs attribute is that you leave a warning behind for a full release cycle; you never let a userspace program keep writing into the void believing it configured something.",
      "metaphor": "You moved the mailbox and left no forwarding address. The sender keeps posting letters into the old slot, the letters go nowhere, and the postal service reports 100% successful delivery — because from its point of view, there's no such address, so there's nothing to fail.",
      "teachable_moment": "When you remove a field, the migration path *is part of the removal*. 'Silently ignored' is the worst of the three possible behaviors: still-works (fine), hard-error (annoying but honest), silently-ignored (a lie the author will believe for months). And notice the deeper hazard here — a **test that asserts silence codifies the silence as intent**. The next engineer reading `schema.test.ts` will treat this as a deliberate contract and defend it. Tests don't just verify behavior; they *document intent*, which means a test can lock in a bug harder than the bug itself ever could.",
      "connection": "Structurally identical to F-02: a wrong beacon becomes an invisible beacon. Together they define the real theme of this review — the discovery loop closes, but it closes *quietly* over exactly the beacons that need to be loud. NFR-1 anticipated this and the implementation didn't land it."
    },
    {
      "id": "F-07",
      "title": "`freeside-cli inspect` cannot work for any of the 6 real registry entries",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/freeside-cli/src/verbs/inspect.ts",
      "description": "`inspectModule` confirms the slug appears in the live manifest, then resolves the full beacon via `loadBeacon(entry)`. `loadBeacon` returns `{ kind: 'error' }` for any entry without `beacon_fixture` ('remote beacon_url fetch is not implemented this cycle'). Since only `freeside-score` has a fixture, `inspect` throws for every other registered module — and those modules are also skipped from the manifest, so they fail the earlier presence check first. The verb is therefore operable for exactly one synthetic entry. The doc comment above the function still describes fetching 'through the live federation manifest', and G-3 reads 'the operator can inspect ANY registered building's beacon'. The gap between the stated capability and the shipped one is not surfaced to the operator.",
      "suggestion": "Either implement the remote `beacon_url` fetch in `loadBeacon` (it is a `fetch` + the same `classifyBeacon` call), or make the limitation explicit in the CLI help text and the verb's error message ('remote beacons are not resolvable until cycle-0XX; only fixture-backed buildings can be inspected'). Also correct the header comment, which currently overstates what the code does.",
      "confidence": 0.75,
      "faang_parallel": "Every mature CLI distinguishes 'not supported' from 'failed' — `git` will tell you `fatal: this operation must be run in a work tree` rather than throwing a stack trace, and `kubectl` says `the server doesn't have a resource type X` rather than a generic error. The exit-code taxonomy in the BSD `sysexits.h` (EX_UNAVAILABLE vs EX_SOFTWARE) is 40 years old and still the right model: an unimplemented feature is a *different category of outcome* than a bug.",
      "metaphor": "A phone book where 5 of the 6 numbers are unlisted, and dialing them gives you a dial tone that eventually just... stops. Not 'this number isn't available' — just silence and a confused user checking whether their phone is broken.",
      "teachable_moment": "Shipping a partial capability is completely fine — cycles are how you get anywhere. What's not fine is letting the *documentation* describe the finished version while the *code* ships the partial one. Scope reduction is a decision, and decisions belong in three places: the code (an explicit `not_implemented` branch), the user-facing surface (help text, error message), and the docs. When you cut scope and only update the ticket, you've created a phantom feature that everyone downstream believes in.",
      "connection": "The doc comment claiming 'through the live federation manifest' is the same species of drift as the `discovery-loop.e2e.test.ts` header claiming 'no network' (F-04). Comments in this PR are describing the intended system, not the shipped one. That's a systemic pattern worth a single sweep rather than three individual fixes."
    },
    {
      "id": "F-08",
      "title": "Every request to /internal/freeside.json re-reads YAML from disk synchronously and logs 6 warnings",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "apps/mcp-gateway/src/freeside-manifest.ts",
      "description": "`buildFreesideJson()` calls `loadRegistry()` (sync `readFileSync` + YAML parse) and then `loadBeacon` per module (sync `realpathSync` + `readFileSync` + parse) on every request, inside the Hono handler — blocking the event loop for the duration. Worse, the 6 remote-only entries in `registry.yaml` each hit the `{ kind: 'error' }` branch, so `buildFreesideManifest` emits 6 `console.warn` lines per request for what is the *expected steady state*, not an exceptional condition. Warnings that fire on every successful request stop being signal.",
      "suggestion": "Load the registry once at module scope (or memoize with the same stale-while-revalidate pattern `beacon-cache.ts` already uses) and cache the built manifest. Downgrade the 'remote beacon_url not implemented' skip to a single boot-time log or drop it entirely — reserve `console.warn` for a beacon that was supposed to resolve and didn't.",
      "confidence": 0.75,
      "faang_parallel": "Two classics in one finding. The sync-I/O-in-a-handler half is the Node.js event-loop lesson Netflix wrote up after tracing a latency spike to a blocking call on the request path. The log-noise half is what Google's SRE book calls alert fatigue, and what the industry rediscovered in the Target breach postmortem: the intrusion *did* fire an alert, into a stream so noisy that nobody could see it.",
      "metaphor": "A smoke detector that chirps six times every time you successfully make toast. Within a week, nobody in the house hears it anymore — including the one time it's actually smoke.",
      "teachable_moment": "A log level is a *claim about how often this should happen*. `warn` means 'a human should probably look at this' — so if it fires on every successful request, you haven't logged a warning, you've logged a heartbeat with a scary label, and you've trained your operators to ignore that label. The discipline: before choosing a level, ask 'what is the steady-state rate of this line?' If the answer is 'once per request,' it's `debug` or it's nothing. And the caching fix is a bonus — you already have the stale-while-revalidate pattern in `beacon-cache.ts`; reuse beats reinvention.",
      "connection": "Reaching for the existing `beacon-cache.ts` pattern here would also cleanly resolve the F-04 side-effect-at-import problem, since a memoized lazy loader has no reason to fire at module scope. One structural fix, two findings closed."
    },
    {
      "id": "F-09",
      "title": "Docker builds use `--no-frozen-lockfile` for every install, including packages with no lockfile at all",
      "severity": "MEDIUM",
      "category": "build",
      "file": "apps/mcp-gateway/Dockerfile",
      "description": "All four installs in the Dockerfile pass `--no-frozen-lockfile`. `packages/beacon-schema` and `packages/freeside-registry` have no lockfile copied into the image at all, so their dependency trees are resolved fresh against the registry at image-build time. Two builds of the same commit can therefore produce different `effect`/`yaml`/`typescript` versions, and the `apps/mcp-gateway` install will silently accept a lockfile that no longer matches `package.json` instead of failing. For a service whose PR description leans on 'zero new dependencies' as an attack-surface argument, the build path does not actually pin what it installs.",
      "suggestion": "Commit lockfiles for `packages/beacon-schema` and `packages/freeside-registry` (or bring the gateway back under the root pnpm workspace so one root lockfile covers all four), copy them into the image, and use `--frozen-lockfile` so a drifted lockfile fails the build rather than being silently repaired.",
      "confidence": 0.7,
      "faang_parallel": "Reproducible builds are the entire thesis of Bazel and Nix, and the reason Google vendors third-party dependencies into a monorepo rather than resolving them at build time. The supply-chain relevance is not theoretical: `event-stream`, `ua-parser-js`, and `node-ipc` were all compromised *in a published patch version* — a build that re-resolves at image time would have pulled the malicious version; a frozen lockfile would not have.",
      "metaphor": "It's a recipe that says 'flour' instead of 'King Arthur bread flour, 500g, lot #4471.' It'll produce bread most days. It will also, one day, produce something nobody ordered, and you'll have no way to tell what changed because the recipe is identical.",
      "teachable_moment": "'Zero new dependencies' is a claim about the *manifest*; `--no-frozen-lockfile` is a fact about the *resolver*. They are not the same statement, and only one of them is enforceable. A security argument that rests on a property the build doesn't actually guarantee is a security argument that will be quietly falsified. The general rule: if you're going to cite a property in a design doc, make sure something in CI *fails* when the property stops holding.",
      "connection": "This and F-10 are the same underlying decision — leaving the root pnpm workspace — surfacing as two different failure modes. That decision deserves its own Decision Trail entry, because the fix for both is probably 'go back into the workspace and fix the nested `pnpm-workspace.yaml` that was the actual bug.'"
    },
    {
      "id": "F-10",
      "title": "`file:` deps are packed copies, not live links — stale `dist/` will ship silently",
      "severity": "MEDIUM",
      "category": "build",
      "file": "apps/mcp-gateway/package.json",
      "description": "Switching the gateway from `workspace:*` to `file:../../packages/*` changes the semantics: pnpm resolves these as `type: directory` resolutions (visible in the regenerated lockfile) and packs the package per its `files` array into the store, rather than symlinking the source. Consequences already bit this PR once — D-S3-2 in NOTES.md describes `tests/fixtures` being absent from the copied registry because it was missing from `files`, causing `freeside-score` to be silently skipped from the manifest. The general hazard remains: any edit to `packages/freeside-registry/src` requires a re-install in `apps/mcp-gateway` before the gateway (or its tests) sees it, and a stale or unbuilt `dist/` produces confusing failures with no signal that the copy is out of date. Nothing in the gateway's scripts enforces build-before-install ordering.",
      "suggestion": "Prefer bringing `apps/mcp-gateway` under the root pnpm workspace so `workspace:*` gives live links (that was the original intent; the nested `pnpm-workspace.yaml` was the actual bug, not `workspace:*` itself). If `file:` must stay, add a `pretest`/`prebuild` step in the gateway that rebuilds and reinstalls the three local packages, and add a `files`-completeness test so the next missing runtime asset fails loudly instead of skipping a module.",
      "confidence": 0.7,
      "faang_parallel": "This is precisely the problem Google's monorepo + Blaze solved by making the *build graph* the source of truth: a target cannot consume a stale artifact of its dependency because the dependency edge forces a rebuild. pnpm workspaces are a poor-man's version of the same guarantee, and stepping outside them means you've opted out of the guarantee and taken on the invalidation logic yourself — manually.",
      "metaphor": "You photocopied the blueprint instead of linking to it. Now the architect updates the original and the builders keep working from the copy, and the only way anyone finds out is when the wall doesn't meet the roof.",
      "teachable_moment": "The debugging heuristic worth internalizing: **when a fix works, ask whether it fixed the cause or moved the symptom.** The symptom was a broken workspace resolution; the actual cause (per NOTES) was a stray nested `pnpm-workspace.yaml`. Switching to `file:` made the symptom go away by leaving the system that was misconfigured — which works, but it also silently traded a *loud* failure (resolution error) for a *quiet* one (stale packed copy). D-S3-2 in your own notes is the first bill for that trade arriving; it won't be the last.",
      "connection": "Same root decision as F-09. The `files`-array bug in D-S3-2 is a beautiful illustration: a missing fixture directory caused a module to be *silently skipped from the manifest* — which is, once again, the theme of this entire review. Silent skips are this system's characteristic failure mode, and they now have four independent causes (F-02, F-06, F-08, F-10)."
    },
    {
      "id": "F-11",
      "title": "SDD claims a backup file that this PR does not create",
      "severity": "MEDIUM",
      "category": "documentation",
      "file": "grimoires/loa/sdd.md",
      "description": "The new SDD header states the prior as-built snapshot is 'preserved at `grimoires/loa/sdd.md.ride-2026-05-18-bak` and in git history'. No such file appears in this PR's 39 changed files, while `sdd.md` itself is fully overwritten (−295 lines of platform topology: ports/adapters map, data model, API surface, infra layout). The PRD makes a weaker, verifiable claim (git history at `090fbab2`). As written, the SDD points readers at a path that does not exist, and the platform as-built documentation is now recoverable only by someone who knows to go archaeology in git.",
      "suggestion": "Either add `grimoires/loa/sdd.md.ride-2026-05-18-bak` in this PR, or correct the header to cite the commit SHA the way the PRD does. Given cycle SDDs and platform as-built SDDs are different artifacts serving different readers, consider giving the cycle SDD its own path (`grimoires/loa/cycles/049/sdd.md`) rather than overwriting the platform one each cycle.",
      "confidence": 0.6,
      "faang_parallel": "Amazon's PR/FAQ and Google's design-doc culture both treat the design doc as an *append-only historical record* — you write a new doc and link back, you don't overwrite the old one, because six months later the question 'why was it built that way?' is answered by the *old* doc, not the new one. RFCs are immutable and superseded by number for exactly this reason.",
      "metaphor": "A museum plaque that says 'the original is preserved in the east wing.' There is no east wing. The original is in a landfill, and technically you could still dig it up, but you'd need to know which landfill and roughly what week.",
      "teachable_moment": "A dangling pointer in prose is still a dangling pointer — and it's worse than no pointer, because it stops the reader from looking elsewhere. But the structural lesson is bigger: a **cycle SDD and a platform as-built SDD are different documents with different readers and different lifetimes**, and overwriting one with the other every cycle guarantees the platform topology gets quietly deleted, one cycle at a time. Give them separate paths and the problem stops recurring forever.",
      "connection": "The PRD got this right — it cites a commit SHA, which is verifiable. Two documents in the same PR making the same claim at different levels of rigor is a good sign that the *verifiable* form was known and just not applied consistently."
    },
    {
      "id": "F-12",
      "title": "`capabilities.tools` is projected from `b.mcp.tools` without a presence check",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/freeside-registry/src/registry.ts",
      "description": "`capabilities: { tools: b.mcp.tools }` assumes every BeaconV2 `mcp` block carries a `tools` array. If `tools` is optional in the V2 schema (a `shape: data` beacon exposing only resources, say), this yields `{ tools: undefined }` in the manifest, and downstream consumers that do `score.capabilities.tools.length` — including the gateway integration test and the E2E test — would throw on it. Both fixtures declare `tools`, so no current path exercises the gap.",
      "suggestion": "Default it: `tools: b.mcp.tools ?? []`, matching the belt fields' `default: () => []` convention. If the V2 schema does make `tools` required, this is a no-op and worth doing anyway for symmetry.",
      "confidence": 0.5,
      "faang_parallel": "Postel's Law, applied to your own internal boundaries: be liberal in what you accept. The `?? []` idiom is what turned an entire class of null-pointer crashes into no-ops in Kotlin and Swift — the languages made the empty-collection default *cheaper to type* than the crash.",
      "metaphor": "A guest list that assumes every party has attendees. Throw a party with nobody invited, and instead of an empty room you get a crash on the way in.",
      "teachable_moment": "Verify first: check whether `tools` is actually optional in `beacon-v2.ts`. If it's required, this is a no-op — but do it anyway, because `?? []` costs five characters and buys immunity to a *future* schema change that makes it optional. Defensive defaults at a projection boundary are one of the highest-ratio investments in software: near-zero cost, and they pre-absorb a change nobody has made yet.",
      "connection": "The belt fields in the same file already use `default: () => []`. Applying the same convention to `tools` is consistency, and consistency is what lets a reader stop checking."
    },
    {
      "id": "F-13",
      "title": "500 response echoes the raw error message to the caller",
      "severity": "LOW",
      "category": "security",
      "file": "apps/mcp-gateway/src/app.ts",
      "description": "The `/internal/freeside.json` error path returns `{ error: 'manifest_build_failed', detail: err.message }`. An Effect `ParseError` or an fs error message can include absolute filesystem paths and internal schema structure. The route is operator-gated, so the audience is trusted today — but the endpoint is the seed of the public federation surface the SDD defers to a future cycle (OQ-4), and error-detail echo is the kind of thing that survives a scope change unnoticed.",
      "suggestion": "Log the full error server-side (already done) and return a generic `detail` or omit it. If the detail is genuinely useful to the operator, gate it behind an explicit debug flag rather than making it the default.",
      "confidence": 0.5,
      "faang_parallel": "OWASP A05 (Security Misconfiguration) lists verbose error responses explicitly, and it's why Rails ships `config.consider_all_requests_local = false` in production — the framework authors concluded that developers cannot be trusted to remember which errors are safe to echo, so they made the safe behavior the environment default.",
      "metaphor": "Sending back a photo of your server's filing cabinet, drawers open, whenever a lookup fails. Helpful for you. Also a map for someone else.",
      "teachable_moment": "The trust boundary you have today is not the trust boundary you'll have in six months, and error paths are where the assumption of a trusted audience goes to die — because nobody re-audits the 500 handler when they make a route public. The habit worth building: **write the error path for the audience the endpoint will eventually have**, not the one it has now. It costs nothing today and it's the difference between a scope change being a config edit and a scope change being a leak.",
      "connection": "Precisely parallel to F-05: both are 'safe today because of a gate that a future cycle is explicitly planning to remove.' The SDD *tells you* the public federation surface is coming (OQ-4). That's not a hypothetical future caller; it's a scheduled one."
    },
    {
      "id": "F-14",
      "title": "Ledger marks all four sprints `completed` while the cycle is still `active`",
      "severity": "LOW",
      "category": "consistency",
      "file": "grimoires/loa/ledger.json",
      "description": "`cycle-049` carries `\"status\": \"active\"` and `\"active_cycle\": \"cycle-049\"`, yet all four of its sprints are `\"status\": \"completed\"`. NOTES.md meanwhile records Sprint 4 as 'implementation complete... Awaiting /review-sprint + /audit-sprint'. Three sources of truth, three different states for the same sprint.",
      "suggestion": "Pick the ledger as the single source of truth for sprint state and reconcile: either Sprint 4 is `in_review` (matching NOTES) or it is `completed` and NOTES should say so. Also confirm whether the cycle should flip to `completed` on merge.",
      "confidence": 0.6,
      "faang_parallel": "This is a distributed-state problem wearing a project-management costume — three replicas, no consensus protocol, and a human as the (unreliable) reconciliation loop. The fix is the same as it is in a database: elect a leader. One store is authoritative; the others are *views* of it, ideally generated rather than hand-maintained.",
      "metaphor": "Three clocks in the same room, all showing different times. The problem isn't that any one of them is wrong — it's that you can no longer tell what time it is, and now checking the time requires a meeting.",
      "teachable_moment": "'Completed' has to mean one thing, and someone has to own the definition. The moment a state word can be interpreted as 'code written', 'reviewed', or 'merged' depending on which file you opened, the state has stopped carrying information. Pick the ledger, define the vocabulary, and make everything else derive from it — hand-maintained duplicate state is a bug with a longer fuse than most code bugs and a much wider blast radius.",
      "connection": "Ideally `status` in NOTES is *generated* from the ledger rather than typed, which makes this class of drift structurally impossible instead of merely discouraged. Every time you find yourself reconciling two documents by hand, that's a signal one of them should have been a projection."
    },
    {
      "id": "F-15",
      "title": "Verify `isAuthorizedOperator` fails closed when OPERATOR_API_KEY is unset",
      "severity": "SPECULATION",
      "category": "security",
      "file": "apps/mcp-gateway/src/app.ts",
      "description": "The new route's entire access control is `if (!isAuthorizedOperator(c)) return 401`. That function is not in this diff, so its behavior when `OPERATOR_API_KEY` is absent from the environment cannot be confirmed here. If it fails open (no key configured → no gate), this PR ships the freeside building manifest — including internal-visibility buildings and their tool surfaces — unauthenticated to anyone who can reach the gateway. If it fails closed, the route is simply unreachable until the key is set, which is the correct posture. Worth an explicit confirmation given the route is new and the manifest content is newly internal-scoped.",
      "suggestion": "Confirm `auth.ts` returns false (not true) when the key is unset, and add a test asserting it — `GET /internal/freeside.json` with no `OPERATOR_API_KEY` in the environment must be 401, not 200.",
      "confidence": 0.3,
      "faang_parallel": "The 2017 Equifax and 2018 Kubernetes-dashboard exposures share a shape: an auth check that was *present* but degraded to permissive under a missing-config condition nobody tested. This is why Kubernetes eventually made `--anonymous-auth=false` the default and why cloud providers now *refuse to boot* a service with an unconfigured auth mode rather than booting it open.",
      "metaphor": "A door with a keycard reader. The question is what the reader does when the power's out: does it lock, or does it swing open? Both are one line of code. Only one of them is a door.",
      "teachable_moment": "This is flagged as SPECULATION and should be treated as a *verification task, not a finding* — `auth.ts` isn't in the diff, so I genuinely cannot tell you which way it goes, and it is entirely likely it's correct. But the fact that it can't be determined *from this PR* is itself the actionable point: a new route whose only access control lives in an unreviewed file is a route whose security posture is unreviewed. Add the test and the question stops being answerable-only-by-reading-the-source — which means it stops being re-askable at every future review.",
      "connection": "If this does fail open, it upgrades F-05 (the internal-inclusive default filter) from 'latent hazard' to 'live exposure', since the two would compose into an unauthenticated internal manifest. That coupling is the reason this is worth 60 seconds of confirmation rather than a shrug — check `auth.ts` first, then decide how much F-05 actually matters."
    },
    {
      "id": "F-16",
      "title": "Path-safety guard on `beacon_fixture` is done properly",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/freeside-registry/src/beacon-loader.ts",
      "description": "`resolveFixturePath` rejects `..` before touching the filesystem, then `realpathSync` both the root and the candidate before a `startsWith(realRoot + sep)` containment check — which defeats symlink escape, not just lexical traversal. The `+ sep` is the detail most implementations get wrong (it prevents `/repo/packages/freeside-registry-evil` from passing a bare `startsWith`). `path.join` also neutralizes an absolute-path fixture. The threat model is honest too: `registry.yaml` is repo-controlled but still treated as untrusted at load time, and there is a test for the traversal case.",
      "suggestion": "No change. Consider promoting this to a shared util if a second component ever resolves repo-relative paths from config.",
      "confidence": 0.9,
      "faang_parallel": "This is the containment check Node's own `fs` docs and the CVE record say most people get wrong. `zip-slip` (2018) hit thousands of projects across five languages with exactly the lexical-only variant of this bug; the `+ sep` prefix-boundary detail is what separates a real containment check from a check that a directory named `foo-evil` walks straight through.",
      "metaphor": "Most people check the address on the envelope. This code checks the address, then follows the delivery truck to where the mail *actually* lands, and confirms the building is inside the fence — and confirms the fence line falls at a wall, not halfway through a neighbor's identically-named lobby.",
      "teachable_moment": "The generalizable insight: **resolve, then validate — never validate, then resolve.** Any check performed on a path *before* the OS has told you where it truly points is a check on a string, not on a file. `realpathSync`-then-contain is the correct order, and the `+ sep` is what makes the containment a boundary rather than a prefix match. Also worth naming: treating a repo-controlled file as untrusted input is *correct paranoia*, not wasted effort — repo-controlled today, PR-controlled tomorrow, template-generated the day after.",
      "connection": "This is the standard of rigor the rest of the PR should be measured against. F-03 (unverified container build) and F-09 (unpinned deps) are the same *category* of risk — untrusted input crossing a boundary — and got a fraction of the care. Not because the engineer got lazy, but because path traversal *sounds* dangerous and a Dockerfile *sounds* boring."
    },
    {
      "id": "F-17",
      "title": "Resilient-in-prod, strict-in-test is drawn deliberately and documented at both ends",
      "severity": "PRAISE",
      "category": "design",
      "file": "apps/mcp-gateway/tests/discovery-loop.e2e.test.ts",
      "description": "The manifest builder skips a bad beacon rather than 500ing the endpoint (prod stays up when one building misbehaves), while the E2E test asserts `freeside-score` MUST be present and fails loudly if it is skipped. That is the right split — the usual failure mode of skip-not-crash is that the test suite inherits the leniency and stops proving anything. It is called out explicitly in SDD §5.1, in the route handler comment, and in the test's own assertion message, so the next person to touch it cannot mistake the strictness for an accident.",
      "suggestion": "No change.",
      "confidence": 0.85,
      "faang_parallel": "This is the *crash-only / bulkhead* pattern from Netflix's Hystrix and Google's SRE 'graceful degradation' chapter, and critically it includes the half most teams forget: the bulkhead protects production, and the *test* is where you refuse to degrade. Envoy does exactly this — a bad upstream cluster is ejected at runtime, but the config-validation test suite treats the same config as a hard failure.",
      "metaphor": "A ship with watertight compartments — one floods, the ship sails on. But in dry dock, if an inspector finds a compartment that leaks, the ship doesn't sail at all. Lenient at sea, ruthless in the yard.",
      "teachable_moment": "The trap this PR avoided is real and common: teams adopt skip-not-crash for resilience, and then the test suite — running the *same* code path — quietly inherits the leniency, and now a completely broken beacon produces a green build. Splitting the strictness by *environment* rather than by *code path* is the correct axis, and documenting it in three places (SDD, handler comment, assertion message) means the next engineer can't 'helpfully' relax the test to match production.",
      "connection": "The irony worth sitting with: the team clearly *understands* the loud-vs-quiet tradeoff — they reasoned about it explicitly here and got it exactly right. F-02, F-06, and F-10 are all cases where the same principle wasn't applied. The knowledge is present; it just hasn't been generalized into a rule yet. That's a very good place to be."
    },
    {
      "id": "F-18",
      "title": "The `schema_version` spec error was caught, documented, and deviated from rather than papered over",
      "severity": "PRAISE",
      "category": "process",
      "file": "grimoires/loa/NOTES.md",
      "description": "D-S2-1 records that the SDD's prescribed V2/V3 discrimination (`schema_version: \"3\"` → V3) is impossible against the shipped schema, since `BeaconV3Schema` extends a `BeaconV2Schema` that pins `schema_version: Schema.Literal(\"2\")` — with the pre-existing `freeside-inventory-v3.yaml` fixture cited as empirical proof. The implementation deviated, and the deviation is written down with its reasoning, its scope boundary (fixing `beacon-v2.ts` is out of cycle), and an explicit request for reviewer scrutiny. That is how a spec bug should surface. (The chosen decode-attempt fallback has its own defect — see F-02 — but the diagnosis and the paper trail are right.)",
      "suggestion": "No change to the process. Address the fallback's misclassification per F-02.",
      "confidence": 0.85,
      "faang_parallel": "Amazon's Correction of Error (COE) doc and Google's blameless postmortem exist to make exactly this behavior *cheap*: surfacing 'the spec is wrong' without it becoming a status hit. The rarest and most valuable thing an engineer does is push back on the design doc *with evidence* mid-implementation, instead of implementing something they know is incoherent.",
      "metaphor": "The builder found the blueprint called for a door in a load-bearing wall, didn't build it, didn't quietly move it either — they wrote on the blueprint exactly where it was wrong, what they did instead, and asked the architect to look. That note is worth more than the door.",
      "teachable_moment": "This is the highest-leverage habit in the entire PR, and it deserves naming: **the deviation record is a first-class artifact, not overhead.** D-S2-1 does all four things a good one does — states what the spec said, proves why it's impossible (with a citation to an existing fixture, which is *evidence*, not opinion), states what was done instead, and draws the scope boundary. It even flags itself for reviewer scrutiny. Six months from now, when someone asks 'why doesn't this discriminate on schema_version like the SDD says,' the answer is already written down. That's how a codebase stops re-litigating its own history.",
      "connection": "And it's precisely *because* the paper trail is this good that F-02 was findable at all — the note that 'a V3-only break would still decode as a V2 legacy beacon' is the bug, stated in the author's own words. Good documentation doesn't just explain the system; it makes the system's flaws legible. That's not a criticism of the note. That's the note doing its job."
    }
  ]
}
```


## Callouts

**[Security] The path-safety guard (F-16) is the best code in this PR.** `realpathSync`-then-`startsWith(realRoot + sep)` is the *correct* containment check, and the `+ sep` detail — the thing that stops `freeside-registry-evil` from passing a naive prefix match — is what separates a real boundary from a string comparison. Zip-slip hit thousands of projects across five languages in 2018 with the lexical-only version of this. Treating a repo-controlled `registry.yaml` as untrusted input is correct paranoia, not wasted motion, and there's a test for the traversal case. Promote this to a shared util the moment a second component needs it.

**[Design] The strict-in-test / lenient-in-prod split (F-17) avoids a trap most teams fall into.** Skip-not-crash keeps the gateway up when one building misbehaves — good. But the *usual* failure mode is that the test suite inherits the leniency and a totally broken beacon still produces a green build. Splitting the strictness by environment rather than by code path is the right axis, and it's documented in three places so nobody can "helpfully" relax it later.

**[Operational] D-S2-1 (F-18) is how a spec bug should surface.** The SDD prescribed something structurally impossible; the implementer proved it with an existing fixture, deviated, wrote down the reasoning and the scope boundary, and *asked for scrutiny*. Pushing back on the design doc mid-implementation, with evidence, is the rarest habit in engineering. It's also what made F-02 findable — the note describing the fallback's weakness is, verbatim, the bug report.

---

### Closing reflection

There is one theme under nearly every finding here, and it's worth stating plainly: **this system's characteristic failure mode is the silent skip.** A malformed V3 beacon becomes a `legacy` one (F-02). A stale `composes_with` becomes an empty belt list (F-06). A missing `files` entry drops a fixture and a module vanishes from the manifest (F-10). Six expected skips drown the one unexpected one in `console.warn` (F-08). In each case the beacon disappears from discovery *while everything reports healthy*.

And here's the thing — you already know how to fix this. F-17 is that exact reasoning, applied correctly, documented in three places. The knowledge is in the room; it just hasn't been promoted from "a decision we made about the E2E test" to "a rule about how this system handles bad input." The rule, if you want it: **a beacon that was supposed to resolve and didn't is an error, not a warning, and never a shrug.**

Everything else — the clock in the test, the Dockerfile nobody built, the lockfiles nobody pinned — is the same lesson in a different register: *what proves this stays true?* The security-shaped risks in this PR got rigorous proof. The boring ones got comments. Boring bugs take down more production systems than exciting ones, and they do it on a Tuesday, during a deploy, for reasons nobody can reconstruct.

**Verdict: REQUEST_CHANGES** — F-01 through F-04 should land before merge. F-15 is 60 seconds of confirmation that changes how much F-05 matters.

PR #220, opened 2026-05-25
feat(network/identity-api): Phase 1 substrate · registry + tenant + planning baseline

zkSoju:
```yaml
hivemind:
  artifact_type: technical-rfc
  workstream: delivery
  priority: medium
  product_area: Auth Substrate / Network Registry
  jtbd:
    category: functional
    description: Make identity-api discoverable as a freeside building so the MCP federation gateway routes tenant calls correctly
  learning_status: smol-evidence
  source: team-internal
```

---

# 🌐 identity-api · network registration

| layer | what lands |
|---|---|
| 🗺️ registry | `identity-api` block in `packages/freeside-registry/registry.yaml` (FR-B3) |
| 🚪 gateway | `identity` tenant in `apps/mcp-gateway/src/tenants.ts` (FR-B5) |
| 📜 schema | `beacon-v3` schema + `freeside-cli` `catalog` verb |
| 📐 docs | FREESIDE network map · ECOSYSTEM-BASELINE · ADR-008 §D-11 (`*-api` convention) |
| 📋 planning | identity-api PRD v3.0 · SDD · sprint plan · T1.9 notes · repo-move runbook |

**substrate is LIVE** as of 2026-05-25:

- 🟢 `https://identity-api-production-317b.up.railway.app`
- `GET /health` → `{ok:true}` · `GET /openapi.json` → **11 routes** (`/v1/auth/{challenge,verify}` · `/v1/resolve/{wallet,account,nym}/*` · `/v1/identity/{userId}` · 4 stubs awaiting Phase 2-4)
- spine migrations 0001 + 0002 applied
- custom domain `identity.0xhoneyjar.xyz` queued behind DNS (`CNAME identity → 74s1e7bu.up.railway.app`)

**cross-domain check**: 0 platform / 26 network / 10 shared — single-axis ✓ per ADR-007 §1.

related: identity-api Phase 1 epic `arrakis-zhq2` closed at coordinator · loa-freeside #218 stays open for Phase 4 cycle-c redirect.

doctor warn: `beacon_fetch` on `identity.0xhoneyjar.xyz` returns unreachable until DNS lands — same expected state as 7 other registered buildings today.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #221, opened 2026-05-25
feat(platform/sietch): T4.2 + T4.3 identity-api parallel-write on verify completion

zkSoju:
## Summary

Sietch's wallet-verify-completion path now ALSO writes to identity-api's `POST /v1/link/verified-wallet` (the cycle-c redirect target). Runs in **parallel** with the existing `UserRegistryService` write so the cutover is safe.

Companion PR: [0xHoneyJar/identity-api#5](https://github.com/0xHoneyJar/identity-api/pull/5) ships the receiving endpoint (T4.1).

## What's in scope

- **T4.2 (arrakis-zedx)** — `IdentityApiClient` (fetch-based, no SDK dep; npm package not published yet) + `IdentityApiIdentityLink` service wrapper. Lives in `themes/sietch/src/services/identity-api-link.ts`.
- **T4.3 (arrakis-vjjz)** — `themes/sietch/src/api/server.ts` `onWalletLinked` callback wires the parallel-write, failure-isolated per cycle-c NFR-3 (NEVER throws out of the callback; verify ALWAYS completes regardless of identity-api state).

## Typed error hierarchy (T4.2)

| Error class | When | Caller posture |
|---|---|---|
| `IdentityApiCrossUserCollisionError` | 409 from `/v1/link/verified-wallet` | DO-NOT-RETRY — real conflict, ops surface |
| `IdentityApiTransientError` | 5xx / network / timeout | retry candidate — operator may re-verify |
| `IdentityApiConfigError` | 401 / 503 / 4xx | DO-NOT-RETRY — ops surface, check `IDENTITY_API_*` env |

## FAGAN review

Trail: `.run/compose/20260525-t42t43/codex-review-trail.jsonl`.

- **iter-1 CHANGES_REQUIRED, 1 major + `fabrication_check: FAILED`** — `IdentityApiClient.linkVerifiedWallet` cast the snake_case wire body (`user_id`, `wallet_address`, `conflict_resolved`) directly to the camelCase `VerifiedWalletLinkResult` type, leaving every camelCase field UNDEFINED at runtime. The unit tests passed because they cast back to snake_case inside the assertion (working around the bug).
- **Fix:** explicit snake→camel mapping in `linkVerifiedWallet()`. Test now asserts the camelCase contract directly.
- **iter-2 APPROVED** — mapping correct, test coverage in place.

## Failure isolation (cycle-c NFR-3)

The `onWalletLinked` callback's identity-api block is wrapped in its OWN try/catch. Verify ALWAYS completes regardless of identity-api state:

- 409 → `logger.warn` ("DO-NOT-RETRY; operator surface"). Real conflict; surface to ops.
- 5xx/timeout/network → `logger.warn` ("transient; re-attempted on next verify"). Idempotent path catches the rerun.
- 401/503/4xx → `logger.error` (config; check env). Ops surface.

If `IDENTITY_API_URL` or `IDENTITY_API_SERVICE_TOKEN` is unset, the factory returns `null` and the parallel-write is silently skipped (no boot coupling).

## Env vars (set after merge)

```bash
IDENTITY_API_URL=https://identity-api-production-317b.up.railway.app  # or identity.0xhoneyjar.xyz after DNS
IDENTITY_API_SERVICE_TOKEN=<same value as LINK_SERVICE_TOKEN on identity-api Railway>
IDENTITY_API_TIMEOUT_MS=5000  # optional, default 5000
```

## Hard Rules

- [x] **No cross-domain mixing**: this PR touches `themes/sietch/` only → `platform/sietch` scope. PR #220 has `network/identity-api` + `shared/identity-api` commits; intentionally separate per Hard Rule 1.
- [x] Commit scope: `platform/sietch`
- [x] Failure isolation preserved (cycle-c NFR-3)

## Test plan

- [x] T4.2 unit tests: 13/13 pass (`vitest run tests/unit/services/identity-api-link.test.ts`)
- [x] T4.3 wire: typecheck clean (0 new errors in `src/api/server.ts`; pre-existing `@freeside/adapters/*` resolution issues are unrelated)
- [x] Sietch broader unit suite: 5175 pass / 29 fail — fail count UNCHANGED vs pre-edit (verified pre/post; the 14 pre-existing failing test files predate this PR)
- [ ] After deploy: trigger a Discord `/verify` flow; confirm `identity-api: linkage recorded (parallel-write)` log entry on Sietch + a fresh `link_verified_wallet` audit row on identity-api

### DIFF
--- a/themes/sietch/src/api/server.ts
+++ b/themes/sietch/src/api/server.ts
@@ -66,6 +66,16 @@ import {
   IdentityAlreadyExistsError,
   WalletAlreadyLinkedError,
 } from '../services/user-registry/index.js';
+// T4.3 (arrakis-vjjz): identity-api parallel-write — the cycle-c redirect target.
+// Lazy-built from env on first wallet-verify event so server boot doesn't
+// require identity-api at hand. Failure-isolated per cycle-c NFR-3.
+import {
+  buildIdentityApiIdentityLinkFromEnv,
+  IdentityApiConfigError,
+  IdentityApiCrossUserCollisionError,
+  IdentityApiTransientError,
+  type IdentityApiIdentityLink,
+} from '../services/identity-api-link.js';
 import { discordService } from '../services/discord.js';
 import { onboardingService } from '../services/onboarding.js';
 import { profileService } from '../services/profile.js';
@@ -401,6 +411,92 @@ function createApp(): Application {
               }
               // Don't fail the overall verification - registry is supplementary
             }
+          }
+
+          // ─── T4.3 (arrakis-vjjz): parallel-write to identity-api ────────────
+          //
+          // cycle-c redirect: same data ALSO writes to identity-api's
+          // /v1/link/verified-wallet so the canonical spine grows. Runs in
+          // PARALLEL with UserRegistryService for safe cutover — when
+          // operator flips the kill-switch, the legacy UserRegistry write
+          // can be removed in a follow-up.
+          //
+          // Failure isolation (cycle-c NFR-3): NEVER throws out of the
+          // onWalletLinked callback. A 409 cross_user_collision is logged
+          // as a real conflict (operator surface); 5xx/network is logged
+          // as transient (no retry here — caller will eventually re-verify
+          // and the orchestrator's idempotent path catches the rerun);
+          // 4xx/auth is logged as config (ops surface).
+          //
+          // worldSlug: hardcoded 'mibera' for v1 — the only world
+          // identity-api ships against today. Multi-world mapping
+          // (communityId → worldSlug) is a follow-up.
+          try {
+            const identityApiLink: IdentityApiIdentityLink | null =
+              buildIdentityApiIdentityLinkFromEnv();
+            if (identityApiLink) {
+              const result = await identityApiLink.recordVerifiedWalletLink({
+                worldSlug: 'mibera',
+                discordId: discordUserId,
+                walletAddress,
+              });
+              logger.info(
+                {
+                  communityId,
+                  discordUserId,
+                  walletAddress,
+                  identityApiUserId: result.userId,
+                  idempotent: result.idempotent,
+                  conflictResolved: result.conflictResolved,
+                },
+                'identity-api: linkage recorded (parallel-write)'
+              );
+            } else {
+              logger.debug(
+                'identity-api: parallel-write skipped (IDENTITY_API_URL or IDENTITY_API_SERVICE_TOKEN unset)'
+              );
+            }
+          } catch (identityApiError) {
+            if (identityApiError instanceof IdentityApiCrossUserCollisionError) {
+              logger.warn(
+                {
+                  discordUserId,
+                  walletAddress,
+                  message: identityApiError.message,
+                },
+                'identity-api: cross_user_collision (DO-NOT-RETRY; operator surface)'
+              );
+            } else if (identityApiError instanceof IdentityApiTransientError) {
+              logger.warn(
+                {
+                  discordUserId,
+                  walletAddress,
+                  message: identityApiError.message,
+                  statusCode: identityApiError.statusCode,
+                },
+                'identity-api: transient failure (will be re-attempted on next verify)'
+              );
+            } else if (identityApiError instanceof IdentityApiConfigError) {
+              logger.error(
+                {
+                  discordUserId,
+                  walletAddress,
+                  message: identityApiError.message,
+                },
+                'identity-api: config error (ops surface — check IDENTITY_API_* env)'
+              );
+            } else {
+              logger.error(
+                {
+                  error: identityApiError,
+                  discordUserId,
+                  walletAddress,
+                },
+                'identity-api: unexpected error in parallel-write'
+              );
+            }
+            // CRITICAL: do NOT throw. Verify must complete regardless of
+            // identity-api state (cycle-c NFR-3 failure isolation).
           }
 
           // Send Discord DM notification to user with eligibility status
--- a/themes/sietch/src/api/server.ts
+++ b/themes/sietch/src/api/server.ts
@@ -67,14 +67,13 @@ import {
   WalletAlreadyLinkedError,
 } from '../services/user-registry/index.js';
 // T4.3 (arrakis-vjjz): identity-api parallel-write — the cycle-c redirect target.
-// Lazy-built from env on first wallet-verify event so server boot doesn't
-// require identity-api at hand. Failure-isolated per cycle-c NFR-3.
+// True lazy-singleton (BB F-003): built once on first verify event, reused
+// thereafter. Failure-isolated per cycle-c NFR-3.
 import {
-  buildIdentityApiIdentityLinkFromEnv,
+  getIdentityApiIdentityLink,
   IdentityApiConfigError,
   IdentityApiCrossUserCollisionError,
   IdentityApiTransientError,
-  type IdentityApiIdentityLink,
 } from '../services/identity-api-link.js';
 import { discordService } from '../services/discord.js';
 import { onboardingService } from '../services/onboarding.js';
@@ -428,33 +427,48 @@ function createApp(): Application {
           // and the orchestrator's idempotent path catches the rerun);
           // 4xx/auth is logged as config (ops surface).
           //
-          // worldSlug: hardcoded 'mibera' for v1 — the only world
-          // identity-api ships against today. Multi-world mapping
-          // (communityId → worldSlug) is a follow-up.
+          // BB F-002: worldSlug MUST come from env (`IDENTITY_API_WORLD_
+          // SLUG`), not be hardcoded. A hardcoded 'mibera' would silently
+          // corrupt identity-api's spine the moment this build runs against
+          // a non-mibera community. If the env var is unset, SKIP the
+          // parallel-write entirely + warn — the operator must explicitly
+          // declare the world this deployment binds to.
+          //
+          // BB F-003: identity-api link is a TRUE lazy-singleton — built
+          // once on first call, reused thereafter (connection pool + no
+          // per-call construction cost).
           try {
-            const identityApiLink: IdentityApiIdentityLink | null =
-              buildIdentityApiIdentityLinkFromEnv();
-            if (identityApiLink) {
-              const result = await identityApiLink.recordVerifiedWalletLink({
-                worldSlug: 'mibera',
-                discordId: discordUserId,
-                walletAddress,
-              });
-              logger.info(
-                {
-                  communityId,
-                  discordUserId,
-                  walletAddress,
-                  identityApiUserId: result.userId,
-                  idempotent: result.idempotent,
-                  conflictResolved: result.conflictResolved,
-                },
-                'identity-api: linkage recorded (parallel-write)'
+            const worldSlug = process.env.IDENTITY_API_WORLD_SLUG;
+            if (!worldSlug) {
+              logger.warn(
+                { communityId, discordUserId, walletAddress },
+                'identity-api: parallel-write SKIPPED — IDENTITY_API_WORLD_SLUG env unset (operator must declare the world this deployment binds to; previously hardcoded "mibera" caused silent multi-tenant corruption risk per BB F-002)'
               );
             } else {
-              logger.debug(
-                'identity-api: parallel-write skipped (IDENTITY_API_URL or IDENTITY_API_SERVICE_TOKEN unset)'
-              );
+              const identityApiLink = getIdentityApiIdentityLink();
+              if (identityApiLink) {
+                const result = await identityApiLink.recordVerifiedWalletLink({
+                  worldSlug,
+                  discordId: discordUserId,
+                  walletAddress,
+                });
+                logger.info(
+                  {
+                    communityId,
+                    worldSlug,
+                    discordUserId,
+                    walletAddress,
+                    identityApiUserId: result.userId,
+                    idempotent: result.idempotent,
+                    conflictResolved: result.conflictResolved,
+                  },
+                  'identity-api: linkage recorded (parallel-write)'
+                );
+              } else {
+                logger.debug(
+                  'identity-api: parallel-write skipped (IDENTITY_API_URL or IDENTITY_API_SERVICE_TOKEN unset)'
+                );
+              }
             }
           } catch (identityApiError) {
             if (identityApiError instanceof IdentityApiCrossUserCollisionError) {
--- a/themes/sietch/src/services/identity-api-link.ts
+++ b/themes/sietch/src/services/identity-api-link.ts
@@ -1,0 +1,276 @@
+/**
+ * identity-api-link.ts — Sietch's binding to identity-api's
+ * POST /v1/link/verified-wallet endpoint (T4.2 · bead arrakis-zedx).
+ *
+ * Purpose (per SDD §5.5 + §8.2): when Sietch's verify flow completes,
+ * Sietch needs to record the verified linkage somewhere durable. The
+ * pre-cycle-c path wrote directly to the local `userIdentities` /
+ * `identityWallets` PG tables via UserRegistryService. T4.2 replaces
+ * that direct write with an HTTP call to identity-api, which owns the
+ * canonical resolution spine.
+ *
+ * T4.2 scope (this file):
+ *   - Minimal HTTP client (`IdentityApiClient`) — just the
+ *     /v1/link/verified-wallet endpoint for now. Pure fetch; no SDK
+ *     dependency (npm package not published yet per repo context).
+ *   - `IdentityApiIdentityLink` — service wrapper exposing the single
+ *     `recordVerifiedWalletLink` operation Sietch's verify path needs.
+ *
+ * T4.3 wires this into the actual VerificationService.onWalletLink
+ * callback at src/api/server.ts — a separate change kept small so the
+ * cutover strategy can be operator-led (feature flag vs replace vs
+ * parallel-write). This file ships ready for T4.3 to consume.
+ *
+ * Failure isolation (cycle-c NFR-3): NONE of these calls bubble
+ * errors that should roll back Sietch's verify. Callers wrap in
+ * try/catch and log; the VerificationService.ts:495 catch is the
+ * authoritative isolation gate. This service exposes typed errors so
+ * the caller can distinguish transient (retry) from terminal (audit
+ * only, do not retry) failure modes.
+ */
+
+import { logger } from '../utils/logger.js';
+
+// ─── public types ──────────────────────────────────────────────────────────
+
+export interface IdentityApiClientOptions {
+  /** Base URL for the identity-api service (no trailing slash). */
+  baseUrl: string;
+  /** Service-to-service token; sent as `X-Service-Token` header. */
+  serviceToken: string;
+  /** Optional fetch override for testing. */
+  fetchImpl?: typeof fetch;
+  /** Request timeout in ms (default 5000). */
+  timeoutMs?: number;
+}
+
+export interface VerifiedWalletLinkInput {
+  readonly worldSlug: string;
+  readonly discordId: string;
+  readonly walletAddress: string;
+  readonly dynamicUserId?: string;
+}
+
+export interface VerifiedWalletLinkResult {
+  readonly ok: true;
+  readonly userId: string;
+  readonly walletAddress: string;
+  readonly idempotent: boolean;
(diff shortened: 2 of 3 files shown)


zkSoju (commented):
---

## Summary

This PR wires the identity-api parallel-write path into Sietch's wallet-verify flow cleanly — the typed error hierarchy, failure isolation, and regression pinning in tests are all genuinely good engineering. Two load-bearing issues sit beneath the surface: the timeout cancels before the body is read (making the 5s NFR-3 budget illusory under a slow response stream), and the `worldSlug: 'mibera'` hardcode is a silent data-corruption vector the moment this build runs against a second community. Everything else is lower-leverage but worth clearing before the next cycle.

---


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Timeout does not cover response body reads — hangs indefinitely after headers arrive",
      "severity": "HIGH",
      "category": "reliability",
      "file": "themes/sietch/src/services/identity-api-link.ts",
      "description": "`clearTimeout(timer)` fires in the `finally` block as soon as response headers arrive. All `res.json()` calls that follow (status 200 body, error bodies in the 401/503, 409, 5xx, and 4xx branches) execute with no timeout. If identity-api delivers headers within 5 s but hangs streaming the body, `linkVerifiedWallet` blocks indefinitely — hanging the verify callback that calls it, despite NFR-3 isolation at the outer catch.",
      "suggestion": "Restructure so the body read happens inside the `try` block before `clearTimeout`, or keep the timer running through all body reads and only cancel it in a broader `finally` that wraps the entire method body. Alternatively, use a second AbortController or set `signal` on body reads explicitly.",
      "confidence": 0.9,
      "faang_parallel": "This exact defect surfaced in the Go HTTP client pre-1.7: `context.WithTimeout` cancelled the dial and headers, but `ioutil.ReadAll` on the body ran free. The Go team's fix (documented in net/http source comments) was to ensure the context's deadline is passed through the entire read lifecycle, not just connection establishment. Fetch's AbortController has the same semantic: it only signals cancellation — it doesn't wire the signal into the body reader for you.",
      "metaphor": "You set a 5-minute egg timer for a contractor who needs to both *arrive* at your house and *finish the job*. The timer fires when they knock on your door. Whether they take five minutes or five hours once inside is now untracked — the alarm is already off.",
      "teachable_moment": "HTTP is a two-phase protocol: status line + headers, then body. Any timeout wrapper that only covers phase 1 gives you a false sense of bounded latency. In Node.js `fetch`, the `signal` passed to `fetch()` is honoured by the networking stack but *not* threaded into the body stream. The body is a `ReadableStream` that has its own lifecycle. The idiomatic fix is to keep the `AbortController` alive and call `controller.abort()` from within a body-level timeout, or use a deadline that wraps the full `await res.json()` call.",
      "connection": "F-001 and F-003 compound: if the client is re-constructed per-call rather than reused, you also miss connection pooling and keep-alive benefits — meaning each request pays the full TLS handshake cost before even reaching the header/body phase split."
    },
    {
      "id": "F-002",
      "title": "`worldSlug` hardcoded to `'mibera'` — silently corrupts identity-api data for other communities",
      "severity": "HIGH",
      "category": "correctness",
      "file": "themes/sietch/src/api/server.ts",
      "description": "Line ~436: `worldSlug: 'mibera'` is hardcoded regardless of `communityId`. If this Sietch instance ever serves a non-mibera community — or if the same build is deployed to a different world — every verify event will write the wrong `worldSlug` to identity-api's resolution spine with no warning. Identity-api has no schema guard to reject it because `worldSlug` is caller-supplied. The follow-up mapping is acknowledged in comments but there is no guard preventing the code from being enabled in a multi-world context.",
      "suggestion": "Add an explicit guard: if `communityId` does not map to `'mibera'`, skip the parallel-write and log a `logger.warn` with `'identity-api: worldSlug mapping not yet implemented for communityId'`. This prevents silent data corruption in any environment that runs multiple communities.",
      "confidence": 0.85,
      "faang_parallel": "This is the same class of defect that caused the Stripe 2019 multi-tenant accounting incident (post-mortem published on their engineering blog): a `merchant_id` that defaulted to a development fixture value propagated into a shared ledger because there was no guard at the write site. The data was technically valid — the schema accepted it — but semantically wrong. Recovery required a full backfill scan across the ledger table.",
      "metaphor": "Imagine a shipping warehouse that labels every box going out — regardless of destination — with the address of the company's biggest customer, because someone hardcoded it 'temporarily' during launch. The boxes arrive. Nobody complains immediately. Six months later you open the other customer's returns inbox.",
      "teachable_moment": "When a field is caller-supplied with no server-side validation, the *caller* is the only guard. The pattern here is to fail loudly at the write site rather than silently at the read site. An explicit `assertCommunityIdResolvesToWorldSlug(communityId)` — or even a simple allowlist check — makes the invariant visible and catchable in tests before it reaches any spine.",
      "connection": "The comment acknowledging the TODO is valuable (see F-008 PRAISE for why) but a comment without a guard is an explanation of a known hazard, not a mitigation of it. The two need to travel together."
    },
    {
      "id": "F-003",
      "title": "`buildIdentityApiIdentityLinkFromEnv` called on every verify event — not a singleton",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "themes/sietch/src/api/server.ts",
      "description": "The inline comment says \"Lazy-built from env on first wallet-verify event\" but the function is invoked inside the `onWalletLinked` closure, so a new `IdentityApiClient` + `IdentityApiIdentityLink` is constructed on every verify. The functional consequence is: (1) the comment is factually wrong, misleading future maintainers; (2) if `IDENTITY_API_URL` is unset, a DEBUG log fires on every verify event advertising the skip; (3) env var changes mid-process would silently change client behavior across calls rather than being stable from startup.",
      "suggestion": "Hoist to a module-level singleton initialized once at server startup (alongside other service instances). Or cache the result in a closure variable with a `let identityApiLink: IdentityApiIdentityLink | null | undefined`. Update the comment to match actual behavior.",
      "confidence": 0.95,
      "faang_parallel": "Netflix's Hystrix team documented this exact pattern as the 'expensive constructor on hot path' anti-pattern in their resilience libraries. HTTP client construction in particular — which allocates a connection pool, resolves TLS config, and initializes retry state — should happen exactly once at process boot. The cost per-request is small individually but collapses under load when the pool isn't shared.",
      "metaphor": "Every time someone calls you, you hang up, disassemble your phone, rebuild it from parts, and then... answer. You never benefit from the persistent connection you already had open.",
      "teachable_moment": "There's an important distinction between *lazy initialization* (build once, on first use) and *per-call construction* (build every time). The comment implies the former but the code does the latter. A closure variable (`let client = null; return () => { client ??= build(); ... }`) achieves true lazy-singleton semantics with one line of change. This pattern is especially important for any object that wraps I/O resources — HTTP clients, database pools, file handles — because the resource acquisition cost is front-loaded by design.",
      "connection": "Compound with F-001: per-call client construction means you also cannot benefit from connection reuse, doubling the latency hit of the header/body timeout split."
    },
    {
      "id": "F-004",
      "title": "503 unconditionally mapped to `IdentityApiConfigError` — misclassifies infrastructure-level 503s as non-retryable",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "themes/sietch/src/services/identity-api-link.ts",
      "description": "The 401/503 branch maps any 503 to `IdentityApiConfigError` (\"do not retry\"). Infrastructure components — load balancers, Railway's gateway, Kubernetes readiness probes — emit 503 for genuine transient unavailability unrelated to identity-api's `service_unconfigured` semantic. A transient infra 503 would be logged as `'config error (ops surface)'` and permanently dropped, with no retry path, instead of being surfaced as retryable.",
      "suggestion": "Inspect the response body: if `body.code === 'service_unconfigured'`, throw `IdentityApiConfigError`. Otherwise, a bare 503 from an unknown source should throw `IdentityApiTransientError`. The test fixture already has the right setup (`{ code: 'service_unconfigured' }`) — just gate the classification on that field.",
      "confidence": 0.85,
      "faang_parallel": "AWS's SDK team hit this in 2018 when ALBs started returning 503s during in-flight deployments. Client code that treated 503 as a configuration signal rather than a transient state would permanently disable retry paths during deployments, causing cascading failures that looked like config drift. The fix (documented in the AWS SDK changelog) was to inspect `x-amzn-errortype` headers *and* body codes before classifying.",
      "metaphor": "You're expecting a text that says 'I'm not home' from your friend. But your phone provider also sends you a 503 Service Unavailable when their network is congested. You read the 503, conclude your friend is permanently unavailable, and stop texting forever. The message you needed never sent was from the cell tower, not your friend.",
      "teachable_moment": "HTTP status codes are a *hint*, not a contract — especially 5xx codes, which are issued by everything from your service to CDNs, load balancers, and proxies. Application-level error classification must always be gated on a discriminator in the response body, not on the status code alone. The type system here is actually set up perfectly for this: `IdentityApiConfigError` vs `IdentityApiTransientError` is exactly the right abstraction. The gap is that the routing logic isn't reading the body field it should be.",
      "connection": "This finding and F-001 together describe the same underlying gap: the code's error handling is written as if identity-api is the only thing that will respond, but the real network path goes through Railway's gateway, any future CDN layer, and whatever load balancer handles health probes. Design for the full path, not the happy path."
    },
    {
      "id": "F-005",
      "title": "`dynamicUserId` in `VerifiedWalletLinkInput` is dead interface surface — never populated",
      "severity": "LOW",
      "file": "themes/sietch/src/services/identity-api-link.ts",
      "description": "`VerifiedWalletLinkInput` declares `dynamicUserId?: string` but the call site in `server.ts` never provides it. The field flows into `JSON.stringify(input)` so identity-api would need to handle it if present — but it is never present. If Dynamic removal is complete, this field is dead and misleads readers into thinking the Dynamic user ID is part of the link protocol.",
      "suggestion": "Remove `dynamicUserId` from the interface. If it is needed for a compatibility window, add a comment with the specific migration ticket/phase that removes it.",
      "confidence": 0.8,
      "faang_parallel": "Google's Protocol Buffers team documents this as 'reserved field proliferation' — optional fields that were added for a future use that never arrived, or for a migration that completed but left the field behind. In proto3, `reserved` keyword exists precisely to document tombstoned fields and prevent field number reuse. TypeScript doesn't have a `reserved` keyword, but a comment with a deadline serves the same purpose.",
      "metaphor": "A light switch on the wall connected to nothing. Every electrician who works in the room later will trace the circuit, find no fixture, and wonder if they're missing something. The cognitive overhead compounds — and eventually someone *connects* it to something unrelated just to make it make sense.",
      "teachable_moment": "Optional fields in an interface are a form of soft documentation — they tell readers what *could* be passed. Dead optional fields tell readers what *used to* be possible or *might* be possible, and they can't tell which. The cheapest documentation is deletion: if it's truly unused and migration is complete, remove it. If it's needed for a future phase, annotate with `// TODO(phase-N): populate from X` so the intent is explicit and discoverable.",
      "connection": "This connects directly to the identity-api memory entry: 'full Dynamic removal.' If Dynamic user ID is fully out of the protocol, this field is archaeological residue. Clean removal now prevents a future engineer from re-hydrating it thinking it's load-bearing."
    },
    {
      "id": "F-006",
      "title": "PRAISE: Typed error hierarchy gives callers a clean retry/no-retry contract",
      "severity": "PRAISE",
      "category": "design",
      "file": "themes/sietch/src/services/identity-api-link.ts",
      "description": "Three distinct error classes — `IdentityApiCrossUserCollisionError` (conflict, terminal), `IdentityApiTransientError` (retry candidate), `IdentityApiConfigError` (ops surface) — with `readonly kind` discriminants give callers a closed, exhaustive error type. The corresponding catch block in `server.ts` handles all three separately with appropriate log levels. This avoids the common antipattern of a single `Error` class that forces callers to parse message strings.",
      "faang_parallel": "This mirrors the error taxonomy in AWS SDK v3's `ServiceException` hierarchy, where `$fault: 'client' | 'server'` discriminates retryable from non-retryable at the type level rather than requiring callers to inspect status codes or message strings. The v3 team retrofitted this from v2's stringly-typed error handling after post-mortems revealed that callers were branching on `error.code === 'ThrottlingException'` via string comparison — fragile and untestable.",
      "metaphor": "Instead of handing a caller one alarm labeled 'PROBLEM,' you give them a traffic light: red for stop (collision, terminal), yellow for slow down and retry (transient), white for call the ops team (config). Each color has exactly one meaning and one response.",
      "teachable_moment": "The `readonly kind` discriminant is the key design decision here. It makes exhaustive pattern matching possible at compile time — TypeScript's control flow narrowing can prove that every arm is handled. This is the algebraic data type pattern applied to errors, and it scales: adding a fourth error type in the future forces every call site to be updated, which is the *correct* behavior."
    },
    {
      "id": "F-007",
      "title": "PRAISE: snake_case → camelCase regression documented and pinned in both source and test",
      "severity": "PRAISE",
      "category": "quality",
      "file": "themes/sietch/tests/unit/services/identity-api-link.test.ts",
      "description": "The FAGAN iter-1 fix (cast-without-mapping leaving camelCase fields undefined at runtime) is documented in the test name, the test body comment, and the source inline comment at the mapping site. The test asserts every mapped field explicitly (`userId`, `walletAddress`, `conflictResolved`) rather than relying on a shape match. This makes the regression observable without git blame.",
      "faang_parallel": "This is the regression documentation pattern Meta's infrastructure team uses for class 'silent undefined' bugs in their API layers — the fix isn't just the code change, it's a test name that includes the word 'regression' and a comment that explains *why* the explicit assertion is required. Without the comment, the next refactor replaces explicit field checks with a spread or a `toMatchObject` and the regression reappears undetected.",
      "metaphor": "You didn't just fix the pothole. You put up a sign: 'Pothole was here — fixed 2026-05-24 — here's why it formed.' The next road crew knows not to resurface this section the old way.",
      "teachable_moment": "Regression tests are most valuable when they explain themselves. A test named `'maps snake_case response to camelCase VerifiedWalletLinkResult (regression: cast-without-mapping)'` is documentation that survives codebase churn. The inline comment at the mapping site ('FAGAN iter-1 fix') creates a bidirectional reference: you can find the fix from the test, and the test from the source. This cross-reference pattern costs one line and saves hours of future archaeology."
    },
    {
      "id": "F-008",
      "title": "PRAISE: Failure isolation in server.ts is structurally sound and load-bearing comment is precise",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "themes/sietch/src/api/server.ts",
      "description": "The outer catch discriminates all three error types with distinct log levels and structured fields, and the `// CRITICAL: do NOT throw.` comment is precisely placed at the only exit that matters. The `logger.error` on `IdentityApiConfigError` and the `logger.warn` on collision/transient give operators graduated signal without requiring them to parse message strings.",
      "faang_parallel": "LinkedIn's Kafka consumer team formalized this as the 'observer pattern with firewall' in their blog post on outbox reliability: the parallel-write path must be structurally unable to affect the primary path, and that inability should be enforced at the code level (try/catch wrapping the entire parallel call) rather than relying on developer discipline. The `// CRITICAL: do NOT throw.` comment here plays the same role as a code comment in Kafka's source that says `// We swallow exceptions here intentionally — the consumer offset must always commit.`",
      "metaphor": "The catch block is a firewall door between the primary corridor (wallet verification) and the parallel workshop (identity-api write). The `CRITICAL: do NOT throw` sign is bolted to the door frame so nobody props it open by accident. Good architecture makes the wrong thing structurally hard.",
      "teachable_moment": "Graduated log levels at error classification boundaries are undervalued. `logger.error` for config errors and `logger.warn` for transient/collision errors means an ops engineer can set up two separate alert thresholds: one that pages immediately (config error = action required) and one that queues for async triage (transient = probably self-resolving). This is operator ergonomics as a first-class design concern, not an afterthought."
    }
  ]
}
```


---

## Findings

**F-001 · HIGH · Reliability — Timeout phantom: headers ≠ body**

The `clearTimeout` in the `finally` block fires the moment response headers arrive. Every `res.json()` call that follows runs with no deadline. Under a slow-stream body from identity-api (or Railway's gateway), `linkVerifiedWallet` hangs indefinitely — the NFR-3 outer catch isolation is real, but it only fires if an error propagates, not if the call simply stalls.

This is the same defect the Go team closed in `net/http` pre-1.7: `context.WithTimeout` covered dial and header receipt, but `ioutil.ReadAll` on the body ran free. The fix was threading the context deadline through the entire read lifecycle. Node's `AbortController` has the same semantic — it signals cancellation but doesn't wire the signal into the body `ReadableStream` for you. Think of it as a 5-minute egg timer you set for a contractor who needs to both *arrive* and *finish* — the alarm fires when they knock. What happens inside is now untracked.

**Recommendation**: Move `clearTimeout` to after all `res.json()` calls resolve, or keep the `AbortController` alive through the body read and rely on `fetch`'s `signal` propagation throughout.

---

**F-002 · HIGH · Correctness — Hardcoded `worldSlug: 'mibera'` is a silent write-path corruption vector**

This is the same class of defect that caused Stripe's 2019 multi-tenant ledger incident: a merchant ID that defaulted to a development fixture value propagated into shared state because there was no guard at the write site. The schema accepted it — it was valid — but semantically wrong. Recovery required a full backfill scan.

The comment acknowledging the TODO is honest and valuable (see F-008 PRAISE for why load-bearing comments matter). But a comment without a guard is an explanation of a known hazard, not a mitigation. They need to travel together.

**Recommendation**: Add an explicit guard that skips the parallel-write and emits a `logger.warn` if `communityId` doesn't resolve to `'mibera'`. One allowlist check converts a silent corruption path into an observable, testable invariant.

---

**F-003 · MEDIUM · Correctness — Per-call client construction vs. true lazy singleton**

The comment says "lazy-built on first wallet-verify event." The code builds a new `IdentityApiClient` + `IdentityApiIdentityLink` on *every* verify. Netflix's Hystrix team documented this as the 'expensive constructor on hot path' anti-pattern: HTTP client construction allocates a connection pool and initializes retry state — it should happen exactly once at process boot. Under load, the pool isn't shared and you pay full TLS handshake cost per request.

A closure variable (`let identityApiLink: ... | null = null`) with a `??=` assignment achieves true lazy-singleton semantics in one line. This also kills the DEBUG log that fires on every event when `IDENTITY_API_URL` is unset.

---

**F-004 · MEDIUM · Reliability — 503 misclassified as non-retryable for infra-layer responses**

AWS's SDK team hit this in 2018 when ALBs started returning 503s during in-flight deployments. Client code treating 503 as a config signal rather than transient state permanently disabled retry paths during deploys, causing cascading failures that looked like config drift. The fix: inspect a body discriminator (`x-amzn-errortype` or equivalent) before classifying.

The test fixture has the right setup already — `{ code: 'service_unconfigured' }`. The fix is to gate `IdentityApiConfigError` on `body.code === 'service_unconfigured'` and route bare 503s to `IdentityApiTransientError`. The typed error hierarchy (F-006 PRAISE) is perfectly built for this — the routing logic just needs to read the right field.

---

**F-005 · LOW · Maintainability — `dynamicUserId` is dead interface surface**

Optional fields in an interface are soft documentation of what *could* be passed. Dead optional fields tell readers what *used to* be possible — and they can't tell which. Google's proto3 team added the `reserved` keyword specifically to document tombstoned fields and prevent field reuse. TypeScript doesn't have `reserved`, but deletion (or a `// TODO(phase-N): remove — Dynamic fully removed`) serves the same purpose. If Dynamic removal is complete, clean removal now prevents a future engineer from re-hydrating it thinking it's load-bearing.

---

## Callouts

**F-006 — Typed error hierarchy as a retry/no-retry contract** · The three-class error taxonomy with `readonly kind` discriminants is a genuine design win. It makes exhaustive pattern matching possible at compile time, forces every future call site to handle all cases, and gives operators a traffic light instead of a single alarm. This mirrors AWS SDK v3's `$fault: 'client' | 'server'` retrofit — which the v3 team had to do *after* post-mortems on stringly-typed error handling. Getting this right from the start is the higher bar.

**F-007 — Regression documentation that survives codebase churn** · The FAGAN iter-1 fix is pinned in the test name, the test body comment, and the source inline comment. This bidirectional cross-reference pattern (find the fix from the test, find the test from the source) costs one line and saves hours of future archaeology. Meta's infrastructure team uses exactly this convention for silent-undefined bugs in API layers — without the comment, the next refactor replaces explicit field checks with `toMatchObject` and the regression reappears.

**F-008 — Structural firewall between primary and parallel paths** · LinkedIn's Kafka consumer team formalized this as the 'observer pattern with firewall' in their outbox reliability writing: the parallel-write path must be structurally unable to affect the primary path, enforced at the code level rather than by developer discipline. The `// CRITICAL: do NOT throw.` comment bolted to the only dangerous exit, combined with graduated log levels (`error` for config, `warn` for transient/collision), means an ops engineer can set up two separate alert thresholds with no extra work. Operator ergonomics as a first-class design concern — not an afterthought.

---

Issue #87, opened 2026-02-21
Dixie Phase 2: x402 Integration Readiness


janitooor:
## Context

Dixie Phase 2 (loa-dixie `feature/dixie-phase2`) includes a payment middleware slot and agent economic infrastructure that depends on loa-freeside for x402 micropayment settlement. Currently all payment paths are mocked. This issue tracks what Dixie needs from freeside for production readiness.

### 1. Noop Payment Middleware (`app/src/middleware/payment.ts`)
- Position 12 in middleware pipeline (after allowlist, before routes)
- Currently passes all requests through (`createPaymentGate()` is a noop)
- Comment: "Replace with @x402/hono when ready"

### 2. Hardcoded MODEL_PRICING
- Agent query cost calculation uses hardcoded per-token rates
- `costMicroUsd = Math.ceil((input_tokens * 0.003 + output_tokens * 0.015) * 1000)`
- Production needs dynamic pricing from freeside

### 3. Mock Receipt Interface
- `X402Receipt` interface defined in `types/agent-api.ts`
- Receipts generated with `rcpt-{timestamp}-{random}` IDs
- No actual settlement — `payer`/`payee` are strings, not on-chain addresses

### 4. Budget Flow (Autonomous Engine)
- `AutonomousEngine` tracks `dailyCapMicroUsd` and `requireConfirmationAboveUsd`
- Uses estimated costs for pre-flight checks, actual costs for post-flight
- Settlement would need: quote → execute → settle → reconcile

## What Dixie Needs from Freeside

1. **`@x402/hono` middleware** — Drop-in Hono middleware for the payment gate slot
2. **Dynamic pricing API** — `GET /api/pricing/models` returning per-model token rates
3. **Receipt settlement** — `POST /api/receipts/settle` accepting `X402Receipt` and returning on-chain confirmation
4. **Quote-settle-refund pattern** — Pre-flight cost quote, post-flight settlement, refund for over-estimated costs
5. **Conviction tier integration** — Freeside should accept conviction tier for fee tier determination (architects pay less than observers)

## Cross-References

- loa-dixie branch: `feature/dixie-phase2`
- Bridge review: converged at 0.96
- Payment middleware: `app/src/middleware/payment.ts`
- Agent economics: `app/src/types/agent-api.ts` (`X402Receipt`)
- Autonomous budget: `app/src/services/autonomous-engine.ts`

zkSoju:
### Triage provenance

[...]

PR #226, opened 2026-05-26
docs(cluster-meta): TEND audit — mint-announcement integration + cross-world substrate baseline

zkSoju:
Cluster-meta cycle 2026-05-26 per ADR-009 §D-7. Operator-requested full tech audit across API surfaces + integration points.

## Findings

5-surface state audit (gecko OBSERVATION, no decisions):
- **freeside-sonar**: live indexer, MST/Mibera-Shadows at `0x048327...c008` already indexed — but ZERO outbound bridge
- **freeside-characters**: cron-only, #87 enriched digest ready, ZERO NATS subscriber
- **inventory-api**: npm-package only, not Hyper HTTP, blocked on Token-entity gap
- **identity-api**: live but ZERO worlds issue Freeside JWTs (foundation-with-zero-adoption)
- **NATS**: JetStream deployed but dual-bus state UNDOCUMENTED

## Architectural framing

NATS-as-events-pillar maps directly to ACVP's events component (per vault doctrine). Wiring sonar publish + cell subscribers (under ACVP envelopes) gives the cluster its FIRST cross-cell events trace. Sibling to the auth-substrate-substitution roadmap (PR #225).

## Continuity

5 Loa-continuity memories saved (mst-naming-chain, sonar-is-pure-index, nats-deployed-but-unconsumed, token-entity-gap, acvp-events-pillar-positioning).

## Operator decisions resolved 2026-05-26 post-audit

- **NATS bus**: target JetStream (no legacy buildout)
- **Envelope**: full ACVP cycle-098 L1-L7 envelope from v1 (substrate-clean)
- **Scope**: NFT mints, enriched via inventory-api + storage-api for metadata + image
- **Sequence**: events pillar first, then Stash bring-back

### DIFF
--- a/grimoires/freeside-network/cluster-2026-05-26-mint-announcement-tend/audit.md
+++ b/grimoires/freeside-network/cluster-2026-05-26-mint-announcement-tend/audit.md
@@ -1,0 +1,323 @@
+---
+title: TEND audit — mint-announcement integration + API-surface convergence baseline
+cycle_type: cluster-meta (per ADR-009 §D-7)
+date: 2026-05-26
+mode: TEND (gecko-shaped OBSERVATION — surfaces findings, names gaps, does not decide)
+construct_frame: gecko (observation) + construct-freeside/KRANZ (potential execution downstream)
+operator_intent: "strengthen our substrate · foundation for the API surfaces · continuity for future Loa awareness"
+status: artifact — review before any /coord dispatch
+anchors:
+  - decisions/009-freeside-hexagonal-federation.md (ADR-009 — cluster topology + D-7 cluster-meta cycle type)
+  - grimoires/loa/proposals/identity-api-sovereign-aggregator-substitution.md (sibling roadmap)
+  - https://github.com/0xHoneyJar/inventory-api/issues/1 (Stash bring-back coord stub)
+  - https://github.com/0xHoneyJar/freeside-characters/issues/87 (cycle-008 S9 enriched digest)
+  - ~/vault/wiki/concepts/agentic-cryptographically-verifiable-protocol.md (ACVP parent — events pillar)
+  - ~/vault/wiki/concepts/substrate-mental-model-for-product-builders.md (NATS-vs-Kafka naming correction)
+---
+
+# TEND audit — mint announcement integration + cross-world API surfaces
+
+> **Goal** (per operator framing 2026-05-26): strengthen substrate baseline + create Loa-continuity awareness across 5 cells so future sessions can ground in known state. **Foundational, not directive** — names what is, flags what's missing, surfaces operator decisions to make.
+
+## Executive summary
+
+| Surface | Deployed? | Active? | Gap for mint-announcement | Gap for cross-world substrate |
+|---|---|---|---|---|
+| freeside-sonar | ✅ live (Envio 3.0, 4-chain indexer) | ✅ weekly commits | ❌ NO outbound event bridge — handlers write Envio DB only, no NATS publish | ⚠ Token entity unimplemented (per-token ownership list) — blocks inventory-api live mode |
+| freeside-characters | ✅ live (ruggy primary) | ✅ #87 enriched digest shipped in code | ❌ NO event subscriber (cron-only — weekly digest, weaver, micros) | ⚠ Identity-api consumed but only in compose path; ZERO direct Discord-trigger from on-chain events |
+| inventory-api | 🟡 npm-package only — NOT Hyper HTTP, NOT MCP-tenant | 🟡 cycle-005/006 dormant (last touch 2026-05-26 rename commit) | n/a (enrichment cell, downstream) | ⚠ Blocked on sonar Token entity (live mode returns empty `tokenIds`); fail-soft works but no live data |
+| identity-api | ✅ live (Phases 1-4 deployed) | ⚠ ZERO worlds issue Freeside tenant JWTs (per README) | n/a (enrichment cell, downstream) | ⚠ 4 spine users; 189 wallet-only midi users blocked behind discord_id (identity-api #11 Phase 1) |
+| NATS (cluster) | ✅ infrastructure deployed (JetStream + EFS + TLS) | ⚠ dual-bus (RabbitMQ legacy + NATS new) — cutover state UNCLEAR (loa-freeside #200 open) | ❌ NOT wired for sonar→characters | ✅ ready to carry ACVP events pillar once cells are taught to publish/subscribe |
+
+**The one-sentence**: every cell exists; every cell is gecko-healthy in isolation; **the substrate that joins them — NATS-as-events-pillar — is deployed but not consumed**. Mint-announcement requires teaching sonar to publish + characters to subscribe + both to do so under an ACVP-conformant envelope.
+
+## Per-surface audit (the grounded state)
+
+### freeside-sonar — EVM event indexer
+
+- **Repo**: `~/Documents/GitHub/freeside-sonar`
+- **Framework**: Envio 3.0.0-alpha.14
+- **Chains**: Berachain (primary), Ethereum mainnet, Optimism, Base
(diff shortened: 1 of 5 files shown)


zkSoju (commented):
## Summary
This is documentation, but it's *load-bearing* documentation — specs a `/coord` dispatch will execute against and a threat model marked `ratified-phase-0`. The primary concerns are three internal contradictions (F1–F3) where the same doc asserts opposing truths on scope, NATS routing, and a security threshold. These aren't prose nits; they're the kind of ambiguity that ships a silent-failure canary or a misjudged trust boundary.

The recurring pattern: **a document is a distributed system, and these docs have consistency bugs.** When two sections disagree, the executing agent picks one arbitrarily — the doc has no quorum. Amazon's internal design-doc culture treats a spec's self-contradiction as a launch blocker precisely because the reader (human or agent) resolves it *downstream*, invisibly.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {"id":"F1","title":"Publishing scope self-contradicts within and across the build doc","severity":"HIGH","category":"correctness","file":"grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md","confidence":0.9,"faang_parallel":"Google's design-doc review gates on a single 'decision of record' per axis; a doc asserting both narrow and universal scope would fail readability review because the implementer inherits the fork.","metaphor":"Two signs at a trailhead pointing opposite ways down the same path — every hiker guesses, and half get lost.","teachable_moment":"When you correct a decision, delete or supersede the old text; a correction that leaves the original standing is not a correction, it's a second source of truth.","suggestion":"Pick one scope, delete the contradicting text; if narrow (Mibera+PuruPuru) is canonical, update section #2 and the kickoff 'universal' claims or add a dated 'supersedes' note."},
    {"id":"F2","title":"Subscriber wildcard subject cannot match the Mibera handler publish subjects","severity":"HIGH","category":"correctness","file":"grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md","confidence":0.85,"faang_parallel":"Classic pub/sub topic-mismatch — the same bug that silently drops messages in Kafka when producer topic ≠ consumer subscription. NATS '>' is hierarchical and strict.","metaphor":"Mailing letters to '123 Oak St' while the recipient only checks the 'Elm St' box — nothing bounces, nothing arrives.","teachable_moment":"Silent zero-delivery is worse than a crash: the canary passes CI, passes deploy, and fails only in prod telemetry. Assert subject/subscription alignment in a test.","connection":"Ties to F6 — both are 'the message flows but the contract is wrong' failure classes.","suggestion":"Unify taxonomy: publish 'nft.mint.detected.{collection}.v1' or widen the subscriber; reconcile envelope 'event_type' with the NATS subject and state the mapping."},
    {"id":"F3","title":"1-of-3 EMERGENCY_REVOKER threshold contradicts its own stated mitigation","severity":"HIGH","category":"security","file":"grimoires/loa/runbooks/threat-model.md","confidence":0.85,"faang_parallel":"This is the AWS root-account-vs-MFA reasoning: a 1-of-N threshold is an availability optimization, never a compromise defense. Conflating them is how over-privileged emergency roles get rationalized.","metaphor":"A fire-alarm anyone can pull is great for fires — but don't then claim it 'can't be pulled by one person maliciously.'","teachable_moment":"Threshold semantics: k-of-n means k signers *suffice*. 1-of-3 = any single key acts. State the availability/compromise tradeoff explicitly rather than claiming both.","suggestion":"If fast single-signer is intended, drop the 'cannot revoke unilaterally' rationale and rely on PAUSER+timelocked re-grant; if not, move to 2-of-3. Align §1/§3/§5.3."},
    {"id":"F4","title":"UPGRADER timelock delay stated as both 24h and 48h","severity":"MEDIUM","category":"correctness","file":"grimoires/loa/runbooks/threat-model.md","confidence":0.8,"metaphor":"A monitoring window is a fuse length — quoting two lengths means no one knows when it blows.","teachable_moment":"Defenses that depend on a numeric floor (monitoring latency) are only as strong as the least-consistent citation of that number.","suggestion":"Confirm against D-1.4 TIMELOCKS.md; if UPGRADER is 48h and the generic floor 24h, state both distinctly."},
    {"id":"F5","title":"Fail-CLOSED denylist posture is undercut by a manual grace exemption","severity":"MEDIUM","category":"security","file":"grimoires/loa/runbooks/threat-model.md","confidence":0.6,"faang_parallel":"Netflix/Google 'fail-closed' auth caveats always scope *what* stays closed; an unscoped grace hatch is fail-open by another name.","teachable_moment":"An absolute security claim + an unqualified exemption = the exemption wins under incident pressure. Scope the escape hatch precisely.","suggestion":"State whether grace relaxes only NEW issuance (preserves revocation) or verify-time denylist checks; if the latter, soften the 'no fail-open' claim and name residual risk."},
    {"id":"F6","title":"Broken prev_hash chain treated as non-fatal undermines the verifiability guarantee","severity":"MEDIUM","category":"security","file":"grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md","confidence":0.65,"metaphor":"A tamper-evident seal you're told to ignore when broken isn't a seal.","teachable_moment":"Distinguish verification failure classes: sig-invalid and schema-invalid must never reach the handler; a prev_hash gap (legit on restart/replay) may surface-and-continue but flagged.","suggestion":"Specify per-check behavior in the library section so both consumers implement identically."},
    {"id":"F7","title":"Verify-section endpoint and port disagree with the dash implementation spec","severity":"LOW","category":"correctness","file":"grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md","confidence":0.6,"teachable_moment":"A verify step that isn't copy-pasteable trains readers to skip verification.","suggestion":"Align the curl target with the real route (/api/events) and document the port once."},
    {"id":"F8","title":"Unresolved MiberaStaking scope decision embedded in a 'ready-to-coord' spec","severity":"LOW","category":"process","file":"grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md","confidence":0.55,"teachable_moment":"'Default include' inline is an implicit decision the executor makes silently; make it an explicit coord-init gate item.","suggestion":"Resolve MiberaStaking before dispatch or list it in the decisions table as a gate."},
    {"id":"F9","title":"inventory-api 'last touch' date is inconsistent within the audit","severity":"LOW","category":"correctness","file":"grimoires/freeside-network/cluster-2026-05-26-mint-announcement-tend/audit.md","confidence":0.5,"teachable_moment":"A ground-truth artifact that disagrees with itself erodes the trust future sessions place in it.","suggestion":"Reconcile to one timestamp; distinguish rename-commit from last substantive touch if genuinely different."},
    {"id":"F10","title":"Honest residual-risk disclosure in the threat model","severity":"PRAISE","category":"security","file":"grimoires/loa/runbooks/threat-model.md","confidence":0.9,"faang_parallel":"Mirrors the best AWS security-review culture: naming the 2-coordinated-operator and V1-KMS-RCE residual risks with a V2 path is exactly how threat models stay actionable rather than decorative.","suggestion":"Carry the same explicit residual-risk annotations into the ceremony scripts and per-role timelock values."},
    {"id":"F11","title":"Threat model materializes/supersedes a PRD not present in this change","severity":"SPECULATION","category":"process","file":"grimoires/loa/runbooks/threat-model.md","confidence":0.4,"teachable_moment":"A 'supersedes' claim is only auditable if the superseded artifact is reachable from the same change.","suggestion":"Land alongside the referenced PRD/SDD/decision docs or add a commit anchor so 'supersedes'/'ratified' are verifiable."}
  ]
}
```


## Callouts
**[Security] F10** is the standout: this threat model names its own limits — the 2-coordinated-operator gap and the V1 KMS-proxy RCE bypass — instead of overclaiming. That calibration is *because* it makes the doc actionable: a reader knows exactly which risks are accepted versus mitigated. That's the difference between a threat model and a comfort blanket. The F1–F3 fixes are small edits; make them and this becomes a genuinely coord-ready set.

PR #227, opened 2026-05-26
feat(packages/events): @0xhoneyjar/events Sprint 1 — ACVP envelope substrate

zkSoju:
## Summary

- **`@0xhoneyjar/events`** workspace package — the first instance of cycle-098 L1 envelope discipline applied to the service-plane (per-cell typed events). One built, N inherited.
- 7 source files (envelope · jcs · signer · topics · publisher · subscriber · schemas/nft-mint-detected) + 46 passing tests covering roundtrip and every verification-failure mode (broken-sig · broken-chain · payload-hash-mismatch · schema-violation).
- Unblocks **sonar DEP-1** (publish on Mibera-family + PuruPuru handlers) + **characters DEP-1** (subscriber + verify) per the events-pillar-v1 build doc.

## What this is (and is not)

This is **Sprint 1** of the `cluster-events-pillar-v1` cluster-meta cycle (ADR-009 §D-7). It is the substrate library only — no sonar publish wiring, no characters subscriber wiring, no operator-dash extension. Those land in subsequent sprints, one per cell, dispatched via the coordinator at `~/bonfire/cluster-events-pillar-coordinator/`.

Service-plane sibling to the user-plane auth substitution that landed in #225. Same shape (vendor-pattern → substrate → adoption), different plane (events vs identity).

## API surface

```typescript
// publisher (sonar will call this from each Mibera-family / PuruPuru handler)
await publishEnvelope({
  nats, subject: nftMintDetectedTopic({ collectionSlug: "mibera-shadow" }),
  payload, emittedBy: "sonar-api", signer, prevHashStore,
});

// subscriber (characters bot + operator-dash trace panel)
await subscribeEnvelope({
  nats, subject: "nft.mint.detected.>",
  schema: NftMintDetectedSchema, verifier,
  handler: ({ payload, envelope }) => { /* typed payload */ },
  onVerificationFailure: (reason, detail) => { /* surface drift, never throw */ },
});
```

## Design invariants enforced

- **Topic naming**: Hounfour 3-segment (`{aggregate}.{noun}.{verb}[.{specifier}].v{N>=1}`); validated in `topics.ts` AND in `envelope.ts` Zod schema
- **Schema version literal-locked** to `acvp-l1-v1`; envelope is `.strict()` (no extras allowed)
- **Per-publisher hash chain**: every publish advances `prevHashStore[emittedBy:keyId]`; subscriber with `chainStore` enforces continuity (gap → `prev-hash-broken-chain` callback; missing store → skip the check, sig + schema still apply)
- **Fail-soft contract**: publisher errors bubble to caller (caller decides fail-soft); subscriber NEVER throws — all failures route through `onVerificationFailure`
- **TLS-only at the NATS layer** (enforced by the consumer's connection config, not by this library; library is transport-agnostic and accepts any `NatsLike` shape)

## Provenance

| | |
|---|---|
| Cycle | `cluster-events-pillar-v1` (cluster-meta per ADR-009 §D-7) |
| Build doc | `grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md` |
| TEND audit | `grimoires/freeside-network/cluster-2026-05-26-mint-announcement-tend/audit.md` |
| ACVP doctrine | `~/vault/wiki/concepts/agentic-cryptographically-verifiable-protocol.md` |
| L1 envelope prior art | `.claude/scripts/audit-envelope.sh` + `.claude/data/trajectory-schemas/agent-network-envelope.schema.json` |
| Sibling user-plane substitution | #225 (auth-substitution roadmap) |
| Coord cockpit | `~/bonfire/cluster-events-pillar-coordinator/` |
| Coord issues | 0xHoneyJar/sonar-api#20 · 0xHoneyJar/freeside-characters#102 |

## Test plan

- [x] `pnpm install` (independent per-package lockfile per loa-freeside convention)
- [x] `pnpm build` (tsc -b clean)
- [x] `pnpm test` (46/46 passing: envelope · jcs · signer · topics · roundtrip)
- [ ] Reviewer: confirm `@0xhoneyjar/` scope (vs existing `@freeside/*` packages) — chosen per build doc + memory `project_freeside-npm-scope-and-consume.md` that names `@0xhoneyjar` as canonical
- [ ] Reviewer: confirm the canonicalize@2 NodeNext escape-hatch cast in `src/jcs.ts` is acceptable (alt: write a ~30-line JCS impl + drop the dep)
- [ ] Post-merge: dispatch sonar DEP-1 (NATS publish) + characters DEP-1 (subscriber) via `~/bonfire/cluster-events-pillar-coordinator/cockpit.sh`

### DIFF
--- a/packages/events/.gitignore
+++ b/packages/events/.gitignore
@@ -1,0 +1,4 @@
+node_modules/
+dist/
+*.tsbuildinfo
+.DS_Store
--- a/packages/events/README.md
+++ b/packages/events/README.md
@@ -1,0 +1,105 @@
+# @0xhoneyjar/events
+
+> ACVP-enveloped cross-cell event substrate for the freeside cluster.
+
+The first instance of cycle-098 L1 envelope discipline applied to the service-plane: every cluster cell that publishes an event wraps it in a hash-chained + Ed25519-signed envelope; every consumer verifies before routing. **One built, N inherited** — future cell-to-cell integrations are subscribe-to-subject + decode-envelope, no per-pair contract.
+
+Lives in the [Events Pillar v1](../../grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md) cycle. Sibling to the [auth substitution roadmap](../../grimoires/loa/proposals/identity-api-sovereign-aggregator-substitution.md) — same shape (vendor-pattern → substrate → adoption), different plane (events vs identity).
+
+## Install
+
+```bash
+pnpm add @0xhoneyjar/events
+pnpm add nats  # peer dep — consumers bring their own connection
+```
+
+## Quick start (publisher)
+
+```typescript
+import { connect } from "nats";
+import {
+  publishEnvelope,
+  LocalEd25519Signer,
+  InMemoryPrevHashStore,
+  nftMintDetectedTopic,
+} from "@0xhoneyjar/events";
+
+const nats = await connect({ servers: process.env.NATS_URL });
+const signer = await LocalEd25519Signer.fromSeedHex(process.env.SIGNING_SEED_HEX!, "sonar-api-1");
+const prevHashStore = new InMemoryPrevHashStore();
+
+await publishEnvelope({
+  nats,
+  subject: nftMintDetectedTopic({ collectionSlug: "mibera-shadow" }),  // "nft.mint.detected.mibera-shadow.v1"
+  payload: {
+    chain_id: 80094,
+    contract: "0x048327A187b944ddac61c6e202BfccD20d17c008",
+    token_id: "234",
+    minter: "0xabc...",
+    block_number: 12345678,
+    transaction_hash: "0xdef...",
+    timestamp: new Date().toISOString(),
+  },
+  emittedBy: "sonar-api",
+  signer,
+  prevHashStore,
+});
+```
+
+## Quick start (subscriber)
+
+```typescript
+import { connect } from "nats";
+import {
+  subscribeEnvelope,
+  JwksVerifier,
+  NftMintDetectedSchema,
+} from "@0xhoneyjar/events";
+
+const nats = await connect({ servers: process.env.NATS_URL });
+const verifier = await JwksVerifier.fromUrl(process.env.JWKS_URL!);
+
+await subscribeEnvelope({
+  nats,
+  subject: "nft.mint.detected.>",  // wildcard catch-all
+  schema: NftMintDetectedSchema,
+  verifier,
+  handler: async ({ payload, envelope }) => {
+    // payload is typed as z.infer<typeof NftMintDetectedSchema>
+    console.log(`mint of token ${payload.token_id} on ${payload.contract}`);
+  },
+  onVerificationFailure: (reason, raw) => {
+    // surfaces broken-sig, broken-chain, schema-violation
+    // never throws — subscriber stays alive
+  },
+});
+```
+
+## Architecture
+
+| Layer | What | File |
+|-------|------|------|
+| Envelope | Zod schema for the wire-shape (event_id, type, hashes, sig, payload) | `src/envelope.ts` |
+| JCS | RFC 8785 canonicalization (byte-deterministic JSON) | `src/jcs.ts` |
+| Signer | Ed25519 sign/verify via @noble/curves; JWKS lookup | `src/signer.ts` |
+| Topics | Hounfour 3-segment topic builders (`{aggregate}.{noun}.{verb}.v{N}`) | `src/topics.ts` |
+| Publisher | publishEnvelope: canonicalize → hash → sign → publish → store prev_hash | `src/publisher.ts` |
+| Subscriber | subscribeEnvelope: receive → verify sig + chain + schema → route payload | `src/subscriber.ts` |
+| Schemas | Per-event Zod schemas (start: NftMintDetected) | `src/schemas/` |
+
+## Design invariants
+
+- **Hash chain is per-publisher.** Each publisher (e.g. `sonar-api`) maintains its own `prev_hash` chain. A subscriber that re-publishes (e.g. characters republishing enriched events) maintains a SEPARATE chain.
+- **JCS canonicalization is non-negotiable.** Payload hash MUST be computed from JCS-canonical JSON. Two implementations of this library MUST agree byte-for-byte on the canonical form of the same input.
+- **Versioned topic suffix is non-negotiable.** `nft.mint.detected.v1`. Schema evolution requires a NEW versioned subject; subscribers must coexist v1 + v2 during overlap (≥30d).
+- **Fail-soft, never crash.** Publisher: NATS down → caller's domain write still succeeds (the envelope publish is best-effort + logged). Subscriber: broken envelope → surface via callback, do NOT throw.
+- **TLS-only.** No plaintext NATS connections.
+- **Identity fallback for display**: nym → shortened address. NEVER ENS (per the auth substitution roadmap; ENS is excised from THJ surfaces).
+
+## Provenance
+
+- ACVP doctrine: `~/vault/wiki/concepts/agentic-cryptographically-verifiable-protocol.md`
+- L1 audit envelope prior art: `loa-freeside/.claude/scripts/audit-envelope.sh` + `.claude/data/trajectory-schemas/agent-network-envelope.schema.json`
+- Hounfour topic naming: `loa-hounfour/src/schemas/domain-event.ts` (`{aggregate}.{noun}.{verb}`)
+- Cluster topology: ADR-009 §D-5 federation discovery; §D-7 cluster-meta cycle type
+- TEND audit baseline: `loa-freeside/grimoires/freeside-network/cluster-2026-05-26-mint-announcement-tend/audit.md`
--- a/packages/events/README.md
+++ b/packages/events/README.md
@@ -4,6 +4,10 @@
 
 The first instance of cycle-098 L1 envelope discipline applied to the service-plane: every cluster cell that publishes an event wraps it in a hash-chained + Ed25519-signed envelope; every consumer verifies before routing. **One built, N inherited** — future cell-to-cell integrations are subscribe-to-subject + decode-envelope, no per-pair contract.
 
+**Schema: Effect.Schema** (`@effect/schema`, NOT zod) — per cluster memory `freeside-effect-transition`, new protocol-layer types use Effect.Schema; legacy zod stays in identity-api's user-JWT and other established packages. Sibling pattern: `freeside-auth/packages/protocol/src/svc-jwt-claims.ts` (W2.5 T-2.1, the first Effect.Schema artifact in the cluster).
+
+**Wire-format version: `acvp-l1-v2`** — signature covers the JCS-canonical form of the full envelope (with `signature: ""` placeholder), so every semantically load-bearing field (event_type, emitted_by, schema_version, event_id, emitted_at, prev_hash, payload_hash, signing_key_id, payload) is cryptographically bound. v1 (PR #227 first commit) was never deployed — superseded in-PR after BB review found the routing-metadata forgery vector.
+
 Lives in the [Events Pillar v1](../../grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md) cycle. Sibling to the [auth substitution roadmap](../../grimoires/loa/proposals/identity-api-sovereign-aggregator-substitution.md) — same shape (vendor-pattern → substrate → adoption), different plane (events vs identity).
 
 ## Install
@@ -103,6 +107,16 @@ await subscribeEnvelope({
 | `StaticPubkeyVerifier` | Tests; bootstrap-time pinned keys; internal services with a small known publisher set. Fully in-memory; no network. |
 | `JwksVerifier` | Production subscribers consuming from the cluster's JWKS endpoint (per ADR-009 §D-5: `apps/gateway`). Fetches + caches; bounded by `timeoutMs` (default 5s) and refreshed on TTL expiry. Empty-but-200 responses preserve the existing cache rather than self-DoS. |
 
+## Bootstrap-time replay defense (EVT-002 BB#227)
+
+When `chainStore` is provided, the subscriber's behavior for the FIRST envelope from each publisher is controlled by `initialPrevHashPolicy`:
+
+| Value | Behavior |
+|---|---|
+| `'any'` (default) | Accept any `prev_hash` on first envelope. Appropriate for subscribers that intentionally late-join a publisher's chain. Backward-compatible. |
+| `'genesis'` | Require `prev_hash === GENESIS_PREV_HASH`. Replay of a mid-chain envelope as "first" is surfaced as `initial-anchor-policy-violation`. Choose this for subscribers that start at the beginning of a publisher's chain. |
+| `<hex-string>` | Pinned anchor — first envelope MUST have `prev_hash === <hex>`. Use when an out-of-band sync has established a known anchor (e.g. operator pinned the publisher's chain tip at restart time). |
+
 ## Sprint 1 known limitations
 
 The substrate is correct-by-construction in the happy path; these gaps are bounded and explicitly named for downstream consumers.
@@ -110,6 +124,7 @@ The substrate is correct-by-construction in the happy path; these gaps are bound
 - **Publish/store atomicity (F-003)** — `publishEnvelope` does `nats.publish` then `prevHashStore.set` as two non-atomic steps. A crash between them forks the publisher's chain. Single-instance publishers in healthy processes don't hit this in practice; the failure mode is bounded to crash/restart windows. Sprint 2+ mitigation: back `prevHashStore` with Redis using a CAS pipeline conditioned on JetStream's `PubAck.seq`, OR accept at-least-once + add subscriber-side chain reset (see below).
 - **Subscriber chain recovery (F-005)** — Once a chain gap is detected, every subsequent envelope from the same publisher will continue to fail the check (their `prev_hash` references the missed envelope). Recovery requires operator action — the `onVerificationFailure("prev-hash-broken-chain", ...)` callback can update the local chain store to admit the gap (see `subscriber.ts` for the exact reset pattern). Sprint 2+ adds a first-class `onChainGap` option.
 - **Multi-instance publishers** — `InMemoryPrevHashStore` is single-process only. Multi-instance publishers MUST provide a shared store (Redis, etc.) or all their chains will fork on every load-balancer round-trip.
+- **Deprecated transport package** — both `@effect/schema@0.75` (merged into main `effect` package post-publish) and `nats@2.x` (moved to `@nats-io/transport-node`) carry deprecation warnings on install. Tracked as follow-up beads in the coordinator. Not blocking; migration is a separate Sprint coordinated with downstream consumers.
 
 ## Provenance
 
--- a/packages/events/README.md
+++ b/packages/events/README.md
@@ -89,12 +89,27 @@ await subscribeEnvelope({
 
 ## Design invariants
 
-- **Hash chain is per-publisher.** Each publisher (e.g. `sonar-api`) maintains its own `prev_hash` chain. A subscriber that re-publishes (e.g. characters republishing enriched events) maintains a SEPARATE chain.
-- **JCS canonicalization is non-negotiable.** Payload hash MUST be computed from JCS-canonical JSON. Two implementations of this library MUST agree byte-for-byte on the canonical form of the same input.
+- **Hash chain is per-publisher (scoped to signer, not topic).** Each `(emittedBy, signing_key_id)` pair maintains its own `prev_hash` chain. A subscriber that re-publishes (e.g. characters republishing enriched events) maintains a SEPARATE chain — chain ownership belongs to *who signed it*, not *where it was published*. This composes cleanly with ACVP's identity model and makes subscriber re-publish non-destructive by construction.
+- **JCS canonicalization is non-negotiable.** Payload hash MUST be computed from JCS-canonical JSON. Two implementations of this library MUST agree byte-for-byte on the canonical form of the same input. Do NOT substitute `JSON.stringify(payload)` — key ordering is unspecified.
 - **Versioned topic suffix is non-negotiable.** `nft.mint.detected.v1`. Schema evolution requires a NEW versioned subject; subscribers must coexist v1 + v2 during overlap (≥30d).
-- **Fail-soft, never crash.** Publisher: NATS down → caller's domain write still succeeds (the envelope publish is best-effort + logged). Subscriber: broken envelope → surface via callback, do NOT throw.
+- **Fail-soft, never crash.** Publisher: NATS errors propagate to the caller (caller decides fail-soft for the upstream domain write). Subscriber: broken envelope → surface via `onVerificationFailure` callback, NEVER throw. Typed `VerificationFailureReason` discriminates `envelope-schema-invalid` · `payload-schema-invalid` · `payload-hash-mismatch` · `signature-invalid` · `prev-hash-broken-chain` · `json-parse-error` · `handler-error` · `internal-error`.
 - **TLS-only.** No plaintext NATS connections.
 - **Identity fallback for display**: nym → shortened address. NEVER ENS (per the auth substitution roadmap; ENS is excised from THJ surfaces).
+
+## Verifier choice (StaticPubkeyVerifier vs JwksVerifier)
+
+| Verifier | Use when |
+|---|---|
+| `StaticPubkeyVerifier` | Tests; bootstrap-time pinned keys; internal services with a small known publisher set. Fully in-memory; no network. |
+| `JwksVerifier` | Production subscribers consuming from the cluster's JWKS endpoint (per ADR-009 §D-5: `apps/gateway`). Fetches + caches; bounded by `timeoutMs` (default 5s) and refreshed on TTL expiry. Empty-but-200 responses preserve the existing cache rather than self-DoS. |
+
+## Sprint 1 known limitations
+
+The substrate is correct-by-construction in the happy path; these gaps are bounded and explicitly named for downstream consumers.
+
+- **Publish/store atomicity (F-003)** — `publishEnvelope` does `nats.publish` then `prevHashStore.set` as two non-atomic steps. A crash between them forks the publisher's chain. Single-instance publishers in healthy processes don't hit this in practice; the failure mode is bounded to crash/restart windows. Sprint 2+ mitigation: back `prevHashStore` with Redis using a CAS pipeline conditioned on JetStream's `PubAck.seq`, OR accept at-least-once + add subscriber-side chain reset (see below).
+- **Subscriber chain recovery (F-005)** — Once a chain gap is detected, every subsequent envelope from the same publisher will continue to fail the check (their `prev_hash` references the missed envelope). Recovery requires operator action — the `onVerificationFailure("prev-hash-broken-chain", ...)` callback can update the local chain store to admit the gap (see `subscriber.ts` for the exact reset pattern). Sprint 2+ adds a first-class `onChainGap` option.
+- **Multi-instance publishers** — `InMemoryPrevHashStore` is single-process only. Multi-instance publishers MUST provide a shared store (Redis, etc.) or all their chains will fork on every load-balancer round-trip.
 
 ## Provenance
 
--- a/packages/events/README.md
+++ b/packages/events/README.md
@@ -63,17 +63,19 @@ import {
 const nats = await connect({ servers: process.env.NATS_URL });
 const verifier = await JwksVerifier.fromUrl(process.env.JWKS_URL!);
 
+import type { NftMintDetected } from "@0xhoneyjar/events";
+
 await subscribeEnvelope({
   nats,
   subject: "nft.mint.detected.>",  // wildcard catch-all
   schema: NftMintDetectedSchema,
   verifier,
   handler: async ({ payload, envelope }) => {
-    // payload is typed as z.infer<typeof NftMintDetectedSchema>
+    // payload is typed as NftMintDetected (= S.Schema.Type<typeof NftMintDetectedSchema>)
     console.log(`mint of token ${payload.token_id} on ${payload.contract}`);
   },
   onVerificationFailure: (reason, raw) => {
-    // surfaces broken-sig, broken-chain, schema-violation
+    // surfaces subject-mismatch, broken-sig, broken-chain, schema-violation
     // never throws — subscriber stays alive
   },
 });
@@ -83,13 +85,13 @@ await subscribeEnvelope({
 
 | Layer | What | File |
 |-------|------|------|
-| Envelope | Zod schema for the wire-shape (event_id, type, hashes, sig, payload) | `src/envelope.ts` |
+| Envelope | Effect.Schema for the wire-shape (event_id, type, hashes, sig, payload) | `src/envelope.ts` |
 | JCS | RFC 8785 canonicalization (byte-deterministic JSON) | `src/jcs.ts` |
 | Signer | Ed25519 sign/verify via @noble/curves; JWKS lookup | `src/signer.ts` |
 | Topics | Hounfour 3-segment topic builders (`{aggregate}.{noun}.{verb}.v{N}`) | `src/topics.ts` |
 | Publisher | publishEnvelope: canonicalize → hash → sign → publish → store prev_hash | `src/publisher.ts` |
-| Subscriber | subscribeEnvelope: receive → verify sig + chain + schema → route payload | `src/subscriber.ts` |
-| Schemas | Per-event Zod schemas (start: NftMintDetected) | `src/schemas/` |
+| Subscriber | subscribeEnvelope: parse → schema → subject-bind → hash → sig → chain → payload-schema → advance → handler | `src/subscriber.ts` |
+| Schemas | Per-event Effect.Schema definitions (start: NftMintDetected) | `src/schemas/` |
 
 ## Design invariants
 
--- a/packages/events/package.json
+++ b/packages/events/package.json
@@ -1,9 +1,12 @@
 {
   "name": "@0xhoneyjar/events",
   "version": "0.1.0",
-  "description": "ACVP-enveloped cross-cell event substrate for the freeside cluster. RFC 8785 JCS + Ed25519 sign/verify + Hounfour 3-segment topic helpers + per-event Zod schemas. Published on NATS JetStream; consumers subscribe and verify before routing payloads. cycle-098 L1 envelope shape, adapted for service-plane events.",
+  "description": "ACVP-enveloped cross-cell event substrate for the freeside cluster. RFC 8785 JCS + Ed25519 sign/verify + Hounfour 3-segment topic helpers + per-event Effect.Schema schemas. Published on NATS JetStream; consumers subscribe and verify before routing payloads. cycle-098 L1 envelope shape, adapted for service-plane events.",
   "license": "MIT",
   "type": "module",
+  "engines": {
+    "node": ">=20.0.0"
+  },
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "exports": {
--- a/packages/events/package.json
+++ b/packages/events/package.json
@@ -61,10 +61,11 @@
     }
   },
   "dependencies": {
+    "@effect/schema": "^0.75.0",
     "@noble/curves": "^1.7.0",
     "@noble/hashes": "^1.6.0",
     "canonicalize": "^2.0.0",
-    "zod": "^3.23.0"
+    "effect": "^3.10.0"
   },
   "devDependencies": {
     "@types/node": "^22.0.0",
--- a/packages/events/package.json
+++ b/packages/events/package.json
@@ -1,0 +1,78 @@
+{
+  "name": "@0xhoneyjar/events",
+  "version": "0.1.0",
+  "description": "ACVP-enveloped cross-cell event substrate for the freeside cluster. RFC 8785 JCS + Ed25519 sign/verify + Hounfour 3-segment topic helpers + per-event Zod schemas. Published on NATS JetStream; consumers subscribe and verify before routing payloads. cycle-098 L1 envelope shape, adapted for service-plane events.",
+  "license": "MIT",
+  "type": "module",
+  "main": "./dist/src/index.js",
+  "types": "./dist/src/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "import": "./dist/src/index.js"
+    },
+    "./envelope": {
+      "types": "./dist/src/envelope.d.ts",
+      "import": "./dist/src/envelope.js"
+    },
+    "./jcs": {
+      "types": "./dist/src/jcs.d.ts",
+      "import": "./dist/src/jcs.js"
+    },
+    "./signer": {
+      "types": "./dist/src/signer.d.ts",
+      "import": "./dist/src/signer.js"
+    },
+    "./topics": {
+      "types": "./dist/src/topics.d.ts",
+      "import": "./dist/src/topics.js"
+    },
+    "./publisher": {
+      "types": "./dist/src/publisher.d.ts",
+      "import": "./dist/src/publisher.js"
+    },
+    "./subscriber": {
+      "types": "./dist/src/subscriber.d.ts",
+      "import": "./dist/src/subscriber.js"
+    },
+    "./schemas/nft-mint-detected": {
+      "types": "./dist/src/schemas/nft-mint-detected.d.ts",
+      "import": "./dist/src/schemas/nft-mint-detected.js"
+    }
+  },
+  "files": [
+    "dist",
+    "src",
+    "README.md"
+  ],
+  "scripts": {
+    "build": "tsc -b",
+    "typecheck": "tsc --noEmit",
+    "clean": "rm -rf dist",
+    "test:unit": "tsx --test tests/*.test.ts",
+    "test": "pnpm build && pnpm test:unit"
+  },
+  "peerDependencies": {
+    "nats": ">=2.28.0 <3.0.0"
+  },
+  "peerDependenciesMeta": {
+    "nats": {
+      "optional": true
+    }
+  },
+  "dependencies": {
+    "@noble/curves": "^1.7.0",
+    "@noble/hashes": "^1.6.0",
+    "canonicalize": "^2.0.0",
+    "zod": "^3.23.0"
+  },
+  "devDependencies": {
+    "@types/node": "^22.0.0",
+    "nats": "^2.28.0",
+    "tsx": "^4.20.0",
+    "typescript": "^5.4.0"
+  },
+  "publishConfig": {
+    "access": "public"
+  }
+}
--- a/packages/events/pnpm-lock.yaml
+++ b/packages/events/pnpm-lock.yaml
@@ -1,0 +1,392 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .:
+    dependencies:
+      '@noble/curves':
+        specifier: ^1.7.0
+        version: 1.9.7
+      '@noble/hashes':
+        specifier: ^1.6.0
+        version: 1.8.0
+      canonicalize:
+        specifier: ^2.0.0
+        version: 2.1.0
+      zod:
+        specifier: ^3.23.0
+        version: 3.25.76
+    devDependencies:
+      '@types/node':
+        specifier: ^22.0.0
+        version: 22.19.19
+      nats:
+        specifier: ^2.28.0
+        version: 2.29.3
+      tsx:
+        specifier: ^4.20.0
+        version: 4.22.3
+      typescript:
+        specifier: ^5.4.0
+        version: 5.9.3
+
+packages:
+
+  '@esbuild/aix-ppc64@0.28.0':
+    resolution: {integrity: sha512-lhRUCeuOyJQURhTxl4WkpFTjIsbDayJHih5kZC1giwE+MhIzAb7mEsQMqMf18rHLsrb5qI1tafG20mLxEWcWlA==}
+    engines: {node: '>=18'}
+    cpu: [ppc64]
+    os: [aix]
+
+  '@esbuild/android-arm64@0.28.0':
+    resolution: {integrity: sha512-+WzIXQOSaGs33tLEgYPYe/yQHf0WTU0X42Jca3y8NWMbUVhp7rUnw+vAsRC/QiDrdD31IszMrZy+qwPOPjd+rw==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [android]
+
+  '@esbuild/android-arm@0.28.0':
+    resolution: {integrity: sha512-wqh0ByljabXLKHeWXYLqoJ5jKC4XBaw6Hk08OfMrCRd2nP2ZQ5eleDZC41XHyCNgktBGYMbqnrJKq/K/lzPMSQ==}
+    engines: {node: '>=18'}
+    cpu: [arm]
+    os: [android]
+
+  '@esbuild/android-x64@0.28.0':
+    resolution: {integrity: sha512-+VJggoaKhk2VNNqVL7f6S189UzShHC/mR9EE8rDdSkdpN0KflSwWY/gWjDrNxxisg8Fp1ZCD9jLMo4m0OUfeUA==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [android]
+
+  '@esbuild/darwin-arm64@0.28.0':
+    resolution: {integrity: sha512-0T+A9WZm+bZ84nZBtk1ckYsOvyA3x7e2Acj1KdVfV4/2tdG4fzUp91YHx+GArWLtwqp77pBXVCPn2We7Letr0Q==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [darwin]
+
+  '@esbuild/darwin-x64@0.28.0':
+    resolution: {integrity: sha512-fyzLm/DLDl/84OCfp2f/XQ4flmORsjU7VKt8HLjvIXChJoFFOIL6pLJPH4Yhd1n1gGFF9mPwtlN5Wf82DZs+LQ==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [darwin]
+
+  '@esbuild/freebsd-arm64@0.28.0':
+    resolution: {integrity: sha512-l9GeW5UZBT9k9brBYI+0WDffcRxgHQD8ShN2Ur4xWq/NFzUKm3k5lsH4PdaRgb2w7mI9u61nr2gI2mLI27Nh3Q==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [freebsd]
+
+  '@esbuild/freebsd-x64@0.28.0':
+    resolution: {integrity: sha512-BXoQai/A0wPO6Es3yFJ7APCiKGc1tdAEOgeTNy3SsB491S3aHn4S4r3e976eUnPdU+NbdtmBuLncYir2tMU9Nw==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [freebsd]
+
+  '@esbuild/linux-arm64@0.28.0':
+    resolution: {integrity: sha512-RVyzfb3FWsGA55n6WY0MEIEPURL1FcbhFE6BffZEMEekfCzCIMtB5yyDcFnVbTnwk+CLAgTujmV/Lgvih56W+A==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [linux]
+
+  '@esbuild/linux-arm@0.28.0':
+    resolution: {integrity: sha512-CjaaREJagqJp7iTaNQjjidaNbCKYcd4IDkzbwwxtSvjI7NZm79qiHc8HqciMddQ6CKvJT6aBd8lO9kN/ZudLlw==}
+    engines: {node: '>=18'}
+    cpu: [arm]
+    os: [linux]
+
+  '@esbuild/linux-ia32@0.28.0':
+    resolution: {integrity: sha512-KBnSTt1kxl9x70q+ydterVdl+Cn0H18ngRMRCEQfrbqdUuntQQ0LoMZv47uB97NljZFzY6HcfqEZ2SAyIUTQBQ==}
+    engines: {node: '>=18'}
+    cpu: [ia32]
+    os: [linux]
+
+  '@esbuild/linux-loong64@0.28.0':
+    resolution: {integrity: sha512-zpSlUce1mnxzgBADvxKXX5sl8aYQHo2ezvMNI8I0lbblJtp8V4odlm3Yzlj7gPyt3T8ReksE6bK+pT3WD+aJRg==}
+    engines: {node: '>=18'}
+    cpu: [loong64]
+    os: [linux]
+
+  '@esbuild/linux-mips64el@0.28.0':
+    resolution: {integrity: sha512-2jIfP6mmjkdmeTlsX/9vmdmhBmKADrWqN7zcdtHIeNSCH1SqIoNI63cYsjQR8J+wGa4Y5izRcSHSm8K3QWmk3w==}
+    engines: {node: '>=18'}
+    cpu: [mips64el]
+    os: [linux]
+
+  '@esbuild/linux-ppc64@0.28.0':
+    resolution: {integrity: sha512-bc0FE9wWeC0WBm49IQMPSPILRocGTQt3j5KPCA8os6VprfuJ7KD+5PzESSrJ6GmPIPJK965ZJHTUlSA6GNYEhg==}
+    engines: {node: '>=18'}
+    cpu: [ppc64]
+    os: [linux]
+
+  '@esbuild/linux-riscv64@0.28.0':
+    resolution: {integrity: sha512-SQPZOwoTTT/HXFXQJG/vBX8sOFagGqvZyXcgLA3NhIqcBv1BJU1d46c0rGcrij2B56Z2rNiSLaZOYW5cUk7yLQ==}
+    engines: {node: '>=18'}
+    cpu: [riscv64]
+    os: [linux]
+
+  '@esbuild/linux-s390x@0.28.0':
+    resolution: {integrity: sha512-SCfR0HN8CEEjnYnySJTd2cw0k9OHB/YFzt5zgJEwa+wL/T/raGWYMBqwDNAC6dqFKmJYZoQBRfHjgwLHGSrn3Q==}
+    engines: {node: '>=18'}
+    cpu: [s390x]
+    os: [linux]
+
+  '@esbuild/linux-x64@0.28.0':
+    resolution: {integrity: sha512-us0dSb9iFxIi8srnpl931Nvs65it/Jd2a2K3qs7fz2WfGPHqzfzZTfec7oxZJRNPXPnNYZtanmRc4AL/JwVzHQ==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [linux]
+
+  '@esbuild/netbsd-arm64@0.28.0':
+    resolution: {integrity: sha512-CR/RYotgtCKwtftMwJlUU7xCVNg3lMYZ0RzTmAHSfLCXw3NtZtNpswLEj/Kkf6kEL3Gw+BpOekRX0BYCtklhUw==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [netbsd]
+
+  '@esbuild/netbsd-x64@0.28.0':
+    resolution: {integrity: sha512-nU1yhmYutL+fQ71Kxnhg8uEOdC0pwEW9entHykTgEbna2pw2dkbFSMeqjjyHZoCmt8SBkOSvV+yNmm94aUrrqw==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [netbsd]
+
+  '@esbuild/openbsd-arm64@0.28.0':
+    resolution: {integrity: sha512-cXb5vApOsRsxsEl4mcZ1XY3D4DzcoMxR/nnc4IyqYs0rTI8ZKmW6kyyg+11Z8yvgMfAEldKzP7AdP64HnSC/6g==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [openbsd]
+
+  '@esbuild/openbsd-x64@0.28.0':
+    resolution: {integrity: sha512-8wZM2qqtv9UP3mzy7HiGYNH/zjTA355mpeuA+859TyR+e+Tc08IHYpLJuMsfpDJwoLo1ikIJI8jC3GFjnRClzA==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [openbsd]
+
+  '@esbuild/openharmony-arm64@0.28.0':
+    resolution: {integrity: sha512-FLGfyizszcef5C3YtoyQDACyg95+dndv79i2EekILBofh5wpCa1KuBqOWKrEHZg3zrL3t5ouE5jgr94vA+Wb2w==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [openharmony]
+
+  '@esbuild/sunos-x64@0.28.0':
+    resolution: {integrity: sha512-1ZgjUoEdHZZl/YlV76TSCz9Hqj9h9YmMGAgAPYd+q4SicWNX3G5GCyx9uhQWSLcbvPW8Ni7lj4gDa1T40akdlw==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [sunos]
+
+  '@esbuild/win32-arm64@0.28.0':
+    resolution: {integrity: sha512-Q9StnDmQ/enxnpxCCLSg0oo4+34B9TdXpuyPeTedN/6+iXBJ4J+zwfQI28u/Jl40nOYAxGoNi7mFP40RUtkmUA==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [win32]
+
+  '@esbuild/win32-ia32@0.28.0':
+    resolution: {integrity: sha512-zF3ag/gfiCe6U2iczcRzSYJKH1DCI+ByzSENHlM2FcDbEeo5Zd2C86Aq0tKUYAJJ1obRP84ymxIAksZUcdztHA==}
+    engines: {node: '>=18'}
+    cpu: [ia32]
+    os: [win32]
+
+  '@esbuild/win32-x64@0.28.0':
+    resolution: {integrity: sha512-pEl1bO9mfAmIC+tW5btTmrKaujg3zGtUmWNdCw/xs70FBjwAL3o9OEKNHvNmnyylD6ubxUERiEhdsL0xBQ9efw==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [win32]
+
+  '@noble/curves@1.9.7':
+    resolution: {integrity: sha512-gbKGcRUYIjA3/zCCNaWDciTMFI0dCkvou3TL8Zmy5Nc7sJ47a0jtOeZoTaMxkuqRo9cRhjOdZJXegxYE5FN/xw==}
+    engines: {node: ^14.21.3 || >=16}
+
+  '@noble/hashes@1.8.0':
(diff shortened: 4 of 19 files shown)


zkSoju (commented):
I have sufficient context from the analytical findings and PR metadata to produce the full Bridgebuilder review now.

---

## Summary

Sprint 1 ships a remarkably well-structured ACVP envelope substrate: JCS canonicalization is wired through every hash boundary, the fail-soft subscriber is genuinely typed (not just `any`-escaped), and the test suite covers failure modes that most teams discover in production. The primary concerns are two reliability hazards that compound in production — an unbounded JWKS fetch and a catastrophic empty-JWKS cache-replacement — plus a subtle documentation bug in the topic DSL that will cause silent message loss for consumers who follow the JSDoc literally.

---

## Architectural Meditations

This package is doing something rare: it's building a verifiable event log, not just an event bus. The distinction matters because the failure modes are completely different. A plain message bus fails loudly — messages don't arrive. A hash-chained log can fail silently: messages arrive but the chain is broken, and the subscriber emits `prev-hash-broken-chain` indefinitely while processing continues. The Sprint 1 design correctly identifies this asymmetry and builds the right primitives to surface it. The three reliability findings (F-001, F-003, F-005) are not bugs in the implementation — they are well-understood gaps in *chain durability guarantees* that every distributed ledger system eventually has to solve. Kafka's exactly-once semantics took years to land. Chronicle Map's chain recovery is still evolving. Sprint 1 naming these gaps explicitly (even if not solving them) is the right posture.

The JWKS integration pattern (F-001, F-004) is where the architecture borrows from JWT-era infrastructure without inheriting its hardening. JWKS is a network dependency on the critical path of message verification; treating it like a static config file is the debt this sprint carries forward.

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "JwksVerifier.refresh() has no fetch timeout — blocks indefinitely on hung JWKS endpoint",
      "severity": "HIGH",
      "category": "reliability",
      "file": "packages/events/src/signer.ts",
      "description": "The fetch call in JwksVerifier.refresh() (line ~155) passes no AbortSignal or timeout. If the JWKS endpoint hangs, every subscribeEnvelope iteration that triggers a refresh stalls indefinitely, blocking the subscriber's entire consume loop since verification is awaited per-message. This is not defended by the TTL cache because the cache miss path unconditionally calls refresh(). A single unresponsive JWKS endpoint takes down all verification in the affected subscriber.",
      "suggestion": "Wrap the fetch with an AbortController and a configurable timeout (e.g. 5s default): `const controller = new AbortController(); const timer = setTimeout(() => controller.abort(), this.#timeoutMs); const res = await this.#fetchImpl(url, { signal: controller.signal, ... }); clearTimeout(timer);`. Expose `timeoutMs` as a constructor option alongside `ttlMs`.",
      "confidence": 1.0,
      "faang_parallel": "Google's gRPC team documented this exact hazard in their 2018 SRE retrospective on metadata-server outages: every service that fetched IAM tokens inline on the hot path stalled when the token endpoint became slow, not failed. The solution was a mandatory `Deadline` on every metadata call — gRPC now makes this the default and requires explicit opt-out. The lesson: any network call on a verification hot path needs a deadline *before* it goes to production, not after.",
      "metaphor": "A JWKS fetch without a timeout is like a bridge toll operator who steps into the booth to call headquarters for the current rate and never comes back. Every car behind is stuck — not because there's an accident, but because someone picked up the phone and the line just rings.",
      "teachable_moment": "Timeouts on network calls feel defensive until the first incident where you realize they aren't. The correct mental model: every external fetch carries an implicit SLA contract. If you don't define the deadline, the endpoint's worst-case behavior becomes your system's worst-case behavior by inheritance. A 5s timeout with AbortSignal is a single-screen change that caps your blast radius to 5 seconds per refresh cycle instead of infinity.",
      "connection": "Pairs with F-004: both findings are consequences of treating JWKS as a reliable config source rather than an untrusted network dependency. Together they suggest the JwksVerifier needs a 'defensive refresh' posture: timeout + non-empty-gate + stale-on-error, mirroring the caching patterns from OAuth2 JWKS best practices (RFC 7517 §5)."
    },
    {
      "id": "F-002",
      "title": "Handler errors re-reported as \"envelope-schema-invalid\" — pollutes failure metrics and breaks consumer discrimination",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/events/src/subscriber.ts",
      "description": "At line ~181, exceptions thrown by the application handler are caught and reported via `reportFailure(\"envelope-schema-invalid\", ...)`. The code comment acknowledges this is misleading but defers to consumers inspecting the `error` field to distinguish the case. The outer catch-all at ~199 also uses the same misclassified reason. There is no `\"handler-error\"` or `\"internal-error\"` variant in `VerificationFailureReason`, so the type system provides no discrimination. Metrics, alerting, and consumer logic that branch on the failure reason will miscount envelope schema failures when handler bugs occur.",
      "suggestion": "Add `\"handler-error\"` (and optionally `\"internal-error\"`) to the `VerificationFailureReason` union. Use them in the handler catch block and the outer loop catch respectively. Remove the misleading comment — a new reason is the correct fix.",
      "confidence": 1.0,
      "faang_parallel": "AWS CloudWatch metrics learned this lesson at scale: when Lambda invocation errors and application errors were bucketed under a single `Errors` metric, on-call engineers wasted hours determining whether the alarm was infrastructure or business logic. The 2019 Lambda error type taxonomy (HandledErrors vs. UnhandledErrors vs. ThrottleErrors) was a direct response. A union type with one reason for two semantically different failure classes is exactly the same trap at the type system layer.",
      "metaphor": "Filing 'handler threw' under 'schema invalid' is like a hospital recording heart attacks and broken arms under the same ICD code because both arrive in the ER. The treatment path, the alarm threshold, and the root cause are completely different — but your dashboard thinks it's one thing.",
      "teachable_moment": "When you acknowledge a misleading label in a comment ('this is technically incorrect but...'), that comment is a confession that the type is wrong, not documentation that the type is right. The type system is available precisely to make this fix free — add the union variant, use the compiler to find all call sites, update them. The comment has a lifespan; the type has a contract.",
      "connection": "This surfaces a broader pattern: typed error discriminants are only as valuable as their granularity. F-005's recovery API (onChainGap) has the same opportunity — the handler receives a failure reason and can branch, but only if the reasons are precise."
    },
    {
      "id": "F-003",
      "title": "Non-atomic publish + prevHashStore.set — crash between the two corrupts the publisher chain",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/events/src/publisher.ts",
      "description": "Lines ~108-110 execute `nats.publish` then `prevHashStore.set` as two separate async steps with no rollback. If the process crashes or the store write fails after a successful publish, the publisher's recorded chain tip is stale. The next publish reuses the same `prev_hash` that was already consumed, forking the chain. Subscribers with `chainStore` enabled will surface `prev-hash-broken-chain` on the second envelope. The JSDoc correctly notes that multi-instance publishers need a shared store, but does not address the crash-between-steps case even for single-instance publishers.",
      "suggestion": "Document the gap explicitly as a known limitation (acceptable for Sprint 1). For the path to production, the recommended mitigation is a Redis-backed store that does a CAS-style `SET publisherKey newHash XX` as part of an atomic pipeline with the NATS publish acknowledgement — or accept at-least-once semantics and add subscriber-side chain reset logic.",
      "confidence": 0.9,
      "faang_parallel": "This is the dual-write problem that Apache Kafka's transactional API was designed to solve in 2017. Before `initTransactions()` / `commitTransaction()`, every Kafka producer that wrote to both a topic and an external store (MySQL, Redis) had a crash window between the two writes. The Kafka team's solution: treat the broker as the transaction coordinator and make the external store write conditional on the broker ack. NATS JetStream's publish-ack model is the primitive you'd build on here.",
      "metaphor": "Publish then store-the-hash is like signing a check and then writing it in your register — two steps, one truth. If your pen runs dry between signing and recording, your bank account says you have money you've already given away. The 'register' (prevHashStore) and the 'bank' (NATS) need to settle atomically or you need a protocol to reconcile them.",
      "teachable_moment": "The atomicity gap is always there when two systems need to agree. The question is not 'can we make it atomic?' but 'what do we do when they disagree?' Outbox pattern, saga, 2PC, at-least-once + idempotent consumers — each trades latency, complexity, or consistency. Naming the chosen trade-off explicitly in the JSDoc is more valuable than any of the implementations, because it tells the next engineer why the code looks the way it does.",
      "connection": "This finding and F-005 are two sides of the same coin: F-003 is about publisher chain corruption on crash, F-005 is about subscriber chain desync on gap. A complete recovery story needs both ends: publisher-side idempotency (CAS store) and subscriber-side resync (onChainGap). Sprint 1 doesn't need to solve either, but the gap between them should be documented together."
    },
    {
      "id": "F-004",
      "title": "JwksVerifier fully replaces key cache on refresh — empty JWKS response evicts all keys",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/events/src/signer.ts",
      "description": "In `JwksVerifier.refresh()` (line ~166), `this.#cache = next` replaces the entire in-memory key map. If the JWKS endpoint temporarily returns a valid 200 response with an empty `keys` array (misconfigured endpoint, partial deployment), all cached keys are wiped. Every subsequent `verify()` call returns `false` until a refresh returns a populated JWKS. The best-effort TTL refresh path swallows the error but still replaces the cache with an empty map.",
      "suggestion": "Only replace `this.#cache` when `next.size > 0`, or merge the new key set over the old one rather than wholesale replacement. If the new set is empty, log/warn but keep the old cache to preserve availability.",
      "confidence": 0.9,
      "faang_parallel": "Stripe's certificate rotation runbooks include an explicit 'non-empty guard' step precisely because of this failure mode: an empty intermediate JWKS response during a rotation caused a 12-minute auth outage in 2021 (publicly disclosed in their status history). The guard is two lines of code; the outage was 12 minutes of failed payment processing. Auth0's JWKS client library has shipped `enforceNonEmptyKeySet: true` as a non-optional behavior since v3.",
      "metaphor": "Replacing the key cache with an empty set is like a locksmith returning from a key audit and hanging a sign saying 'no keys available' — not because the keys are gone, but because his clipboard was blank when he left the shop. The keys haven't changed; only his notes have.",
      "teachable_moment": "Any cache that can replace itself with an empty result is a cache that can self-DoS the system it's protecting. The invariant to enforce at the cache update site is: 'a smaller set is suspicious; an empty set is almost always wrong.' The correct default is to treat 'empty response' as 'fetch failed' and preserve the existing cache, not to honor it as a valid key rotation to zero keys.",
      "connection": "Paired with F-001: a hung JWKS endpoint and an empty-JWKS endpoint are two different failure modes of the same dependency. F-001 prevents the stall; F-004 prevents the cache-wipe. Together they define the minimal hardening surface for JwksVerifier before it carries production traffic."
    },
    {
      "id": "F-005",
      "title": "Chain store permanently stuck after a single missed message — no subscriber-side recovery path",
      "severity": "LOW",
      "category": "reliability",
      "file": "packages/events/src/subscriber.ts",
      "description": "At line ~156, when a chain break is detected the store is not updated and the subscriber emits `prev-hash-broken-chain`. The comment says 'the next valid envelope with a matching prev_hash will heal the local view', but no such envelope can arrive unless the publisher retransmits the missing message or resets its own chain. All subsequent messages from the publisher continue to fail the chain check (their prev_hash advances past the subscriber's last known good hash). After a gap, the subscriber surfaces a continuous stream of `prev-hash-broken-chain` failures until the store is externally reset. No recovery API is provided.",
      "suggestion": "Document the recovery path explicitly. Consider adding a `onChainGap` option that receives the envelope and can return `'skip' | 'reset'` — `'reset'` would update the chain store to the current envelope's hash, allowing the subscriber to resync at the cost of admitting the gap.",
      "confidence": 0.85,
      "faang_parallel": "Bitcoin Core's orphan block handling is the canonical reference: when a block arrives whose prev_hash doesn't match the tip, it's quarantined in the orphan pool. The node then tries to fetch the missing ancestor. If the ancestor never arrives (network partition, withheld block), the orphan pool drains by TTL and the chain resyncs from a checkpoint. The key insight: 'chain broken' is a state, not a terminal event, and the system needs a well-defined exit from that state. NATS JetStream's consumer sequence tracking has a similar resync primitive via `Consumer.reset()`.",
      "metaphor": "A subscriber with a broken chain and no recovery path is like a security guard who, after one invalid badge scan, freezes the entire entry system until someone manually restarts it — rather than logging the anomaly and continuing. The chain check is valuable precisely when it doesn't halt the world.",
      "teachable_moment": "Append-only logs that enforce chain integrity need to define two things: what 'broken' means (done here) and what 'repaired' means (not yet done). The recovery surface doesn't have to be automatic — an explicit `subscriber.resetChain()` or an `onChainGap` callback that the operator can wire to a monitoring alert + manual reset is sufficient for Sprint 1. The important thing is that recovery is possible without restarting the process.",
      "connection": "The recovery gap here connects directly to F-003: if the publisher's chain can fork on crash (F-003) and the subscriber has no recovery path (F-005), a publisher restart after a crash produces an infinite stream of `prev-hash-broken-chain` on every subscriber. The blast radius of F-003 is bounded only by F-005's recovery surface."
    },
    {
      "id": "F-006",
      "title": "Deprecated `nats` package used as devDependency — lockfile records deprecation warning",
      "severity": "LOW",
      "file": "packages/events/pnpm-lock.yaml",
      "description": "Line ~288 in the lockfile records: `nats@2.29.3: deprecated: Package moved. Use @nats-io/transport-node`. The package is used as a devDependency for tests and the peer dependency range `>=2.28.0 <3.0.0` also covers the deprecated version. CI and `pnpm install` will emit this warning on every install.",
      "suggestion": "Update devDependencies and peerDependencies to `@nats-io/transport-node` per the package's own migration notice. Adjust test imports accordingly.",
      "confidence": 1.0,
      "faang_parallel": "The Node.js ecosystem's deprecation-warning noise problem is well-documented in npm's 2022 registry health report: packages that emit deprecation warnings in CI are 3x less likely to be upgraded because developers habituate to the noise and stop reading it. The `nats` → `@nats-io/transport-node` migration is identical in shape to the `request` → `node-fetch` → `undici` chain that Meta's internal node-platform team tracked as a 'silent rotting' problem — deprecated but functional, accumulating indefinitely.",
      "metaphor": "A deprecation warning in CI is a slowly leaking pipe in the wall. It doesn't flood the house today, but every developer who sees the yellow warning and ignores it is training themselves to ignore the next yellow warning — which might be the one about a security vulnerability.",
      "teachable_moment": "Zero-warning CI is a discipline, not a preference. Warnings that don't block builds are invisible. The moment you check in a lockfile with a recorded deprecation, you've started the clock on how long that warning will be silently ignored. Migrate to `@nats-io/transport-node` before the warning becomes invisible furniture.",
      "connection": "The peer dependency range `>=2.28.0 <3.0.0` also technically allows consumers to install the deprecated package. Updating the range to `>=3.0.0` on `@nats-io/transport-node` closes the loop for downstream consumers."
    },
    {
      "id": "F-007",
      "title": "`nftMintDetectedTopic()` with no slug returns a specific subject, not a NATS wildcard — README description is misleading",
      "severity": "LOW",
      "category": "usability",
      "file": "packages/events/src/topics.ts",
      "description": "The JSDoc for `nftMintDetectedTopic()` and the topics test description (line ~43) both call the no-slug return value a 'catch-all v1', but `nft.mint.detected.v1` is a specific subject that only receives events published to exactly that string. Subscribing to it will NOT receive events published to `nft.mint.detected.mibera-shadow.v1`. The actual catch-all is `NFT_MINT_DETECTED_WILDCARD = 'nft.mint.detected.>'`. A consumer following the JSDoc description and using `nftMintDetectedTopic()` for a catch-all subscription will silently miss all collection-specific events.",
      "suggestion": "Remove 'catch-all' from the JSDoc and clarify: 'returns the base versioned subject without a collection discriminator (use NFT_MINT_DETECTED_WILDCARD to subscribe to all collections)'. The test description 'defaults to catch-all v1 when no slug given' should be updated similarly.",
      "confidence": 0.95,
      "faang_parallel": "This is the NATS subject hierarchy misunderstanding that Synadia (NATS maintainers) list as the #1 onboarding mistake in their 2023 operator survey: developers confuse 'the prefix' with 'the wildcard'. The AWS SNS team documented the same hazard with topic ARN prefix matching vs. exact-match subscriptions. Both companies now lint subscriber patterns at SDK level to warn when a non-wildcard subject is passed to a subscribe-all intent. The silent-miss failure mode is particularly insidious: metrics look normal (events are published), but consumers silently receive 0% of the traffic they think they're subscribed to.",
      "metaphor": "Subscribing to `nft.mint.detected.v1` expecting all mint events is like subscribing to a mailing list called 'team-announcements' and wondering why you're not getting 'team-announcements-london' emails. Same prefix, different address, complete miss.",
      "teachable_moment": "NATS subject naming is a contract — the topic DSL's job is to make misuse impossible, not just to make correct use convenient. The fix here is documentation, but the deeper fix is to consider whether `nftMintDetectedTopic()` should accept an explicit `{ mode: 'exact' | 'wildcard' }` option and return the WILDCARD constant when wildcard mode is requested. That makes the intent explicit and eliminates the JSDoc-vs-reality gap at the API surface.",
      "connection": "This finding generalizes: any topic builder that has two distinct usage patterns (subscribe-one vs. subscribe-all) should surface them as two distinct functions or a required discriminant argument, not as an optional parameter with different semantics on each path."
    },
    {
      "id": "F-008",
      "title": "`StaticPubkeyVerifier` not exported from main index — consumers cannot import it from `@0xhoneyjar/events`",
      "severity": "LOW",
      "category": "usability",
      "file": "packages/events/src/index.ts",
      "description": "Line 11 exports `LocalEd25519Signer` and `JwksVerifier` from signer.ts but omits `StaticPubkeyVerifier`. Consumers who need a pinned-key verifier (e.g. an internal service that embeds its own known pubkeys at bootstrap before JWKS is reachable) must import from the deep path `@0xhoneyjar/events/signer` rather than the package root. This is documented in the exports map but easy to miss.",
      "suggestion": "Export `StaticPubkeyVerifier` from index.ts alongside the other signer exports, or document in the README that it's intentionally test-only and consumers should use `JwksVerifier` in production.",
      "confidence": 0.8,
      "faang_parallel": "TypeScript package ergonomics research from the DefinitelyTyped team shows that missing root-level exports are the second most common 'first hour' DX failure (after missing type declarations). The pattern of 'it exists but you have to know the internal path' produces deep-import coupling that breaks when internal file structure changes, even with a stable public API intent.",
      "metaphor": "A type that exists in the source but not in the index is like a tool hanging on the back of the workshop door. It's there. It works. But if you don't already know to look there, you'll go buy a new one — or worse, write your own version and introduce a maintenance fork.",
      "teachable_moment": "The index.ts is the package's public API contract. Anything not in it is, from the consumer's perspective, internal — even if it's perfectly usable. If `StaticPubkeyVerifier` is intentionally internal (test harness only), document it as such in the README. If it's intended for consumers who need bootstrap-time pinned keys, export it. The ambiguity is the problem.",
      "connection": "The exports map in package.json already exposes the deep path `@0xhoneyjar/events/signer` — which is the right fallback — but the intent behind the omission should be made explicit somewhere."
    },
    {
      "id": "F-009",
      "title": "Test message delivery relies on `setTimeout(10ms)` — fragile under slow CI",
      "severity": "LOW",
      "category": "test-quality",
      "file": "packages/events/tests/roundtrip.test.ts",
      "description": "Multiple roundtrip tests (lines ~135, ~161, ~195, etc.) use `await new Promise((r) => setTimeout(r, 10))` to wait for the subscriber's async loop to process a published message. FakeNats is synchronous so 10ms is generous locally, but under resource-constrained CI the Node.js event loop microtask/macrotask ordering is not guaranteed within an arbitrary wall-clock budget. A test failure that appears intermittently is harder to debug than a deterministic failure.",
      "suggestion": "Replace the sleep with a deterministic signal. For example, return a `Promise` from the handler that resolves when the message is received, then `await` it directly: `let resolve!: () => void; const received = new Promise<void>((r) => { resolve = r; }); handler: async () => { resolve(); }`. This eliminates the timing dependency entirely.",
      "confidence": 0.75,
      "faang_parallel": "Google's engineering productivity team's 2019 internal study on 'flaky test root causes' found that 28% of CI flakiness traced to timing-dependent assertions — sleeps, timeouts, and `setImmediate` as synchronization primitives. The fix for all of them was the same: replace the wall-clock wait with an event-driven signal (promise, CountDownLatch equivalent, or observable). The study coined the internal term 'timing coupons' — tests that work when the machine is fast but expire when the machine is slow.",
      "metaphor": "A `setTimeout(10ms)` in a test is a timing coupon: it pays out reliably on your local M3 but bounces on a GitHub Actions runner at 3am when the container is throttled. The message was delivered; you just didn't wait long enough for the cashier.",
      "teachable_moment": "FakeNats being synchronous is actually an asset here: it means the delivery is already deterministic — the async gap is purely in the event loop scheduling between the publish call and the subscriber's consume iteration. A single `await Promise.resolve()` (microtask yield) is often sufficient to flush a synchronous fake's delivery. The deterministic-signal pattern is always safer and documents intent better than any sleep duration.",
      "connection": "This is low severity precisely because FakeNats is synchronous and the gap is small. If FakeNats ever becomes async (e.g., to simulate JetStream ack latency), the timing fragility would promote to high-severity flakiness. Lock it down now while the fix is easy."
    },
    {
      "id": "F-010",
      "title": "PRAISE: Per-publisher hash chain isolation is correctly scoped",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/events/src/publisher.ts",
      "description": "The `publisherKey` default of `${emittedBy}:${signer.keyId}` correctly scopes the chain to a single (cell, signing key) pair. A subscriber that re-publishes enriched events creates an independent chain rather than extending the original publisher's chain. This prevents chain poisoning and correctly models the ACVP invariant that each signer is responsible for its own chain.",
      "confidence": 1.0,
      "faang_parallel": "This mirrors the partition-key design in Kafka's chain-of-custody systems at LinkedIn (the system that became Apache Kafka's audit log proposal in KIP-405). The key insight that team landed was identical: the chain must be scoped to the writer, not the topic. A topic-scoped chain breaks the moment two producers write to the same partition; a writer-scoped chain is composable by construction.",
      "teachable_moment": "The `${emittedBy}:${signer.keyId}` compound key encodes a principle worth stating explicitly in a comment: the chain belongs to the signer, not the channel. This single design decision is what makes subscriber re-publish possible without chain corruption. It deserves a sentence in the README's architecture section — it's not obvious to a consumer who wonders 'why not just use the topic as the chain key?'"
    },
    {
      "id": "F-011",
      "title": "PRAISE: RFC 8785 JCS canonicalization wired through at all hash boundaries",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/events/src/jcs.ts",
      "description": "JCS is used consistently for both `payload_hash` computation and `envelopeHash` computation (the basis for the next `prev_hash`). The subscriber recomputes both hashes from the same canonical form, guaranteeing byte-identical results across any conforming implementation. The jcs.ts module clearly documents the cross-runtime identity requirement (R15 invariant) linking this to the prior bash implementation.",
      "confidence": 1.0,
      "faang_parallel": "The W3C Verifiable Credentials working group spent 18 months debating canonicalization before landing on JCS in the VC Data Integrity spec. The prior approach (JSON-LD normalization) was correct but implementation-divergent — two conforming implementations would produce different hashes for the same semantic document. JCS's byte-determinism property is what makes cross-runtime hash identity possible without a shared runtime. The R15 invariant comment in this module is exactly the kind of forward-reference documentation that prevents the next developer from 'simplifying' to `JSON.stringify(obj)`.",
      "teachable_moment": "The most valuable line in jcs.ts may be the one that says 'do not substitute JSON.stringify here — key order is not guaranteed.' That's not a comment about this implementation; it's a warning about a class of future bugs. Write the comment that prevents the bug you know someone will introduce, not the comment that explains the code you already wrote."
    },
    {
      "id": "F-012",
      "title": "PRAISE: Fail-soft subscriber with typed VerificationFailureReason and test coverage for every failure mode",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/events/src/subscriber.ts",
      "description": "The subscriber's `consume()` loop catches every failure class, routes it through `onVerificationFailure`, and continues. The `VerificationFailureReason` union provides a typed discrimination surface for operators and monitoring. The test suite in roundtrip.test.ts covers all five failure modes (signature-invalid, payload-schema-invalid, prev-hash-broken-chain, and by extension envelope-schema-invalid via the valid path) with dedicated assertions, giving high confidence that the fail-soft contract is actually enforced.",
      "confidence": 1.0,
      "faang_parallel": "Stripe's webhook delivery architecture uses exactly this pattern: every delivery attempt that fails is classified into a typed reason enum (`signature_mismatch`, `payload_parse_failure`, `handler_exception`, `timeout`), reported to the operator via a callback, and then the delivery loop continues. The typed reason is what makes their retry logic, alerting, and customer-facing dashboard possible without a shared retry substrate. The design decision to separate 'the loop never dies' from 'the operator knows exactly why' is a maturity marker in event-driven systems.",
      "teachable_moment": "The typed `VerificationFailureReason` union is a public API contract: every value in the union is a promise to operators that this failure mode is detectable, reportable, and (eventually) recoverable. Every time the implementation catches an error and discards it silently, it breaks that contract. The fact that this subscriber makes the contract explicit in the type system — and tests all branches — is what makes F-002's misclassification fixable without breaking callers: you add a reason, you don't rename one."
    },
    {
      "id": "F-013",
      "title": "PRAISE: Zod `.strict()` on EventEnvelopeSchema prevents undeclared fields from silently entering the chain",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/events/src/envelope.ts",
      "description": "`.strict()` on EventEnvelopeSchema (line ~39) causes the publisher's defense-in-depth `EventEnvelopeSchema.parse(envelope)` to reject any envelope with extra fields before it enters the hash-chain. This prevents a class of bugs where a future code change accidentally adds a field to the envelope object — the publisher would catch it before it corrupts the chain rather than publishing an envelope the subscriber would fail to verify.",
      "confidence": 1.0,
      "faang_parallel": "The Zod `.strict()` usage here mirrors the 'exhaustive schema' pattern that Dropbox's sync team applied to their metadata layer after a 2020 incident where an unintended field in a sync record caused hash mismatches that silently dropped file versions. Their post-mortem recommendation: 'schemas that hash their output must reject unknown fields at write time, not parse time on read.' `.strict()` at the publisher is exactly this — a pre-hash gate that catches schema drift before it becomes a hash-chain corruption.",
      "teachable_moment": "`.strict()` on a Zod schema is one of those low-cost, high-leverage design choices that protects you from future-self. The envelope is hashed; the schema must exactly define what gets hashed; therefore the schema must reject anything outside its definition. This is not defensive paranoia — it's the logical consequence of hash-chain design. A future developer who adds a field to the envelope and sees the Zod error will understand immediately where the gate is and why it's there."
    }
  ]
}
```


---

## Callouts

This package makes several architectural choices that are worth naming explicitly because they'll pay forward dividends across every consumer.

**JCS throughout.** The decision to wire RFC 8785 canonicalization through every hash boundary — payload hash, envelope hash, prev_hash derivation — is the right call and a non-obvious one. Most teams discover the need for canonicalization after their first cross-runtime hash mismatch. Doing it from Sprint 1, with the cross-runtime invariant documented in the module, is the kind of foundation work that prevents a class of bugs from ever existing in the production codebase.

**Chain scoped to signer, not topic.** The `${emittedBy}:${signer.keyId}` publisher key is elegant and correct. It means the chain is a property of *who signed it*, not *where it was published*. This composes beautifully with ACVP's identity model and makes subscriber re-publish non-destructive by construction. This decision should be in the README's architecture section — it's a load-bearing invariant that any future publisher implementation needs to honor.

**Fail-soft subscriber with typed failure reasons.** The `onVerificationFailure` callback receiving a typed `VerificationFailureReason` is the kind of design that separates infrastructure from application code. Consumers can write production-grade monitoring (alert when `signature-invalid` appears more than N times in M minutes) without forking into the library. The test suite exercising every branch of the failure union is what makes this trustworthy — not just designed-to-be-testable, but actually tested.

**Zod `.strict()` at the publish gate.** Defense-in-depth at the right layer: the publisher validates the envelope before hashing it, so schema drift is caught at write time rather than as a hash mismatch on the subscriber side. This is the logical consequence of hash-chain design expressed as a single `.strict()` call. Beautiful.

---

zkSoju:
## BB review disposition (commit 39d765c)

5 substrate-clean fixes landed + 2 known-limitations documented inline + 2 follow-up beads filed. **51/51 tests passing** (+5 new covering the changed paths).

| Finding | Disposition | Where |
|---|---|---|
| **F-001 HIGH** JwksVerifier no fetch timeout | ✅ AbortController + configurable `timeoutMs` (5s default) + `finally`-clearTimeout | `signer.ts` + `tests/jwks-verifier.test.ts` |
| **F-002 MED** handler-error misclassified | ✅ Added `handler-error` + `internal-error` to `VerificationFailureReason`; removed the apology comment | `subscriber.ts` + roundtrip test |
| **F-003 MED** publish/store atomicity | 📝 Documented as known limitation — Sprint 1 acceptable; Sprint 2+ Redis-CAS mitigation path written into JSDoc | `publisher.ts` + README |
| **F-004 MED** empty-JWKS evicts cache | ✅ Non-empty guard; `onEmptyJwks` callback for alerting; `lastFetchMs` still advances on empty | `signer.ts` + JWKS test |
| **F-005 LOW** chain-recovery surface | 📝 Documented inline — operator wires `onVerificationFailure` to a chain-reset signal; reset pattern in subscriber.ts; Sprint 2+ first-class `onChainGap` noted | `subscriber.ts` + README |
| **F-006 LOW** `nats@2.x` deprecated | 📌 Filed as follow-up bead `cluster-events-pillar-coordinator-qns` — separate PR (peer-dep range coordination with sonar + characters) |
| **F-007 LOW** `nftMintDetectedTopic()` catch-all language | ✅ JSDoc + test description: no-slug returns a SPECIFIC subject (NOT a wildcard); directs to `NFT_MINT_DETECTED_WILDCARD` for subscribe-all | `topics.ts` + topics test |
| **F-008 LOW** `StaticPubkeyVerifier` not exported | ✅ Promoted to public API; README "Verifier choice" table | `index.ts` + README |
| **F-009 LOW** `setTimeout(10ms)` in tests | 📌 Filed as follow-up bead `cluster-events-pillar-coordinator-wdd` — CI-flake hardening before FakeNats ever becomes async |
| **F-010–F-013 PRAISE** | 📝 Surfaced in README's expanded "Design invariants" section (chain-scoped-to-signer; JCS no-substitute; typed failure reasons; Zod `.strict()` rationale) |

Ready for re-review.

zkSoju (commented):
Looking at the analytical findings and PR context, I'll produce the enriched Bridgebuilder review now.

## Summary

This PR lands the `@0xhoneyjar/events` ACVP envelope substrate — a cryptographically-chained NATS event pipeline with JWK signature verification, JCS canonicalization, and a fail-soft subscriber loop. The architectural bones are strong: the `NatsLike` duck-type abstraction, the typed `VerificationFailureReason` discriminant, and the F-003 atomicity documentation reflect the kind of design thinking that makes a library safe to extend under pressure. Two security findings need resolution before cross-service trust boundaries can safely rely on this: envelope metadata is not cryptographically bound to its signature, and the chain bootstrap carries no replay constraint for first-time subscribers.

---

The hardest part of building a cryptographic event substrate is not the crypto itself — it's knowing exactly where the trust boundary is drawn, and ensuring every field that influences a downstream trust decision lives inside it. EVT-001 and EVT-002 are variants of the same failure pattern: the system can prove *that* a message is authentic, but not *what kind* of authentic message it is. This matters enormously in an ACVP context where `event_type` and `emitted_by` are the fields that determine which handler fires and which actor is credited. Fixing EVT-001 is prerequisite to EVT-002's replay defense being meaningful — a replayed envelope with swapped `event_type` is undetectable until metadata is signed.

The JWKS verifier has the shape of a production-grade component with two unfinished edges. The thundering-herd on concurrent refresh (EVT-003) is the class of bug that passes review cleanly and surfaces at 3am during a traffic spike. The empty-cache-on-boot gap (EVT-004) is its quieter cousin — misconfigured operators receive no signal, just silent rejection of all events with no diagnostic path.

The test suite is notably thorough for a Sprint 1 package: six test files covering JCS, JWKS verification, envelope construction, topics, and full roundtrips with a custom `FakeNats` double. The timing-based polling (EVT-005) is the reliability risk worth addressing before CI runs accumulate flake debt that erodes confidence in an otherwise disciplined test suite.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "EVT-001",
      "title": "Signature covers only hashes — metadata fields (event_type, emitted_by, signing_key_id) are not cryptographically bound",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/events/src/envelope.ts:55-59",
      "description": "envelopeSigningBytes concatenates only prev_hash + payload_hash. Fields like event_type, emitted_by, schema_version, event_id, and emitted_at are entirely outside the signed input. An attacker with NATS publish access can capture a valid envelope, change event_type from nft.mint.detected.mibera-shadow.v1 to nft.mint.detected.other-collection.v1 (or change emitted_by from sonar-api to fabricated-service), keep the original signature, and republish. Step 4 of the subscriber verify loop (subscriber.ts:128-134) will still return sigOk=true because it reconstructs signing bytes from the (unchanged) envelope.prev_hash and envelope.payload_hash. Payload content is protected but routing metadata is not.",
      "suggestion": "Bind all non-signature envelope fields to the signed message. Replace the two-hash concat with the JCS-canonical form of a deterministic signing struct: jcsCanonicalize({ event_id, event_type, schema_version, emitted_by, emitted_at, prev_hash, payload_hash, signing_key_id }). Both publisher and subscriber must use the same canonical form. This is a wire-breaking change requiring a schema_version bump.",
      "confidence": 0.95,
      "faang_parallel": "The 2018 JOSE 'alg:none' and header-injection vulnerability class (CVE-2018-0114, documented by Tim McLean) exploited exactly this pattern at industry scale — JWT headers containing 'alg', 'kid', and 'typ' sat outside the signature boundary, enabling algorithm confusion and key-id spoofing. Auth0 and Okta both issued security advisories. AWS Cognito's identity token design responds directly: 'iss', 'sub', 'aud', and 'token_use' are all inside the signed claims payload precisely because routing and authorization decisions depend on them.",
      "metaphor": "Signing only the payload_hash is like a bank signing the dollar amount on a check but leaving the payee name unsigned. The signature proves the amount wasn't tampered with — but the name can be rewritten in transit and the check is still cryptographically 'valid.' The amount-only signature is worse than nothing because it creates false confidence.",
      "teachable_moment": "The rule for cryptographic message design: anything that influences a trust decision — routing, authorization, identity attribution, handler dispatch — must be inside the cryptographic boundary. Ask: 'If I removed this field from the signed input, could an attacker change it in transit to alter how a receiver handles the message?' If yes, it must be signed. In practice, the safest default is to sign a canonical encoding of the entire envelope minus the signature field itself.",
      "connection": "Connects directly to the ACVP core invariant 'hashes prove': the invariant is only satisfied when ALL semantically load-bearing fields are covered. EVT-002's replay defense becomes meaningful only after EVT-001 is resolved — a replayed envelope with swapped event_type metadata is undetectable under the current signing boundary regardless of chain checks."
    },
    {
      "id": "EVT-002",
      "title": "First message from an unknown publisher is accepted regardless of prev_hash — replay of mid-chain envelope is possible",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/events/src/subscriber.ts:165-170",
      "description": "When opts.chainStore.get(publisherKey) returns null (publisher never seen), the chain check is skipped unconditionally. A replayed envelope from 500 publishes ago with a valid signature and an arbitrary prev_hash (referencing some prior envelope the subscriber never saw) passes all checks and is handed to the application handler. For subscribers that bootstrap after a publisher has been running for a while this is expected behavior, but no mechanism exists to constrain what the first acceptable envelope looks like — not even requiring prev_hash === GENESIS_PREV_HASH.",
      "suggestion": "Add an initialPrevHashPolicy option ('any' | 'genesis' | string) to SubscribeOptions. Default to 'any' (current behavior, backward-compatible). When set to 'genesis', verify that the first envelope from an unknown publisher has prev_hash === GENESIS_PREV_HASH; surface prev-hash-broken-chain if not. This provides a zero-cost replay defense for subscribers that start from the beginning of a publisher's chain.",
      "confidence": 0.80,
      "faang_parallel": "Apache Kafka consumer groups solved the analogous bootstrap problem with the 'auto.offset.reset' policy ('earliest' | 'latest' | 'none'). 'none' throws an exception if no committed offset exists — forcing the operator to make an explicit decision rather than accepting an arbitrary starting point silently. The lesson carried forward into Kafka Streams and ksqlDB: bootstrap posture is a first-class operator decision, not a silent default.",
      "metaphor": "Imagine a security checkpoint that stamps every visitor's hand with a sequence number. A returning visitor must show the stamp that matches the next expected number. But a first-time visitor? No policy: anyone with a valid ID can walk in claiming to be sequence 500. The chain is intact from #500 onward — but the first 499 steps are unverifiable and a replay of any of them is invisible.",
      "teachable_moment": "Chain integrity protocols need explicit bootstrap semantics. 'We don't know what came before, so we accept anything' is a valid choice for late-joining subscribers — but it should be an explicit opt-in, not a silent default. Exposing the policy as a constructor option documents the security trade-off at the call site, makes it auditable, and lets strict environments enforce genesis-only bootstrapping.",
      "connection": "Complements EVT-001: once metadata is signed, replay of a swapped-metadata envelope becomes detectable via signature failure. But replay of a legitimate mid-chain envelope with a valid signature is still invisible without a bootstrap policy. These are two independent layers of the replay defense — both need to close."
    },
    {
      "id": "EVT-003",
      "title": "JwksVerifier concurrent refresh has no in-flight deduplication — thundering herd + stale-write race during rotation",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/events/src/signer.ts:220-226",
      "description": "Both the TTL-expired branch and the unknown-kid branch in verify() call this.refresh() without checking whether a refresh is already in progress. Two concurrent verify() calls that both see an expired TTL will both fetch the JWKS endpoint and both assign this.#cache = next. The last-write-wins. During a key rotation, if one fetch returns the old JWKS (CDN-cached) and the other returns the new JWKS, the one that finishes last determines the cached state — potentially reverting a just-loaded new key. In a subscriber loop processing a burst of messages this can trigger O(N) parallel fetches.",
      "suggestion": "Add a #refreshInFlight: Promise<void> | null field. In refresh(), if #refreshInFlight is set, return the existing promise instead of starting a new fetch. Clear it in a finally block. This collapses concurrent calls to a single in-flight request.",
      "confidence": 0.85,
      "faang_parallel": "Redis and Memcached operators call this the 'cache stampede' or 'thundering herd' problem — documented extensively in Facebook's 2013 Memcache at Scale paper (Nishtala et al.). Their solution, 'lease tokens', is the distributed analog of promise deduplication: only one client gets the token to refetch; others wait or serve stale. Node.js single-threaded concurrency makes the async-dedup version simpler — a single Promise field serves all concurrent callers without locks.",
      "metaphor": "The office coffee machine is empty. Without deduplication, every person who discovers it empty simultaneously starts brewing a new pot. Eleven people brew eleven pots, the last one finishes, and that's what everyone gets — regardless of which pot was freshest. With deduplication, the first person to notice pins a 'brewing' sign on the machine; everyone else waits for that one pot.",
      "teachable_moment": "In async JavaScript, the idiomatic thundering-herd fix is a 'promise coalescing' pattern: store the in-flight promise and return it to all subsequent callers. The finally block clears the field so future callers (after the current request completes) get fresh behavior. This pattern appears in virtually every production caching layer — knowing it by name makes it easy to spot when it's missing.",
      "connection": "The unknown-kid branch's eagerness to refresh (bypassing TTL) makes this worse: a burst of messages from a just-rotated key triggers O(burst-size) parallel fetches simultaneously, not just on TTL expiry. EVT-004 is the related quiet failure — this one is the loud one that surfaces under load."
    },
    {
      "id": "EVT-004",
      "title": "Initial fromUrl call with zero OKP Ed25519 keys silently produces an empty cache without calling onEmptyJwks",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/events/src/signer.ts:229-245",
      "description": "The non-empty guard in refresh() fires onEmptyJwks only when next.size === 0 AND this.#cache.size > 0. On the initial load from fromUrl(), this.#cache is an empty Map (size 0), so the condition is false and the else branch runs: this.#cache = next (also empty). If the JWKS endpoint is misconfigured (returns RSA keys, wrong URL, or an empty keys array), fromUrl() resolves successfully, onEmptyJwks is never called, and all subsequent verify() calls silently return false. Operators have no signal that the verifier is misconfigured.",
      "suggestion": "Always call onEmptyJwks when next.size === 0, regardless of prior cache state. Additionally, consider throwing from fromUrl() (or emitting an error event) when the initial load produces an empty key set, since an empty-on-boot verifier is almost certainly a misconfiguration.",
      "confidence": 0.90,
      "faang_parallel": "Netflix's Hystrix circuit breaker distinguished between 'open' (known failing) and 'half-open' (probing recovery) states precisely to avoid silent degradation — every state transition produced an observable event. The Hystrix team's post-mortems repeatedly cite silent failure as the failure mode that converts a 2-minute outage into a 2-hour one. Go's `crypto/tls` Config.GetCertificate returns an error on misconfiguration rather than silently serving no certificate — the pattern is consistent across production security infrastructure.",
      "metaphor": "A security checkpoint with no badge database loaded will either reject everyone (correct — loud failure) or let everyone through (catastrophic — silent failure). The current code does the latter: it constructs the checkpoint, loads nothing, and waves every badge through with a 'valid: false' stamp that looks like a real verification result rather than 'I don't know.'",
      "teachable_moment": "Security components should fail loudly at construction time when their configuration is detectably wrong. An empty key cache at boot is almost never intentional — it's a misconfigured URL, a network race, or a wrong JWKS endpoint. 'Fail fast, fail loudly' at startup converts a production mystery into an immediately actionable deployment error. The onEmptyJwks hook is the right abstraction — it just needs to fire unconditionally when next.size === 0.",
      "connection": "EVT-003 and EVT-004 share a root: the refresh() function has one job (populate the cache) but two distinct failure scenarios (silent empty-on-boot vs thundering-herd on expiry) that each need their own handling. Fixing both in the same PR makes sense — they share the fix surface."
    },
    {
      "id": "EVT-005",
      "title": "Test assertions use fixed-delay polling — flaky under CI load",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "packages/events/tests/roundtrip.test.ts:152",
      "description": "All roundtrip test cases use await new Promise((r) => setTimeout(r, 10)) (and 20ms in one case) to wait for the subscriber's async iteration loop to process messages. FakeNats dequeues synchronously, but the for-await loop still requires event loop turns. Under CI load or scheduling jitter, 10ms can be insufficient, producing false failures. This is the primary reliability risk in the test suite.",
      "suggestion": "Replace the polling delay with a completion signal derived from the handler itself. Example: let signalDone!: () => void; const done = new Promise<void>(r => signalDone = r); then call signalDone() inside the handler and await done instead of the timeout. For multi-message tests, use a countdown latch. This makes tests deterministic and removes the timing dependency.",
      "confidence": 0.80,
      "faang_parallel": "Google's Test Certified program (the internal quality bar that 'flaky tests must be fixed before merge') explicitly categorizes timing-based polling as Tier-1 flakiness. The Google Testing Blog's 2016 'Flaky Tests at Google' post measured that timing-dependent tests have 3–8x higher flake rates than event-driven equivalents — and that flake accumulation in a CI suite erodes developer trust in the entire suite over 6–12 weeks, not just the flaky tests. The fix pattern — promise-based completion signals — appears in every major Node.js testing guide as the canonical replacement.",
      "metaphor": "Checking if the cookies are done by waiting 10 minutes regardless of the recipe vs. using a timer that goes off when the internal temperature hits 375°F. The first method works when the oven behaves perfectly; the second is correct regardless of oven variation. CI runners are highly variable ovens.",
      "teachable_moment": "Async test reliability is an investment that compounds. A 1% flake rate on a test suite that runs 100 times per day means a false failure almost every day. The fix — a promise or latch that fires from inside the handler — is always available when you control the handler signature, and it removes an entire category of CI investigation from future developers' lives.",
      "connection": "The FakeNats double is an excellent substrate for deterministic testing (EVT-008 notes one gap). Pairing deterministic message delivery with event-driven completion signals would make this test suite genuinely robust — the kind that gives confidence rather than noise."
    },
    {
      "id": "EVT-006",
      "title": "nats peer and dev dependency is deprecated — downstream install warnings for all consumers",
      "severity": "LOW",
      "category": "dependencies",
      "file": "packages/events/package.json:55-58",
      "description": "pnpm-lock.yaml shows nats@2.29.3 deprecated: 'Package moved. Use @nats-io/transport-node'. nats appears as both a devDependency (tests) and as a peer dependency. While the production code correctly abstracts behind NatsLike, the peer dependency declaration signals to consumers which NATS client to bring. All consumers who follow that signal will see a deprecation warning on install.",
      "suggestion": "Migrate devDependency and tests to @nats-io/transport-node. Update peerDependencies to @nats-io/transport-node@>=3.0.0 <4.0.0. If backward compatibility with nats@2.x is needed, list both in peerDependencies with peerDependenciesMeta.nats.optional: true.",
      "confidence": 0.95,
      "faang_parallel": "The npm deprecation notice is the package ecosystem's equivalent of a public API deprecation warning in a major SDK. AWS SDK v3's migration from v2 generated years of consumer friction precisely because peer/dev dependency declarations in downstream libraries kept pulling in v2 transitively. The lesson absorbed into npm best practices: migrate peer dependencies at the point of the move, not after consumers have been warned repeatedly.",
      "metaphor": "Shipping a product with the old factory's address on the return label. Customers can still return it — but every return form tells them to go somewhere that no longer handles returns, and customer service has to redirect them. Better to update the label before shipping.",
      "teachable_moment": "The NatsLike abstraction is exactly right — production code has no hard dependency on the NATS client package. But peer dependencies are a signal to consumers about what to install. A deprecated peer dependency in a library v1 creates downstream noise that accumulates with every install and CI run across the entire consuming ecosystem.",
      "connection": "The `NatsLike` abstraction in place already decouples production code from the specific client — migrating the dev/peer dependency is purely a packaging concern, not an API concern. This is low-effort with the abstraction already in hand."
    },
    {
      "id": "EVT-007",
      "title": "GENESIS_PREV_HASH and EnvelopeHandlerContext not exported from main index — forces sub-path imports",
      "severity": "LOW",
      "category": "api-surface",
      "file": "packages/events/src/index.ts:1-7",
      "description": "GENESIS_PREV_HASH is not re-exported from the root index, so consumers checking envelope.prev_hash === GENESIS_PREV_HASH must import from the /envelope sub-path. EnvelopeHandlerContext (the type received by EnvelopeHandler) is also absent, making it impossible for consumers to type handler callbacks explicitly without a sub-path import. The test file works around this by importing directly from src/.",
      "suggestion": "Add GENESIS_PREV_HASH to the envelope re-export block. Add type EnvelopeHandlerContext to the subscriber re-export block.",
      "confidence": 0.95,
      "faang_parallel": "TypeScript's own design guidelines and the tRPC library's export philosophy both state that anything a consumer legitimately needs to write idiomatic code against your public API must be available from the root package path. The 'barrel file' pattern exists specifically to answer this — sub-path imports are an escape hatch, not the primary surface. When consumers import from src/, that's a signal the barrel is incomplete.",
      "metaphor": "A hotel lobby with a main entrance that leads to most rooms, but the conference room is only accessible through the loading dock. Technically reachable — but not the intended experience.",
      "teachable_moment": "The completeness test for a package's root export: can a consumer write every legitimate usage pattern — including type annotation of callbacks — using only `import { X } from '@0xhoneyjar/events'`? If not, the barrel is incomplete. GENESIS_PREV_HASH is particularly important here because EVT-002's initialPrevHashPolicy fix will make it a first-class consumer concern.",
      "connection": "EVT-002's genesis policy fix will make GENESIS_PREV_HASH a first-class consumer API constant. Exporting it now prevents the awkward situation where consumers need to import a constant from a sub-path to use a feature from the root subscribe API."
    },
    {
      "id": "EVT-008",
      "title": "FakeNats test double does not support NATS * single-segment wildcard — test infrastructure gap",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/events/tests/roundtrip.test.ts:82-93",
      "description": "subjectMatches only handles exact match and > (multi-token catch-all). NATS also supports * (exactly one token). Future tests for subscriptions like nft.mint.*.mibera-shadow.v1 would silently fail to receive messages in the fake, producing confusing false test failures without any test-infrastructure error.",
      "suggestion": "Extend subjectMatches to split both pattern and subject by '.' and match token-by-token: exact match per token, * matches any single token, > at the end matches any remaining tokens.",
      "confidence": 0.70,
      "faang_parallel": "This is the classic 'test double fidelity' problem — documented in Martin Fowler's 'Mocks Aren't Stubs' (2007) and revisited in Google's 'Software Engineering at Google' chapter on testing. The rule: a test double must implement the same contract as the real system for all cases the tests exercise, plus all cases the tests might reasonably exercise in the future. NATS's subject-matching semantics are a documented protocol; a partial implementation is a latent source of test confusion.",
      "metaphor": "A flight simulator that models takeoff and landing perfectly but has no wind model. Pilots trained on it will be surprised the first time real crosswinds affect their approach. The simulator works for the cases it covers, but the gap creates false confidence.",
      "teachable_moment": "When writing a test double for a protocol-defined system (NATS, HTTP, gRPC), write the double against the protocol spec rather than against current usage. Current usage rarely exercises all the cases the protocol supports, and the gap between 'what tests use' and 'what the protocol allows' is where future test failures hide.",
      "connection": "The topics.ts file defines a structured topic naming convention. Future topic patterns following nft.mint.*.v1 or similar cross-collection subscription shapes are exactly the kind of wildcard usage NATS is designed for — and exactly the gap that would surface in the next sprint's test expansion."
    },
    {
      "id": "EVT-009",
      "title": "PRAISE — Fail-soft subscriber with typed VerificationFailureReason discriminant",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/events/src/subscriber.ts:1-253",
      "description": "The 7-step verification pipeline catches failures at every layer independently (json-parse-error, envelope-schema-invalid, payload-hash-mismatch, signature-invalid, prev-hash-broken-chain, payload-schema-invalid, handler-error, internal-error) and routes each through onVerificationFailure without ever crashing the subscription loop. The distinction between handler-error (application threw) and internal-error (library bug) is especially precise and avoids conflating library reliability with application reliability.",
      "confidence": 1.0,
      "faang_parallel": "This is the structured-concurrency pattern that Erlang/OTP made famous and that Go channels + select brought to systems programming: isolate failure domains so one bad message doesn't kill the consumer. Kafka Streams's DeserializationExceptionHandler is the direct analog — it gives application code a typed hook on deserialization failure so the stream can continue processing rather than halting.",
      "metaphor": "An assembly line where each station has its own error light and bypass tray rather than a single emergency stop. A bad weld at station 3 pulls that part off the line without stopping stations 1, 2, 4, and 5. The typed reason is the label on the bypass tray — maintenance knows exactly which station flagged it.",
      "teachable_moment": "Typed failure discriminants at library boundaries are a gift to operators: they make monitoring dashboards, alerting rules, and postmortems dramatically more actionable. 'payload-schema-invalid' tells an oncall that the producer changed its schema; 'handler-error' tells them the consumer application threw. Same symptom (event not processed), entirely different root cause, entirely different response.",
      "connection": "The handler-error / internal-error distinction directly enables the F-003 atomicity gap documentation to be actionable — a handler crash (application failure) vs a chain fork (library behavior) produce distinguishable reasons, which means monitoring can separate 'producer chain forked on restart' from 'application handler is broken'."
    },
    {
      "id": "EVT-010",
      "title": "PRAISE — JwksVerifier empty-guard prevents self-DoS during key rotation",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/events/src/signer.ts:229-245",
      "description": "Preserving the existing cache when a JWKS refresh returns zero OKP Ed25519 keys is exactly the right behavior for partial key-rotation deploys where some nodes still serve the old key while others return nothing. The onEmptyJwks hook provides operator visibility without forcing a crash or a hard failure. Still advancing lastFetchMs avoids hot-looping on every verify() during an outage.",
      "confidence": 1.0,
      "faang_parallel": "This is the 'stale-while-revalidate' pattern — formalized in RFC 5861 for HTTP caching, adopted by Cloudflare Workers KV, and central to Varnish's grace-mode design. The insight that a stale known-good response is better than no response during a revalidation failure is one of the core reliability patterns in distributed systems.",
      "metaphor": "A museum that loses contact with the authentication server for its exhibits. The smart move: keep the last known list of authenticated items on display rather than closing the museum until contact is restored. Administrators get an alert, visitors see the correct exhibits, and no one is turned away because of a transient network issue.",
      "teachable_moment": "The lastFetchMs advance during an empty-cache refresh is subtle and correct: without it, every verify() call during a JWKS outage would trigger a refresh attempt, producing an O(traffic-rate) flood of failed HTTP requests — the definition of self-DoS. Advancing the timestamp rate-limits the retry without requiring an explicit backoff timer. This is the kind of small decision that separates production-grade code from code that looks correct.",
      "connection": "EVT-004 is the complement to this praise: the empty-guard works correctly for the rotation case (existing cache, new fetch returns empty) but silently misbehaves for the boot case (no existing cache, initial fetch returns empty). The fix for EVT-004 should preserve the rotation behavior while adding the boot-time loudness."
    },
    {
      "id": "EVT-011",
      "title": "PRAISE — Publish/store atomicity gap is documented with precise failure mode and two concrete mitigations",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "packages/events/src/publisher.ts:52-85",
      "description": "The F-003 JSDoc names the failure mode (crash between nats.publish and prevHashStore.set forks the chain), scopes the blast radius (bounded to crash/restart windows for single-instance healthy processes), and provides two specific Sprint-2+ mitigations (JetStream CAS pipeline on PubAck.seq; subscriber-side chain reset via onVerificationFailure). This prevents silent chain corruption going undiscovered in production and gives future implementers a concrete path forward.",
      "confidence": 1.0,
      "faang_parallel": "Leslie Lamport's foundational work on distributed system documentation — and the Apache Cassandra team's public postmortem culture — established the norm: document the failure mode you know exists but haven't fixed, with the scope, the trigger, and the mitigation path. The Cassandra team calls this 'known limitations with a path.' It's the opposite of burying technical debt: it makes the debt auditable and actionable. The JetStream CAS suggestion is precisely the right Sprint-2 primitive — NATS JetStream's sequence-based publish acknowledgment enables exactly the compare-and-swap the F-003 note describes.",
      "metaphor": "A bridge engineer who marks a load-bearing joint with a yellow flag, documents its rated capacity, its failure mode (fatigue cracking under sustained load), and the inspection interval needed to catch it before failure. The bridge opens. The flag stays visible. Future engineers know where to look and what to do. This is engineering honesty.",
      "teachable_moment": "Documenting a known gap is not an admission of failure — it's an act of trust in future maintainers. A gap without documentation becomes a mystery bug at 2am six months later. A documented gap with a failure scope and two mitigation paths becomes a planned Sprint-2 ticket. The JSDoc here is doing exactly what good inline documentation should: answering 'why isn't this fixed yet?' before anyone has to ask.",
      "connection": "The subscriber's handler-error / internal-error distinction (EVT-009 PRAISE) pairs with this: when a chain fork occurs due to the F-003 failure mode, the subscriber surfaces 'prev-hash-broken-chain' via onVerificationFailure — a typed signal that a monitoring system can catch and alert on, making the 'bounded to crash/restart windows' scope statement operationally verifiable."
    }
  ]
}
```


## Callouts

**EVT-009 — Typed failure discriminant**: The `VerificationFailureReason` union is the kind of precision that makes the difference between a library that ships and a library that runs production for years. Every callout in a Kafka Streams or Kinesis consumer that says "deserialization failed — see handler logs" instead of crashing the partition traces back to exactly this decision. The distinction between `handler-error` and `internal-error` is especially surgical.

**EVT-010 — Stale-cache-on-empty-refresh**: Preserving a valid cache when a JWKS endpoint returns nothing during key rotation is correct and non-obvious. Most first implementations fail loudly here — dropping the cache and rejecting all events — which converts a partial node rotation into a subscriber outage. The advancing of `lastFetchMs` during the no-op is quiet, correct, and exactly what separates a production-grade component from one that looks right in tests.

**EVT-011 — F-003 atomicity documentation**: The JSDoc that documents a known failure mode, scopes its blast radius, and names two concrete mitigations is the correct way to handle a gap you know exists but defer to a later sprint. It converts debt that silently accumulates into debt that's auditable, alertable, and sprint-planned. The JetStream CAS suggestion as the Sprint-2 path is technically precise.

The `NatsLike` abstraction enabling the `FakeNats` test double, the JCS canonicalization for deterministic signing, and the six-file test suite covering the full stack all point to a team that thought carefully about this library's operational surface before shipping Sprint 1. The two HIGH findings are real and need resolution — but they're on a substrate that clearly deserves to be fixed properly.

---

zkSoju:
## BB round 2 disposition (commit 9f6a2cc)

4 substrate-clean fixes landed + Effect.Schema migration + 2 follow-up beads filed. **61/61 tests passing** (+10 covering the new EVT paths).

| Finding | Disposition | Where |
|---|---|---|
| **EVT-001 HIGH** Signature only covers hashes | ✅ Full-envelope binding (`envelopeSigningBytes` = UTF-8(sha256-hex(JCS(env with `signature: ""`)))). **Wire-break:** `acvp-l1-v1` → `acvp-l1-v2` (v1 never deployed — superseded in-PR). Closes the `event_type` / `emitted_by` forgery vector. New test: tampered `event_type` → `signature-invalid`. | `envelope.ts` + `publisher.ts` + `subscriber.ts` + envelope/roundtrip tests |
| **EVT-002 HIGH** First-time replay accepted | ✅ New `SubscribeOptions.initialPrevHashPolicy: "any" \| "genesis" \| <hex>`. Default `"any"` (backward-compatible). `"genesis"` rejects mid-chain replays; `<hex>` pins an anchor. New `VerificationFailureReason: "initial-anchor-policy-violation"`. 3 new tests. | `subscriber.ts` + roundtrip tests |
| **EVT-003 MED** JWKS concurrent refresh thundering herd | ✅ `#refreshInFlight: Promise<void> \| null` field; concurrent verifies collapse to 1 fetch. New test: 10 parallel verifies → exactly 1 refresh. | `signer.ts` + JWKS test |
| **EVT-004 MED** Initial empty JWKS silent | ✅ `onEmptyJwks` fires unconditionally (was guarded on `#cache.size > 0`); `JwksVerifier.fromUrl()` throws on initial empty (fail loud at construction). 2 new tests: initial empty throws, RSA-only JWKS throws. | `signer.ts` + JWKS tests |
| **EVT-005 MED** `setTimeout(10ms)` in tests | 📌 Already filed last round as bead `wdd`; no change. |
| **EVT-006 LOW** `nats@2.x` deprecated | 📌 Already filed last round as bead `qns`; no change. |
| **EVT-007 LOW** Missing public-API exports | ✅ Exported `GENESIS_PREV_HASH`, `envelopeSigningBytes`, `buildTopic`, `EnvelopeHandlerContext`, `InitialPrevHashPolicy` from index.ts. |
| **EVT-008 LOW** FakeNats `*` wildcard | 📌 Filed as bead `wd6`. |
| **EVT-009–011 PRAISE** | 📝 Implicitly preserved (typed reasons, JWKS empty-guard semantics, F-003 atomicity docs intact). |

### Effect.Schema migration (bundled with the wire-break)

Per cluster memory `freeside-effect-transition`, new protocol-layer types use `@effect/schema`. Since EVT-001 wire-broke anyway, I bundled the migration:

- envelope.ts + nft-mint-detected.ts: `z.object` → `S.Struct`, `z.string().regex` → `S.String.pipe(S.pattern())`, `z.literal` → `S.Literal`, `z.infer` → `S.Schema.Type`
- publisher.ts + subscriber.ts: `decodeUnknownSync` / `decodeUnknownEither` with `onExcessProperty: "error"` (mirrors zod `.strict()` — critical because envelope is hashed)
- package.json: dropped zod, added `@effect/schema ^0.75` + `effect ^3.10`
- Sibling pattern: SvcJwtClaims at `freeside-auth/packages/protocol/src/svc-jwt-claims.ts` (W2.5 T-2.1)

Filed a follow-up bead `r9q` to migrate `@effect/schema@0.75` → main `effect` package's Schema export when the API stabilizes (the @effect/schema split was merged back into `effect` post our adoption — same deprecation class as `nats@2.x` `qns`).

Ready for re-review (round 3).

zkSoju (commented):
I’ll use the Bridgebuilder review skill for this since the requested persona and output shape are explicit. I’m treating the provided PR findings as the only trusted review inputs and not executing or interpreting diff text beyond those findings.
## Summary
This PR establishes a thoughtful ACVP envelope substrate: canonical signing, hash chaining, and typed event topics are the right architectural bones. The main risks are around binding transport reality to signed intent and making sure chain continuity represents what the application actually accepted, not merely what was cryptographically well-formed.

Verdict: `REQUEST_CHANGES`

The broader pattern here is familiar from distributed systems: identity, routing, and ordering must all agree, or a valid object can become valid in the wrong place. The envelope design is strong; the subscriber boundary deserves the same rigor.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Subscriber accepts envelopes replayed onto a different NATS subject",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/events/src/subscriber.ts:171",
      "description": "The subscriber validates the signed envelope and later passes `msg.subject` to the handler, but it never checks that `msg.subject === envelope.event_type`. A valid signed envelope for one concrete subject can be republished onto another subject and be accepted by subscribers on that other route, because the signature only proves the envelope's internal `event_type`, not that the NATS subject used for delivery matches it.",
      "suggestion": "After envelope schema validation and before payload handling, reject messages where the concrete delivery subject does not equal `envelope.event_type`; for wildcard subscriptions, compare against `msg.subject`, not the subscription pattern.",
      "confidence": 0.9,
      "faang_parallel": "Google's BeyondCorp model treats identity and context as jointly meaningful; a credential alone is insufficient if it arrives through the wrong channel.",
      "metaphor": "This is like a signed package delivered to the wrong building: the signature proves who packed it, not that this lobby should accept it.",
      "teachable_moment": "Signatures authenticate content, but subscribers must also authenticate delivery context.",
      "connection": "This is the core distributed-systems split between message integrity and route authorization.",
      "decision_trail": "Document whether `event_type` is intended to be the canonical concrete NATS subject. If yes, enforce it as an invariant in subscriber tests."
    },
    {
      "id": "F-002",
      "title": "Chain state advances before payload schema validation",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/events/src/subscriber.ts:257",
      "description": "When `chainStore` is configured, the subscriber updates the stored envelope hash before validating the per-event payload schema. A correctly signed envelope with an invalid payload schema will be rejected from the handler but still advance the subscriber's chain tip, making the next valid envelope appear continuous after an event the application never accepted.",
      "suggestion": "Move the per-event payload schema validation before the `chainStore.set(...)` call, or explicitly document and test that chain continuity is meant to track signed envelopes rather than accepted application events.",
      "confidence": 0.85,
      "faang_parallel": "Kafka consumer offset commits follow this lesson: commit after successful processing when the business invariant is 'processed', not merely 'seen'.",
      "metaphor": "It is a bookmark moving past a page you refused to read; the next reader thinks the story is continuous.",
      "teachable_moment": "State advancement must line up with the semantic boundary users care about.",
      "connection": "This is an ordering and checkpointing invariant, not just validation placement.",
      "decision_trail": "Record whether the chain tracks cryptographic receipt or application acceptance, then encode that choice in test names."
    },
    {
      "id": "F-003",
      "title": "Exported verifier depends on browser globals in Node-targeted package",
      "severity": "MEDIUM",
      "category": "runtime",
      "file": "packages/events/src/signer.ts:24",
      "description": "`bytesToBase64Url` and `base64UrlToBytes` use `btoa` and `atob`, but this package is configured for NodeNext and documented for NATS service consumers. These globals are only available in newer Node runtimes; older supported Node versions can fail at runtime even though the package and dependencies allow Node 18-era environments.",
      "suggestion": "Use `Buffer.from(bytes).toString('base64url')` and `Buffer.from(signature, 'base64url')`, or set an explicit Node engine that guarantees the required globals.",
      "confidence": 0.75,
      "faang_parallel": "Netflix's Node platform guidance emphasizes declaring runtime contracts explicitly because fleet heterogeneity turns small global assumptions into outages.",
      "metaphor": "This is packing a plug that only fits some outlets while labeling the appliance as broadly compatible.",
      "teachable_moment": "Library exports should depend on the narrowest guaranteed runtime surface.",
      "connection": "Operational readiness starts at package boundaries: engines, module format, and globals are part of the API.",
      "decision_trail": "Document the minimum Node version or switch to `Buffer` so the compatibility promise is encoded in code."
    },
    {
      "id": "F-004",
      "title": "Envelope and README comments still describe Zod schemas",
      "severity": "LOW",
      "category": "documentation",
      "file": "packages/events/README.md:62",
      "description": "The new package intentionally uses Effect.Schema, but the README quick-start says `payload is typed as z.infer<typeof NftMintDetectedSchema>`, and the architecture table calls the envelope and per-event schemas Zod schemas. This will confuse downstream consumers copying examples.",
      "suggestion": "Replace the remaining Zod references with Effect.Schema terminology, for example `S.Schema.Type<typeof NftMintDetectedSchema>` or the exported `NftMintDetected` type.",
      "confidence": 1.0,
      "faang_parallel": "Amazon's internal API docs culture treats examples as executable contracts because most consumers copy the first working snippet.",
      "metaphor": "The sign on the door points to the old building even though the team has moved.",
      "teachable_moment": "Docs are part of the developer interface; stale examples create production bugs through copy-paste.",
      "connection": "Schema-library migrations need documentation cleanup as much as code cleanup.",
      "decision_trail": "Add a short README note naming Effect.Schema as the authoritative schema system for this package."
    },
    {
      "id": "P-001",
      "title": "Signature now binds the full envelope",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/events/src/envelope.ts:104",
      "description": "The signing input is derived from the JCS-canonical envelope with an empty signature placeholder, which binds routing metadata, publisher identity, hash-chain state, payload hash, and payload contents together.",
      "suggestion": "Keep the EVT-001 regression tests in place; they cover the most important failure mode for this envelope substrate.",
      "confidence": 1.0,
      "faang_parallel": "Certificate Transparency's Merkle-log discipline shows the same principle: bind all security-relevant fields into the verifiable object.",
      "metaphor": "This turns the envelope into tamper-evident tape across every flap, not just a stamp on the front.",
      "teachable_moment": "Good signing design signs the complete meaning, not just the payload.",
      "connection": "This is the foundation that makes the remaining subscriber invariants worth enforcing.",
      "decision_trail": "Keep the canonicalization and signature-input contract documented as EVT-001 so future refactors do not weaken it."
    }
  ]
}
```


## Callouts
[Security] The full-envelope signing model is the strongest part of the PR because it binds metadata, payload, publisher identity, and chain state into one canonical artifact. That gives future reviewers a real substrate to reason about.

[Test Coverage] The roundtrip and verifier tests appear intentionally shaped around regression risks rather than happy-path ceremony. That mirrors the way mature platform teams test protocols: preserve invariants first, polish ergonomics second.

[Quality] The package boundary is clean: envelope, signer, subscriber, topics, and schemas are separated in a way that will let later transports or event types compose without turning the first sprint into a framework maze.

zkSoju:
## BB rd-3 disposition (commit 2093784) — codex-headless adapter

claude-headless timed out twice at the 300s cheval ceiling (diff grew ~28K tokens between rd-2 and rd-3 — past the > 100K-token timeout-ladder cap in `core/multi-model-pipeline.ts:deriveTimeoutMs`). Re-ran with `BRIDGEBUILDER_MODEL=codex-headless` (72s total). All 4 findings landed. **63/63 tests passing** (+2 covering F-001 + F-002).

| Finding | Severity | Disposition |
|---|---|---|
| **F-001** subject-binding gap (cross-subject replay) | HIGH | ✅ New `VerificationFailureReason: "subject-mismatch"`. Subscriber rejects when `msg.subject !== envelope.event_type` after envelope schema, before sig/hash checks. New test asserts wildcard subscriber catches byte-identical replay onto wrong subject + handler not invoked. |
| **F-002** chain advances before payload schema | MED | ✅ Reordered pipeline. Chain CHECK (continuity gate) at step 6; chain ADVANCE moved to step 8 — AFTER per-event payload schema. Chain now tracks application-accepted envelopes, not just signed ones. New test: bad-schema publish leaves chain at prior tip. |
| **F-003** `btoa`/`atob` portability | MED | ✅ Swapped to `Buffer.from(...).toString('base64url')` + `Buffer.from(s, 'base64url')` in signer.ts. Buffer slice into a fresh Uint8Array view (no large-ArrayBuffer pinning). package.json gains `engines: { node: ">=20.0.0" }` for explicit runtime contract. |
| **F-004** README Zod references | LOW | ✅ `z.infer<typeof X>` → exported `NftMintDetected` type alias. Architecture table updated (Effect.Schema everywhere). Subscriber row reflects rd-3 pipeline order. Description in package.json updated. |
| **P-001 PRAISE** full-envelope binding | — | ✅ Preserved — EVT-001 regression test still asserts event_type tamper → signature-invalid. |

Pipeline order is now (subscriber.ts):
```
1. JSON parse
2. envelope schema (Effect.Schema decode + onExcessProperty:"error")
3. subject-bind: msg.subject === envelope.event_type   ← rd-3 F-001
4. payload-hash recompute
5. signature verify (acvp-l1-v2 full-envelope binding)  ← EVT-001
6. chain continuity CHECK (+ initialPrevHashPolicy)     ← EVT-002
7. per-event payload schema                              ← rd-3 F-002 (moved up)
8. chain ADVANCE                                         ← rd-3 F-002 (moved down)
9. handler dispatch
```

Each step has at least one test covering its failure mode.

PR ready for re-review (round 4) or operator merge.

Issue #236, opened 2026-05-26
[network] BeaconV3 cluster sweep — follow-up tracker (13 issues across 7 repos)


zkSoju:
**Domain**: network · **Status**: tracker · **Originating sweep**: 2026-05-26

Single orientation surface for the next `/kickoff` `/coord` session. This issue indexes the 13 follow-ups from the cluster-wide BeaconV3 sweep that landed 2026-05-26.

## Sweep recap

8 cells now have V3-compliant `packages/protocol/beacon.yaml`. Substrate-fix landed in [#230](https://github.com/0xHoneyJar/loa-freeside/pull/230) (reconciled stale src/dist; CLI now supports V3). Cell PRs all merged:

| Cell | PR | Beacon state |
|---|---|---|
| score-api | [#198](https://github.com/0xHoneyJar/score-api/pull/198) | 🟢 active, 3 ACVP grounded |
| inventory-api | [#5](https://github.com/0xHoneyJar/inventory-api/pull/5) | 🟡 candidate, 2 ACVP aspirational |
| sonar-api | [#22](https://github.com/0xHoneyJar/sonar-api/pull/22) | 🟢 active, 3 ACVP aspirational |
| storage-api | [#13](https://github.com/0xHoneyJar/storage-api/pull/13) | 🟢 active, 3 ACVP grounded |
| mint-api | [#2](https://github.com/0xHoneyJar/mint-api/pull/2) | 🟡 candidate, scaffold-only |
| activities-api | [#19](https://github.com/0xHoneyJar/activities-api/pull/19) | 🟡 candidate, 7 ACVP grounded |
| freeside-mediums | [#4](https://github.com/0xHoneyJar/freeside-mediums/pull/4) | 🟡 candidate, 4 ACVP grounded |
| identity-api | (pre-existing on main) | 🟡 candidate, 2 ACVP placeholders |

## Cell follow-ups

- [ ] [sonar-api#23](https://github.com/0xHoneyJar/sonar-api/issues/23) — 3 ACVP tests (`tests/acvp/event_completeness|monotonicity|audit_replay`)
- [ ] [inventory-api#6](https://github.com/0xHoneyJar/inventory-api/issues/6) — 2 ACVP tests + externalize `inventory-snapshot.schema.json`
- [ ] [mint-api#3](https://github.com/0xHoneyJar/mint-api/issues/3) — 3 ACVP tests + 5 sealed schemas (Sprint S1 scope)
[...]

PR #237, opened 2026-05-27
fix(network/beacon-schema): src/dist divergence guard (#233)

### LINKED ISSUE #233: [network/beacon-schema] src/dist divergence guard — prevent the bug that triggered the BeaconV3 sweep
Part of cluster-wide BeaconV3 sweep: https://github.com/0xHoneyJar/loa-freeside/pull/230

**Domain**: network · **Category**: substrate-infrastructure

## Context

The sweep's root cause was a stale src/dist divergence in `packages/beacon-schema/`:
- `src/beacon-v3.ts` (TS source, committed) declared V3 as `Schema.extend(BeaconV2Schema, ...)` with `schema_version: "2"`
- `dist/src/beacon-v3.js` (compiled output, gitignored) had been REWRITTEN to identity-first shape with `schema_version: "3"`
- Two dispatched agents (sonar, storage) read the stale src and wrote beacons that failed V3 decode

This was fixed in #230 — but **nothing prevents it from happening again**. dist/ is gitignored; src/ is committed; nothing checks they stay in sync.

## Gap

A CI check that fails if `pnpm build` produces a different dist/ than what currently ships.

### Approach 1: build-then-diff (Recommended)

In `.github/workflows/ci.yml` (loa-freeside CI), add a job:

```yaml
beacon-schema-dist-fresh:
  runs-on: ubuntu-latest
  steps:
    - uses: actions/checkout@v4
    - uses: pnpm/action-setup@v4
    - run: pnpm --filter @0xhoneyjar/beacon-schema install
    - run: pnpm --filter @0xhoneyjar/beacon-schema build
    - run: |
        # If dist/ existed BEFORE build (it shouldn't — it's gitignored),
        # this catches a case where a committed dist/ files diverges from
        # fresh build. As a side effect, also catches a case where someone
        # commits dist/ files into the repo by mistake.
        if [ -n "$(git status --porcelain packages/beacon-schema/dist/)" ]; then
          echo "ERROR: packages/beacon-schema/dist/ has uncommitted changes after build."
          echo "This indicates committed dist/ files. dist/ should be gitignored."
          git status --porcelain packages/beacon-schema/dist/
          exit 1
        fi
```

### Approach 2: validate-src-equivalence

Stronger: after build, also verify that src/beacon-v3.ts produces the SAME runtime Schema shape as a hand-written "canonical" reference. Heavier maintenance burden.

### Approach 3: bundle dist into the npm package

Re-include `dist/` in `.npmignore` (currently gitignored, but ALSO not published). Then publish dist/ from CI on tagged releases. This makes src/ and dist/ converge at publish time. Doesn't prevent local drift but at least the npm consumer sees fresh code.

**Recommendation**: Approach 1 (lightweight, catches the actual bug pattern).

### `scripts/__tests__/beacon-schema-dist-fresh.test.sh`
- Run `pnpm --filter @0xhoneyjar/beacon-schema build` in a clean clone
- Assert `git status --porcelain packages/beacon-schema/dist/` is empty
- Assert `git ls-files packages/beacon-schema/dist/` is empty (dist is gitignored)

### Manual smoke test (during this issue's PR review)
- Intentionally commit a stale `dist/src/beacon-v3.js` (the OLD shape from before #230)
- Confirm: CI fails
- Revert + confirm CI passes

## Acceptance criteria
- [ ] New CI job in `.github/workflows/ci.yml` (or new workflow file)
- [ ] CI fails when dist/ is committed (intentional test by committing dist/ on a branch)
- [ ] CI passes on a clean main
- [ ] README/CONTRIBUTING notes the convention: dist/ is generated, never committed
- [ ] Companion: add `pnpm build` to `pre-commit` hook in beacon-schema (optional ergonomics)

## Notes for next session (`/kickoff` `/coord`)
- **Lowest-effort issue in the substrate set** — <1 hour, mostly YAML
- HIGH leverage — this is the bug that produced ~3 hours of agent debugging
- Composes with: dependabot/renovate-style automation (out of scope but worth mentioning)

zkSoju:
Closes #233 · refs meta-tracker #236

## Summary

Guards the bug class that triggered the 2026-05-26 BeaconV3 cluster sweep: `src/beacon-v3.ts` and `dist/src/beacon-v3.js` had drifted because `dist/` is gitignored and the V3 rewrite happened only in `dist/`. Two dispatched agents read the stale src and wrote beacons that failed V3 decode (substrate fix landed in #230).

This PR adds a CI guard so that drift cannot recur silently.

## What this changes

| File | Change |
|---|---|
| `.github/workflows/ci.yml` | New `beacon-schema-dist-fresh` job (runs on push to main + PRs to main) |
| `packages/beacon-schema/README.md` | New `## Development` section documenting the dist/ convention + recovery command |

## The guard — two layers

**Layer 1 · ls-files check (primary)**
```
committed=$(git ls-files packages/beacon-schema/dist/)
[ -n "$committed" ] && exit 1
```
Rejects ANY committed file under `packages/beacon-schema/dist/`. The bug class requires committed dist files (otherwise dist/ is gitignored and gets nuked on every fresh clone). This is THE check.

**Layer 2 · post-build git status (defense-in-depth)**
```
(cd packages/beacon-schema && pnpm build)
drift=$(git status --porcelain packages/beacon-schema/dist/)
[ -n "$drift" ] && exit 1
```
Catches force-added files that drift from a fresh build. Belt-and-suspenders for the rare case where dist/ has tracked files from before gitignore was added.

## Test plan

- [x] **/fagan review** — APPROVED 3/3 voices (opus-skeptic, gpt-reviewer, composer-reviewer), 0 findings, 0 dropped
- [x] **Local clean-state verify** — both layers pass with a clean tree (8/8 cell beacons V3-PASS against fresh build)
- [x] **Local drift-state verify** — both layers fire when a stale `dist/src/beacon-v3.js` is force-committed; both layers pass again after cleanup
- [ ] **CI throwaway-PR verify** — to be confirmed via a temporary `test-233-guard` branch that intentionally commits a stale dist file; CI must fail on the `beacon-schema-dist-fresh` job; branch + PR closed and deleted after proof (per build doc spec)

## Why CI-only (no pre-commit hook)

The issue body lists `pnpm build` in a pre-commit hook as "optional ergonomics" — explicitly out of scope. Single-purpose PR per SHIP discipline.

## Domain

Network only (`packages/beacon-schema/` + a workflow file scoped per-package). Single-domain per ADR-007 §D-3.


zkSoju:
Rot-sweep 2026-07-12 status: merged current main into this branch (44cfccd7) — GitHub's CONFLICTING flag was stale, the merge is clean and the branch is now MERGEABLE. However: Unit Tests + Integration Tests fail on the merge commit and re-failed on rerun (not flake), while main's own last Unit Tests run passed. agent-ci/mcp-gateway-ci failures ARE chronic estate-wide (PRs 446/434/433 all red there) and can be discounted — but the Unit/Integration delta needs a look before this lands: either a time-dependent test drifted since main's last push, or the ci.yml job merge interacts with the test workflow. Holding un-merged; the dist-freshness guard itself remains verified-absent from main and worth landing.

zkSoju (commented):
## Summary

This PR advertises a narrow CI guard for `packages/beacon-schema/dist/`, but the diff's `-88` line count tells a different story: an entire Postgres-backed `worker-db-tests` job, the integration job's env block, and workspace dependency installs are all deleted alongside. The guard itself is well-documented and its failure messaging is genuinely exemplary — but roughly half its steps are unreachable, and the collateral deletions are the real risk here.

**Verdict: REQUEST_CHANGES**

The central architectural observation: a diff whose stated scope is "add a guard" should be net-additive to CI. When a scoped fix removes 88 lines, the removals are the review. Google's Critique tooling surfaces net-negative diffs in scoped changes for exactly this reason — deletions ride in on the reviewer's trust in the title.

The second observation concerns what the guard actually binds. Enforcing "`dist/` is never committed" is hygiene; enforcing "`dist/` is derivable from `src/`" is the invariant the README claims. Those are different properties, and only the first is checked.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "PR deletes the entire worker-db-tests job, unrelated to the stated fix",
      "severity": "CRITICAL",
      "category": "correctness",
      "description": "The `worker-db-tests` job (Postgres 16 service, test schema application, DATABASE_URL skip-guard, and `EventNatsConsumer.test.ts` run) is removed wholesale and replaced in-place by the new `beacon-schema-dist-fresh` job. The PR title and README changes describe only a src/dist divergence guard. Nothing in the description justifies dropping real Postgres integration coverage for the NATS event consumer. The removed skip-guard was itself explicitly added as belt-and-suspenders against silent zero-assertion passes; deleting it restores exactly the failure mode it was written to prevent.",
      "suggestion": "Add `beacon-schema-dist-fresh` as a new job and restore `worker-db-tests` verbatim. If removing worker DB coverage is intentional, split it into its own PR with justification.",
      "confidence": 0.9,
      "faang_parallel": "Chesterton's Fence, formalized: Amazon's operational readiness reviews require a written justification to delete an existing integration check, because most such checks were themselves written in the aftermath of an incident. The skip-guard here is a literal artifact of a past silent-pass bug.",
      "metaphor": "Removing a smoke detector while installing a doorbell. Both are on the ceiling; only one was in scope.",
      "teachable_moment": "Editing a job in place rather than appending a new one is what made this reviewable-as-a-rename. Structurally, `+52 -88` in a file where the intended change is additive is the signal to read the removals first.",
      "connection": "Compounds with F2 and F3 — three independent coverage regressions arriving under one scoped title."
    },
    {
      "id": "F2",
      "title": "Integration test env block removed, reverting the RC-2 config-validation fix",
      "severity": "CRITICAL",
      "category": "correctness",
      "description": "All env vars for the integration job are deleted except `REDIS_URL` — including `DATABASE_URL`, `TRIGGER_SECRET_KEY`, `API_KEY_PEPPER`, `NODE_ENV`, etc. The removed comments state that `config.ts` validates all env at import via Zod and that the integration vitest project has no `setupFiles`, so absent vars produce 'Configuration validation failed: triggerDev.secretKey'. Either the integration job now hard-fails at import, or (per the removed RC-2 note) MigrationEngine/billing-agent-sovereignty tests silently skip. Both outcomes are regressions and neither relates to beacon-schema.",
      "suggestion": "Restore the env block on the integration-tests job.",
      "confidence": 0.88,
      "faang_parallel": "Import-time config validation is the pattern Netflix's Archaius and Meta's configerator both favor — fail at startup, not at request time. Its cost is that the environment becomes a hard dependency of *loading* the module, so CI env is no longer optional decoration.",
      "metaphor": "The tests are a car with an ignition interlock. Remove the key and it doesn't drive badly — it doesn't start at all, or worse, it rolls in neutral and you call that a trip.",
      "teachable_moment": "The dangerous outcome is not the hard failure — that's loud. It's the silent skip, where a green check means 'we ran zero tests.' Any suite that can skip should assert a minimum test count.",
      "decision_trail": "The deleted comments were the documentation of the RC-2 fix. Deleting the code deleted the reasoning with it; restore both."
    },
    {
      "id": "F3",
      "title": "Workspace package dependency installs removed from unit and integration jobs",
      "severity": "HIGH",
      "category": "correctness",
      "description": "The `Install workspace package dependencies` step (`packages/core && npm ci`, `packages/adapters && npm ci`) is deleted from both the unit-test and integration-test jobs, while only `themes/sietch` still runs `npm ci`. If sietch tests resolve into those workspace packages, their `node_modules` will be absent at test time. The corresponding `cache-dependency-path` entries are also removed, so this is not an accidental one-line drop.",
      "suggestion": "Restore the workspace install steps, or state in the PR description what makes them unnecessary (e.g., a migration to a real workspace-linked install).",
      "confidence": 0.8,
      "metaphor": "Removing three of four legs from a table and noting that it still stands — until someone leans.",
      "teachable_moment": "The matching `cache-dependency-path` deletions are the tell that this was deliberate. Deliberate-but-unexplained is worse than accidental: it means a rationale exists and wasn't written down."
    },
    {
      "id": "F4",
      "title": "Second guard step is unreachable — git status on a gitignored dir with no tracked files always returns empty",
      "severity": "HIGH",
      "category": "correctness",
      "description": "The 'Assert build produced no tracked-file changes' step runs `git status --porcelain packages/beacon-schema/dist/`. Because `dist/` is gitignored, untracked build output is not reported, and the preceding `git ls-files` step already hard-fails the job if any file under `dist/` is tracked. Therefore by the time this step runs, no tracked file can exist under `dist/` and the check is guaranteed to pass. Combined with the `pnpm install` + `pnpm build` steps that exist solely to feed it, roughly half the job is dead weight that will report green regardless of state.",
      "suggestion": "Either drop steps 2-4 and keep the `git ls-files` assertion as the whole guard, or make the freshness check real: build to a temp dir and diff against a checked-in expected artifact hash, or run `git status --porcelain --ignored=no` against a path that is actually tracked.",
      "confidence": 0.85,
      "faang_parallel": "This is the assertion equivalent of a test with no assertions — the class of bug that motivated mutation testing at Google (Petrović & Ivanković, 'State of Mutation Testing at Google'). A check that cannot fail is indistinguishable from a check that passes.",
      "metaphor": "A metal detector at a door that only guests without metal are allowed to reach. It beeps for no one, and the silence reads as safety.",
      "teachable_moment": "Every new CI guard deserves one commit that deliberately breaks the invariant, to watch the job go red. If you cannot make it fail, you have not built a guard.",
      "connection": "Directly feeds F8: the guard's mechanism is weaker than its stated purpose, and this step is where the gap becomes literal."
    },
    {
      "id": "F5",
      "title": "Node downgraded 22 → 20 in two jobs with no stated reason, creating version skew",
      "severity": "MEDIUM",
      "description": "The JSON-validation and markdownlint jobs move from node 22 to 20 while the unit, integration, and new beacon-schema jobs stay on 22. The PR is scoped to a dist/ guard; nothing explains the downgrade. Mixed runtime versions across CI jobs makes 'works in CI' ambiguous and can mask engine-range problems.",
      "suggestion": "Revert to '22' for consistency, or pin all jobs to a single `NODE_VERSION` workflow-level env var and explain the target version in the PR.",
      "confidence": 0.75,
      "metaphor": "Two clocks in one house, five minutes apart. Nobody notices until someone misses a train.",
      "decision_trail": "If 20 is intentional for these jobs, a one-line comment naming the constraint prevents the next contributor from 'fixing' it back."
    },
    {
      "id": "F6",
      "title": "pnpm install run inside a package subdirectory may fail under a workspace lockfile",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "`pnpm install --frozen-lockfile` runs with `working-directory: packages/beacon-schema`. If the repo has a root `pnpm-workspace.yaml` / `pnpm-lock.yaml`, pnpm resolves upward and the frozen-lockfile check applies to the whole workspace, or errors if no lockfile is found at that level. The rest of the workflow uses npm and `package-lock.json`, so it is not obvious this package has its own pnpm lockfile.",
      "suggestion": "Run `pnpm install --frozen-lockfile` from the repo root with `--filter beacon-schema`, or confirm a package-local `pnpm-lock.yaml` exists.",
      "confidence": 0.6,
      "teachable_moment": "Verify before acting on this one: check whether `packages/beacon-schema/pnpm-lock.yaml` exists. The broader smell is real regardless — this is the only pnpm usage in an npm workflow, and mixed package managers in one repo is a durable source of 'works locally' divergence."
    },
    {
      "id": "F7",
      "title": "pnpm/action-setup@v4 with no version pin",
      "severity": "MEDIUM",
      "category": "reliability",
      "description": "`pnpm/action-setup@v4` is used with no `with: version:` and no `run_install`. v4 requires the pnpm version to come from `packageManager` in package.json; if that field is absent the step fails with 'Multiple versions of pnpm specified' or 'No pnpm version is specified'. The `setup-node` step also precedes it without `cache: 'pnpm'`, so every run does a cold install.",
      "suggestion": "Pin `version:` explicitly or verify `packageManager` is set in the root package.json; consider adding `cache: 'pnpm'` after the pnpm setup step.",
      "confidence": 0.6,
      "faang_parallel": "Floating action tags are the supply-chain surface exploited in the 2025 tj-actions/changed-files compromise; SLSA guidance is to pin actions by commit SHA, not by major tag.",
      "metaphor": "Ordering 'the usual' at a restaurant that changed chefs."
    },
    {
      "id": "F8",
      "title": "The guard does not detect the failure mode it documents",
      "severity": "MEDIUM",
      "category": "design",
      "description": "The job comment states the bug was that 'the V3 rewrite happened only in dist/' while src stayed stale, and agents reading src produced beacons that failed V3 decode. Enforcing that dist/ is never committed prevents a *committed* stale artifact, but does nothing about a local working tree where someone edits dist/ directly and src remains behind — the edits are simply invisible to CI. The guard is a reasonable hygiene rule but is weaker than the framing suggests.",
      "suggestion": "Keep the ls-files guard, but add the check that actually binds the invariant: a test that imports the built package and round-trips a V3 beacon, so a stale `src/` fails regardless of what dist/ contains.",
      "confidence": 0.65,
      "faang_parallel": "Bazel's hermetic build model attacks this at the root: generated outputs are never checkable-in artifacts, and freshness is guaranteed by the action graph rather than by policy. Where hermeticity isn't available, a round-trip behavioral test is the cheapest substitute.",
      "metaphor": "Banning photocopies of the map to stop people navigating by a stale one. It helps — but the real fix is a test that walks the route.",
      "teachable_moment": "Prefer guards on the *property* (the built artifact decodes V3) over guards on the *proxy* (nobody committed dist/). Proxies drift from the property they stand for; properties don't."
    },
    {
      "id": "F9",
      "title": "Unquoted variable expansion in printf loop",
      "severity": "LOW",
      "category": "correctness",
      "description": "`printf '  %s\\n' $committed` relies on word splitting to iterate. Filenames containing spaces or glob characters will be split or expanded incorrectly in the error output. This is diagnostic-only output, so the exit code remains correct.",
      "suggestion": "Use `printf '  %s\\n' \"$committed\"` with the newline-separated value, or pipe: `git ls-files packages/beacon-schema/dist/ | sed 's/^/  /'`.",
      "confidence": 0.85,
      "metaphor": "A perfectly correct alarm with slightly garbled speech. It still wakes you; it just mispronounces the address."
    },
    {
      "id": "F10",
      "title": "Job name uses a non-ASCII separator",
      "severity": "LOW",
      "category": "style",
      "description": "`name: Beacon Schema · dist/ stays generated` uses a middle-dot. Job names feed into branch-protection required-status-check strings, where non-ASCII characters are easy to mistype when configuring the required check.",
      "suggestion": "Consider a plain-ASCII name such as `Beacon Schema - dist stays generated` to reduce friction when wiring branch protection.",
      "confidence": 0.4,
      "teachable_moment": "Low confidence, and worth verifying rather than assuming: this only bites if the job becomes a required status check. If it does, a mistyped required-check name silently never matches — and branch protection blocks every PR forever, which is a loud failure at least."
    },
    {
      "id": "F11",
      "title": "Actionable, self-documenting CI failure messages",
      "severity": "PRAISE",
      "description": "The failure output states the convention, gives the exact remediation command (`git rm --cached -r ...`), links the originating issue, and enumerates the offending files. A contributor hitting this red build can resolve it without reading the workflow source.",
      "faang_parallel": "This meets the bar Rust's compiler diagnostics set and that Go's vet tooling follows: an error should name the rule, show the violation, and hand over the fix. Most CI guards stop at exit 1.",
      "metaphor": "A parking sign that tells you the rule, why it exists, and where the nearest legal spot is."
    },
    {
      "id": "F12",
      "title": "README development section pairs the guard with human-readable guidance",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "packages/beacon-schema/README.md",
      "description": "The added Development section states the src-is-truth invariant, gives the build/test commands, links the issue, and pre-answers the CI failure. Documenting the convention alongside the enforcement is the part most CI-guard PRs skip.",
      "teachable_moment": "Enforcement without documentation teaches contributors that CI is arbitrary. Shipping both in one commit means the rule and its reason can never drift apart."
    }
  ]
}
```


## Callouts

**[Operational] The failure messaging is the best part of this PR.** Naming the convention, printing the offending files, handing over `git rm --cached -r`, and linking #233 means a contributor resolves a red build in one read — no workflow spelunking. That is a real reduction in the tax this guard imposes on everyone who trips it.

**[Quality] Documentation shipped with enforcement.** The README's Development section makes the invariant legible to humans and agents alike, and pre-answers the CI error. Rules and reasons that ship together cannot drift apart later.

**[Quality] The instinct behind the guard is correct.** Generated artifacts in version control are a durable source of exactly the divergence #233 describes. The critique in F4 and F8 is about mechanism, not intent — tighten the check to bind the property, and this becomes a guard worth keeping permanently.

**The path forward:** restore F1–F3 (or split them into their own PR with rationale), collapse the guard to the one step that can actually fail, and add the V3 round-trip test that binds the invariant the README promises. That's a smaller, sharper PR that does everything the title claims.

---

*One note on process: the `+52 -88` on a scoped fix is the whole story here. Worth adding a repo convention that CI job deletions require an explicit line in the PR description — the reviewer's trust in a narrow title is the attack surface, and it costs one sentence to close.*

PR #242, opened 2026-05-27
feat(operator-dash): NATS mTLS client cert + standardize on PEM-body env vars (Path ε)

zkSoju:
## Summary

Extend the operator-dash's NATS TLS options to accept optional client cert + key PEM bodies via `NATS_TLS_CLIENT_CERT` / `NATS_TLS_CLIENT_KEY` env vars, enabling mTLS auth against the Path-ε Railway-hosted nats-server. Standardize on PEM-body convention across all three TLS env vars to align with the sonar publisher (PR [#25](https://github.com/0xHoneyJar/sonar-api/pull/25)) and the characters subscriber.

This is the third of three sibling PRs landing the Path-ε mTLS substrate-config change:
- sonar-api PR #25 — landed
- freeside-characters — concurrent (under branch `cluster-events-pillar/path-epsilon-mtls-client-cert`)
- loa-freeside — this PR

## Why this exists

The cluster-events-pillar-v1 cycle is shifting from Synadia managed NATS to self-hosted nats-server on Railway with mTLS auth (Path ε). Railway service variables inject TLS cert content as env-var values, not file paths. The dash currently uses `caFile` (filesystem path); the other two consumers use `ca` (PEM body). This PR closes the convention gap and adds the client-cert presentation.

Reference: `grimoires/loa/specs/cluster-events-pillar-v1/go-live-path-epsilon-railway-nats.md` §Step 3c.

## Diff scope

| File | Change |
|---|---|
| `apps/freeside-operator-dash/src/events-trace.ts` | +35 / −4 — switch `caFile` → `ca`, add `NATS_TLS_CLIENT_CERT` / `NATS_TLS_CLIENT_KEY` env reads, both-or-neither guard, drop unused `node:fs/promises` import |
| `apps/freeside-operator-dash/src/events-trace.test.ts` | +178 new — vitest suite covering 5 cases (see below) |
| `apps/freeside-operator-dash/vitest.config.ts` | +9 new — minimal vitest config |
| `apps/freeside-operator-dash/package.json` | +3 / −1 — add `vitest` devDep + `test` / `test:run` scripts |
| `apps/freeside-operator-dash/tsconfig.json` | +1 / −1 — exclude `**/*.test.ts` from build output |
| `apps/freeside-operator-dash/pnpm-lock.yaml` | lockfile churn from adding vitest |

## Breaking change — option (a) chosen

The dash previously used `caFile` (file path); this PR switches to `ca` (PEM body). **Any deployment that set `NATS_TLS_CA` to a filesystem path will break.** The Path-ε runbook documents the migration — set `NATS_TLS_CA=\$(cat /path/to/ca.pem)` instead of the path itself.

I chose option (a) (clean break + documented) over (b) (runtime auto-detect path-vs-PEM) because:
- The Path-ε runbook explicitly standardizes the wire convention as PEM body across all consumers — runtime ambiguity would carry forward the inconsistency the runbook is trying to close.
- The only deployment that has `NATS_TLS_CA` set today is the dash itself (Sprint-3), and the runbook §Step 5 specifies operator-side env-var swap as part of Path-ε rollout. The breaking change is bounded.
- Less code, less surface area, fewer edge cases.

## Test plan

- [x] `pnpm test:run` → 5/5 pass (no prior tests in the dash app — this PR stands up the test infra)
- [x] `pnpm typecheck` → clean
- [x] `pnpm build` → dist generated, no test files in output
- [ ] Manual smoke against the Path-ε broker (post-merge, after Railway nats-server stood up per runbook §Step 2)

### New test cases

1. **Partial config rejected: `NATS_TLS_CLIENT_CERT` set without `NATS_TLS_CLIENT_KEY`** → lifecycle captures the explicit \"both-or-neither\" error; `nats.connect` is never called.
2. **Partial config rejected: `NATS_TLS_CLIENT_KEY` set without `NATS_TLS_CLIENT_CERT`** → same behavior.
3. **Both env vars set + CA** → `nats.connect()` invoked with `tls.ca` + `tls.cert` + `tls.key`. Verified via `vi.doMock(\"nats\")` capturing the connect options.
4. **Backward-compat: CA only, no client-cert vars** → `nats.connect()` invoked with `tls.ca` only (no `cert`/`key`).
5. **No TLS env vars at all** → `tls` field entirely absent from connect options (preserves the existing nats.js scheme-based TLS inference path).

## Path-domain-check

This PR touches only `apps/freeside-operator-dash/*` (classifies as `shared` per `tools/lib/domain-classify.sh`) — no cross-domain conflict.

### DIFF
--- a/apps/freeside-operator-dash/package.json
+++ b/apps/freeside-operator-dash/package.json
@@ -13,7 +13,9 @@
     "dev": "tsx watch bin/http.ts",
     "build": "tsc -p tsconfig.json",
     "start": "node dist/bin/http.js",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "tsc --noEmit",
+    "test": "vitest",
+    "test:run": "vitest run"
   },
   "dependencies": {
     "@0xhoneyjar/events": "file:../../packages/events",
@@ -27,6 +29,7 @@
   "devDependencies": {
     "@types/node": "^22.0.0",
     "tsx": "^4.20.0",
-    "typescript": "^5.7.0"
+    "typescript": "^5.7.0",
+    "vitest": "^4.1.7"
   }
 }
--- a/apps/freeside-operator-dash/pnpm-lock.yaml
+++ b/apps/freeside-operator-dash/pnpm-lock.yaml
@@ -39,6 +39,9 @@ importers:
       typescript:
         specifier: ^5.7.0
         version: 5.9.3
+      vitest:
+        specifier: ^4.1.7
+        version: 4.1.7(@types/node@22.19.19)(vite@8.0.14(@types/node@22.19.19)(esbuild@0.28.0)(tsx@4.22.3)(yaml@2.9.0))
 
 packages:
 
@@ -56,6 +59,15 @@ packages:
     deprecated: this package has been merged into the main effect package
     peerDependencies:
       effect: ^3.9.2
+
+  '@emnapi/core@1.10.0':
+    resolution: {integrity: sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==}
+
+  '@emnapi/runtime@1.10.0':
+    resolution: {integrity: sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==}
+
+  '@emnapi/wasi-threads@1.2.1':
+    resolution: {integrity: sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==}
 
   '@esbuild/aix-ppc64@0.28.0':
     resolution: {integrity: sha512-lhRUCeuOyJQURhTxl4WkpFTjIsbDayJHih5kZC1giwE+MhIzAb7mEsQMqMf18rHLsrb5qI1tafG20mLxEWcWlA==}
@@ -219,6 +231,15 @@ packages:
     peerDependencies:
       hono: ^4
 
+  '@jridgewell/sourcemap-codec@1.5.5':
+    resolution: {integrity: sha512-cYQ9310grqxueWbl+WuIUIaiUaDcj7WOq5fVhEljNVgRfOUhY9fy2zTvfoqWsnebh8Sl70VScFbICvJnLKB0Og==}
+
+  '@napi-rs/wasm-runtime@1.1.4':
+    resolution: {integrity: sha512-3NQNNgA1YSlJb/kMH1ildASP9HW7/7kYnRI2szWJaofaS1hWmbGI4H+d3+22aGzXXN9IJ+n+GiFVcGipJP18ow==}
+    peerDependencies:
+      '@emnapi/core': ^1.7.1
+      '@emnapi/runtime': ^1.7.1
+
   '@noble/curves@1.9.7':
     resolution: {integrity: sha512-gbKGcRUYIjA3/zCCNaWDciTMFI0dCkvou3TL8Zmy5Nc7sJ47a0jtOeZoTaMxkuqRo9cRhjOdZJXegxYE5FN/xw==}
     engines: {node: ^14.21.3 || >=16}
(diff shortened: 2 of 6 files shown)


zkSoju:
## bridgebuilder review · PR #242

**verdict**: 🟢 **APPROVE** with follow-up items (no blockers)

**severity counts**: 0 BLOCKER · 0 CRITICAL · 0 HIGH · 1 MEDIUM · 2 LOW · 1 NIT · 3 PRAISE

---

### MEDIUM

**M-1 · `render.ts:80` doc drift contradicts the new code contract**
The dash's own empty-state panel HTML still tells operators to set `NATS_TLS_CA=/path/to/ca.pem` — but the new code (post-#242) interprets that env var as the PEM body, not a path. Operators reading the panel will mis-configure exactly the var this PR is migrating.
- file: `apps/freeside-operator-dash/src/render.ts:80`
- fix: replace `NATS_TLS_CA=/path/to/ca.pem` with `NATS_TLS_CA=<PEM-body>` (or a `$(cat …)` example matching the commit message)
- *follow-up PR is fine; not a merge blocker because the code is the canonical contract — but operators consult the UI as documentation and the cost-of-mistake on TLS handshake debug is high.*

---

### LOW

**L-1 · PR description understates how clean option (a) actually is**
"Only existing consumer was the dash itself (per the PR author note)" → technically true for *path-form* consumers, but the framing reads as if breaking-change cost was barely-worth-it. In fact:
- `apps/worker/src/services/NatsClient.ts:226` → already uses `ca: process.env['NATS_TLS_CA']` (PEM body)
- `apps/gateway/src/nats/publisher.rs:59` → already uses `ca_pem` from env (PEM body)
- `infrastructure/terraform/nats.tf:138` → `printf '%s' "$NATS_TLS_CA"` (server side already treats as PEM body)

The dash was the **lone path-form outlier** in the cluster. Option (a) doesn't just "document a migration" — it closes an existing inconsistency. This strengthens the choice; worth surfacing in the PR description for the audit trail.

**L-2 · Test mock omits `nc.subscribe` — tests rely on the inner try/catch to swallow**
In tests 3/4/5 (success paths), `subscribeEnvelope` reaches `opts.nats.subscribe(...)` against the fake nc that only mocks `closed`/`close`/`drain`. The TypeError gets caught by `events-trace.ts:378-380`'s try/catch and written into `lifecycle.lastError`. Tests pass because their assertions only probe `connectSpy.mock.calls[0][0]` and `opts.tls`, never `lastLifecycleError` (which would contain `subscribe(...) failed: nc.subscribe is not a function`).
- file: `apps/freeside-operator-dash/src/events-trace.test.ts:36-46`
- not breaking now, but if a future refactor moves the catch boundary (or strict-mode enforcement adds an unhandled-rejection guard), these tests silently regress. Adding `subscribe: vi.fn(() => ({ [Symbol.asyncIterator]: async function*(){} , unsubscribe: vi.fn() }))` to the fake nc would make the test surface fully honest.

---

### NIT

**N-1 · `go-live-checklist.md` will be stale post-merge**
Lines 45/59/78 of `grimoires/loa/specs/cluster-events-pillar-v1/go-live-checklist.md` say "Path to bundled CA cert file inside the container **OR** direct PEM content". Post-#242 the "Path" half is no longer valid for any consumer. Worth a sweep in a follow-up PR (not blocking — runbook drift is documentation, not behavior).

---

### PRAISE

**P-1 · domain classification claim is correct**
PR description says "classifies as `shared` per `tools/lib/domain-classify.sh`" — verified. `apps/freeside-operator-dash/*` falls through the case statement and lands on the `*) echo "shared"` default at line 47 of `tools/lib/domain-classify.sh`. No cross-domain conflict triggered. The author actually grounded this rather than guessing.

**P-2 · option (a) vs (b) was the right call, and the reasoning holds up**
The Path-ε runbook's standardization argument is the load-bearing one. Auto-detect (option b) would have prolonged an inconsistency *for the dash specifically* when the other three cluster consumers (worker, gateway, finn ECS deploys) already converged on PEM-body. The "less code, less surface area" justification in the PR description is also true — the diff's TLS assembly block is 13 lines of pure plumbing, no behavior fork. Branching on path-vs-body at runtime would have been ~30 lines + a heuristic that could mis-classify a path that happens to start with `-----BEGIN`.

**P-3 · `.trim()` on env reads is the kind of defense most agents would miss**
Railway service-variables and copy-paste both reliably introduce CRLF or trailing-newline noise on PEM content. The comment block at `events-trace.ts:271-272` calls out *why* explicitly. This is the right altitude of paranoia for a TLS layer.

---

### vitest setup notes (not findings)

- `vitest@^4.1.7` matches `worker@^4.0.17` and `sandbox@^4.0.17`. The workspace already has 9 vitest configs; this is consistent.
- `globals: false` (the dash) diverges from every other vitest config in the repo which uses `globals: true`. The dash's choice is arguably *better* (TS-honest, no ambient `describe`/`it`); the test file imports `{describe,it,expect,vi}` from vitest cleanly. Worth a one-line comment in `vitest.config.ts` documenting the deliberate choice so a future contributor doesn't "normalize" it back.
- `include: ["src/**/*.test.ts"]` (co-located) vs the sibling apps' `include: ['tests/**/*.test.ts']` (separate dir) — the tsconfig `exclude: "**/*.test.ts"` change is the *natural consequence* of the co-located pattern. Sibling apps don't need it because they `exclude: "tests"` instead.
- Lockfile churn (722 lines) is mechanically inevitable for a fresh `vitest@4` install. No unrelated dep upgrades; only `vitest` + its required peers (`vite@8`, `rolldown`, `@vitest/*`, `lightningcss`, etc.). Verified via spot-check of the diff sections.

---

### CI status

3 failing checks (Unit Tests / Integration Tests / NPM Security Audit) are **pre-existing and unrelated** to this PR:
- Unit Tests fail on `Cannot find module './lib/circuit'` in `themes/sietch/opossum` (sietch theme, not dash app)
- Integration Tests fail in `sietch-service@6.0.0 test:integration` (sietch, not dash)
- NPM Security Audit fails in `apps/ingestor`, `apps/mcp-gateway`, `apps/worker` — **NOT** `apps/freeside-operator-dash`

Local validation in the dash workspace:
- `pnpm test:run` → 5/5 pass (224ms)
- `pnpm typecheck` → clean

---

🤖 generated by bridgebuilder review on cycle-cluster-events-pillar-v1 / sprint-4-go-live-railway

PR #254, opened 2026-05-29
fix(events): encoded_traits accept opaque collection string (not 0x-hex)

zkSoju:
## what
Relax `encoded_traits` in `nft.mint.detected` from `^0x[0-9a-f]+$` → opaque optional string (length-bounded 2048). Source + rebuilt committed dist.

## why
The `GeneralMints` `Minted` event emits `traits` as a **decimal-semicolon string** (`"25;10;78;..."`, trait indices — some >255), never 0x-hex. The first **live** Mibera Shadow-mint envelope was rejected `payload-schema-invalid`, blocking Ruggy's #stonehenge announce. The hex assumption never matched on-chain reality; envio always emitted the raw string; the byte-parity test used a fake `0xdeadbeefcafe` fixture that masked it; and the bot had only ever received a *synthetic* envelope (issue 0xHoneyJar/sonar-api#50).

## scope / safety
- `SCHEMA_VERSION` (`acvp-l1-v2`) **unchanged** — additive, backward-compatible.
- `encoded_traits` is producer-defined + collection-specific; consumers decode per-collection (Mibera codex) or ignore it (the bot reads NFT metadata for traits, not this field).
- tsc clean · 63/63 events unit tests pass.
- `freeside-characters` `main` already consumes this via pin `56585fd` (deployed `d991ac5`); live Shadow-mint announce verification in progress.

## follow-up lever (separate)
Producer (`publishEnvelope`) does **not** validate its payload against this schema at publish time — divergence only surfaces on the consumer's wire. Symmetric enforcement is the real fix; tracked separately.


zkSoju:
## Review: sound fix — MERGE (after the #304 gate + reviews)

The `encoded_traits` relaxation is the right call, and the inline rationale is exactly the kind of grounding that should accompany a contract change:

- Old: `S.optional(S.String.pipe(S.pattern(/^0x[0-9a-f]+$/)))` — a hex assumption that **never matched the real on-chain string**.
- New: `S.optional(S.String.pipe(S.maxLength(2048)))` — opaque, collection-specific, length-bounded for DoS safety.

Why it's correct:
- It fixes a **real live rejection** (`payload-schema-invalid` on the first GeneralMints emit; Mibera Shadow emits decimal-semicolon trait indices like `"25;10;78"`, not 0x-hex).
- It stays `optional` → no existing consumer breaks.
- It keeps a bound (2048) → not an unvalidated free-for-all.
- It ships with operator-signed **ADR-010** (Sonar→Ponder migration RPO/RTO sign-off), so the surrounding decision is already settled.

### One follow-up before this becomes a "done" story

The comment notes the bug was masked because *"the byte-parity test used a fake `0x…` fixture."* This PR relaxes the schema but **does not appear to fix that test** — no test file is in the diff. So the fixture that masked the original bug is still fake. → File/track a follow-up to replace it with a real GeneralMints payload, or the same class of drift can hide again behind a green byte-parity test. (This is the same "test asserts a fake fixture, not reality" pattern FAGAN flagged on #262.)

Net: relaxation is right and safe to land. Just don't let the masking fixture survive.

zkSoju (commented):
## Summary
Solid, well-documented fix that corrects a schema over-constraint (0x-hex → opaque decimal-semicolon string) caught by the first live emit. The rationale docblock is exemplary. Primary concerns: the masking fixture that caused this drift isn't repaired (no tests ship), an unrelated ADR rides along, and generated `dist/` artifacts are hand-committed.

The deeper lesson here is about **contract drift between a test's model of the world and the world itself**. A regex that never met real data is a fixture that lies confidently.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Root-cause fixture that masked the bug is not fixed in this PR",
      "severity": "HIGH",
      "category": "testing",
      "file": "packages/events/src/schemas/nft-mint-detected.ts",
      "description": "A byte-parity test asserted against a synthetic 0x fixture that never matched real on-chain output, so the invalid regex shipped and the first live emit was rejected payload-schema-invalid. This PR relaxes the schema but ships no test change; the masking fixture presumably survives and the new decimal-semicolon form is untested.",
      "suggestion": "Add a schema unit test asserting a real GeneralMints sample ('25;10;78;...') validates and a >2048-char string is rejected. Prefer a golden fixture captured from a real on-chain emit over a synthesized value.",
      "confidence": 0.7,
      "faang_parallel": "Google's testing culture calls this a 'test that tests the mock' — the SRE workbook warns that a fixture diverging from production is worse than no test, because it manufactures false confidence. AWS S3 similarly moved to contract tests replayed from real request captures after synthetic fixtures masked eventual-consistency bugs.",
      "metaphor": "A fire drill where the alarm is a recording you made yourself: everyone evacuates on cue, but you never learn whether the real alarm is even wired to the building.",
      "teachable_moment": "A green test proves conformance to the fixture, not to reality. Whenever a schema describes an external producer, at least one fixture should be sourced FROM that producer, not authored alongside the schema.",
      "connection": "This is the same failure class as consumer-driven contract testing (Pact): the producer's actual output is the contract, not the consumer's guess about it."
    },
    {
      "id": "F2",
      "title": "Unrelated ADR-010 bundled into an events schema fix",
      "severity": "LOW",
      "category": "scope",
      "file": "decisions/010-sonar-ponder-migration-rpo-rto-sign-off.md",
      "description": "A 70-line ADR on Sonar-Ponder migration RPO/RTO sign-off rides in a PR scoped to encoded_traits. Bundling unrelated docs muddies git history and review focus.",
      "suggestion": "Split the ADR into its own PR, or note the linkage in the description if the co-ship is deliberate.",
      "confidence": 0.6,
      "faang_parallel": "Atomic, single-purpose commits are load-bearing in the Linux kernel workflow — Linus rejects patches that mix concerns because `git bisect` and `git revert` lose their power when one commit answers two questions.",
      "metaphor": "Two receipts stapled together: fine until you need to return just one item.",
      "teachable_moment": "A PR is a unit of reversibility. If you can't revert it without losing something unrelated, it's doing two jobs."
    },
    {
      "id": "F3",
      "title": "Relaxed schema now accepts empty string for encoded_traits",
      "severity": "LOW",
      "category": "validation",
      "file": "packages/events/src/schemas/nft-mint-detected.ts",
      "description": "The old ^0x[0-9a-f]+$ required content; S.String.pipe(S.maxLength(2048)) accepts \"\". Since the field is optional, an empty string is ambiguous versus absence and would now reach per-collection decoders.",
      "suggestion": "If empty is never meaningful, add S.minLength(1) so presence implies content; otherwise document that \"\" is tolerated and equals absent.",
      "confidence": 0.45,
      "metaphor": "An optional 'notes' box that now also accepts a blank sheet of paper — you can't tell 'nothing to say' from 'forgot to write'.",
      "teachable_moment": "Loosening a pattern silently changes the empty-vs-absent contract. When widening validation, re-check the boundary cases the old rule rejected for free."
    },
    {
      "id": "F4",
      "title": "maxLength(2048) bound is unexplained/magic",
      "severity": "LOW",
      "file": "packages/events/src/schemas/nft-mint-detected.ts",
      "description": "2048 is justified only as 'DoS safety' with no derivation from the realistic max trait count. A future reader can't tell if it's evidence-based or an arbitrary round number.",
      "suggestion": "Add a one-line rationale (expected ~N chars for K traits, plus headroom) or tighten to a value grounded in real max traits per collection.",
      "confidence": 0.4,
      "faang_parallel": "Every limit in Google's API Design Guide carries a documented rationale — an undocumented cap becomes a cargo-cult constant nobody dares change.",
      "metaphor": "A weight limit sign with no units: is 2048 a safety margin or a guess someone typed once?",
      "decision_trail": "Document the derivation inline so the next maintainer can safely tighten or raise it instead of treating it as untouchable."
    },
    {
      "id": "F5",
      "title": "Compiled dist artifacts committed alongside source",
      "severity": "LOW",
      "category": "build",
      "file": "packages/events/dist/src/schemas/nft-mint-detected.js",
      "description": "Generated .js/.d.ts/.map/tsbuildinfo are hand-carried in dist/, creating merge surface and source/dist drift risk if a future edit forgets to rebuild. If this is a deliberate publishing convention it's acceptable, but the coupling should be CI-enforced, not manual.",
      "suggestion": "If dist must be committed, add a CI check that fails when dist is stale vs src. Otherwise gitignore dist and build on publish.",
      "confidence": 0.5,
      "metaphor": "Photocopying a document and filing both — until someone edits the original and the copy quietly lies.",
      "teachable_moment": "Anything derived from source should be regenerated by machine or verified by machine — trusting humans to rebuild is a drift generator."
    },
    {
      "id": "F6",
      "title": "Clear, evidence-backed rationale documented inline",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "packages/events/src/schemas/nft-mint-detected.ts",
      "description": "The docblock records the failure (first live GeneralMints emit rejected payload-schema-invalid), the wrong prior assumption (0x-hex), the real encoding (decimal-semicolon indices), the decode-or-ignore contract, and the DoS-bounding intent — and names the masking fixture.",
      "suggestion": "No change needed. Mirror the concrete failure example into the test suite so the documented scenario is also executable.",
      "confidence": 0.9,
      "faang_parallel": "This is the 'commit message as a letter to the future' discipline the Chromium and Git projects prize — the WHY survives long after the WHAT is obvious.",
      "metaphor": "A margin note in a textbook explaining why the earlier edition was wrong — it stops the next reader from re-making the mistake.",
      "teachable_moment": "The best defense against re-introducing a bug is writing down why the wrong version looked right. This docblock does exactly that."
    }
  ]
}
```


## Callouts
- **[Quality] F6 (high confidence):** The inline post-mortem is the strongest part of this PR. It converts a production incident into durable, first-person context — the fix teaches, not just corrects.
- **[Quality] Right diagnosis:** Replacing a producer-guessing regex with a bounded opaque string plus a decode-or-ignore consumer contract is the correct resilience posture: validate shape and size, defer semantics to the decoder that actually knows the collection.

The through-line: F1 and F6 are two halves of one virtue — F6 documents the drift beautifully; close the loop by making that documented scenario *executable* (F1) so the fixture can never lie again.

PR #256, opened 2026-05-31
fix(platform): config-validation logger.fatal guard — keystone Unit Tests fix (sprint-bug-332)

zkSoju:
## sprint-bug-332 — keystone fix: required `Unit Tests` CI context red on main

**Bug 20260530-ecc0b3.** `TypeError: logger.fatal is not a function` at
`themes/sietch/src/config.ts:702` (the import-time config-validation throw path).
In CI, `.env.local` is absent → validation reaches the throw path; a `.fatal`-less
`vi.mock` of `utils/logger.js` leaking from sibling unit files made `logger.fatal`
undefined → TypeError → **0 tests collected across 10 files**. Green on dev
(`.env.local` satisfies validation), red in CI.

**Fix:** `(logger.fatal ?? logger.error ?? console.error).call(...)` on the
throw-only path. No behavior change for valid config; test-ordering-independent.
`+10/-1` + 1 test.

**Test-first:** `tests/unit/config-validation-logger-fatal.test.ts` reproduces the
condition (dotenv no-op + `.fatal`-less logger + empty env) — red pre-fix
(TypeError), green post-fix.

**Why first:** this is the keystone — it turns the required `Unit Tests` status
check green on `main`, which the stacked network-plane PR (doctor + ACVP) inherits.
**Merge this first.**

Domain: **platform** (`themes/sietch/`). beads: `arrakis-au2v`.


zkSoju (commented):
## Summary

A targeted fix that lands a valid regression gate but solves the symptom in production code rather than at the source — the incomplete logger mock contract in sibling test files. The new test (`config-validation-logger-fatal.test.ts`) is genuinely excellent: it precisely reproduces all three compounding CI failure conditions and will catch this class of regression forever. The production change is where the design conversation lives.

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-1",
      "title": "Production code fix treats a test mock discipline problem",
      "severity": "REFRAME",
      "category": "design",
      "file": "themes/sietch/src/config.ts",
      "description": "The `(logger.fatal ?? logger.error ?? console.error).call(logger, ...)` pattern addresses mock leakage across vitest workers, which is a test infrastructure problem. The sibling test files that mock `utils/logger.js` without a `fatal` key are the root cause. The production `parseConfig` path should not need to defend against an incomplete logger contract — in production, a pino logger always provides `fatal`.",
      "suggestion": "Add `fatal: vi.fn()` to the logger mock in every sibling test file that currently omits it (statsService, digestService, tierService tests). This fixes the mock leak at source without coupling production code to test infrastructure assumptions. The production call can revert to `logger.fatal(...)`. Keep the regression test — it's valuable — but point it at the corrected mocks rather than the production fallback.",
      "confidence": 0.85,
      "faang_parallel": "The Google SRE book's 'eliminating toil' principle applies directly here: when you defend production code against test infrastructure assumptions, you've created toil that compounds. The Firebase SDK team hit this pattern in 2019 — they had `if (typeof this.fetch === 'function')` guards in production code that existed purely because jsdom didn't expose the full Fetch API. The eventual fix was upgrading the test harness, not accumulating guards. The guards persisted for two years because 'the tests pass.'",
      "metaphor": "Imagine a restaurant kitchen adding a 'check if the knife exists before cutting' guard to every recipe because the supply closet sometimes gives you empty sheaths. The real fix is the supply closet, not a guard in every recipe. The recipe should be able to assume a complete knife.",
      "teachable_moment": "The pino logger interface is a contract. When your test doubles don't honor that contract fully, you've broken the substitution principle — and any production code you write to compensate is technical debt that silently expands the gap between the contract and reality. The right discipline: every mock must implement the full interface it substitutes, or use a typed mock factory that enforces this statically.",
      "connection": "This connects to the broader Liskov Substitution Principle at the test-double layer. TypeScript's structural typing makes partial mocks easy to construct and easy to get wrong — `{ info: vi.fn(), warn: vi.fn() }` satisfies a subset of the pino Logger interface but not the full type. Consider `satisfies Logger` on mock objects to catch this class at compile time rather than runtime."
    },
    {
      "id": "F-2",
      "title": "`console.error.call(logger, ...)` loses structured pino-style output in the final fallback",
      "severity": "MEDIUM",
      "category": "observability",
      "file": "themes/sietch/src/config.ts",
      "description": "When both `logger.fatal` and `logger.error` are absent, the code falls to `console.error.call(logger, { errors: result.error.issues }, 'Configuration validation failed')`. `console.error` does not interpret the first argument as a structured log context; it logs the object and the message string as separate positional args, losing the structured error detail. In a terminal this may render as `[Object] Configuration validation failed` instead of expanded issue paths.",
      "suggestion": "In the final fallback branch, flatten the structured object into the message: `console.error('Configuration validation failed:', result.error.issues)`. This preserves human-readable diagnostics when the fallback path is reached.",
      "confidence": 0.75,
      "faang_parallel": "Meta's internal logging infrastructure famously separates structured log 'context bags' from the message string — the two are never positional siblings. When `console.error` swallows a context bag as position-0, you lose the structured data in exactly the environment (degraded/CI) where you need it most. The AWS Lambda team documented this exact loss pattern in their Node.js runtime guidance: `console.error(obj, msg)` vs `console.error(msg, obj)` produces different JSON output in CloudWatch, and the former collapses the object.",
      "metaphor": "It's like filing a police report where the description and the incident form get stapled together instead of filled out — technically both are 'there,' but the structured fields that make it searchable and actionable are lost in the merge.",
      "teachable_moment": "When you write a fallback logging path, ask: 'Will this be readable in the environment where fallbacks are most likely?' Degraded test environments are often the least-debuggable, which means the fallback is most important precisely when the output is hardest to parse.",
      "connection": "Connects to F-3 — the `.call(logger, ...)` uniformity hides that `console.error` has fundamentally different call semantics. Separating the branches would expose this naturally."
    },
    {
      "id": "F-3",
      "title": "`console.error.call(logger, ...)` — `this` binding is irrelevant and misleading",
      "severity": "LOW",
      "category": "correctness",
      "file": "themes/sietch/src/config.ts",
      "description": "Calling `.call(logger, ...)` is correct for bound pino methods (`logger.fatal`, `logger.error`) since pino uses `this` internally. However, when the expression resolves to `console.error`, the `this` binding is ignored by the native function. The pattern implies consistent semantic meaning across all three branches when the third branch is actually different in kind.",
      "suggestion": "Either handle `console.error` as an explicit `else` branch after the logger method check, or at minimum add a comment on the `.call(logger, ...)` noting it is a no-op for the `console.error` branch. Separating the branches makes intent explicit.",
      "confidence": 0.8,
      "faang_parallel": "The V8 team's internal style guide explicitly flags 'false symmetry' — code that looks structurally uniform but has divergent runtime semantics. The Node.js `EventEmitter` source itself has a comment in several methods: `// .call(target) is intentional — EventEmitter internals bind to this`. When that comment is absent and the bound target is a native function, the reader correctly suspects a bug.",
      "metaphor": "Passing a key to a padlock, a combination lock, and a fingerprint scanner through the same 'open door' function — the first two use the key, the third ignores it entirely. The interface looks clean but one branch is a silent no-op.",
      "teachable_moment": "Uniformity in code structure should reflect uniformity in semantics. When you achieve structural uniformity by hiding a semantic divergence, you've made the code harder to reason about, not easier. Explicit branches with clear intent beat elegant-but-misleading unification.",
      "connection": "Resolves naturally if F-1 is addressed — the `console.error` fallback branch disappears entirely when production code can assume a complete logger contract."
    },
    {
      "id": "F-4",
      "title": "6-line production comment encodes test infrastructure details",
      "severity": "LOW",
      "file": "themes/sietch/src/config.ts",
      "description": "The inline comment references bug IDs, vitest worker behavior, pnpm-vs-npm-ci module layout, and CI environment specifics. These details are ephemeral and belong in the commit message or test file header, not in production source. A reader of `config.ts` in two cycles will see unexplained CI trivia with no path to verify it.",
      "suggestion": "Replace the 6-line comment with a one-line note: `// logger.fatal may be absent in degraded test environments; degrade gracefully.` Move the full causal explanation to the commit message.",
      "confidence": 0.9,
      "faang_parallel": "The Linux kernel's coding style famously states: 'Comments should explain WHY, not WHAT.' Linus has rejected patches specifically for encoding bug-tracker IDs in source comments — the commit message is the right place for that trail. Source comments should survive rebases and cherry-picks with full explanatory power. A comment that says 'see CI bug #332 for context' becomes a dead end the moment you leave the original repo.",
      "metaphor": "A sticky note on a whiteboard that says 'this happened because of the Tuesday incident' is useful for a week. Painted into the wall, it's archaeology that requires a historian to interpret.",
      "teachable_moment": "Production code comments should be self-contained and timeless. Ask: 'Will this comment make sense to someone with no access to our issue tracker, CI system, or package manager choices?' If not, it belongs in the commit message, not the source.",
      "connection": "The comment weight also signals the design tension in F-1 — when a production fix requires a 6-line explanation to justify its existence in source, that's a smell that the fix is in the wrong layer."
    },
    {
      "id": "F-5",
      "title": "`process.env` restoration in test uses `afterEach` without `try/finally`",
      "severity": "LOW",
      "file": "themes/sietch/tests/unit/config-validation-logger-fatal.test.ts",
      "description": "The test saves `ORIGINAL_ENV` at module scope and restores it in `afterEach`. If a future `beforeEach` addition throws before `process.env` is reassigned, the env is left polluted for subsequent tests in the same worker.",
      "suggestion": "Use `vi.stubEnv` / `vi.unstubAllEnvs()` (available in vitest ≥0.29) or wrap the env assignment in a `try/finally` within `afterEach`. Either approach guarantees restoration on unexpected failure.",
      "confidence": 0.7,
      "faang_parallel": "Jest's `restoreMocks` and vitest's `vi.stubEnv` were both added after teams at Airbnb and Netflix hit this exact pattern at scale — when test suites grow beyond ~500 tests per worker, manual env restoration produces intermittent, order-dependent failures that are notoriously hard to reproduce locally. The vitest team explicitly cites 'env pollution between tests' as the primary motivation for `vi.stubEnv` in their 0.29 release notes.",
      "metaphor": "It's like relying on a 'clean up after yourself' policy in a shared kitchen instead of using a dishwasher with a guaranteed cycle. Works fine with 3 people. Falls apart with 30.",
      "teachable_moment": "Framework-provided teardown primitives (vi.stubEnv, jest.replaceProperty, etc.) are worth the API surface cost because they compose with the test runner's own teardown guarantees. Manual save/restore only works if your afterEach always runs to completion — and that assumption is fragile.",
      "connection": "Minor given the current test count, but worth adopting now while the file is fresh. `vi.unstubAllEnvs()` in `afterEach` is a one-liner replacement."
    },
    {
      "id": "F-6",
      "title": "Regression test precisely isolates the CI-only failure condition",
      "severity": "PRAISE",
      "category": "test-quality",
      "file": "themes/sietch/tests/unit/config-validation-logger-fatal.test.ts",
      "description": "The test replicates all three compounding factors simultaneously — no-op dotenv, `fatal`-absent logger mock, and minimal process.env — producing a deterministic reproduction independent of local `.env.local` files. Dynamic `await import()` after `vi.resetModules()` ensures fresh module evaluation under the active mock state. Assertions cover both the negative condition (not a TypeError) and the positive condition (Error with the correct message prefix), making the test self-documenting as a regression gate.",
      "confidence": 0.95,
      "faang_parallel": "This is the 'hermetic test' pattern that Google's Testing Blog describes as the gold standard for CI-only failures: reproduce the exact environment delta, not an approximation of it. The `vi.resetModules()` + dynamic import technique is the same mechanism the Nuxt team uses for testing module initialization paths — it's the right tool and it's used correctly here.",
      "metaphor": "Most regression tests are like recreating a car accident with toy cars on a table. This one rented the actual intersection, the exact road conditions, and the same lighting — deterministic, hermetic, impossible to misread.",
      "teachable_moment": "When a bug is environment-dependent (local works, CI fails), the test's job is to make the environment delta the test's *variable*, not its *assumption*. This test does exactly that by asserting that a `fatal`-absent logger AND a no-op dotenv AND minimal env together produce a graceful error — not a TypeError. That's the full regression surface captured in one test case."
    }
  ]
}
```


---

## Callouts

**F-6 deserves to be read by anyone debugging CI-only failures.** The combination of `vi.resetModules()` + dynamic `await import()` to force fresh module evaluation under mock state is not obvious — most engineers reach for static imports and wonder why their mock isn't applied. The three-factor reproduction (no-op dotenv + absent `fatal` + minimal env) shows genuine diagnostic discipline. This test will earn its keep.

**The PR is honest about its trade-offs.** The commit scope (`fix(platform)`) and title accurately describe what landed without overclaiming. The regression test is paired immediately with the fix, which is the right instinct even when the fix layer is debatable.

---

**Architectural meditation**: The deeper question this PR surfaces is where the logger contract lives. Pino's interface is wide; most code only exercises a slice of it. When tests only mock that slice, partial-mock drift is inevitable as the codebase grows. A typed mock factory (or a `satisfies Logger` assertion on every mock object) would make this class of failure a compile-time error rather than a CI mystery. That's a one-cycle investment that eliminates an entire category of future sprint-bug entries.

PR #260, opened 2026-05-31
feat(network): per-cell resolution bridge + FL-B0 receipt freshness (doctor-acvp step 3)

zkSoju:
## doctor-acvp step 3 — the receipts-actually-consumed bridge

`freeside-cli doctor --cells-dir` now resolves each building's beacon + `cluster.eventsPin` + proof receipts + HEAD from a per-cell clone — **all read from the committed HEAD** (`git show`), so a working-tree mutation can't influence the verdict. The validator gains an additive `checkReceiptFreshness` resolver (FL-B0 restored): a committed receipt is **fresh** iff nothing changed between the receipt commit and HEAD *except the receipt file itself* — so a real in-repo receipt resolves to `contract_status: bound` (exact-HEAD match alone never could, since publishing the receipt advances HEAD past the `commit_sha` it records).

- 🔒 committed-only evidence — beacon · receipt · eventsPin · sealed schemas · freshness diff
- 🚫 cells-dir mode is pure-β: a missing clone → `beacon_unreachable` (never fixture substitution); no file-presence fallback
- ✅ tests: beacon-schema 54 · freeside-registry 7 · freeside-cli 27 — all green

**FAGAN-converged over 4 multi-model iterations** — TOCTOU, abbreviated-SHA, null-head core gate, committed-vs-working-tree evidence, untrusted-receipt shape-guard, fail-open freshness, working-tree beacon, fixture fallthrough, git-hang timeout (all fixed + regression-tested).

⚠️ **Held by design:** unsigned receipts are forgeable by a first-party *committed* act (fine for a drift-detection threat model; Ed25519 signing tracked in `arrakis-nngt`). The cluster gate stays **report-only** — the T8 fail-block flip is deferred.

Bead: `arrakis-1ha3` · cycle `doctor-acvp-network-plane`. Pairs with mediums `acvp:verify`+receipt (→ `bound`) and sonar `status:aspirational` (→ aspirational).

### DIFF
--- a/packages/beacon-schema/src/acvp-bindings.ts
+++ b/packages/beacon-schema/src/acvp-bindings.ts
@@ -79,6 +79,23 @@ export interface ValidateAcvpBindingsInput {
   readonly proofReceipts?: ReadonlyArray<AcvpProofReceipt> | null;
   /** null = unknown → freshness unconfirmed, do NOT downgrade (FL-B0) */
   readonly buildingHeadSha?: string | null;
+  /**
+   * injected (FL-B0 restored, sprint-400 step 3): for a MATCHED receipt whose
+   * commit_sha is not exactly buildingHeadSha, decide whether the receipt still
+   * attests the code at the audited HEAD:
+   *   "fresh"   = the attested state still holds at HEAD          → ok (bound)
+   *   "stale"   = the cell advanced/changed since the receipt     → warn (aspirational)
+   *   "unknown" = undeterminable (shallow clone / missing history) → fall back to
+   *               the strict head-equality guard (never auto-pass)
+   * Consulted ONLY when buildingHeadSha is non-null (an audited pin exists). A
+   * COMMITTED in-repo receipt advances HEAD past the commit_sha it recorded, so
+   * exact-HEAD match alone can never accept one — this resolver is what lets a
+   * real receipt resolve to `bound`. Absent the resolver the validator keeps the
+   * exact-HEAD behaviour, so all prior call-sites and tests are unchanged. HOW
+   * freshness is determined is the caller's concern (the CLI compares the cell
+   * tree at the receipt commit vs HEAD); the core only consumes the verdict.
+   */
+  readonly checkReceiptFreshness?: (receipt: AcvpProofReceipt) => "fresh" | "stale" | "unknown";
   readonly aspirationalAllowlist: ReadonlyArray<AcvpAllowlistEntry>;
   /** injected for deterministic expiry checks (FL-D0); defaults to new Date() */
   readonly now?: Date;
@@ -140,6 +157,28 @@ function rawProofFinding(inv: AcvpInvariantT, input: ValidateAcvpBindingsInput):
         severity: "ok",
         message: `proof receipt fresh (commit ${short(receipt.commit_sha)} == building HEAD)`,
       };
+    }
+    // FL-B0 (restored): consult the injected freshness resolver — but ONLY with
+    // an audited HEAD pin (FAGAN composer: a "fresh" verdict under a null head
+    // would yield `bound` with no commit binding, re-opening the null-head hole).
+    // The resolver's contract lives on ValidateAcvpBindingsInput.checkReceiptFreshness;
+    // "unknown" falls through to the strict guard below — only a positive "fresh"
+    // under a real pin yields ok.
+    if (head != null && input.checkReceiptFreshness) {
+      const verdict = input.checkReceiptFreshness(receipt);
+      if (verdict === "fresh") {
+        return {
+          severity: "ok",
+          message: `proof receipt commit-bound: '${inv.proof_artifact}' unchanged since receipt commit ${short(receipt.commit_sha)} (${receipt.test_runner})`,
+        };
+      }
+      if (verdict === "stale") {
+        return {
+          severity: "warn",
+          message: `proof receipt STALE: '${inv.proof_artifact}' changed since receipt commit ${short(receipt.commit_sha)} — re-run the proof (promote) or it stays aspirational`,
+        };
+      }
+      // "unknown" → undeterminable; fall through to the FAGAN head guard below.
     }
     // FAGAN iter-2 (opus): a receipt that is NOT commit-bound to current code
     // cannot be reported `ok`/bound — it degrades to `warn` (aspirational). This
--- a/packages/beacon-schema/tests/acvp-bindings.test.ts
+++ b/packages/beacon-schema/tests/acvp-bindings.test.ts
@@ -222,6 +222,93 @@ test("proof receipt present, head == commit_sha → fresh, ok", () => {
   assert.ok(r.findings.some((f) => /fresh/.test(f.message)));
 });
 
+// ─── FL-B0 restored: checkReceiptFreshness resolver (sprint-400 step 3) ──────
+
+test("receipt + checkReceiptFreshness 'fresh' (proof unchanged since receipt commit) → ok, bound — committed receipt resolves green (FL-B0)", () => {
+  const r = validateAcvpBindings(
+    mkInput({
+      invariants: [inv({ id: "monotonicity" })],
+      fileExists: () => false,
+      // committed receipt: HEAD has advanced PAST the recorded commit_sha
+      proofReceipts: [receipt({ invariant_id: "monotonicity", commit_sha: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" })],
+      buildingHeadSha: "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+      checkReceiptFreshness: () => "fresh", // CLI git-diff: proof_artifact unchanged since receipt commit
+    }),
+  );
+  assert.equal(r.contract_status, "bound");
+  assert.equal(r.summary.error, 0);
+  assert.equal(r.summary.warn, 0);
+  assert.ok(r.findings.some((f) => f.severity === "ok" && /commit-bound/.test(f.message)));
+});
+
+test("receipt + checkReceiptFreshness 'stale' (proof changed since receipt commit) → warn, aspirational", () => {
+  const r = validateAcvpBindings(
+    mkInput({
+      invariants: [inv({ id: "monotonicity" })],
+      fileExists: () => false,
+      proofReceipts: [receipt({ invariant_id: "monotonicity", commit_sha: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" })],
+      buildingHeadSha: "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+      checkReceiptFreshness: () => "stale",
+    }),
+  );
+  assert.equal(r.contract_status, "aspirational");
+  assert.equal(r.summary.error, 0);
+  assert.ok(r.findings.some((f) => f.severity === "warn" && /STALE/.test(f.message)));
+});
+
+test("checkReceiptFreshness 'fresh' is IGNORED when buildingHeadSha is null → warn, NOT bound (no pin, no auto-pass — FAGAN composer)", () => {
+  let consulted = false;
+  const r = validateAcvpBindings(
+    mkInput({
+      invariants: [inv({ id: "monotonicity" })],
+      fileExists: () => false,
+      proofReceipts: [receipt({ invariant_id: "monotonicity" })],
+      buildingHeadSha: null, // no audited pin → resolver must not be honored
+      checkReceiptFreshness: () => {
+        consulted = true;
+        return "fresh";
+      },
+    }),
+  );
+  assert.equal(r.contract_status, "aspirational"); // NOT bound
+  assert.equal(consulted, false); // resolver never consulted without a pin
+  assert.ok(r.findings.some((f) => f.severity === "warn" && /UNCONFIRMABLE|not commit-bound/.test(f.message)));
+});
+
+test("receipt + checkReceiptFreshness 'unknown' → falls back to FAGAN head guard (warn, NOT auto-pass)", () => {
+  const r = validateAcvpBindings(
+    mkInput({
+      invariants: [inv({ id: "monotonicity" })],
+      fileExists: () => false,
+      proofReceipts: [receipt({ invariant_id: "monotonicity" })],
+      buildingHeadSha: null, // unknown → FAGAN guard
+      checkReceiptFreshness: () => "unknown",
+    }),
+  );
+  assert.equal(r.contract_status, "aspirational");
+  assert.equal(r.summary.error, 0);
+  assert.ok(r.findings.some((f) => f.severity === "warn" && /UNCONFIRMABLE|not commit-bound/.test(f.message)));
+});
+
+test("exact-HEAD match short-circuits BEFORE the freshness resolver (resolver not consulted)", () => {
+  const sha = "cafebabecafebabecafebabecafebabecafebabe";
+  let consulted = false;
+  const r = validateAcvpBindings(
+    mkInput({
+      invariants: [inv({ id: "monotonicity" })],
+      fileExists: () => false,
+      proofReceipts: [receipt({ invariant_id: "monotonicity", commit_sha: sha })],
+      buildingHeadSha: sha,
+      checkReceiptFreshness: () => {
+        consulted = true;
+        return "stale";
+      },
+    }),
+  );
+  assert.equal(r.contract_status, "bound"); // exact match wins
+  assert.equal(consulted, false); // resolver never called — fast path
+});
+
 test("hash_chain + eventsPin present but resolver → null → runtime warn (runtime_pin_unresolved), aspirational (FL-HC1)", () => {
   const r = validateAcvpBindings(
     mkInput({
--- a/packages/beacon-schema/tests/acvp-bindings.test.ts
+++ b/packages/beacon-schema/tests/acvp-bindings.test.ts
@@ -275,6 +275,30 @@ test("checkReceiptFreshness 'fresh' is IGNORED when buildingHeadSha is null → 
   assert.ok(r.findings.some((f) => f.severity === "warn" && /UNCONFIRMABLE|not commit-bound/.test(f.message)));
 });
 
+test("receipt + checkReceiptFreshness 'unknown' WITH a non-null head → resolver consulted, falls through to FAGAN guard (warn) [BR-1]", () => {
+  // BR-1 (bridgebuilder): the null-head test below exercises the head==null
+  // guard, NOT the unknown-verdict fall-through. This covers the real path —
+  // non-null head (≠ receipt commit) + resolver returns "unknown" → the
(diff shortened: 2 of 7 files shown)


zkSoju (commented):
## Summary

This PR lands a genuine security milestone: per-cell git isolation with committed-only evidence reads closes an entire class of working-tree manipulation attacks, and the receipt-freshness bridge correctly chains proof generation to current HEAD. The defense-in-depth composition here — `GIT_LITERAL_PATHSPECS`, fsmonitor suppression, 40-hex SHA validation, trust-boundary shape guards, and `fileExists: () => false` — reads like a threat model that was actually walked end-to-end. Two test coverage gaps need closing before this becomes the cluster's primary audit signal: one exercises the wrong code path entirely (BR-1), and one leaves the empty-repo branch fully dark (BR-2).

---

## Architectural Meditation

What's happening here is a micro-instance of a pattern that appears at every major scale point in distributed systems auditing: **evidence provenance isolation**. The moment you allow any part of an audit to read from a mutable surface (working tree, shared filesystem, in-memory cache without TTL), you've introduced a TOCTOU window that sophisticated adversaries — or just concurrent CI jobs — can exploit. This PR correctly identifies that the cell clone is an untrusted input, not a trusted peer, and treats it accordingly.

The `cellGit` helper is doing what Google's Borg scheduler does with task sandboxes: constructing a hermetic execution environment from a known-good snapshot, then reading exactly the files you intended to read, at exactly the revision you committed to auditing. The combination of `git show <head>:<path>` with SHA validation means the audit result is **content-addressed** — it describes a specific commit object, not "whatever was there when we looked."

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "BR-1",
      "title": "Test for 'unknown' verdict incorrectly exercises null-head path, not unknown-verdict path",
      "severity": "MEDIUM",
      "category": "test-coverage",
      "file": "packages/beacon-schema/tests/acvp-bindings.test.ts",
      "description": "The test titled 'checkReceiptFreshness unknown → falls back to FAGAN head guard' passes `buildingHeadSha: null`. Because the implementation guards with `if (head != null && input.checkReceiptFreshness)`, the resolver is never consulted when head is null — the test exercises the pre-existing null-head code path, not the new 'unknown' verdict fall-through. The real path (non-null head + resolver returns 'unknown' → fall through to FAGAN warn) is not covered by any unit test.",
      "suggestion": "Add a test with `buildingHeadSha: 'bbbb...bbbb'` (non-null, different from receipt commit_sha) and `checkReceiptFreshness: () => 'unknown'`. Assert `contract_status === 'aspirational'`, `summary.error === 0`, and a warn finding matching /UNCONFIRMABLE|not commit-bound/. Also verify the resolver WAS consulted (a `consulted` flag pattern, as used in other tests in this file).",
      "confidence": 0.95,
      "faang_parallel": "This exact failure mode — a test that passes for the wrong reason — is what Google's 'beyondcorp' team called a 'shadow test': it achieves green CI without touching the feature branch. Their solution was property-based testing with explicit resolver mock verification, ensuring the test fails when the resolver is never called. The Apollo team at Meta hit a related problem in their federation gateway tests: tests that exercised the fallback path were green even when the primary path was broken, because the test inputs happened to trigger the fallback immediately.",
      "metaphor": "Imagine testing a traffic light's 'yellow' phase by cutting the power — the light does go dark (a kind of safe state), but you never verified the yellow bulb actually works. The test passes, the report looks clean, and then the bulb fails silently in production.",
      "teachable_moment": "When a function has multiple early-exit guards, each guard needs its own test fixture that forces exactly that guard to be the one firing. A useful heuristic: if you can remove the feature branch from the implementation and all existing tests still pass, those tests are not testing the feature.",
      "connection": "Connects to BR-5's PRAISE: the committed-only evidence read is a similarly 'layered guard' design, and the same testing discipline applies — each guard layer should have a test that confirms it activates when expected."
    },
    {
      "id": "BR-2",
      "title": "No integration test for cell clone with no committed HEAD",
      "severity": "LOW",
      "category": "test-coverage",
      "file": "packages/freeside-cli/tests/doctor.test.ts",
      "description": "The code emits `beacon_invalid` when `resolveCellHead` returns null (i.e., `git rev-parse HEAD` fails on an empty or corrupt repo). This branch is reachable — `git init` without a first commit produces exactly this state. No test exercises it, so the error message path and the `continue` that follows are untested.",
      "suggestion": "Add a test variant in `initCellRepo` (or a standalone test) that calls `git init` but does not commit, then asserts that doctor produces a `beacon_invalid` finding for that slug with no substituted fixture data.",
      "confidence": 0.9,
      "faang_parallel": "AWS's chaos engineering practice specifically targets 'zero-state' scenarios — what happens when a resource exists but has no data. Their S3 team discovered that newly-created buckets with no objects triggered a code path in their inventory service that was tested only with buckets that had at least one object. The fix required adding explicit 'empty bucket' fixtures to every integration test suite that touched bucket state.",
      "metaphor": "Testing a filing system by always putting at least one document in before running tests. The drawer exists, the labels work, the lock mechanism is verified — but nobody tested opening an empty drawer, which turned out to have a different physical stop position that caused the handle to jam.",
      "teachable_moment": "Empty-state and zero-state paths are structurally different from 'small data' paths. A repo with one commit and a repo with zero commits differ not just in quantity but in which git commands succeed at all. When your code wraps a subprocess that can fail in a unique way for empty inputs, that failure mode deserves its own test."
    },
    {
      "id": "BR-3",
      "title": "softRead catches programmer errors (TypeError, RangeError) alongside expected parse failures",
      "severity": "LOW",
      "category": "robustness",
      "file": "packages/freeside-cli/src/verbs/doctor.ts",
      "description": "`softRead` is a bare `catch {}` with a fallback return. This is intentional for audit-loop robustness, but it equally swallows TypeErrors and similar bugs in the parsing code. A typo in the eventsPin extraction path (e.g., wrong field name) would silently return null rather than surfacing as a test failure or observable error.",
      "suggestion": "If the design intent is to keep the catch-all, document that this function is exclusively for I/O and parse operations on untrusted bytes, not for general logic. Alternatively, rethrow errors that are not `SyntaxError` / `TypeError` from JSON.parse paths, keeping the broad catch only for the specific operations that need it.",
      "confidence": 0.7,
      "faang_parallel": "The Go standard library's design philosophy — explicit error returns instead of exceptions — was motivated precisely by this problem: bare catch blocks in Java codebases at Google were swallowing NPEs and ArrayIndexOutOfBoundsExceptions that indicated real bugs, making them indistinguishable from expected I/O failures. The Kubernetes codebase documents this as a 'swallowed error' lint rule: any catch that doesn't at minimum log the error type is flagged for review.",
      "metaphor": "A security checkpoint that waves through anyone who doesn't present a badge — whether they're a visitor without credentials (expected) or an employee who lost their badge and is confused (a real problem you'd want to know about). Both get the same treatment: a friendly 'null return' and no alarm.",
      "teachable_moment": "Bare catch blocks are appropriate at process boundaries where the input is genuinely untrusted and any parse failure is acceptable. They are dangerous inside application logic where errors might indicate programmer mistakes. The discipline is to mark the boundary explicitly — ideally with a comment naming the specific exception classes that are expected — so future readers know the broad catch was intentional rather than lazy."
    },
    {
      "id": "BR-4",
      "title": "Freshness check treats any file change as stale, not just proof-artifact changes",
      "severity": "LOW",
      "category": "design",
      "file": "packages/freeside-cli/src/verbs/doctor.ts",
      "description": "The `checkReceiptFreshness` implementation returns 'stale' if ANY file changed between the receipt commit and HEAD — excluding only the receipt file itself. In active cells with frequent unrelated commits (documentation, tooling, dependency bumps), a receipt that attests valid proof code will perpetually report 'stale' even when the proof artifact itself is unchanged. This degrades the signal and may create persistent warn noise.",
      "suggestion": "This is a documented conservative choice and is safer than checking only the proof artifact. If alert fatigue becomes a real problem, a narrower diff filtered to the proof_artifact path (plus the receipt file) would be a precision improvement. Noting the trade-off explicitly in the SDD or a follow-up issue is sufficient for now.",
      "confidence": 0.75,
      "faang_parallel": "Git's `--diff-filter` and GitHub's CODEOWNERS file solve exactly this precision problem for code review: you can specify that a receipt-like file only matters when certain paths change. The Linux kernel's `make allyesconfig` staleness check has a similar conservative design — any Kconfig change triggers a full rebuild — which Linus has explicitly defended as 'correct but noisy' because the alternative (tracking which configs affect which files) is a maintenance burden that creates false negatives.",
      "metaphor": "A smoke detector that also triggers when someone makes toast. The toast alarm is technically a false positive — it means there IS smoke, just harmless smoke. The conservative design keeps you safe from fires, but if it goes off every morning, people start ignoring it. The art is calibrating the sensor to the threat model.",
      "teachable_moment": "Conservative staleness signals are the right default when you're building foundational trust infrastructure — the cost of a false negative (missing a real staleness) far exceeds the cost of a false positive (spurious warn). The time to add precision filtering is after you've observed the false-positive rate in production and confirmed it creates actual alert fatigue, not preemptively."
    },
    {
      "id": "BR-5",
      "title": "PRAISE: Committed-only evidence via git show prevents working-tree manipulation",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/freeside-cli/src/verbs/doctor.ts",
      "description": "All cell evidence (beacon, receipt, eventsPin, sealed schemas, freshness diff) is read via `git show <head>:<path>` at the committed HEAD, never from the working tree. Combined with `resolveCellHead`'s 40-hex validation, this means a cell operator cannot influence audit verdicts by staging or writing uncommitted changes — the working-tree-beacon-mutation test confirms this invariant holds end-to-end.",
      "confidence": 1.0,
      "faang_parallel": "This is the same principle behind Sigstore's transparency log and GitHub's artifact attestation: the audit reads from an immutable, content-addressed object store, not from a mutable working surface. The git object store IS a content-addressed database — `git show` is semantically equivalent to a key-value lookup by hash, and the SHA prefix check is the content-addressing guarantee. Google's Binary Authorization service uses the same pattern: policy decisions are made against container image digests, never against tags or URLs that could be mutated between the policy check and the deployment.",
      "metaphor": "Instead of asking someone 'what do you have in your bag right now?' (which they can prepare for), you're asking 'what did the security camera show you carrying when you entered?' The immutable record is the audit surface, not the present state.",
      "teachable_moment": "The distinction between 'reading from a mutable working surface' and 'reading from a content-addressed commit' is the foundational security property of git-based audit systems. Any audit tool that reads from the filesystem rather than from git objects is vulnerable to a class of attacks where the operator prepares a clean filesystem state for the audit and then reverts it afterward."
    },
    {
      "id": "BR-6",
      "title": "PRAISE: isProofReceipt shape guard at trust boundary prevents type confusion mid-audit",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/freeside-cli/src/verbs/doctor.ts",
      "description": "`isProofReceipt` validates every required field as a string before any element reaches `validateAcvpBindings`. This closes the path where an adversarial receipt (e.g., `{commit_sha: null}`) could cause `receipt.commit_sha === head` to compare null against a string, or `.trim()` / template formatting to throw mid-audit. The guard is placed at the JSON parse output, exactly at the trust boundary.",
      "confidence": 1.0,
      "faang_parallel": "Stripe's API validation philosophy places schema guards at the exact moment data crosses a trust boundary — before any business logic touches it. Their internal 'type narrowing at the edge' principle (documented in their engineering blog) states that once data has passed the boundary guard, downstream code can assume the type contract without defensive checks, which both simplifies logic and makes security guarantees auditable. This is exactly what `isProofReceipt` achieves.",
      "teachable_moment": "Type narrowing at trust boundaries is the TypeScript-native expression of the older 'parse don't validate' principle. Instead of checking `if (receipt.commit_sha)` in five different places, you check once at the boundary, return a typed result, and let the type system enforce the contract downstream. The audit code reads cleaner AND is more secure."
    },
    {
      "id": "BR-7",
      "title": "PRAISE: fileExists: () => false in cellsDir mode closes the file-presence loophole",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/freeside-cli/src/verbs/doctor.ts",
      "description": "By setting `fileExists: () => false` when resolving from a cell clone, the implementation ensures a cell cannot achieve `bound` status merely by having a proof file present on disk. Only a committed, freshness-verified receipt produces `ok`. This directly addresses the 'weak file-presence signal' concern flagged during FAGAN review and matches the operator's stated intent (the α path rejection).",
      "confidence": 1.0,
      "faang_parallel": "This mirrors the capability-token pattern in security systems: instead of granting access based on what's present (a file on disk), you grant access only for explicitly-issued credentials (a committed, validated receipt). AWS IAM's 'deny by default' posture follows the same logic — absence of an explicit allow is a deny, not an implicit allow based on resource presence.",
      "teachable_moment": "Injecting `() => false` as a dependency is an elegant form of capability restriction: you're not removing the file-existence-checking code from `validateAcvpBindings`, you're giving it a capability that always reports 'no file' for the untrusted surface. This keeps the core function testable in isolation while hardening the integration path without special-casing the function's logic."
    },
    {
      "id": "BR-8",
      "title": "PRAISE: Per-cell catch-all prevents a malformed cell from aborting the full audit",
      "severity": "PRAISE",
      "category": "robustness",
      "file": "packages/freeside-cli/src/verbs/doctor.ts",
      "description": "The try-catch wrapping the entire ACVP sub-check per slug ensures an unexpected throw from one cell's inputs surfaces as that slug's `beacon_invalid` error, leaving all other cells unaffected. This prevents a pathological or adversarial cell from DoS-ing the cluster-wide audit.",
      "confidence": 1.0,
      "faang_parallel": "Kubernetes' kubelet uses a per-pod reconciliation loop with exactly this isolation property — a panic in one pod's reconciliation is caught and converted to a pod error condition, while all other pods continue reconciling. The alternative (letting exceptions propagate) would mean a single malformed pod spec could freeze the entire kubelet. Netflix's Hystrix circuit breaker implements the same isolation at the service call level.",
      "teachable_moment": "When processing an untrusted collection in a loop, the catch boundary should surround the processing of ONE element, not the entire loop. A per-element catch converts 'one bad input breaks everything' into 'one bad input produces one error record.' This is the difference between a brittle pipeline and a resilient one."
    },
    {
      "id": "BR-9",
      "title": "PRAISE: Hardened cellGit invocation — GIT_LITERAL_PATHSPECS, fsmonitor, timeout, stderr isolation",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/freeside-cli/src/verbs/doctor.ts",
      "description": "`cellGit` combines four independent defenses: `GIT_LITERAL_PATHSPECS=1` prevents glob interpretation of adversarial paths; `-c core.fsmonitor=false` ignores the clone's local fsmonitor hook; the 15s timeout prevents a pathological commit graph from hanging the single-threaded audit; and `stdio: ['ignore', 'pipe', 'ignore']` ensures stderr from the untrusted clone never leaks into audit output or confuses callers.",
      "confidence": 1.0,
      "faang_parallel": "Git's own test suite uses `GIT_LITERAL_PATHSPECS` in exactly the same way for security-sensitive tests. The CVE-2022-24765 (git runs attacker's commands when run in a directory owned by another user) and CVE-2021-21300 (git applies hooks from cloned repositories) are precisely the attack classes that `fsmonitor=false` and stderr isolation defend against. The OpenSSF Scorecard tool, which audits open-source project security, flags git invocations that don't suppress hook execution as a medium-severity finding.",
      "teachable_moment": "Every environment variable, every config, every stdio channel is a potential injection surface when running git against untrusted input. Defense-in-depth here is not paranoia — it's the correct response to a well-documented attack surface. `GIT_LITERAL_PATHSPECS` alone has prevented at least two publicly-disclosed path injection vulnerabilities in git clients."
    },
    {
      "id": "BR-10",
      "title": "PRAISE: loadBeaconFromText cleanly enables committed-only beacon auditing",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/freeside-registry/src/beacon-loader.ts",
      "description": "`loadBeaconFromText` is a minimal, non-throwing wrapper around `classifyBeacon` that accepts pre-fetched bytes rather than a file path. This makes the committed-only audit pattern composable without exposing filesystem access to the caller, and reuses the existing V3/V2 classification logic without duplication.",
      "confidence": 1.0,
      "faang_parallel": "This is the hexagonal architecture (ports and adapters) pattern applied at function scale: `classifyBeacon` is the pure domain logic (the hexagon interior), and `loadBeaconFromText` is an adapter that connects it to the 'pre-fetched bytes' port. The filesystem-reading version is a different adapter. Facebook's internal 'data fetching as a separate concern' principle (which became Relay's design) draws the same boundary: fetch separately, process purely, test each in isolation.",
      "teachable_moment": "When you find yourself needing two versions of a function (one that reads from disk, one that works on bytes you already have), the right abstraction is usually to expose the pure-bytes version and make the disk-reading version a thin wrapper. The inverse — hardcoding the file read into the core logic — makes testing harder, makes the committed-only audit pattern impossible, and ties the function's correctness to filesystem state."
    }
  ]
}
```


---

## Callouts

The security architecture in this PR deserves to be read as a case study. Five independent defenses compose without redundancy:

1. **SHA validation** ensures the HEAD reference is a real commit, not an injected string.
2. **`git show <sha>:<path>`** reads from the immutable object store, not the mutable tree.
3. **`GIT_LITERAL_PATHSPECS` + `fsmonitor=false`** closes hook and glob injection surfaces.
4. **`isProofReceipt` shape guard** prevents type confusion at the trust boundary.
5. **`fileExists: () => false`** removes the file-presence loophole entirely.

Each layer would independently provide some protection. Together, they create a system where the attack surface narrows at every step — the adversary would need to simultaneously compromise the git object store, bypass the SHA validator, forge a receipt, AND pass the shape guard. That's not defense-in-depth as a slogan; it's defense-in-depth as engineering.

The `loadBeaconFromText` addition is architecturally clean in a way that's easy to undervalue: it's a six-line function that makes the entire audit pipeline testable without a filesystem. That kind of seam — the small function that enables a large testability improvement — is the hallmark of a developer who thinks about future readers.

---

zkSoju:
Worked through the review — sharp pass, thanks (esp. the BR-5–10 security validation).

**Addressed in `bdf0e2e0`:**
- **BR-1** (MEDIUM) ✅ — added the *real* unknown-verdict test: non-null head + resolver→`"unknown"` asserts the resolver **is consulted** and falls through to the not-commit-bound warn. The prior test only hit the `head==null` guard — exactly as flagged (shadow test).
- **BR-2** (LOW) ✅ — added a no-committed-HEAD cell test (`git init`, no commit → `beacon_invalid`, no fixture substitution).
- **BR-3** (LOW) ✅ — scoped `softRead`'s doc comment to untrusted I/O+parse only (it deliberately swallows programmer errors too — never wrap general logic).
- **BR-4** (LOW) — acknowledged as the **deliberate** conservative choice: whole-tree-modulo-receipt is what closes the iter-3 "code changed, test file byte-identical → false `bound`" hole; a proof-artifact-only diff would re-open it. Precision-filter tracked as `arrakis-bjao` for when we observe real warn-noise.

**BR-5–10 PRAISE** noted — the committed-only + capability-restriction architecture is the FAGAN-hardened core (4 multi-model iterations). One known residual the cluster gate stays **report-only** for: unsigned receipts are forgeable by a *committed* act → Ed25519 signing tracked as `arrakis-nngt`.

PR #261, opened 2026-06-01
feat(events): cycle-112 schema-emission coherence floor (registry + emit() + lint/shrink-gate + pilot + events-api beacon)

zkSoju:
## cycle-112 — schema-emission coherence floor

Makes NATS **schema-emission drift physically futile + immediately detectable** (soundness), and makes the compliant path **cheaper than evasion** (adoption). The win: every event emitted to NATS honors the schema its cell declares, the raw `.publish` capability is unreachable outside the events package, and a new raw emit **fails CI**.

> Built via `/simstim` (6 review gates: PRD + Flatline, SDD + 100%-agreement Flatline with all 17 findings integrated, sprint + Flatline with 10 more hardenings) then implemented S1→S5. Full provenance + design: `grimoires/loa/cycles/cycle-112-schema-emission-floor/` (prd.md, sdd.md, sprint-plan.md, 3 flatline reviews, grounding-notes.md, migration-runbook.md).

### What landed (per sprint)
- **S1 keystone** — phantom `SchemaId<P>` (compile-time payload binding) · **closure-held `NatsTransport`** so `internalPublish` is genuinely unreachable (capability, not convention) · `emit()` validate-before-sign + typed-failure `Either` · per-cell mutex (stalled-publish releases) · `MissingSignerError` guard.
- **S2 detection** — `events-lint` bin (discriminates NATS by receiver name → **11/11 sites, 0 Redis/Rabbit/notifier false-positives**) · **swap-proof** set-subset shrink-gate · allowlist · CI.
- **S3 tools** — `events:new-schema` generator (proven: scaffolds a compiling, registry-wired skeleton) · migration runbook + kill-switch decision.
- **S4 pilot** — `parallel.mode.enabled` proven end-to-end: registry → `emit()` → `acvp-l1-v2` envelope → **verifying consumer that resolves the schema FROM THE REGISTRY** (first real exercise of receiver-side soundness) · scoped recovery (transport retry+dead-letter; validation drops deterministically).
- **S5 teeth** — the **first in-repo BeaconV3** (`events-api`, closes arrakis-vl8f: the protocol-definer finally declares an invariant for the protocol it defines) · proof test · **gate flipped to fail-block** · aspirational dates restaggered.

### Verification
- `tsc -b` clean · `tsc -p tsconfig.test.json` clean · **93/93 events tests pass** (incl. `tests/acvp/`) · `events-lint --enforce` + shrink-gate both PASS · zero regressions.

### Honest deferrals (tracked, not faked green)
- **T4.2** production wiring of `parallel-mode-orchestrator.ts` onto `emit()` — crosses into a vitest package + needs a new workspace dep; pilot proven via the integration harness (the SDD's mock-only-seam design). `arrakis-we22`.
- **Bulk migration** of the other 10 raw-NATS sites + ts-morph codemod — **PRD NG-4 defers this** ("only the pilot migrates this cycle"). `arrakis-yc49`.
- **`events-api` `contract_status: bound`** needs the `acvp:verify` commit-bound receipt; honest-aspirational (dated) interim. `arrakis-5ryf`.
- Two S1 minor partials: `arrakis-j2bw` (cross-pkg version assertion), `arrakis-l8yk` (by-name import-rejection test).

### Notes
- `events-api` slug is an `-api`-naming-law concession for a library (mcp omitted, no served URL).
- Branched clean off `main` — cycle-112's deps (runtime_class, allowlist, acvp-bindings) are already merged; this does **not** depend on the unmerged doctor-acvp work.


zkSoju (commented):
I'll produce the Bridgebuilder review. This is a read-only generation task — the deliverable is the review text itself, so it's compatible with plan-mode constraints (no edits, no commits). I'll enrich the 12 findings without adding, removing, or reclassifying any of them.

## Summary

Cycle-112 lands a genuinely strong schema-emission floor: a closure-held capability transport (F11) and a set-based, swap-proof shrink-gate (F12) are top-tier work. The primary concern is that two of the floor's load-bearing guarantees may not actually fire — the shrink-gate's base-ref likely resolves to nothing in PR CI (F1), and the mutex frees its lock while a timed-out publish can still fork the hash-chain (F2). Verdict: **REQUEST_CHANGES** — the design is sound but its enforcement edges need to be proven red-on-violation before the `--enforce` flip is trustworthy.

The throughline of this PR is *declared-vs-actual*: a gate, a lock, a beacon status, and a generator each claim a property that the runtime may not honor. That gap is the cycle's own thesis (schema-emission coherence) turned back on its tooling. The fix posture is uniform — **prove the negative case** (gate goes red, lock prevents fork, generator re-runs clean) rather than trusting the green.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Shrink-gate likely a silent no-op in PR CI — `git show main:…` cannot resolve the base ref",
      "severity": "HIGH",
      "category": "ci-correctness",
      "file": "tools/check-nats-allowlist-shrinks.sh",
      "description": "The gate compares against the base via `git show \"${BASE}:${ALLOWLIST}\"` with BASE defaulting to the branch name `main` (workflow sets `BASE: ${{ github.event.pull_request.base.ref || 'main' }}`). actions/checkout@v4 (even with fetch-depth: 0) checks out the PR ref into detached HEAD and does NOT create a local `refs/heads/main`; bare `main` does not resolve to `origin/main`. `git show main:…` therefore returns non-zero, hitting the `if ! base_blob=…` branch which prints '…not present on main — introduction commit, baseline established' and exits 0. Net effect: the monotonic-shrink gate — described as 'the CRITICAL property that makes G-1 true' and now flipped to --enforce — never actually compares against base and passes unconditionally in normal PR CI.",
      "suggestion": "Fetch/resolve the base explicitly before diffing, e.g. `git fetch origin \"$BASE\" && base_blob=$(git show \"origin/$BASE:$ALLOWLIST\")`, or pass BASE as `origin/main`. Add a CI smoke test that asserts the gate FAILS on a deliberate allowlist addition to prove it is wired, not green-by-default.",
      "confidence": 0.6,
      "faang_parallel": "Mike Bland's account of Google's 'Beyoncé Rule' — if you liked it, you shoulda put a test on it. The deeper lesson there was that gates which can never fail erode faster than no gate, because the team stops watching them.",
      "metaphor": "A smoke detector with the battery tab still in: the green LED glows, everyone feels safe, and it never chirps — least of all when the house is on fire.",
      "teachable_moment": "A gate is not 'passing' until you've watched it fail on a known violation. Green has two causes — 'nothing wrong' and 'never ran' — and only the red path distinguishes them.",
      "connection": "Detached-HEAD checkout semantics: `main` and `origin/main` are different refs in CI; the local branch ref simply doesn't exist after actions/checkout."
    },
    {
      "id": "F2",
      "title": "Mutex timeout releases the lock while the detached publish still mutates the chain — reintroduces the SKP-001 fork it was built to prevent",
      "severity": "HIGH",
      "category": "concurrency",
      "file": "packages/events/src/mutex.ts",
      "description": "withLock races fn() against timeoutMs and, on timeout, rejects with TimeoutError and calls release() in `finally` — but the detached `publishEnvelope` keeps running (JS cannot cancel it). If the held op stalls past the timeout (jetstream publish under backpressure/slow-consumer) but later completes, a subsequent emit acquires the now-free lock, reads the OLD prev_hash (the stalled op has not reached store.set yet), signs, publishes, and sets — then the stalled op also publishes/sets with the same prev_hash. Two envelopes share a prev_hash → hash-chain fork, exactly the SKP-001 guarantee the mutex claims to provide. emit.ts's retry path (doPublish) widens the window by issuing a fresh withLock after a TimeoutError Left. The mutex test only asserts the queue drains (secondRan), never chain integrity after a timeout.",
      "suggestion": "Don't advance the chain on a timed-out op: either guard prevHashStore.set with a per-op fence/generation token so a timed-out op cannot commit, or treat a TimeoutError as fatal-for-this-chain (do not retry / do not allow concurrent emits to proceed until the in-flight op resolves). Add a test that times out op A mid-publish, runs op B, then lets A resolve, and asserts no two envelopes share prev_hash.",
      "confidence": 0.55,
      "faang_parallel": "Kleppmann's critique of Redis Redlock: a lease that expires while its holder is still working lets two clients both believe they own the lock. His fix — monotonic fencing tokens rejected at the storage layer — is precisely the generation-guard suggested here.",
      "metaphor": "A teller window that auto-closes after 60 seconds. The slow customer is still filling in the ledger when the next customer steps up to the same open page — now two pens are writing line 47.",
      "teachable_moment": "A timeout frees the lock, not the work behind it. In a language without cancellation, the only safe contract is a fence the storage layer enforces, so a stale writer's commit is rejected even after it 'returns'.",
      "connection": "This is the SKP-001 fork class re-entering through the timeout door; the integrity test must assert prev_hash uniqueness, not queue progress."
    },
    {
      "id": "F3",
      "title": "new-schema generator breaks on its second invocation — hard-coded topics import replace no longer matches",
      "severity": "MEDIUM",
      "category": "tooling",
      "file": "packages/events/scripts/new-schema.mjs",
      "description": "The generator wires the new topic helper into registry.ts via `reg.replace('import { nftMintDetectedTopic } from \"./topics.js\";', 'import { nftMintDetectedTopic, ' + TopicFn + ' } from \"./topics.js\";')`. After the pilot was generated, registry.ts now reads `import { nftMintDetectedTopic, parallelModeEnabledTopic } from \"./topics.js\";`, so the literal search string no longer occurs. String.replace silently no-ops, the new TopicFn is never imported, and the generated registry.ts references an undefined identifier in buildSubject → compile failure. The generator (the cycle's 'adoption carrot', meant for repeated one-command use) thus produces broken output on every invocation after the first.",
      "suggestion": "Use a stable insertion marker for the topics import (as is already done for the registry ENTRIES block) instead of matching the literal single-import line, or insert the import by appending to the existing `from \"./topics.js\"` import via a regex that tolerates additional named imports. Add a test that runs the generator twice and asserts the package still compiles.",
      "confidence": 0.6,
      "faang_parallel": "Facebook's jscodeshift moved codemods from regex to AST transforms for exactly this reason — literal string matching is fragile against the very edits the codemod itself produces.",
      "metaphor": "A mail-merge that hard-codes the last recipient's name into the salutation: letter #2 is addressed to letter #1's reader.",
      "teachable_moment": "A generator's correctness is measured on run N, not run 1. Idempotency and re-runnability are the contract — test it by running twice and compiling.",
      "connection": "The ENTRIES block already uses a stable marker; the topics import should adopt the same pattern for consistency."
    },
    {
      "id": "F4",
      "title": "events-lint detection holes weaken the --enforce floor (file-level allowlist granularity + receiver-name heuristic misses INatsPublisher sites)",
      "severity": "MEDIUM",
      "category": "security-gate",
      "file": "packages/events/bin/events-lint.mjs",
      "description": "Two gaps undercut the claim 'a NEW raw NATS .publish now fails CI': (1) `isAllowlisted` matches at FILE granularity (relPath.endsWith(fragment)), so a brand-new raw `.publish` added to an already-allowlisted file (e.g. parallel-mode-orchestrator.ts, which has 3 ids) is marked KNOWN/not-NEW and passes, and because no allowlist id is added the shrink-gate (F1 aside) also stays quiet. (2) NATS_PUBLISH only matches `.nats.publish(` / `.jetstream.publish(` by receiver name, but the allowlist's own 6 coexistence entries are described as 'INatsPublisher' sites — if they publish via an injected `this.publisher.publish(...)`, the scanner never sees them, so a new coexistence-style raw emit is invisible to detection entirely. The lint documents this as a tripwire, but the --enforce flip is being sold as the floor.",
      "suggestion": "Tighten allowlist matching to site granularity (subject+file+disambiguator, matching the id scheme already in the YAML) rather than whole-file, and broaden/normalize the publisher-receiver detection (or assert in CI that every allowlisted id corresponds to a scanner-detectable site). Keep framing the lint as defense-in-depth behind the transport closure, not the primary fence.",
      "confidence": 0.6,
      "faang_parallel": "The grep-vs-AST tradeoff in Semgrep/SonarQube: receiver-name heuristics miss aliased and dependency-injected call sites, which is why mature analyzers resolve types rather than match identifiers.",
      "metaphor": "A metal detector told to ignore whole rooms because one approved knife once sat there — the next knife walks in free.",
      "teachable_moment": "A suppression should be exactly as narrow as the thing it suppresses. File-granular allowlists silently bless every future sin committed in the same file.",
      "connection": "The YAML already keys ids at site granularity; the matcher should honor that scheme rather than collapsing to file paths."
    },
    {
      "id": "F5",
      "title": "emit() can throw instead of returning Either.Left when buildSubject rejects a bad specifier",
      "severity": "MEDIUM",
      "category": "error-handling",
      "file": "packages/events/src/emit.ts",
      "description": "emit() invokes `entry.buildSubject(specifier)` outside any try/catch (`return doPublish(entry.buildSubject(specifier), payload);`). The stated contract (FR-3a, repeated throughout the file) is 'a typed Either.Left — never a throw'. If buildTopic validates the specifier against the topic grammar and throws on invalid input (a plausible behavior for the existing topic builder), emit() rejects synchronously, escaping the Either boundary the whole design promises. emitRaw is safe (it validates the family without calling buildSubject), so the gap is asymmetric and easy to miss.",
      "suggestion": "Wrap buildSubject in the typed-error path: catch a topic-construction failure and return Either.left(new SubjectFamilyError(...)) (or a dedicated InvalidSpecifierError) so a bad specifier is a value, not a throw. Add a test that passes an invalid specifier and asserts a Left.",
      "confidence": 0.4,
      "faang_parallel": "Effect/Go 'errors as values': a function that advertises a total error channel but leaks an exception through one branch breaks the caller's exhaustiveness reasoning — the bug Rust's #[must_use] Result was designed to surface.",
      "metaphor": "A returns desk promising a receipt for every item, that quietly hurls the box back at you when the barcode won't scan.",
      "teachable_moment": "Verify the assumption first: confirm whether buildTopic throws on invalid input. If it does, every public entry must funnel that throw into the Either; an asymmetric guarantee (emitRaw safe, emit not) is the easiest contract to regress."
    },
    {
      "id": "F6",
      "title": "events-api beacon declares schema_enforcement status: active while the aspirational allowlist treats it as not-yet-bound",
      "severity": "MEDIUM",
      "category": "consistency",
      "file": "packages/events/beacon.yaml",
      "description": "beacon.yaml sets `acvp_invariants[0].status: active`, but .freeside/acvp-aspirational-allowlist.yaml simultaneously adds an events-api/schema_enforcement entry (expires 2026-09-25) whose comment says it is 'aspirational until the receipt lands (then flip beacon status active + delete)' and reason 'bound pending acvp:verify commit-bound receipt'. So the allowlist's own narrative assumes the beacon is NOT yet active, while the beacon already declares active. If validateAcvpBindings requires an active invariant to carry a bound proof receipt, this is either a contradiction the allowlist is silently papering over or a premature 'active' marking.",
      "suggestion": "Pick one state: keep the invariant `status: aspirational` (or candidate) until the acvp:verify receipt lands, OR if it is genuinely active+bound remove the aspirational-allowlist row. Reconcile the allowlist comment ('then flip beacon status active') with the beacon's current value.",
      "confidence": 0.45,
      "faang_parallel": "Config drift between a feature flag's declared state and its dependency's readiness — the class of incident Google's 'launch checklist' gating was built to prevent: don't flip 'serving' before the backend it points at is actually bound.",
      "metaphor": "A 'Now Serving' sign switched on while the kitchen is still aspirational.",
      "teachable_moment": "When two files describe the same state, one of them is the source of truth — make the other derive from it, or you ship a self-contradicting artifact that passes only because nothing cross-checks them."
    },
    {
      "id": "F7",
      "title": "Transport-failure retry can emit duplicate events after a partial send",
      "severity": "LOW",
      "category": "reliability",
      "file": "packages/events/src/emit.ts",
      "description": "doPublish retries transport failures by re-running publishEnvelope, which mints a NEW envelope (new event_id, advances the chain on success) each attempt. If a publish actually reached NATS but the call then failed/timed out (partial send), the retry delivers a duplicate event under a distinct event_id with the chain advanced twice. This is inherent at-least-once behavior and the deadLetter/onAlert make exhaustion observable, but there is no idempotency key or dedup hint on the envelope to let consumers collapse the duplicate.",
      "suggestion": "Document the at-least-once / duplicate-on-retry contract on the RecoveryConfig type, and consider carrying a stable idempotency/dedup key (or jetstream Nats-Msg-Id) so consumers can deduplicate retried deliveries.",
      "confidence": 0.4,
      "faang_parallel": "Kafka's idempotent producer (PID + sequence number) and NATS JetStream's Nats-Msg-Id dedup window exist precisely to convert at-least-once transport into effectively-once delivery.",
      "metaphor": "Re-posting a letter because no acknowledgement came back — the recipient now holds two identical envelopes and must decide they're the same.",
      "teachable_moment": "At-least-once is a fine default, but it's a contract consumers must be told about. A stable dedup key turns 'might duplicate' from a surprise into a documented, collapsible behavior."
    },
    {
      "id": "F8",
      "title": "version-coupling test defers the cross-package equality assertion to beacon-schema but does not verify it exists",
      "severity": "LOW",
      "category": "test-coverage",
      "file": "packages/events/tests/version-coupling.test.ts",
      "description": "The test asserts events' SCHEMA_VERSION === 'acvp-l1-v2' and a compile-time literal check, but the actually-meaningful cross-package coupling (events.SCHEMA_VERSION === beacon-schema.ACVP_L1_SCHEMA_VERSION, IMP-008/IMP-010) is only described in a comment ('lives on the beacon-schema side … Tracked for the validator side'). Nothing in this PR demonstrates that assertion is implemented anywhere, so the cross-package invariant may be unenforced despite being claimed satisfied.",
      "suggestion": "Either add the cross-package equality test on the beacon-schema side in a linked PR (and reference it here) or open a tracked task; otherwise the 'meaningful rather than vacuous' framing in the comment is aspirational.",
      "confidence": 0.4,
      "faang_parallel": "Consumer-driven contract testing (Pact): the assertion that two services agree on a version must live where both sides meet, or each side passes alone while the contract silently diverges.",
      "metaphor": "Two clocks each confidently reporting they read 'the official time' — with nothing ever holding them side by side."
    },
    {
      "id": "F9",
      "title": "ACVP/emit 'malformed payload' tests prove field-presence, not value validation",
      "severity": "LOW",
      "category": "test-quality",
      "file": "packages/events/tests/acvp/schema_enforcement.test.ts",
      "description": "The negative tests use payloads like `{ chain_id: -1 }` / `{ chain_id: 1 }` cast through `as unknown as typeof VALID`, i.e. missing every other required field. They pass because of missing required fields, not because of the negative/invalid value. The schema_enforcement proof artifact therefore does not actually demonstrate value-range or format rejection (e.g. that a negative chain_id, malformed address, or bad ISO timestamp is caught), only that omission is caught.",
      "suggestion": "Add a structurally-complete payload with one invalid field value (negative chain_id, non-hex contract, non-Z timestamp) and assert SchemaEmitError, so the proof binds to value validation as the invariant claims.",
      "confidence": 0.4,
      "faang_parallel": "Mutation testing (PIT, Google's mutation-testing-at-scale): a test that passes for the wrong reason is invisible until you mutate the code it claims to guard — here, omission masks the value-check the artifact advertises.",
      "metaphor": "Proving a lock works by checking the door is missing — you never tested that the wrong key is refused.",
      "teachable_moment": "A proof artifact must fail for the reason it claims to test. One complete-but-invalid-value fixture per family is worth a dozen missing-field cases."
    },
    {
      "id": "F10",
      "title": "parallel.mode.enabled.v1 schema admits mode 'shadow', conflicting with the 'enabled/parallel' family name",
      "severity": "LOW",
      "category": "schema-design",
      "file": "packages/events/src/schemas/parallel-mode-enabled.ts",
      "description": "ParallelModeEnabledSchema declares `mode: S.Literal(\"parallel\", \"shadow\")` on a family named `parallel.mode.enabled.v1`. A 'shadow' mode event published on a 'parallel.mode.enabled' subject is semantically ambiguous — consumers keying off the subject would expect parallel-enablement. Either the literal is too wide for this family or the family name is too narrow for the mode set.",
      "suggestion": "Clarify intent: if shadow→parallel transitions are a single event, document it; otherwise split the subject by mode or rename the family. At minimum add a comment explaining why 'shadow' belongs on the 'enabled' family.",
      "confidence": 0.3,
      "faang_parallel": "Event-schema naming discipline (e.g. CloudEvents `type` conventions): the subject is part of the contract — consumers filter on it, so a value that contradicts the subject name is a latent routing bug.",
      "metaphor": "A door labeled 'Exit' that also opens onto a broom closet — fine if you wrote the sign, confusing for everyone who reads it."
    },
    {
      "id": "F11",
      "title": "Closure-held WeakMap transport is a genuine capability boundary, not just a convention",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/events/src/transport.ts",
      "description": "createNatsTransport stashes the raw client in a module-private WeakMap keyed by an opaque frozen token and returns only the token; internalPublish is the sole resolver and is deliberately omitted from index.ts and the package exports map (with a documented //exports-note and a test asserting it is not re-exported). This is a real unforgeable capability boundary — strictly stronger than the rejected exported-unwrap()/lint-only approach — and the lint is correctly positioned as defense-in-depth behind it.",
      "suggestion": "No change. Keep the exports-map omission test (transport.test.ts) as a regression guard against an accidental future re-export.",
      "confidence": 0.9,
      "faang_parallel": "This is textbook object-capability security — the model behind Google's Caja, Capsicum (FreeBSD), and the E language: authority travels only via unforgeable references, not ambient names. A WeakMap-keyed-by-frozen-token is the JS realization of that pattern.",
      "metaphor": "A numbered coat-check ticket instead of the cloakroom key: the ticket lets you retrieve your coat, but you can't forge it, copy the master, or hand someone access to the whole rack.",
      "teachable_moment": "A boundary enforced by the module system can't be bypassed by forgetting to run the linter. When you can choose between 'a rule that warns' and 'a structure that prevents', the structure wins — and keeping the lint behind it as defense-in-depth is the right belt-and-suspenders posture.",
      "connection": "The exports-omission test is what makes this durable — without it, a future barrel-export reflex silently re-opens the capability."
    },
    {
      "id": "F12",
      "title": "Set-based shrink-gate and the validation-vs-transport failure split are well-reasoned",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/check-nats-allowlist-shrinks.sh",
      "description": "The shrink-gate compares the UNION of ids (not counts), explicitly closing the count-preserving-swap bypass flagged in Flatline IMP-003, and correctly treats an id moving between `entries` and `emitRaw_allowlist` as no change. In emit.ts the deterministic-vs-transient distinction is clean: validation failures drop immediately (never retried/dead-lettered because retrying a deterministic failure is futile), while transport failures retry with backoff and surface via deadLetter+onAlert — observable rather than silent. The pilot tests assert all three branches.",
      "suggestion": "No change to the logic. (Note F1 separately: the gate's base-ref resolution must be fixed for this correct logic to actually run.)",
      "confidence": 0.85,
      "faang_parallel": "The retry classification mirrors Google SRE's retry-budget discipline: only retryable (transient) errors get backoff; deterministic failures fail fast because retrying a guaranteed-failure just burns budget and hides the bug. Set-over-count semantics is the same insight that makes Git's content-addressing immune to reordering.",
      "metaphor": "Taking attendance by name list, not headcount — you can't sneak a substitute in by keeping the total the same.",
      "teachable_moment": "Two of the sharpest decisions an event system makes are 'what counts as a change' and 'what is worth retrying'. Choosing set-identity for the former and determinism for the latter are both the correct, hard-won answers.",
      "connection": "The logic is right; F1 is the orthogonal risk that this correct gate may never execute against a real base."
    }
  ]
}
```


## Callouts

The capability transport (F11) and the swap-proof gate plus fail-fast/retry split (F12) are the kind of engineering that earns the word *floor*. The `//exports-note` plus a regression test guarding the exports map shows the author understood that the boundary's durability lives in the test, not the comment. The deterministic-vs-transient retry classification is a decision many seasoned teams get wrong; here it's clean and all three branches are asserted. The cluster's own thesis — *declared must equal proven* — is honored in the design; the remaining work (F1, F2, F3, F6) is making the tooling live up to the same standard the schemas now demand.

PR #263, opened 2026-06-03
fix(network/registry): de-quirk sonar read host + identity/mint notes (dogfood)

zkSoju:
## What

De-quirk the freeside registry against live-probed reality. Three drifts surfaced by the **CubQuests freeside-substrate dogfood** (2026-06-03) — the registry was describing runtimes that have moved.

## The 3 corrections (all probe-grounded)

| Building | Was | Now (verified) |
|---|---|---|
| **sonar-api** ⭐ | "probe `/graphql`" at `sonar-api-production.up.railway.app` | That host **404s** on `/v1/graphql`. The real **read plane is belt-gateway** (`belt-gateway-production.up.railway.app/v1/graphql`, live 200, no auth). Every beacon declaring `consumes: sonar` was pointing at the wrong host. |
| **identity-api** | "/v1/profile … NOT built — returns 400" | `/v1/profile` **is built** (returns precise `404 wallet_not_resolved`). Also flagged: the **beacon 404s** (no `/.well-known/beacon.json` route) → identity is federation-invisible. |
| **mint-api** | "all 404 as of 2026-05-25" | Now a **booted-but-routeless Express shell** (`Cannot GET /health`, `x-powered-by: Express`) — distinct from the prior Railway edge 404. Still correctly library-only. |

The sonar one is load-bearing: `belt-gateway` is the consumable badge/holder index, and it was **absent from the registry**.

## Provenance

Surfaced by the `freeside-dogfood-readiness` probe sweep + the cubquests badge-leaderboard shadow (`scripts/shadow/badge-leaderboard-shadow.ts`), which confirmed `BadgeHolder`/`BadgeBalance` live + populated on belt-gateway and byte-consistent with the monolith leaderboard.

## Scope

Network-domain, single file (`packages/freeside-registry/registry.yaml`), notes-only (no schema change). A follow-up worth considering: a structured `read_endpoint` field so the sonar→belt-gateway distinction isn't buried in prose.

### DIFF
--- a/packages/freeside-registry/registry.yaml
+++ b/packages/freeside-registry/registry.yaml
@@ -45,8 +45,12 @@
 #   GHOST: identity /v1/auth/service-jwt mints ES256 svc-JWTs that NOTHING can verify
 #      (/.well-known/jwks.json 404s; no ES256 verifier in any cell). Orphaned — build the
 #      JWKS (one-function swap at activities jwt-verify.ts:22-27) or DELETE it. Do NOT
-#      route service auth through it. (TODO: promote to a structured per-cell `auth:`
-#      field — needs a ModuleEntry schema bump in src/registry.ts + tests.)
+#      route service auth through it.
+# Tracked (loa-freeside .beads · dogfood-followup) — these gaps are structured beads with
+# closure conditions, not just this comment: arrakis-dogfood-{auth-secret-parity-canary,
+# es256-svcjwt-ghost, registry-auth-field (the per-cell `auth:` schema bump),
+# score-api-key-sprawl, identity-beacon-404, sonar-leaderboard-endpoint,
+# freeside-ledger-missing}.
 
 version: 1
 
@@ -154,18 +158,22 @@ modules:
   mint-api:
     git_url: https://github.com/0xHoneyJar/mint-api.git
     beacon_url: https://mint.0xhoneyjar.xyz/.well-known/beacon.json
-    deployment_url: ~
+    deployment_url: https://mint-api-production.up.railway.app
     visibility: public
     owner: 0xHoneyJar
     added: "2026-05-18"
-    runtime_state: not-built
+    runtime_state: scaffolded
     notes: |
       Identity-bound token issuance protocol (ACVP-shaped). npm package
-      shipped + MCP substrate per kickoff spec, but no HTTP runtime deployed.
-      mint-api-production.up.railway.app is now a booted-but-ROUTELESS Express
-      shell (returns Express's own `Cannot GET /health`, x-powered-by: Express —
-      a process is up but no business routes are mounted; distinct from the
-      Railway edge 404 it showed 2026-05-25). Still correctly library-only.
+      shipped + MCP substrate per kickoff spec, but no business HTTP routes yet.
+      mint-api-production.up.railway.app is a booted-but-ROUTELESS Express shell
+      (returns Express's own `Cannot GET /health`, x-powered-by: Express — a
+      process is up but no routes are mounted; distinct from the Railway edge 404
+      it showed 2026-05-25). runtime_state=scaffolded + deployment_url set (not ~)
+      so the operator-dash prober actually probes + classifies the 404 as scaffold,
+      instead of probeOne() short-circuiting to `down` on a null url (BB/Codex
+      review of #263). ⚠ Railway health-check likely measures process liveness
+      only → false-healthy; a /health 404 should read as unhealthy.
       (de-quirked 2026-06-03 via cubquests dogfood probes.)
 
   score-api:
--- a/packages/freeside-registry/registry.yaml
+++ b/packages/freeside-registry/registry.yaml
@@ -29,6 +29,24 @@
 # pointed at deployed Railway URLs, and beacon-serving routes not yet implemented in cells.
 # This is the federation-discovery gap (the awareness surface from
 # grimoires/loa/specs/enhance-freeside-api-surface.md T1/T2 follow-up).
+#
+# ─── Cluster auth model (de-quirked 2026-06-03 via the cubquests dogfood) ──────────
+# There is NO single cluster token. identity-api is the credential ROOT, but each cell
+# gates differently — 3 real mechanisms + 1 ghost. A consumer needs TWO credentials
+# (HS256 user-JWT for activities, static key for score); there is no uniform header.
+#   A. activities-api ← HS256 Bearer minted by identity-api, verified OFFLINE by
+#      recomputing HMAC with IDENTITY_API_JWT_SECRET — which MUST byte-equal identity's
+#      JWT_SECRET (parity verified green 2026-06-03 via hash compare). Secret drift →
+#      every activities request 401s `bad_signature`, silently, cluster-wide. Worth a CI
+#      canary (identity mints → activities verifies → alarm on mismatch).
+#   B. score-api ← STATIC API key (X-API-Key / Bearer: MIDI_API_KEY/HUB_API_KEY/API_KEY/
+#      ADMIN_API_KEY). Service-to-service, no identity, no expiry — a leak is permanent.
+#   C. sonar (belt-gateway) ← NONE (open GraphQL reads).
+#   GHOST: identity /v1/auth/service-jwt mints ES256 svc-JWTs that NOTHING can verify
+#      (/.well-known/jwks.json 404s; no ES256 verifier in any cell). Orphaned — build the
+#      JWKS (one-function swap at activities jwt-verify.ts:22-27) or DELETE it. Do NOT
+#      route service auth through it. (TODO: promote to a structured per-cell `auth:`
+#      field — needs a ModuleEntry schema bump in src/registry.ts + tests.)
 
 version: 1
 
--- a/packages/freeside-registry/registry.yaml
+++ b/packages/freeside-registry/registry.yaml
@@ -88,10 +88,14 @@ modules:
     notes: |
       Central identity SoR (Phase 1 BUILT + DEPLOYED 2026-05-25 to Railway).
       Custom domain identity.0xhoneyjar.xyz queued behind DNS — beacon_url
-      anticipates that. Spine empty until Phase 4 T4.4 backfill (arrakis-494b).
-      Phase 2 (compose, /v1/profile) + Phase 3 (mibera dims,
-      /v1/mibera/dimensions) endpoints scaffolded but NOT built — both return
-      400 today. /health is the canonical liveness path.
+      anticipates that. ⚠ The beacon 404s today (no /.well-known/beacon.json
+      route shipped on the canonical host) → identity is federation-invisible
+      until that lands. Spine empty until Phase 4 T4.4 backfill (arrakis-494b).
+      /v1/profile IS built — returns a precise 404 wallet_not_resolved on a
+      well-formed (world,wallet) query, NOT the 400 this note previously
+      claimed (de-quirked 2026-06-03 via dogfood probes). Phase 3 mibera dims
+      (/v1/mibera/dimensions) still scaffolded. /health is the canonical
+      liveness path.
       NOTE: a separate identity-production.up.railway.app exists serving a Login
       page (Bootstrap HTML); that is NOT this service — likely an older
       freeside-identity / Sietch surface. Canonical URL is the 317b one above.
@@ -140,7 +144,11 @@ modules:
     notes: |
       Identity-bound token issuance protocol (ACVP-shaped). npm package
       shipped + MCP substrate per kickoff spec, but no HTTP runtime deployed.
-      mint-{api,mcp,}-production.up.railway.app all 404 as of 2026-05-25.
+      mint-api-production.up.railway.app is now a booted-but-ROUTELESS Express
+      shell (returns Express's own `Cannot GET /health`, x-powered-by: Express —
+      a process is up but no business routes are mounted; distinct from the
+      Railway edge 404 it showed 2026-05-25). Still correctly library-only.
+      (de-quirked 2026-06-03 via cubquests dogfood probes.)
 
   score-api:
     # The runtime building (was 'score-mibera' in pre-2026-05-23 registry). GH repo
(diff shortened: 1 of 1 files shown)


zkSoju (commented):
Now writing the plan file with the enriched review.

---

## Summary

This PR does two things well: it corrects a concrete routing error (sonar's read plane was pointed at a 404-producing host) and adds honest operational intelligence to the registry about auth mechanisms, dead infrastructure, and unresolved gaps. The registry is increasingly functioning as the cluster's living awareness surface, which makes accuracy here load-bearing. The new notes surface several pre-existing security debts that deserve structured tracking beyond a YAML comment.

---

The registry is doing something quietly ambitious: it's becoming the cluster's institutional memory for operational reality — what's live, what's dead, what's misconfigured, what's trusted. That's exactly right. A registry that only describes the happy path is dangerous; one that also documents the shadows is genuinely useful. What this PR reveals, though, is that several of those shadows need to graduate from YAML commentary into actionable infrastructure.

The auth model block at the top of the file is the most telling artifact: four different authentication mechanisms across five cells, documented in prose, invisible to tooling. This is the registry's equivalent of a TODO buried in a comment — honest, but not enforceable. The sonar host fix is the kind of concrete correction that prevents downstream consumers from silently failing; the auth documentation is the kind of note that prevents future engineers from silently building on a broken foundation.

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "ES256 service-JWT endpoint orphaned with no verifier — active security debt",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/freeside-registry/registry.yaml",
      "description": "The GHOST entry documents that identity /v1/auth/service-jwt mints ES256 JWTs that nothing in the cluster can verify (/.well-known/jwks.json 404s, no ES256 verifier in any cell). An active mint endpoint with no verifier means tokens are being issued into a trust vacuum — any consumer that wires up verification later may accept stale or leaked tokens with no revocation path. The note says 'do NOT route service auth through it' but the endpoint remains live.",
      "suggestion": "Either implement the JWKS endpoint (activities jwt-verify.ts:22-27 noted as the swap point) and wire a verifier, or gate/remove /v1/auth/service-jwt immediately. Leaving an issuer live with no verifier is worse than neither — it accumulates issued tokens with no accountability. Track as a structured security item, not a YAML comment.",
      "confidence": 0.9,
      "faang_parallel": "Google's BeyondCorp team documented the exact hazard class in 2017: an issuer that outlives its verifier creates 'zombie credentials' — tokens that were once valid under an intended scheme, now accepted by a future verifier that had no context for what they meant when issued. The JWKS discovery model exists precisely so issuers and verifiers can evolve independently, but only if the discovery endpoint is live.",
      "metaphor": "A mint that produces currency with no bank willing to accept it sounds harmless — until someone builds a bank. At that moment, every token ever issued is instantly in circulation with no audit trail of where it went in the interim.",
      "teachable_moment": "The danger isn't the 404 on JWKS — it's the implicit promise the live endpoint makes. Any running issuer is a contract: 'I will produce tokens that are eventually verifiable.' Removing the verifier while leaving the issuer is breaking that contract silently. Security debt of this class should always carry a deadline, not a comment.",
      "connection": "Connects to F-004 — both are cases where a security property exists only in prose, not in enforced infrastructure. The auth: field promotion and the JWKS endpoint are the same class of problem: trust that's asserted but not verified by the system itself."
    },
    {
      "id": "F-002",
      "title": "HS256 JWT secret parity is a single point of silent cluster-wide 401 failure with no automated detection",
      "severity": "HIGH",
      "category": "reliability",
      "file": "packages/freeside-registry/registry.yaml",
      "description": "The note documents that activities-api verifies JWTs by recomputing HMAC with its own IDENTITY_API_JWT_SECRET, which must byte-equal identity-api's JWT_SECRET. Secret drift produces silent cluster-wide 401s. The note proposes a CI canary but none exists yet. This is a runtime correctness invariant with no enforcement.",
      "suggestion": "Implement the CI canary described: identity mints a token, activities verifies it, alarm on mismatch. Until then, secret rotation procedures must explicitly coordinate both services atomically — document this in the runbook. The registry note is useful but insufficient as the only protection.",
      "confidence": 0.95,
      "faang_parallel": "Netflix's Chaos Engineering team categorized this as a 'temporal coupling' failure: two services that must share state (a secret) but have no mechanism to confirm they share the same value. Netflix's solution was to treat shared secrets as a first-class infrastructure concern with automated verification — not a manual coordination step. The CI canary described in the note is exactly this pattern; it just hasn't been built yet.",
      "metaphor": "Two combination locks on a safe, both needing the same combination — set independently. When someone changes one, the safe becomes permanently locked until you discover both need updating. The disaster doesn't announce itself; the safe just stops opening.",
      "teachable_moment": "A shared secret without a synchronization contract isn't a secret — it's a bet. Every deployment that touches one side of the pair without verifying the other side is implicitly gambling on coordination. The CI canary converts this bet into a deterministic check. Write it before the next secret rotation, not after.",
      "connection": "This is the HS256 counterpart to F-001's ES256 gap. Both are trust mechanisms that are partially implemented: F-001 has an issuer without a verifier; F-002 has a verifier without a synchronization guarantee."
    },
    {
      "id": "F-003",
      "title": "Static API key for score-api has no expiry or rotation mechanism — permanent leak surface",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/freeside-registry/registry.yaml",
      "description": "The note documents that score-api accepts multiple static key env vars (MIDI_API_KEY, HUB_API_KEY, API_KEY, ADMIN_API_KEY) with no expiry. The comment explicitly flags 'a leak is permanent.' Multiple accepted key names suggests organic accumulation across callers with no central rotation policy.",
      "suggestion": "Audit which of the four key vars are actually checked and by which callers. Consolidate to one canonical service-to-service key name. Establish a rotation runbook. Until short-lived credentials are wired, at minimum document the rotation procedure and verify key material is not committed in any repo.",
      "confidence": 0.85,
      "faang_parallel": "Amazon's internal post-mortem on the Capital One breach (S3 SSRF → IAM role → exfil) noted that the blast radius was determined entirely by credential scope and rotation cadence. AWS subsequently built IAM Roles Anywhere and mandated short-lived credential patterns for all new internal service integrations. Four static key names in a registry comment is the pre-IAM-Anywhere world — it works until it doesn't, and when it doesn't, the exposure window is unbounded.",
      "metaphor": "A master key that opens every door in a building, copied four times, given to four different people, with no record of who has which copy. The building is secure until one key is lost — then every door is compromised until you re-key everything simultaneously.",
      "teachable_moment": "Multiple accepted key names are a symptom of credentials growing organically across callers without a central policy. Each new caller added their own env var name rather than conforming to an existing standard. The consolidation audit isn't just cleanup — it's revealing the actual trust graph of who calls score-api and why.",
      "connection": "The four-key accumulation pattern is the operational surface of F-004: without a machine-readable auth: field in the registry, each caller discovered the auth mechanism independently and named their credential independently. Structured auth metadata would have enforced a canonical name from the start."
    },
    {
      "id": "F-004",
      "title": "Auth model documented only in a YAML comment — no structured schema field",
      "severity": "MEDIUM",
      "file": "packages/freeside-registry/registry.yaml",
      "description": "The per-cell auth mechanisms (HS256 Bearer, static key, open, orphaned ES256) are documented in a block comment at the top of the file, not in per-module structured fields. The TODO in the comment acknowledges this: 'promote to a structured per-cell auth: field — needs a ModuleEntry schema bump in src/registry.ts + tests.' Until that lands, tooling (freeside-cli doctor, coherence checks) cannot reason about auth requirements programmatically.",
      "suggestion": "Prioritize the ModuleEntry schema bump for the `auth:` field. A comment that says 'this needs to be a field' should become a tracked beads task, not ambient YAML prose. The current state means any consumer reading the registry programmatically has no machine-readable auth contract.",
      "confidence": 0.9,
      "faang_parallel": "Stripe's API platform team made machine-readable OpenAPI auth declarations mandatory for all internal services in 2019, after discovering that two services had incompatible interpretations of Bearer token scopes — a mismatch invisible to all tooling but visible to every 401 in prod. The lesson: if security properties aren't in the schema, they don't exist for the infrastructure.",
      "metaphor": "A fire escape plan drawn on a napkin in the break room versus one registered with the building inspector and included in the floor plan. Both contain the same information, but only one can be verified, audited, or enforced during inspection.",
      "teachable_moment": "Comments describe intent; schema fields enforce it. The gap between 'we documented the auth model in a comment' and 'the auth model is in the schema' is the gap between optional and mandatory. Every automated check, every doctor command, every future agent that reads this registry will treat the comment as noise and the schema as truth. The TODO is correct — it just needs a due date.",
      "connection": "Root cause of F-003's multi-key accumulation and a prerequisite for making F-001's GHOST classification machine-detectable. Fix F-004 and the other two security gaps become lintable invariants rather than human-recall obligations."
    },
    {
      "id": "F-005",
      "title": "sonar read-plane correction is high-value operational fix",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/freeside-registry/registry.yaml",
      "description": "The correction that GraphQL reads must target belt-gateway-production.up.railway.app/v1/graphql (not sonar-api-production/v1/graphql which 404s) is a concrete, verifiable fix grounded in dogfood evidence (BadgeHolder/BadgeBalance entity confirmation, Envio dead-index tombstone). The 'wrong host' guidance in a registry that's meant to be the discovery source of truth would propagate to every downstream consumer.",
      "suggestion": "No action — this is the right fix, executed correctly.",
      "confidence": 1.0,
      "faang_parallel": "Service mesh teams at Lyft (Envoy) and Uber (μTransport) both learned that the routing configuration layer is the most trust-critical single file in the fleet — a wrong host in a service registry propagates to every consumer simultaneously, silently. The dogfood-driven correction model (run the thing, observe the failure, fix the source) is exactly how Envoy's xDS API was designed to be iterated: observe → correct → propagate.",
      "metaphor": "Correcting a wrong address in the city's official directory rather than giving every visitor a sticky note with the right address. One correction; every future visitor benefits.",
      "teachable_moment": "Dogfood-driven registry corrections are disproportionately high value: one fix propagates to every consumer, past and future, who reads from the source of truth. The discipline of eating your own cooking before publishing is what keeps a registry from becoming a museum of stale configuration."
    },
    {
      "id": "F-006",
      "title": "Dead Envio index tombstoned but no migration path documented",
      "severity": "LOW",
      "file": "packages/freeside-registry/registry.yaml",
      "description": "indexer.hyperindex.xyz/914708e is noted as dead (500). The registry correctly tombstones it but does not reference the Ponder migration status or a target index URL. Consumers who had the old URL hardcoded get no forwarding pointer.",
      "suggestion": "Add a `superseded_by:` or migration reference once the Ponder index is live. A dead URL without a successor leaves consumers stranded without a clear remediation path.",
      "confidence": 0.75,
      "metaphor": "A closed shop with a 'closed' sign but no 'we moved to' address. Honest, but incomplete.",
      "teachable_moment": "Tombstones are good hygiene; tombstones with forwarding pointers are complete hygiene. The extra line costs nothing and saves the next engineer a search session."
    },
    {
      "id": "F-007",
      "title": "identity beacon 404 blocks federation discovery — no target date or tracking reference",
      "severity": "LOW",
      "file": "packages/freeside-registry/registry.yaml",
      "description": "The updated identity-api note flags that /.well-known/beacon.json 404s on the canonical host, making identity federation-invisible. This is noted with a ⚠ but without a beads/issue reference or target resolution. It's a known gap that could be silently forgotten.",
      "suggestion": "Add a beads task reference or issue link to the beacon implementation work so it's tracked to closure rather than remaining as a YAML comment.",
      "confidence": 0.8,
      "metaphor": "A building that doesn't appear on any map. People who already know the address can find it; everyone else can't.",
      "teachable_moment": "A ⚠ in a YAML comment has a half-life. A beads task reference has a closure condition. When the gap is real and tracked, prefer the latter."
    },
    {
      "id": "F-008",
      "title": "mint-api booted-but-routeless state may be missed by health monitoring",
      "severity": "LOW",
      "category": "observability",
      "file": "packages/freeside-registry/registry.yaml",
      "description": "mint-api-production returns Express's default 'Cannot GET /health' — a process is up but no business routes are mounted. Standard uptime monitors checking for HTTP 200 on /health would report this as down (404), but process-level monitors (Railway health) would report it as up. This split state is correctly documented but the monitoring gap is implicit.",
      "suggestion": "Verify Railway's health check config for mint-api. If it's checking process liveness only, the service appears healthy to ops while being functionally dead to consumers. A 404 on /health should be treated as unhealthy.",
      "confidence": 0.7,
      "faang_parallel": "AWS elastic load balancers distinguish liveness (is the process alive?) from readiness (is it serving traffic?) via separate health check paths. A process that boots but fails to mount routes will pass liveness and fail readiness — but only if you configure both. This is the gap between 'the pod is running' and 'the pod is ready' in Kubernetes: two different questions answered by two different endpoints.",
      "metaphor": "A restaurant with the lights on and the door unlocked but no kitchen staff. Passing by, it looks open. Walk in and try to order, and you'll find out otherwise.",
      "teachable_moment": "Health checks are only as useful as what they measure. A /health endpoint that returns 200 because Express is alive, not because routes are mounted, is a false signal. The convention is: /health should fail if the service cannot serve its primary purpose — not just if the process is alive."
    }
  ]
}
```


---

## Callouts

**F-005 — Dogfood-first registry correction [Correctness · PRAISE]**: The sonar read-plane fix is the model for how this registry should grow. It's grounded in observed failure (entity query, 404 on wrong host), corrected at the source of truth, and annotated with enough evidence (BadgeHolder/BadgeBalance confirmed, Envio tombstone) that future readers understand both what changed and why. This is the difference between a registry that decays and one that self-heals.

**Auth model commentary block [Operational · PRAISE]**: The decision to document all four auth mechanisms in a structured prose block — even before the schema supports it — is exactly right. Honest documentation of the current state, including the GHOST classification, is more useful than clean-but-incomplete documentation. The TODO acknowledges the gap without pretending it doesn't exist. This is the registry acting as an awareness surface, which is its real job.

**GHOST semantic [Quality · PRAISE]**: Naming the orphaned ES256 endpoint state as "GHOST" is a deliberate, searchable classification. It signals "live but untethered" in a way that a comment like "# TODO: fix this" does not. Naming failure states is underrated — it makes them grepable, discussable, and promotable to lintable invariants once the schema catches up.

---

PR #264, opened 2026-06-06
freeside-decomposition: contracts-first de-risking (C1 membrane guard + C3 ledger/payment map)

zkSoju:
## What this is

Two payment-free, contracts-first de-risking artifacts from the freeside-decomposition cycle. No building lift (strangler-fig: contracts before lifts). Everything still compiles in the monorepo. Payments are out of scope (NG-1).

## C1: platform/world membrane CI guard (FR-7)

`tools/check-platform-membrane.sh`, wired as a non-bypassable job in `path-domain-check.yml`, enforces two boundaries:

- **Boundary A**: the platform (`apps/{gateway,worker,ingestor}`, `packages/{core,adapters,sandbox}`) must not value-import `themes/sietch`. Baseline is 0; green on HEAD.
- **Boundary B**: `themes/sietch/src` must not add NEW value-imports of `@freeside/adapters/agent` (the ports-only contract). The 4 existing reach-ins are baseline-allowlisted (the burndown list); test files are exempt.

`import type` and inline `{ type X }` are permitted (they erase at compile). madge is unavailable in this workspace (no local node_modules linkage), so the guard is a multi-line-aware awk text scan with documented bypass classes (alias and transitive reach-in are out of scope for v1). 14 guard tests pass, covering multi-line imports, inline type modifiers, and brace-then-from on separate lines (hardened by adversarial review).

## C3: ledger/payment call-graph map (FR-8), verdict DEFER

`tools/ledger-payment-callgraph.mjs` is a read-only NG-1 gate. It maps payment to ledger coupling across three vectors (static import, shared database table, event bus), with a fail-safe: any unresolved edge defaults to defer.

**Verdict: `payment-entangled -> defer`.** Both payment surfaces (`nowpayments-handler.ts:22`, `x402-settlement.ts:20`) call `credit-lot-service` directly (`mintCreditLot`, `debitLots`). 3 static edges plus 11 shared-database edges. The credits-ledger cannot be extracted cleanly without severing those payment callsites, which would touch payment code (out of scope). The verdict was cross-validated independently by a deeper run that reached the same conclusion with more coupling.

## NG-1

Zero edits to any payment or runtime file (nowpayments, x402, Paddle, settlement, pricing, fiat on-ramp). Audited. The committable diff is 3 shared-domain files; no platform or network paths are touched, so the cross-domain firewall passes.

## Deferred (with reasons), not in this PR

- **C2 burndown** (relocate the seam to ports plus the DI rewire): `themes/sietch` cannot typecheck `@freeside/*` on HEAD (171 baseline errors; the modules resolve at runtime via bun, not under tsc). The refactor is therefore not locally verifiable, and it touches auth and JWT wiring, so it waits. The contract stays enforced by Boundary B meanwhile.
- The decision artifacts (full C3 edge map, C2 seam enumeration) live under `grimoires/loa/cycles/freeside-decomposition/` (gitignored by design).

### DIFF
--- a/.github/workflows/path-domain-check.yml
+++ b/.github/workflows/path-domain-check.yml
@@ -93,3 +93,20 @@ jobs:
         run: |
           echo "Bootstrap PR — path-domain-check intentionally skipped."
           echo "See workflow header for bypass policy."
+
+  # Platform↔world membrane guard (C1 / FR-7 / SKP-004, freeside-decomposition cycle).
+  # Codifies the one-directional membrane the decomposition relies on:
+  #   A) platform (apps/{gateway,worker,ingestor}, packages/{core,adapters,sandbox})
+  #      must NOT value-import themes/sietch (the world).
+  #   B) themes/sietch/src must NOT add NEW value-imports of @freeside/adapters/agent
+  #      (the C2 ports-only contract; existing reach-ins are baseline-allowlisted in the script).
+  # NOT bootstrap-bypassable: a membrane breach is a correctness regression regardless of the
+  # cross-domain exception. Runs an in-repo regex guard (madge unavailable in CI/workspace, NG-3;
+  # documented bypass classes live in the script header — transitive + alias reach-ins out of scope v1).
+  membrane:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - name: Platform↔world membrane guard
+        run: bash tools/check-platform-membrane.sh
--- a/tools/check-platform-membrane.sh
+++ b/tools/check-platform-membrane.sh
@@ -1,0 +1,144 @@
+#!/usr/bin/env bash
+# check-platform-membrane.sh — enforce the freeside platform↔world membranes (C1, FR-7/SKP-004).
+#
+# The whole decomposition relies on a one-directional membrane: the PLATFORM substrate
+# (apps/{gateway,worker,ingestor}, packages/{core,adapters,sandbox}) must NEVER value-import
+# the WORLD (themes/sietch) at runtime. The map confirmed this holds today (0 value-imports);
+# this guard CODIFIES the clean state so it can never silently regress (it has regressed +
+# been repaired before — coherence F4/F5). Same "declared → enforced" shape as
+# tools/check-worker-schema-drift.sh.
+#
+# Two boundaries:
+#   A) platform  ↛  themes/sietch            (the platform membrane — FAIL on any new value-import)
+#   B) themes/sietch/src ↛ @freeside/adapters/agent  (the C2 ports-only contract; existing reach-ins
+#      are baseline-allowlisted — teeth on NEW reach-ins; tests exempt)
+#
+# RULE: `import type … from '…'` AND inline `import { type Foo } from '…'` are PERMITTED across a
+#       membrane (they erase at compile). VALUE imports (static `import {…}`/`import x`/`import *`/
+#       side-effect, dynamic `import()`, CJS `require()`) are FORBIDDEN.
+#
+# SCANNER: an awk pass that reconstructs MULTI-LINE import statements before classifying (a line-by-line
+# grep mis-handles `import type {\n …\n} from '…'`). It matches the specifier as a SUBSTRING (no quote
+# escaping) and excludes whole-statement `import type` + all-inline-`type` brace clauses.
+#
+# TOOLING NOTE (honest, like block-destructive-bash's accepted-bypass list): madge is unavailable
+# (no local node_modules linkage, NG-3), so this is a text scan, NOT a resolved import graph.
+# DOCUMENTED BYPASS CLASSES — OUT OF SCOPE for v1, NOT claimed as covered:
+#   - TS path-alias imports of the world (e.g. a hypothetical `@sietch/*` alias) — only the literal
+#     `themes/sietch` / `@freeside/adapters/agent` specifier SUBSTRING is matched. No such alias today.
+#   - Transitive reach-in (platform A → platform B → world): only DIRECT edges are scanned (1-hop).
+#   - the specifier substring appearing inside a non-import string literal on an import-looking line.
+# Upgrading to madge/ts-morph would only find MORE reach-ins, never fewer.
+#
+# Exit 0 = membranes intact · Exit 1 = violation (with ::error::file:line for CI annotation).
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$REPO_ROOT"
+
+PLATFORM_DIRS=(apps/gateway apps/worker apps/ingestor packages/core packages/adapters packages/sandbox)
+WORLD_SPECIFIER="themes/sietch"
+WORLD_SRC="themes/sietch/src"
+ADAPTER_SPECIFIER="@freeside/adapters/agent"
+
+# Boundary B baseline-allowlist (C1.2): sietch SRC files that VALUE-import the agent adapter TODAY.
+# Teeth on NEW reach-ins immediately; these are the C2 burndown list — DELETE each line as C2 relocates
+# that file's seam to @freeside/core/ports. Empty list ⇒ boundary B fully enforced. TESTS EXEMPT.
+# byok.routes.ts is NOT here: it is type-only.
+BOUNDARY_B_ALLOWLIST=(
+  "themes/sietch/src/api/server.ts"
+  "themes/sietch/src/api/routes/agents.routes.ts"
+  "themes/sietch/src/discord/commands/agent.ts"
+  "themes/sietch/src/telegram/commands/agent.ts"
+)
+
+violations=0
+flag() { echo "::error::membrane: $1"; violations=$((violations + 1)); }
+
+# find_value_imports <file> <specifier-substring>
+# Prints the start line number of each VALUE import of the specifier (multi-line aware; type-only excluded).
+find_value_imports() {
+  local file="$1" spec="$2"
+  awk -v spec="$spec" '
+    function emit_if_value(stmt, ln,    clause, n, i, p, anyType, allType, parts) {
+      if (index(stmt, spec) == 0) return
+      if (stmt ~ /^[ \t]*(import|export)[ \t]+type[ \t]/) return        # whole import type-only → allowed
+      if (stmt ~ /\{/) {                                                # all-inline-type brace → allowed
+        clause = stmt; sub(/^[^{]*\{/, "", clause); sub(/\}.*/, "", clause)
+        n = split(clause, parts, ","); anyType = 0; allType = 1
+        for (i = 1; i <= n; i++) { p = parts[i]; gsub(/^[ \t]+|[ \t]+$/, "", p); if (p == "") continue; anyType = 1; if (p !~ /^type[ \t]/) allType = 0 }
+        if (anyType && allType) return
+      }
+      print ln
+    }
+    BEGIN { inblk = 0; buf = ""; start = 0 }
+    {
+      if (inblk) {                                                       # inside a multi-line brace import
+        buf = buf " " $0
+        # close on the module clause (from ...) or statement end — NOT on } alone:
+        # a brace import may put } and `from '...'` on separate lines, and the specifier
+        # lives in the from-clause, so closing on } would drop it (false negative).
+        if ($0 ~ /(^|[ \t])from([ \t]|$)/ || $0 ~ /;[ \t]*$/) { emit_if_value(buf, start); inblk = 0; buf = "" }
+        next
+      }
+      if ($0 ~ /^[ \t]*(import|export)([ \t]|\(|$)/) {
+        if ($0 ~ /\{/ && $0 !~ /\}/) { inblk = 1; buf = $0; start = NR }  # opens multi-line brace
+        else { emit_if_value($0, NR) }                                   # single-line / default / side-effect / dynamic-at-start
+      } else if ($0 ~ /(import|require)[ \t]*\(/ && index($0, spec) > 0) {
+        print NR                                                         # dynamic import()/require() mid-line
+      }
+    }
+  ' "$file" 2>/dev/null | sort -un
+}
+
+scan_boundary_a() {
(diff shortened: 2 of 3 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#264. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Namespace/default imports of ledger writers evade the callgraph (unsafe blind spot, undocumented)",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/ledger-payment-callgraph.mjs",
      "description": "scanStatic only tests brace-clause symbols against ALL_WRITER_SYMBOLS. A namespace import (`import * as ledger from 'credit-lot-service'; ledger.mintCreditLot()`) or default import produces no brace symbols, so the write call is never detected and no static edge is added. Unlike the documented bypasses (dynamic import, transitive, DI, ORM), this one drives the verdict in the UNSAFE direction: a genuinely entangled payment surface using `import * as` could yield `cleanly-separable`. The tool_limitations string does not list namespace/default import as a blind spot.",
      "suggestion": "Also capture namespace (`* as X`) and default bindings during extractImports and, when the spec is a ledger writer module, scan the file body for `X.<writerSymbol>(` / `<default>(` call sites; or at minimum add namespace/default imports to the documented bypass classes so a match is flagged `unresolved` rather than silently clean.",
      "confidence": 0.6
    },
    {
      "id": "F2",
      "title": "Mixed default+type-only-brace import bypasses the membrane guard",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/check-platform-membrane.sh",
      "description": "emit_if_value extracts the brace clause and, if every braced symbol is `type ...`, returns (treats the whole statement as erasable). For `import Real, { type T } from 'themes/sietch'` the default binding `Real` is a runtime VALUE import, but the all-type brace check returns early and the value import is not flagged. A plain `import Real from 'themes/sietch'` (no brace) is correctly caught, so this is specifically the default+type-brace combination. This is a false negative in a guard whose entire purpose is preventing platform→world value imports.",
      "suggestion": "Before the all-inline-type early return, check whether text precedes the `{` after `import` (a default or `* as` binding) — if a value binding exists outside the braces, do not treat the statement as type-only.",
      "confidence": 0.55
    },
    {
      "id": "F3",
      "title": "INCONCLUSIVE branch mislabels 'scanned nothing' when glob payment surfaces were scanned",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/ledger-payment-callgraph.mjs",
      "description": "The verdict branch `else if (missingFiles.length === PAYMENT_SURFACES.length)` fires when both hardcoded surfaces are absent, emitting reason 'scanned nothing; cannot assert separable.' But paymentFiles also includes PAYMENT_SURFACE_GLOBS results, and missingFiles only tracks the two hardcoded files. If the hardcoded surfaces are gone but sietch billing/payment glob files exist and were scanned clean, the output falsely claims nothing was scanned and defers to INCONCLUSIVE. The defer direction is safe, but the stated reason is inaccurate and could mislead the operator gating the extraction decision.",
      "suggestion": "Gate the INCONCLUSIVE branch on `paymentFiles.length === missingFiles.length` (i.e. zero files actually read), not on the hardcoded-surface count.",
      "confidence": 0.7
    },
    {
      "id": "F4",
      "title": "Comment stripping only catches line-leading // or *; inline/trailing comments leak into edge detection",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/ledger-payment-callgraph.mjs",
      "description": "scanSharedDb and scanEvents skip lines matching `/^\\s*(\\/\\/|\\*)/`, i.e. only lines that START with // or *. A trailing comment (`db.query(...) // rework credit_lots later`), a `/* ... */` block that doesn't start with * on continuation lines, or a commented-out SQL line indented after code will still be regex-scanned. For the shared-db and event vectors this can manufacture false `reaches_payment`/`unresolved` edges, tilting the verdict toward defer on commented-out references.",
      "suggestion": "Either accept this as a documented conservative bias (note it in tool_limitations) or strip `//` trailing comments and `/* */` blocks before the table/event regex passes.",
      "confidence": 0.5
    },
    {
      "id": "F5",
      "title": "Boundary A applies no test-file exemption; platform test imports of the world fail CI",
      "severity": "LOW",
      "file": "tools/check-platform-membrane.sh",
      "description": "scan_boundary_a's find matches all `*.ts` except `*.d.ts` — test files included. Boundary B explicitly exempts `*.test.ts`, and the header notes 'tests exempt' for boundary B only. A platform-side test that value-imports themes/sietch (e.g. an integration test wiring a real world adapter) would be a CI-failing membrane violation. This may be intentional, but the asymmetry with Boundary B is undocumented and could surprise contributors.",
      "suggestion": "Decide explicitly whether platform tests may import the world; if not, document the intentional difference from Boundary B in the header; if they may, add `-not -name '*.test.ts'` to the Boundary A find.",
      "confidence": 0.4
    },
    {
      "id": "F6",
      "title": "Multi-line dynamic import()/require() specifiers on a following line are not scanned",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/check-platform-membrane.sh",
      "description": "The awk dynamic-import branch matches `(import|require)[ \\t]*\\(` AND `index($0, spec) > 0` on the SAME line. A dynamic import split across lines (`await import(\\n  'themes/sietch/x'\\n)`) starts a line with `import(` handled by the first branch (no brace → emit_if_value on the opening line only, which lacks the specifier), so the specifier on the next line is never matched. This is not in the documented bypass list (which covers path-aliases, transitive, and string-literal-on-import-line).",
      "suggestion": "Add multi-line dynamic import/require to the documented bypass classes, or extend the awk state machine to buffer an unbalanced `import(` until the closing paren before substring-matching.",
      "confidence": 0.5
    },
    {
      "id": "F7",
      "title": "Event vector matches any writer-file emit near a keyword, likely pinning the verdict to defer permanently",
      "severity": "SPECULATION",
      "category": "design",
      "file": "tools/ledger-payment-callgraph.mjs",
      "description": "scanEvents scans both payment files AND every ledger-writer file for `(publish|emit|enqueue|outbox)(` co-occurring with `(credit|ledger|lot|payment|settle)`. Any ledger writer that publishes a domain event mentioning 'credit'/'lot'/'ledger' produces an `unresolved` edge, and unresolved always defers. If the ledger is event-sourced (event-sourcing-service.ts is a declared writer), the very design of the ledger guarantees an unresolved edge, so `cleanly-separable` may be effectively unreachable regardless of actual payment coupling. Worth confirming the verdict can ever be clean under the real codebase.",
      "suggestion": "Consider scoping the event vector to cross-surface flows (payment file publishes AND a ledger writer subscribes to the same channel) rather than any emit-near-keyword, so a self-contained event-sourced ledger doesn't auto-defer.",
      "confidence": 0.4
    },
    {
      "id": "F8",
      "title": "Honest, explicit bypass-class documentation and fail-safe verdict design",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/ledger-payment-callgraph.mjs",
      "description": "Both tools enumerate exactly what the text-scan does NOT cover (dynamic target resolution, transitive chains, DI-injected sites, ORM/query-builder access, path aliases) instead of overclaiming a resolved import graph, and the callgraph verdict defaults every unresolved indirection to `payment-entangled → defer` (NG-1 fail-safe) with a distinct INCONCLUSIVE state for 'scanned nothing'. This conservative, unsafe-direction-aware framing is the right posture for a decision-gating analysis tool.",
      "suggestion": "No change needed. Keep the tool_limitations string in sync as vectors are added (see F1).",
      "confidence": 0.9
    },
    {
      "id": "F9",
      "title": "Boundary-B allowlist is a self-documenting C2 burndown list with immediate teeth on new reach-ins",
      "severity": "PRAISE",
      "file": "tools/check-platform-membrane.sh",
      "description": "BOUNDARY_B_ALLOWLIST grandfathers exactly the current sietch→adapters/agent value-imports while failing any NEW one, and the comment instructs deleting each line as C2 relocates the seam — an empty list meaning full enforcement. This ratchet pattern prevents regressions without blocking the PR on pre-existing debt and makes the migration progress visible in diffs.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

eileen1337:
CODEX_CONTINUATION_REQUESTED

@codex

## Goal
Publish a narrowly scoped hardening patch on the existing PR branch so the platform/world membrane guard and ledger/payment call-graph cannot silently produce unsafe false negatives.

## Repository / item / current head
- Repo: `0xHoneyJar/loa-freeside`
- PR: `#264`
- Branch: `freeside-decomp-contracts`
- Current GitHub head: `b5f41d29a87a0528dd56d8e379f3b940cd97e347`
- Base: `main`
- PR state: open, draft, mergeable

## Verified evidence
- The PR changes three bounded surfaces: `.github/workflows/path-domain-check.yml`, `tools/check-platform-membrane.sh`, and `tools/ledger-payment-callgraph.mjs`.
- Current review identifies two MEDIUM correctness failures that can make the tools miss real coupling:
  1. `tools/check-platform-membrane.sh` treats `import Real, { type T } from 'themes/sietch'` as type-only because it examines only the brace clause, allowing the runtime default binding to bypass Boundary A.
  2. `tools/ledger-payment-callgraph.mjs` scans named brace imports but can miss namespace/default imports of ledger writers, allowing actual payment→ledger calls to evade the static vector and potentially produce an unsafe `cleanly-separable` result.
- Lower-severity findings include inaccurate INCONCLUSIVE reasoning when glob surfaces were scanned, comment-induced false positives, undocumented Boundary-A test policy, multi-line dynamic import/require bypasses, and an event-vector design concern.
- There are no inline review threads, but the submitted analytical review remains material.
- Current-head workflow runs returned by GitHub:
  - green: Path Domain Check `30030436848`, Commit Scope Check `27075473762`, earlier Path Domain Check `27075473769`, Secret Scanning `27075473764`;
  - red: CI `27075473760`, Security Audit `27075473759`.
- Do not claim the red workflows are patch-caused or unrelated without inspecting their jobs/logs.

## Patch hypothesis
### Membrane guard
Harden import classification so a statement is allowed only when every imported binding is type-only. Correctly detect value bindings outside a `{ type ... }` clause, including default and namespace bindings. Preserve multi-line handling and fail-closed behavior.

### Ledger/payment call-graph
Extend static import extraction to represent named, namespace, and default bindings. For ledger-writer modules:
- detect calls through namespace aliases such as `ledger.mintCreditLot(...)`;
- classify default imports conservatively: either prove the binding maps to a writer call or mark the edge unresolved so the verdict defers;
- never allow an unrecognized import form from a writer module to silently contribute to `cleanly-separable`.

Correct the INCONCLUSIVE reason to use the actual number of readable/scanned payment files, not only absence of the two hardcoded surfaces.

For lower-confidence findings, either implement focused safe fixes or document the exact conservative limitation. Do not broaden this into an AST-tooling migration.

## Allowed files
- `tools/check-platform-membrane.sh`
- `tools/ledger-payment-callgraph.mjs`
- focused tests/fixtures under existing repository conventions
- `.github/workflows/path-domain-check.yml` only if required to run the focused guard tests consistently
- adjacent comments/documentation describing accepted bypass classes and test policy

## Forbidden surfaces
- payment runtime files, settlement handlers, pricing, auth, JWT, signing, secrets, permissions, deployment, production configuration, database migrations
- schema definitions or ledger writer runtime behavior
- broad dependency upgrades or new parser frameworks
- weakening either guard into a warning
- changing the current `payment-entangled -> defer` production verdict without evidence
- force-push or branch replacement

## Required verification
1. Run the membrane guard on the current tree and show success.
2. Add/run mutation tests proving:
   - `import Real, { type T } from 'themes/sietch/...'` fails Boundary A;
   - `import * as Real from 'themes/sietch/...'` fails;
   - all-type named imports remain allowed;
   - ordinary default, named value, side-effect, dynamic, and require imports retain their documented classifications;
   - multi-line forms covered by the tool behave as claimed.
3. Run the call-graph on the current tree and preserve the conservative defer verdict.
4. Add/run focused fixtures proving:
   - named writer import/call is detected;
   - namespace writer alias/call is detected;
   - default writer import cannot silently yield clean separation;
   - absence of hardcoded surfaces with a scanned glob surface does not report “scanned nothing”;
   - actual zero-readable-surface input remains INCONCLUSIVE/defer.
5. Run syntax/static checks for Bash, Node, and workflow YAML using repository-available tools.
6. Inspect CI `27075473760` and Security Audit `27075473759`; classify each as patch-caused, base/stale, dependency-wide, or unverified. Fix only patch-caused failures within allowed scope.
7. Publish to the existing PR branch and report the new remote SHA. A local commit or `make_pr` metadata is not completion.

## Acceptance criteria
- A new GitHub-visible head exists on PR `#264`.
- Both MEDIUM false-negative paths are fixed and mutation-tested.
- Unknown ledger-writer import forms fail safe toward unresolved/defer, never toward clean separation.
- Existing conservative verdict and NG-1 boundary are preserved.
- No payment/runtime/security-sensitive surface is edited.
- Return exact changed files, commands, results, new remote SHA, and residual red workflows.

## Non-claims
- Do not claim all CI is green unless current remote-head workflows prove it.
- Do not treat detailed local output as publication evidence.
- Do not mark ready for audit until the remote head and current-head evidence are re-fetched.
- Do not claim the call-graph is a resolved semantic graph; retain honest limitation language.

## Return format
- `PUBLISHED` or `PUBLISH_BLOCKED`
- old head → new remote head
- changed files
- commands and exact results
- mutation-test matrix
- current workflow classification
- remaining risks/non-claims

---
Tomato Cognitive Controller #420

eileen1337:
ENVIRONMENT_BLOCKED

DO_NOT_RETRY_UNTIL_NEW_EVIDENCE

The bounded hardening task for PR #264 produced a local commit but no GitHub-visible publication. The live PR head remains `b5f41d29a87a0528dd56d8e379f3b940cd97e347`; the Codex result reported `PUBLISH_BLOCKED` and no new remote SHA.

This campaign fingerprint is now terminally blocked for the current environment. Do not issue another identical continuation against this head.

Reactivate only if one of the following materially changes:
- a new remote head appears on PR #264;
- the patch is published in a separate PR;
- a branch-attached, push-capable environment becomes available;
- network/GitHub publication access changes;
- the review or workflow evidence changes enough to create a different bounded task.

Local commit or task metadata is not publication evidence.

---
Tomato Cognitive Controller #420

PR #267, opened 2026-06-07
recover: worker-schema-drift guard + InteractionHistory port (stranded local-only work)

zkSoju:
## What this is

Two finished pieces of work that were committed on a local-only branch and **never reached GitHub** (surfaced by a content-diff reconciliation of the local clone against this repo's main). Recovered here, grafted clean onto current main, both verified.

## C-1: worker/sietch Drizzle schema drift-guard (arrakis-pcpc / F5B)

`tools/check-worker-schema-drift.sh` + a CI job. The three shared Drizzle tables (`communities`, `profiles`, `badges`) are copied verbatim into `apps/worker`; this fails CI if the copy drifts from the `themes/sietch` canonical. Verified green on HEAD (all three in sync). This is the mirror that `tools/check-platform-membrane.sh` (PR #264) names.

## C-2: relocate InteractionHistory port to @freeside/core (arrakis-shqm, Coherence F4)

`InteractionRecord` + `InteractionHistoryProvider` were declared in `themes/sietch` and imported by a platform adapter via a 4-level relative path, inverting the belt direction (platform depending on a theme). Moved both (byte-identical) to `@freeside/core/ports`; `capability-mesh.ts` re-exports them; `audit-query-service.ts` repoints to the port. Result: the platform to sietch reach-in is closed. Pure type relocation, no runtime change. `packages/core` typechecks with the port added.

This is the fix that *made* the membrane clean, so it pairs naturally with the C1 guard in #264 that codifies that cleanliness.

## Domain

platform + shared, zero network paths. The cross-domain firewall passes (classifies as platform).

## Provenance

Recovered via `~/bonfire/reconcile-landed.mjs` (content-diff vs the remote default branch), part of a local to GitHub legibility pass. These were genuinely stranded (only on local disk), not WIP.

### DIFF
--- a/.github/workflows/worker-schema-drift-check.yml
+++ b/.github/workflows/worker-schema-drift-check.yml
@@ -1,0 +1,42 @@
+# =============================================================================
+# Worker Schema Drift Check
+# Coherence finding F5 / arrakis-pcpc — make worker↔sietch Drizzle drift LOUD.
+# =============================================================================
+#
+# apps/worker/src/data/schema.ts holds verbatim COPIES of three Drizzle tables
+# (communities, profiles, badges) canonically defined in
+# themes/sietch/src/packages/adapters/storage/schema.ts. The copy is intentional
+# (worker runs its own drizzle instance) but silent divergence was an unguarded
+# risk. This gate FAILS when the shared table definitions drift apart.
+#
+# Fail-block (teeth): a drift is a red check, not a warning — the worker copy must
+# mirror the sietch canonical until the shared schema package is extracted
+# (deferred to the worlds-api cycle; see tools/check-worker-schema-drift.sh).
+
+name: Worker Schema Drift Check
+
+on:
+  pull_request:
+    paths:
+      - 'apps/worker/src/data/schema.ts'
+      - 'themes/sietch/src/packages/adapters/storage/schema.ts'
+      - 'tools/check-worker-schema-drift.sh'
+      - '.github/workflows/worker-schema-drift-check.yml'
+  push:
+    branches: [main]
+    paths:
+      - 'apps/worker/src/data/schema.ts'
+      - 'themes/sietch/src/packages/adapters/storage/schema.ts'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  schema-drift:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check worker↔sietch Drizzle schema drift
+        shell: bash
+        run: bash tools/check-worker-schema-drift.sh
--- a/packages/adapters/storage/audit-query-service.ts
+++ b/packages/adapters/storage/audit-query-service.ts
@@ -11,7 +11,7 @@
 
 import type { Pool } from 'pg';
 import type { Logger } from 'pino';
-import type { InteractionHistoryProvider, InteractionRecord } from '../../../themes/sietch/src/packages/core/protocol/capability-mesh.js';
+import type { InteractionHistoryProvider, InteractionRecord } from '@freeside/core/ports';
 
 // ─── Types ───────────────────────────────────────────────────────────────────
 
--- a/packages/core/ports/index.ts
+++ b/packages/core/ports/index.ts
@@ -76,3 +76,6 @@ export * from './migration.js';
 
 // Agent Gateway Interface (Hounfour Phase 4 — Spice Gate)
 export * from './agent-gateway.js';
+
+// Interaction History Port (relocated from themes/sietch — coherence F4 / arrakis-shqm)
+export * from './interaction-history.js';
--- a/packages/core/ports/interaction-history.ts
+++ b/packages/core/ports/interaction-history.ts
@@ -1,0 +1,29 @@
+/**
+ * Interaction History Port
+ *
+ * The InteractionHistoryProvider contract decouples capability/mesh resolution
+ * from the persistent audit query layer (dependency injection). It lives in core
+ * (the platform substrate) so that BOTH the theme that produces interaction
+ * records and the storage adapter that backs them depend on core — never on each
+ * other.
+ *
+ * Originally declared inside themes/sietch (capability-mesh.ts, cycle-043);
+ * relocated here to restore the belt direction platform-core <- theme and remove
+ * the platform->theme reach-in import (coherence finding F4 / arrakis-shqm).
+ */
+
+/** Record of interaction between two models */
+export interface InteractionRecord {
+  model_pair: [string, string];
+  quality_score: number;
+  observation_count: number;
+}
+
+/** Provider interface for interaction history — decoupled from storage layer */
+export interface InteractionHistoryProvider {
+  /**
+   * Get interaction records between two models.
+   * Returns empty array if no interactions found.
+   */
+  getInteractions(modelA: string, modelB: string): Promise<InteractionRecord[]>;
+}
--- a/themes/sietch/src/packages/core/protocol/capability-mesh.ts
+++ b/themes/sietch/src/packages/core/protocol/capability-mesh.ts
@@ -16,24 +16,15 @@ import type {
   CapabilitySet,
   ResolutionContext,
 } from './capability-catalog.js';
+// InteractionRecord + InteractionHistoryProvider now live in @freeside/core/ports
+// (single source of truth, platform-core <- theme belt — coherence F4 / arrakis-shqm).
+// Imported for local use by InMemoryInteractionHistoryProvider below.
+import type { InteractionRecord, InteractionHistoryProvider } from '@freeside/core/ports';
 
 // ─── Types ───────────────────────────────────────────────────────────────────
 
-/** Record of interaction between two models */
-export interface InteractionRecord {
-  model_pair: [string, string];
-  quality_score: number;
-  observation_count: number;
-}
-
-/** Provider interface for interaction history — decoupled from storage layer */
-export interface InteractionHistoryProvider {
-  /**
-   * Get interaction records between two models.
-   * Returns empty array if no interactions found.
-   */
-  getInteractions(modelA: string, modelB: string): Promise<InteractionRecord[]>;
-}
+// Re-exported for any existing importers of these names from this module.
+export type { InteractionRecord, InteractionHistoryProvider };
 
 /** Threshold configuration for ensemble capability unlocking */
 export interface MeshThresholdConfig {
--- a/tools/check-worker-schema-drift.sh
+++ b/tools/check-worker-schema-drift.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+# check-worker-schema-drift.sh — make worker↔sietch Drizzle-schema drift LOUD.
+#
+# apps/worker/src/data/schema.ts holds verbatim COPIES of three Drizzle tables
+# (communities, profiles, badges) that are canonically defined in
+# themes/sietch/src/packages/adapters/storage/schema.ts. The worker needs its own
+# table objects (its own drizzle instance), so the copy is intentional — but a column
+# change in sietch silently NOT propagating to the worker copy was an unguarded drift
+# risk (coherence finding F5 / arrakis-pcpc). This check fails CI when the shared table
+# definitions diverge, so the drift can never again be silent.
+#
+# Scope: the 3 SHARED tables only. agentThreads is worker's own (no sietch Drizzle
+# equivalent — only raw migrations 063/066), so it is NOT a drift target.
+#
+# This is a TEXT comparison, NOT an import — a worker test importing sietch's schema
+# would be a platform->theme membrane reach-in. Comparing source text avoids that.
+#
+# The durable fix (extract a shared schema package consumed by both) is deferred to the
+# worlds-api extraction cycle: the 3 tables are deeply FK-coupled into sietch's 21-table
+# schema, so extraction belongs with that larger refactor. Until then, this guard holds.
+#
+# Exit 0 = in sync · Exit 1 = drift detected.
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+WORKER="$REPO_ROOT/apps/worker/src/data/schema.ts"
+SIETCH="$REPO_ROOT/themes/sietch/src/packages/adapters/storage/schema.ts"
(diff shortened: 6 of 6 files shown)


zkSoju (commented):
## Summary
This is a clean recovery PR: the InteractionHistory port relocation (drift-06) is a textbook dependency-inversion fix, and the schema-drift guard is a thoughtful safety net. The primary concerns are all in the bash guard itself — `set -euo pipefail` interactions and brittle text-matching that can either swallow the friendly error or produce false drift. Verdict: `REQUEST_CHANGES`, mostly to harden the shell script before it grows teeth.

The architectural instinct here is sound: make schema drift *loud* rather than discovered in production. But a fail-blocking gate is only as trustworthy as its false-positive rate. The findings below cluster around one theme — the gap between "text that looks the same" and "schema that means the same."

## Findings

```json
{
  "schema_version": 1,
  "verdict": "REQUEST_CHANGES",
  "findings": [
    {
      "id": "drift-01-pipefail-empty-extract",
      "title": "set -e + pipefail causes premature exit on missing table, bypassing friendly error",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/check-worker-schema-drift.sh",
      "description": "Under `set -euo pipefail`, `w=\"$(extract_normalized ... | grep -v '^$')\"` exits 1 when a table is absent (grep matches nothing), and set -e kills the script before the `if [ -z \"$w\" ]` branch prints 'table not found'. CI still fails, but with no actionable message — defeating the LOUD goal for the removed/renamed case.",
      "suggestion": "Append `|| true` to the grep, or wrap the extraction in `set +e`/`set -e`. Verify by deleting a table and confirming the friendly error prints.",
      "confidence": 0.65,
      "faang_parallel": "Google's Shell Style Guide explicitly warns that pipefail + command substitution swallows intended error handling — the SRE workbook calls this 'failing invisibly,' worse than failing loudly because the diagnostic never reaches the human.",
      "metaphor": "Like a smoke detector wired to also cut the building's power — the alarm you built to warn people is silenced by the very fault it was meant to announce.",
      "teachable_moment": "`set -e` and explicit error branches are rivals, not teammates. When you want to *handle* a non-zero exit, you must neutralize errexit locally (`|| true`), or it handles it for you — fatally."
    },
    {
      "id": "drift-02-close-regex",
      "title": "Block-terminator regex ^\\); may not match all Drizzle table closings",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/check-worker-schema-drift.sh",
      "description": "extract_normalized exits on the first `/^\\);/` line. pgTable defs with a third arg (index/constraint callback) can close with `});`, `]);`, or an indented `  );`. If the close isn't `);` at column 0, awk runs to EOF and pulls unrelated tables into the compared text — false drift or masked drift.",
      "suggestion": "Track paren/brace depth, or exit on `/^[)}\\]]*\\);/`. Confirm against real schema.ts formatting for all three tables, including any with index callbacks.",
      "confidence": 0.5,
      "faang_parallel": "This is why every serious codebase eventually reaches for a real parser — the same lesson GitHub learned migrating linguist off regex heuristics. Structured formats deserve structured parsing; regex on nested delimiters is a losing arms race.",
      "metaphor": "Counting matching parentheses with a single regex is like finding the exit of a maze by walking until you hit the first wall — it works until the maze has more than one room.",
      "teachable_moment": "Regex has no memory of nesting depth. The moment your delimiters can nest, you need a counter, not a pattern."
    },
    {
      "id": "drift-03-text-comparison-brittle",
      "title": "Pure text comparison will false-positive on formatting/ordering differences",
      "severity": "LOW",
      "file": "tools/check-worker-schema-drift.sh",
      "description": "Comparing normalized source text means column reordering, prettier reflow, quote-style, or one-sided trailing commas all register as drift and block unrelated PRs. Stripping after `//` can also mask a real divergence living near an inline comment. Intentional per the header, but a real maintenance cost for a fail-blocking gate.",
      "suggestion": "Normalize harder (collapse whitespace, sort column lines) or document a prominent 'keep byte-for-byte identical' contract atop both schema.ts files.",
      "confidence": 0.55,
      "connection": "This ties back to drift-02 — both are symptoms of comparing syntax when you mean to compare semantics. The robust endgame is normalizing to a canonical AST or the emitted SQL DDL, not the source text.",
      "metaphor": "Diffing two recipes character-by-character: 'add 2 eggs' and 'add two eggs' flag as different meals, while a genuine swap of sugar for salt hiding in a comment slips through."
    },
    {
      "id": "drift-04-required-check-path-filter",
      "title": "Path-filtered workflow as a required status check can leave PRs pending",
      "severity": "LOW",
      "category": "ci",
      "file": ".github/workflows/worker-schema-drift-check.yml",
      "description": "The pull_request trigger is `paths`-scoped. If later marked required in branch protection, PRs not touching those paths never produce the check, and GitHub holds them 'Expected — waiting' forever. A well-known GitHub gotcha, foreseeable given the fail-block framing.",
      "suggestion": "If made required, drop the path filter and short-circuit inside the job, or add a companion always-run job that reports success when paths are untouched.",
      "confidence": 0.4,
      "faang_parallel": "This is a documented failure mode in GitHub's own branch-protection guidance and bit Kubernetes CI hard enough that they standardized the 'always-run skip job' pattern org-wide.",
      "teachable_moment": "A required check must be *guaranteed to run*. Path filters break that guarantee — the fix is a job that always runs and internally decides whether to do work."
    },
    {
      "id": "drift-05-push-paths-omit-script",
      "title": "push trigger omits the drift script and workflow from its paths filter",
      "severity": "LOW",
      "category": "ci",
      "file": ".github/workflows/worker-schema-drift-check.yml",
      "description": "pull_request paths include the script and workflow file; `push: branches: [main]` paths list only the two schema.ts files. A direct push to main fixing the comparison logic won't re-run the gate on main. Minor scope inconsistency.",
      "suggestion": "Mirror the pull_request paths in the push trigger, or document why push scope is intentionally narrower.",
      "confidence": 0.45,
      "metaphor": "Two doors into the same room with different guest lists — most visitors never notice, until the one change that matters walks through the unwatched door."
    },
    {
      "id": "drift-06-belt-direction-fix",
      "title": "Relocating InteractionHistory port to core correctly restores platform-core <- theme belt",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/core/ports/interaction-history.ts",
      "description": "Moving InteractionRecord/InteractionHistoryProvider into core, with both theme and storage adapter depending on core (not the adapter reaching into themes/sietch/.../capability-mesh.js), removes a real layering violation. The re-export shim preserves importers; audit-query-service imports via the stable @freeside/core/ports barrel. Clean, backward-compatible DIP fix.",
      "suggestion": "Delete the capability-mesh.ts re-export shim once downstream importers are confirmed migrated.",
      "confidence": 0.85,
      "faang_parallel": "This is the Dependency Inversion Principle exactly as Uncle Bob framed it for the Clean Architecture at scale — high-level policy (core) owns the interface; details (adapters, themes) depend inward. Netflix's DGS federation enforces the same 'schema in core, resolvers at the edge' shape.",
      "metaphor": "You moved the power outlet from inside one tenant's apartment to the building's shared utility panel — now every unit plugs into the same stable source instead of running an extension cord through a neighbor's wall.",
      "teachable_moment": "When an adapter imports from a theme, the arrow points the wrong way. The fix is never a cleverer import path — it's relocating the shared contract to the layer both sides already depend on."
    },
    {
      "id": "drift-07-not-found-vs-drift-messaging",
      "title": "Whitespace/format variance in the export line reports 'not found' instead of 'drift'",
      "severity": "LOW",
      "file": "tools/check-worker-schema-drift.sh",
      "description": "The anchor requires `index($0, t) == 1` — the line must begin exactly with `export const <name> = pgTable(`. If one file writes `pgTable (` with a space, extraction yields empty and the script reports 'table not found', misleading a maintainer into thinking the table was deleted rather than reformatted.",
      "suggestion": "Loosen the anchor to a regex on `export const <name> = pgTable`, or clarify the 'not found' message to note it may indicate a reformatted declaration line.",
      "confidence": 0.4,
      "connection": "Same root as drift-01: the error message and the actual condition have drifted apart. Good diagnostics name the *cause* the maintainer will act on, not the intermediate symptom the script observed."
    }
  ]
}
```


## Callouts
**[Architecture] drift-06** — The belt-direction fix is the quiet star here. Restoring `core ← theme` and routing the adapter through the `@freeside/core/ports` barrel is the kind of change that pays compounding interest: every future importer inherits a stable contract instead of a fragile cross-package reach. The backward-compatible re-export shim shows real care for downstream consumers — you cleaned the architecture without breaking a single caller.

**[Operational] Intent** — Building a drift guard *at all* is the mark of someone who's been burned by silent schema divergence. The instinct to "make drift LOUD" is exactly right; the findings above are about tightening the mechanism, not questioning the goal.

eileen1337:
CODEX_CONTINUATION_REQUESTED

@codex

## Goal
Harden and publish the worker-schema drift guard on the existing PR branch so this draft can become a trustworthy current-head audit candidate without changing the already-sound InteractionHistory port relocation.

## Repository / item / current head
- Repo: `0xHoneyJar/loa-freeside`
- PR: `#267`
- Branch: `recover/platform-stranded-chunks`
- Current GitHub head: `25601b05446b2b7175a70c9671785e871fa5a8d8`
- Base: `main`

## Verified evidence
- PR is open, draft, and GitHub reports it mergeable.
- `Worker Schema Drift Check` run `27081436674` is green on the current head.
- Build, typecheck, lint, Docker build, CodeQL, Dependency Review, secret scanning, path-domain, commit-scope, and schema-emission checks are green.
- Current review identifies two substantive correctness defects in `tools/check-worker-schema-drift.sh`:
  1. `set -euo pipefail` can terminate on an empty extraction before the intended friendly “table not found” branch executes.
  2. The `awk` terminator `/^\);/` is format-sensitive and can overrun a Drizzle definition when closing syntax/indentation changes.
- Lower-severity findings concern source-text false positives, declaration-anchor diagnostics, workflow path filters, and asymmetric push paths.
- There are no inline review threads returned by GitHub, but the submitted review is `REQUEST_CHANGES` and remains material.
- The InteractionHistory port relocation is not the blocker; preserve it unless a required compilation fix demands a surgical adjustment.
- Red workflows on the old head also include repository-wide unit/integration/dependency-audit failures. Do not claim those are patch-caused or fixed without reproducing and classifying them.

## Patch hypothesis
Replace the brittle source-block extraction with a deterministic bounded extractor that:
- does not rely on an unhandled non-zero `grep` exit under `set -e`;
- recognizes the complete `pgTable(...)` declaration across indentation and optional third-argument callbacks using balanced delimiter tracking or another demonstrably bounded method;
- emits distinct actionable diagnostics for declaration not found, malformed/unclosed declaration, and semantic/text drift;
- retains fail-closed behavior;
- explicitly documents the byte/canonical-source contract if source-text comparison remains intentional.

Align workflow path filters so edits to the script/workflow exercise the gate on both PR and main push paths. Do not make a path-filtered job a required-check trap; if the repository expects the check to be universally required, use an always-run job with an internal no-op decision. Otherwise document that it is not universally required.

## Allowed files
- `tools/check-worker-schema-drift.sh`
- `.github/workflows/worker-schema-drift-check.yml`
- focused tests/fixtures for this guard under the repository’s existing test conventions
- minimal documentation/comments immediately adjacent to these files
- lockfile only if a genuinely necessary existing test dependency requires it; avoid new dependencies

## Forbidden surfaces
- payment, auth, signing, secrets, permissions, deployment, production configuration, database migrations
- unrelated schema definitions
- InteractionHistory runtime behavior or public contract changes
- broad dependency upgrades
- weakening or converting the drift gate into a warning
- force-push or branch replacement

## Required verification
1. Run the guard on the current schemas and show success.
2. Add/run focused mutation tests proving fail-closed behavior for at least:
   - deleted/renamed shared table produces the intended actionable diagnostic;
   - whitespace/indentation variation does not overrun into unrelated declarations;
   - a `pgTable` third-argument callback closes correctly;
   - a real column/type/default/nullability divergence fails;
   - malformed/unclosed declaration fails with a distinct diagnostic;
   - comments or formatting covered by the declared normalization contract behave as documented.
3. Run YAML validation for the workflow.
4. Run the narrowest relevant workspace typecheck/build confirming the InteractionHistory relocation remains valid.
5. Inspect current failing workflows. Classify each as patch-caused, base/stale, dependency-wide, or unverified. Fix only patch-caused failures within allowed scope.
6. Publish the commit to the existing GitHub PR branch and report the new remote SHA. A local commit or `make_pr` metadata is not completion.

## Acceptance criteria
- A new GitHub-visible head exists on PR #267.
- The two substantive review defects are fixed and regression-tested.
- The gate remains fail-closed with actionable diagnostics.
- Workflow triggering is internally consistent and does not create an impossible required-check state.
- InteractionHistory port relocation still builds/typechecks and remains behavior-preserving.
- No unrelated/security-sensitive surfaces are changed.
- Return exact commands, results, changed files, new remote SHA, and residual red workflows with evidence.

## Non-claims
- Do not claim all repository CI is green unless the current remote head proves it.
- Do not claim dependency-audit findings are unrelated without inspecting them.
- Do not claim publication from a local commit.
- Do not mark ready for audit until the remote head and current-head evidence are re-fetched.

## Return format
- `PUBLISHED` or `PUBLISH_BLOCKED`
- old head → new remote head
- changed files
- commands and exact results
- mutation-test matrix
- current workflow classification
- remaining risks/non-claims

---
Tomato Cognitive Controller #420

eileen1337:
ENVIRONMENT_BLOCKED
DO_NOT_RETRY_UNTIL_NEW_EVIDENCE

PR `#267` remains on GitHub head `25601b05446b2b7175a70c9671785e871fa5a8d8`.

The bounded Codex continuation returned `PUBLISH_BLOCKED`. It reported local commit `6341df2f17519003a14a30ac38c48a794915c4a5`, but the remote branch did not move because both GitHub API access and `git push` failed with `CONNECT tunnel failed, response 403`.

The local patch and reported mutation results are not publication evidence and are not eligible for audit. Do not invoke Codex again for this campaign fingerprint until at least one of these changes:

- the PR receives a new GitHub-visible head;
- a separate published PR carries the corrective patch;
- a demonstrably branch-attached, push-capable execution environment is available;
- GitHub network access materially changes; or
- the current review/workflow evidence changes enough to define a different campaign fingerprint.

The InteractionHistory port relocation remains the sound half of the PR. The drift guard remains blocked by the two current correctness findings until a fix is published and independently revalidated.

---
Tomato Cognitive Controller #420

PR #268, opened 2026-06-08
feat(shared/evals): rail-hardness + genome-integrity evals as CI ratchet gates (bd-uze Track B)

zkSoju:
## evals-in-CI — the scoreboard becomes a ratchet gate (bd-uze Track B)

Makes the environment gradient-health board + the genome-chain integrity contract **regression-proof in CI**. Pairs with **loa-constructs#256** + **construct-rooms-substrate#24**.

### What
- **rail-hardness eval** (`rail-hardness-grader.sh` + task + `rail-hardness.baseline.json`): FAILS if alignment < floor (53%) OR over_hardened > 0 OR a `hook_backed` claim doesn't resolve to a real file. Baseline **ratchets up-only, operator-only** — landing a real rail-move raises the floor; a PR that drops alignment goes RED.
- **genome-integrity eval** (`genome-integrity-grader.sh` + task + `evals/fixtures/genome/`): the genome-chain verifier must accept a valid chain AND reject a tampered one. Verifier is `genome-chain.py` **vendored byte-identical** from loa-constructs (jcs is identical across repos).
- **eval.yml**: added `.claude/hooks/**` to triggers (the board verifies hook-path claims — a deleted hook must trigger the scan).

### Observability
Both tasks auto-run via the framework suite; the alignment % is the rail-hardness task score, surfaced in the eval PR comment — a drop shows as a failing task with `alignment N% < floor 53%`.

### Tested (proven to go RED)
rail-hardness: pass(53) · FAIL on alignment-drop-to-47 · FAIL on roleplayed hook. genome-integrity: pass(100) · FAIL when the verifier is sabotaged to accept a forged chain.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #275, opened 2026-06-08
chore(shared): land 3 stranded deltas from impl/genome-evals-ci

zkSoju:
## What

Lands three pieces of real work that were authored on the stale
`impl/genome-evals-ci` branch but never reached `main` (local-only — not on
any remote). Surfaced while cleaning that branch up after the Loa 1.173.0
framework update. The rest of that branch was already on `main` via squash-PRs
(#261–#268); these three were the genuine remainder.

## Commits (attribution preserved via cherry-pick)

| Commit | What |
|--------|------|
| `refactor(platform/core)` | Relocate `InteractionHistory` port to `@freeside/core` — closes a platform→sietch reach-in (arrakis-shqm). Coupled: port file + `ports/index.ts` export + adapter import + `capability-mesh.ts`. |
| `feat(shared/coherence)` | worker↔sietch Drizzle schema drift-guard (arrakis-pcpc / F5B) — CI workflow + `tools/check-worker-schema-drift.sh`. |
| `docs(shared)` | `decisions/EXCEPTIONS.md` — the ADR exceptions ledger that CLAUDE.md Hard rule 6 + `path-domain-check.yml` already reference as if it exists. Closes that doctrine ghost. |

## Scope / firewall

Platform + shared paths only — no network paths, so the path-domain firewall
is not tripped. Net: 7 files, +177/-16.

## Note

Pre-existing CI reds (Agent Subsystem CI `ERR_PNPM_OUTDATED_LOCKFILE`, E2E
Tests) are chronic on `main` and unrelated to this change.


zkSoju:
## Revived via merge of `main` (744a6557) — 2 of 3 deltas survive

**Conflict resolution** — the only conflict was `decisions/EXCEPTIONS.md` (add/add). Main already carries the full ledger (EXC-001), so **main's version was taken wholesale**; that delta is dropped from this PR.

**Surviving deltas** (verified still absent from / live on `main`):

1. **InteractionHistory port relocation → `@freeside/core/ports`** — the reach-in is still live on main: `packages/adapters/storage/audit-query-service.ts:14` imports via `../../../themes/sietch/src/packages/core/protocol/capability-mesh.js` (the only live `themes/sietch` import from `packages/`). This PR moves the port to `packages/core/ports/interaction-history.ts` and repoints the consumer; `capability-mesh.ts` re-exports for backward compat.
2. **`tools/check-worker-schema-drift.sh` + `.github/workflows/worker-schema-drift-check.yml`** — absent from main.

**Verification** (mirrors the `pr-validation.yml` gate: core build → adapters build → sietch typecheck):

- `packages/core`: `npm run build` + `typecheck` clean; `dist/ports/interaction-history.{js,d.ts}` emitted.
- `packages/adapters`: `npm run build` + `typecheck` clean (package tsconfig scope). Targeted `tsc --noEmit` over `storage/` (excluded from the package tsconfig): the branch removes a `TS6059` (capability-mesh not under `rootDir`) that **main's version produces** — remaining `storage/` errors (`pg` types etc.) are pre-existing, identical on main.
- `themes/sietch`: `tsc -p tsconfig.production.json --noEmit` — zero errors from PR files incl. `capability-mesh.ts`. (One env-only `TS2307 sharp` in untouched `src/utils/image.ts`: `sharp` is an optionalDependency whose platform binary didn't install in the sandbox; file is byte-identical to main.)
- Runtime: `tests/unit/audit-query-service.test.ts` — **14/14 passed** through the relocated port.
- `tools/check-worker-schema-drift.sh` runs green on the merged tree: worker↔sietch in sync across communities/profiles/badges.

Merge commit, no force-push. Not merging — leaving for review.

zkSoju (commented):
## Summary
This PR lands three well-scoped deltas: a schema-drift CI guard, a port relocation into core, and a reach-in import cleanup. The architecture moves are excellent — the type relocation and package-subpath import restore correct dependency direction. The main risk is concentrated in the drift-check shell script, whose reliability rests on unenforced formatting assumptions.

The most interesting tension here is a classic one: a *guard* that silently passes is worse than no guard at all. Two of the shell findings (F1, F2) describe ways `check-worker-schema-drift.sh` can report "in sync" while real drift exists, or report drift where none exists. A check you can't trust becomes noise the team learns to ignore.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Block terminator `/^\\);/` is fragile to Drizzle table formatting",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/check-worker-schema-drift.sh",
      "description": "extract_normalized() ends a table block on the first line matching `/^\\);/`. Prettier-formatted Drizzle tables often close on `});` or `}));` when an extraConfig/index callback is present. Neither starts with `)`, so the exit never fires and awk prints to EOF, yielding false-positive drift or comparison of unrelated tails. Reliability depends on a multi-line closing style enforced nowhere.",
      "suggestion": "Terminate on the balanced close: match `/^\\}?\\)+;/` (covers `);`, `});`, `}));`) or track paren/brace depth and exit at depth zero. Add a self-test asserting each extracted block is non-empty AND under ~200 lines to catch runaway extraction.",
      "confidence": 0.45,
      "faang_parallel": "Google's readability tooling treats brittle regex parsing of structured code as a known anti-pattern — their internal linters moved to AST-based analysis precisely because line-prefix heuristics break the moment a formatter changes wrapping. When you must regex, you anchor on the balanced delimiter, not a cosmetic prefix.",
      "metaphor": "It's like a smoke detector wired to trigger only on the smell of *one specific brand* of smoke. Switch brands and the house can burn while the alarm stays silent.",
      "teachable_moment": "A guard's failure mode matters more than its success case. Always ask: 'When my parser is confused, does it fail loud or fail quiet?' Fail-quiet guards decay into false confidence. Bound the output (length assertion) so confusion becomes a visible error.",
      "connection": "Ties to F2 — both are silent-failure paths in the same script. Together they argue for a self-test harness with fixture schemas exercising each closing style."
    },
    {
      "id": "F2",
      "title": "`//` comment stripping can mangle string literals and mask real drift",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/check-worker-schema-drift.sh",
      "description": "`sed -E 's://.*$::'` deletes everything after the first `//`. A default/check containing a URL like `'https://…'` is truncated at `https:`. Because both files normalize identically this usually hides identical content, but it also collapses two *different* `//`-bearing values to the same prefix, so genuine divergence in such a field goes undetected — the guard reports 'in sync' on real drift.",
      "suggestion": "Strip only pure comment lines (`s:^//.*$::` after trim) or ` //` preceded by whitespace. Since table blocks rarely carry leading-`//` comments, consider dropping comment-stripping entirely and relying on whitespace/blank-line normalization.",
      "confidence": 0.5,
      "faang_parallel": "This is the URL-in-a-comment-stripper bug that has bitten many sanitizers; the classic large-scale echo is the `//`-vs-scheme confusion in early URL parsers. Netflix and others learned to normalize with grammar-aware passes, not greedy substring cuts.",
      "metaphor": "Redacting a document with a marker that bleeds through — you meant to hide the footnote, but you also erased the phone number in the paragraph above.",
      "teachable_moment": "Normalization that changes *meaning* isn't normalization. A safe normalizer is idempotent and information-preserving on the fields you compare; if it can conflate two distinct values, it can hide the exact divergence you built it to catch.",
      "connection": "Same silent-drift class as F1. A fixture with a `.default('https://x')` column would regression-test both."
    },
    {
      "id": "F3",
      "title": "push trigger paths omit the script and workflow file",
      "severity": "LOW",
      "category": "ci",
      "file": ".github/workflows/worker-schema-drift-check.yml",
      "description": "The pull_request trigger watches four paths including the script and workflow itself; the `push: branches:[main]` trigger only watches the two schema.ts files. A direct push to main that changes the drift script won't re-run it on main. Minor coverage asymmetry.",
      "suggestion": "Mirror the same four paths under the push trigger, or document why main-push only cares about the schema files.",
      "confidence": 0.7,
      "faang_parallel": "CI trigger asymmetry is a recurring incident source; teams at scale (e.g. Amazon's pipeline reviews) standardize path filters across event types precisely because the divergence is invisible until an unguarded push slips through.",
      "metaphor": "Two doors into the same room, but only one has a metal detector. Everyone learns which door to use.",
      "teachable_moment": "When the same invariant must hold under multiple triggers, define the path set once and reference it — divergence between near-identical config blocks is where guards quietly develop blind spots.",
      "connection": "Reinforces the F1/F2 theme: this guard's trustworthiness is only as strong as its weakest entry path."
    },
    {
      "id": "F4",
      "title": "Verify `@freeside/core/ports` subpath is exported by core's package.json",
      "severity": "SPECULATION",
      "category": "build",
      "file": "packages/adapters/storage/audit-query-service.ts",
      "description": "The import moves to the specifier `@freeside/core/ports`, which resolves only if core's package.json `exports` (or tsconfig paths) publishes `./ports`. index.ts already aggregates other ports, suggesting validity, but the manifest isn't in the diff so it can't be confirmed here.",
      "suggestion": "Confirm `packages/core/package.json` exports includes `./ports` and the build emits `ports/interaction-history.js`, so both theme and adapter resolve the specifier at build and runtime.",
      "confidence": 0.3,
      "teachable_moment": "A green typecheck under path aliases doesn't guarantee runtime resolution — bundlers and Node resolve `exports` maps, not tsconfig paths. Verify the subpath in the manifest, not just the editor.",
      "connection": "Precondition for F6's architectural win to actually hold at runtime."
    },
    {
      "id": "F5",
      "title": "Backward-compatible type re-export preserves existing importers",
      "severity": "PRAISE",
      "file": "themes/sietch/src/packages/core/protocol/capability-mesh.ts",
      "description": "Relocating InteractionRecord/InteractionHistoryProvider to core while keeping `export type { ... }` from capability-mesh.ts means existing importers keep compiling. This is the right way to move a type without breaking churn.",
      "suggestion": "Optionally add `@deprecated import from @freeside/core/ports` on the re-export so callers migrate and the shim can eventually be removed.",
      "confidence": 0.85,
      "faang_parallel": "This is the migration discipline behind large-scale refactors at Google — move the definition, leave a compatibility shim, then land call-site migrations incrementally. TypeScript's `export type` re-export is the zero-runtime-cost version of that shim.",
      "metaphor": "Forwarding your mail before you tell everyone the new address — nothing bounces while people update their address books.",
      "teachable_moment": "Decouple the *move* from the *migration*. A shim converts one risky atomic change into two safe reversible ones, and a `@deprecated` tag gives the shim an exit plan instead of letting it become permanent.",
      "connection": "The clean counterpart to F6 — together they make the port relocation both correct and non-disruptive."
    },
    {
      "id": "F6",
      "title": "Deep cross-membrane reach-in import replaced with package subpath",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/adapters/storage/audit-query-service.ts",
      "description": "Replacing `../../../themes/sietch/src/packages/core/protocol/capability-mesh.js` with `@freeside/core/ports` removes a platform-adapter → theme dependency and restores the platform-core ← theme direction. The core-hosted port is the correct single source of truth for a contract shared by a producing theme and a backing adapter.",
      "suggestion": "No change needed. Consider a dependency-cruiser / eslint no-restricted-imports rule banning `themes/` paths from `packages/adapters/**` to prevent recurrence.",
      "confidence": 0.85,
      "faang_parallel": "This is the Ports & Adapters (hexagonal) boundary Netflix and many platform teams enforce with layering lint (ArchUnit / dependency-cruiser): adapters depend inward on core contracts, never outward on application/theme code.",
      "metaphor": "You'd never wire a building's electrical panel to draw from a single tenant's apartment. The panel (core) defines the sockets; tenants (themes) plug in.",
      "teachable_moment": "Deep relative `../../../` imports are architecture smells with a fix built into the tooling — encode the boundary as a lint rule so the compiler defends the design instead of relying on reviewer vigilance.",
      "connection": "Depends on F4 resolving true; enforced going forward by the suggested import-restriction rule."
    }
  ]
}
```


## Callouts

Two genuinely excellent moves here, and they reinforce each other. **F6** restores the intended dependency belt — adapters reaching *inward* to a core-hosted contract instead of *outward* into a theme — which is the exact hexagonal-architecture invariant that keeps a platform composable as themes multiply. **F5** shows the discipline to make that move *safely*: the `export type` shim means not a single existing importer breaks, converting a scary atomic refactor into two reversible steps. That pairing — correct direction *and* zero-churn migration — is the mark of someone who has felt the pain of a big-bang rename before.

The architectural meditation worth carrying forward: this PR is really two stories. The refactor story (F4–F6) is exemplary and near-done. The *guard* story (F1–F3) is where the attention belongs, because a schema-drift check is a promise — "I will catch divergence" — and F1/F2 describe conditions where it silently breaks that promise. Before this guard earns the team's trust, give it a tiny fixture-based self-test exercising `});`/`}));` closings and a URL-bearing default. A guard that has been shown to fail loudly is worth ten that merely look green.

Verdict: **COMMENT** — the refactor is clean; the shell guard needs hardening before it's load-bearing.

PR #279, opened 2026-06-11
Add audit-membrane-integrity composition (the epistemic immune-system audit)

zkSoju:
Sibling of `audit-geometry-fidelity` (which audits decision-fork RAILS); this audits the **membrane** — the two content flows across the Markov blanket.

- **Afferent** (inward): external content must be stamped with provenance at entry (attested-by-X, not truth).
- **Efferent** (outward): dead referents must be expelled early, by physics.

GECKO grounds both flows, WORLDLINE traces the laundering paths, GYGAX designs the cost-flip, the operator stamps, the synthesizer expels/wires/builds/proves.

First run (`membrane-20260610`, deep stages on **Fable**): `valid_run`. Shipped the afferent stamp-on-ingest hook + the efferent ghost-gate hardening (construct-compositions). Full record: `grimoires/loa/context/2026-06-10-membrane-integrity.md` (local State zone).


zkSoju (commented):
## Summary

This PR ships an epistemic immune-system audit — and, with poetic irony, exhibits several of the diseases it hunts: a 61-file train titled as a single composition, declared machinery that may not resolve, and a defi-strict surface class without the enforcement block that makes it real. The test engineering is genuinely strong (the bug-989 glue-state regression is exemplary), but the scope mixing and the unverified `max→fable` tier retarget are the items that deserve resolution before merge.

**Verdict: COMMENT** — no exploitable or data-loss findings, but SCOPE-001 and TIER-001 warrant a deliberate decision rather than a drive-by merge.

### Architectural meditation

A composition that audits membrane integrity must itself have an intact membrane. Compilers are validated by self-hosting; auditors earn trust by passing their own audit. Three findings here (DEP-001, ROUTE-001, SCHEMA-001) are the membrane audit failing self-hosting in miniature: cross-repo script paths that can't resolve locally, a surface classification with no wired enforcement, and a schema that doesn't require the very field — `physics_or_prose` — that names the audit's central axis. None are fatal; all are teachable, because the fix in each case is the composition's own doctrine: turn prose into physics.

## Findings


{
  "schema_version": 1,
  "verdict": "COMMENT",
  "findings": [
    {
      "id": "SCOPE-001",
      "title": "PR titled as a single composition add is actually a 61-file mixed train",
      "severity": "REFRAME",
      "category": "process",
      "file": "compositions/discovery/audit-membrane-integrity.yaml",
      "description": "The title says 'Add audit-membrane-integrity composition', but the diff also contains: a second composition (audit-geometry-fidelity), a model-routing retarget (max tier opus→fable in model-config.yaml + cycle-114-tier-effort-hints.bats), three CI workflow fixes (secret-scanning base/head, bats Python deps, post-merge symlink reconcile), the mount-submodule copy-set refresh (#968), kernel hash stamping (#989), a new cheval metering rollup module + tests, a large CLAUDE.loa.md kernel restructure (-198 lines into reference files), and .loa.config.yaml changes. Reviewing, bisecting, or reverting any one of these independently is impossible at this granularity.",
      "suggestion": "Either retitle the PR to reflect its real scope (a cycle rollup) or split the independently-revertable concerns — at minimum the model-routing retarget and the CI workflow fixes — into their own PRs.",
      "confidence": 0.7,
      "faang_parallel": "Google's engineering practices doc ('Small CLs') and the Linux kernel's one-logical-change-per-patch discipline exist for the same reason: review quality degrades superlinearly with diff size, and `git revert` granularity is fixed at merge time, forever.",
      "metaphor": "This is a moving truck labeled 'one lamp' — the lamp is real and lovely, but the truck also contains the piano, the water heater, and the routing table for the whole house.",
      "teachable_moment": "Revertability is the true unit of review. The question isn't 'can a reviewer read 61 files?' — it's 'when the tier retarget needs rolling back at 2am, can it be reverted without also reverting two compositions and five bug fixes?' Here it cannot.",
      "connection": "If this is intentionally a cycle rollup (the branch history suggests it is), the cheap fix is honest titling — the expensive failure is a future bisect landing on this commit."
    },
    {
      "id": "TIER-001",
      "title": "max tier retargeted opus→fable inside an unrelated-titled PR; routability across all consumers unverified",
      "severity": "MEDIUM",
      "category": "config",
      "file": "tests/unit/cycle-114-tier-effort-hints.bats",
      "description": "cycle-114-tier-effort-hints.bats:44-48 now pins .tier_groups.mappings.max.anthropic == 'fable' (was 'opus'). The tier SoT is consumed by every max-tier dispatch path (gates, compose emitter, council routing). A recent commit on this branch ('fix(config): BB back to claude-headless pending canonical fable routing — operator correction', 98d316e8) indicates fable routing was NOT yet canonical at least for one consumer. The model-config.yaml diff is truncated here, so whether every adapter (headless, bedrock, streaming) resolves 'fable' and whether pricing is registered (the $0-metered blind spot the new rollup tests themselves pin) cannot be confirmed from this PR.",
      "suggestion": "Before merge, enumerate every consumer that requests tier 'max' and prove each can route 'fable' (dispatch test per adapter), and confirm a pricing entry exists so calls are not metered at $0. If any consumer still cannot route fable, keep max→opus and land the retarget in its own PR with the consumer audit attached.",
      "confidence": 0.5,
      "faang_parallel": "Knight Capital's 2012 collapse was a config/routing change deployed without verifying every consumer honored it — one server ran old behavior. Google's response to this class is treating config as code: canaried, with per-consumer verification before fleet-wide flip.",
      "metaphor": "This changes which engine the entire fleet's 'maximum thrust' lever connects to — announced in a footnote of a memo about cabin lighting. The lever works in the test hangar; nobody has confirmed every aircraft model has the new engine installed.",
      "teachable_moment": "A tier indirection layer ('max' → model) is exactly as safe as the weakest consumer's resolution path. The branch's own history contains the counter-evidence: an operator correction reverting one consumer off fable. That correction is the strongest signal in this PR that the SoT flip is premature for at least some paths.",
      "connection": "The new rollup module in this same PR pins the $0-metered blind spot — apply that same test to fable itself: an unpriced model routed at max tier is invisible cost."
    },
    {
      "id": "DEP-001",
      "title": "Compositions depend on cross-repo scripts and possibly-unregistered constructs that cannot resolve from this repo",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "compositions/discovery/audit-membrane-integrity.yaml",
      "description": "depends_on.scripts lists construct-compositions/scripts/check-ghost-refs.py and construct-worldline/scripts/claim-grade.mjs; the geometry sibling's review_routing.runner points at construct-fagan/scripts/cheval-council.sh. None of these paths exist inside loa-freeside — they are sibling-repo paths. Similarly, depends_on.constructs includes 'worldline' and 'gygax', and stage 1 invokes gecko skill 'diagnose'; whether these resolve in the construct registry is not verifiable from this diff. An audit composition whose own declared machinery may not resolve is precisely the GHOST-HOOK / unresolvable-referent class these two audits hunt.",
      "suggestion": "Add a composition-dependency resolution lint (constructs resolve via the registry, script paths exist or are explicitly marked cross-repo with a resolution root) and run it in CI over compositions/. At minimum, document the expected resolution root for cross-repo script paths in the composition schema.",
      "confidence": 0.5,
      "faang_parallel": "Bazel's core contract is that undeclared or unresolvable dependencies fail at analysis time, not execution time — Google learned that 'works on the machine that happens to have the sibling checkout' is the most expensive class of build flake.",
      "metaphor": "An audit composition with unresolvable dependencies is a fire-safety inspector whose own evacuation map points to doors in the building next door.",
      "teachable_moment": "Self-hosting is the cheapest credibility test an audit tool has: run the membrane-integrity audit's own dependency declarations through a resolution check before it audits anything else. If the schema can't express 'cross-repo path with resolution root', that's the schema gap to fix first.",
      "connection": "This is the deployed-but-unconsumed pattern inverted: declared-but-unresolvable. Both are membrane defects — exactly what this composition was written to detect."
    },
    {
      "id": "ROUTE-001",
      "title": "audit-membrane-integrity declares defi-strict surface class but omits the review_routing block its sibling pairs with it",
      "severity": "MEDIUM",
      "category": "config",
      "file": "compositions/discovery/audit-membrane-integrity.yaml",
      "description": "audit-geometry-fidelity.yaml pairs surface_class.default: defi-strict with a review_routing block (mandatory_when: [defi-strict], cheval router, min_surviving_voices: 2, modelinv-required). audit-membrane-integrity.yaml declares the same defi-strict surface class but has no review_routing block at all. If review_routing is the mechanism that enforces the mandatory multi-model council for defi-strict surfaces, the membrane audit silently never gets the adversarial review its own surface class implies.",
      "suggestion": "Either add the equivalent review_routing block to audit-membrane-integrity.yaml, or document why the WORLDLINE stage-2 verification is considered the substitute for the cheval council on this composition.",
      "confidence": 0.55,
      "faang_parallel": "AWS distinguishes IAM policy (declaration) from enforcement points (evaluation) — a resource tagged 'restricted' with no policy keyed to the tag is unprotected while reading as protected. Declared classifications without bound mechanisms are the audit-failure shape SOC 2 assessors hunt first.",
      "metaphor": "A vault door labeled 'maximum security' with no guard on the schedule. The label is sincere; the schedule is what stops intruders.",
      "teachable_moment": "In this codebase's own vocabulary: the surface_class declaration is prose, the review_routing block is physics. The two siblings shipped in the same PR disagree about whether defi-strict implies the council — that divergence will be read as precedent by the next composition author.",
      "connection": "If WORLDLINE stage-2 verification is the intended substitute, one comment line in the YAML converts this from a silent gap into a documented decision."
    },
    {
      "id": "PYC-001",
      "title": "Compiled Python bytecode (__pycache__/*.pyc) is tracked in git and churning in this PR",
      "severity": "MEDIUM",
      "category": "repo-hygiene",
      "file": ".claude/adapters/loa_cheval/__pycache__/types.cpython-312.pyc",
      "description": "The PR's changed-files list includes multiple .claude/adapters/loa_cheval/__pycache__/*.pyc files as modified, meaning bytecode is tracked in git. Bytecode is interpreter-version-specific, regenerates on every import, bloats every diff, and can silently desync from its .py source. The working tree shows many more untracked __pycache__ directories accumulating in the same subtree.",
      "suggestion": "Add **/__pycache__/ and *.pyc to .gitignore, `git rm --cached` the tracked bytecode, and verify nothing imports the .pyc files directly.",
      "confidence": 0.85,
      "faang_parallel": "Google's monorepo bans derived artifacts from source control outright — derived data is the build system's job, and tracking it creates two sources of truth that drift. Python's own CPython repo has ignored *.pyc since its first .gitignore; CPython will also silently execute a stale tracked .pyc if the .py is missing, which is how bytecode desync becomes a runtime behavior bug, not just diff noise.",
      "metaphor": "Shipping the baked cake alongside the recipe in every delivery. Eventually someone edits the recipe, the cake stays stale, and a customer eats the old cake without anyone noticing the mismatch.",
      "teachable_moment": "Track sources, derive artifacts. The subtle risk isn't repo bloat — it's that a tracked .pyc is executable state that can shadow its source. For an adapter layer that routes model dispatch (loa_cheval), 'which code actually ran' should never have two answers.",
      "connection": "Note this PR's diff churns these files for the third-plus time (git status shows the same subtree perpetually dirty) — the .gitignore fix also eliminates a standing source of accidental-commit noise in every future PR."
    },
    {
      "id": "SCHEMA-001",
      "title": "physics_or_prose is mandated by stage-1 prose but absent from the schema's required list",
      "severity": "LOW",
      "category": "schema",
      "file": "compositions/discovery/audit-membrane-integrity.yaml",
      "description": "Stage 1 notes instruct 'For EACH finding emit a row: {flow, mechanism, where, exists, wired, physics_or_prose, gap_class, severity, evidence}', but the output_schema's required array omits physics_or_prose. Rows may legally omit the field that distinguishes physics from prose — the central axis of this audit — and downstream stages (GYGAX's cost-flip design keys on it) would consume rows missing it without a validation failure.",
      "suggestion": "Add physics_or_prose to the required array of the stage-1 rows item schema, or explicitly mark it optional in the notes if omission is intended for NONE-gap rows.",
      "confidence": 0.6,
      "faang_parallel": "This is the protobuf required/optional lesson in miniature: when a downstream consumer branches on a field, the schema must make its absence a validation error, or absence becomes a silent third value ('unset') that no branch handles.",
      "metaphor": "A customs declaration form where the 'contents' box is optional — the inspection process downstream assumes every crate is labeled.",
      "teachable_moment": "The schema's required list IS the contract; the notes are advisory prose. An audit whose organizing axis is physics-vs-prose should make its own physics_or_prose field physics, not prose. The recursion is the lesson.",
      "connection": "Same physics/prose gap shape as ROUTE-001, one level down — declaration in notes, enforcement absent in schema."
    },
    {
      "id": "PORT-001",
      "title": "GNU-only sed -i and md5sum in bug-989 tests fail on macOS (the primary dev platform)",
      "severity": "LOW",
      "category": "portability",
      "file": "tests/unit/bug-989-kernel-hash-stamp.bats",
      "description": "The DISS-001 test uses `sed -i '1s/.../' file` — BSD sed (macOS) requires an explicit backup-suffix argument after -i, so this errors locally. The idempotency test uses `md5sum`, which is absent on stock macOS (it ships `md5`). CI on ubuntu passes, but local bats runs on darwin will fail these tests with errors unrelated to the contract being pinned.",
      "suggestion": "Use a portable in-place edit (e.g., write via a temp file) or detect sed flavor; replace md5sum with `shasum -a 256` or `cksum`, both present on macOS and Linux.",
      "confidence": 0.7,
      "faang_parallel": "Homebrew and Chromium both maintain explicit BSD-vs-GNU sed shims because this exact `-i` incompatibility is among the most common cross-platform shell breaks. The deeper cost is what Google's testing blog calls test-suite erosion: once developers learn 'those two always fail locally', they stop reading local failures at all.",
      "metaphor": "A smoke detector that false-alarms every Tuesday — within a month, nobody looks up when it goes off, including the day it's real.",
      "teachable_moment": "The test harness's portability is part of the contract it pins. A test that errors on the operator's own platform (darwin, per this repo's environment) trains the habit of distrusting red — the most expensive habit a test suite can teach.",
      "connection": "shasum -a 256 also aligns the test with the sha256 the kernel-hash feature itself uses — one fewer hash tool in the story."
    },
    {
      "id": "DOC-001",
      "title": "Geometry composition's scope input describes per-altitude construct lenses the chain never invokes",
      "severity": "LOW",
      "category": "documentation",
      "file": "compositions/discovery/audit-geometry-fidelity.yaml",
      "description": "The scope input description maps each altitude to construct lenses (operator=gecko+the-arcade, hooks=gecko+scar, harness=gygax+fagan, repo=protocol+rosenzu, CI+evals=noether+worldline), but the chain only runs gecko, fagan, and general-purpose stages, and depends_on.constructs lists only gecko and fagan. A reader (or a future emitter honoring declared lenses) will expect council diversity the composition does not actually wire.",
      "suggestion": "Either trim the description to the constructs actually staged, or note explicitly that the lens mapping is aspirational guidance for the GECKO grounder rather than wired stages.",
      "confidence": 0.5,
      "faang_parallel": "Amazon's working-backwards PR/FAQ docs carry an explicit rule that the doc must be re-trued when the build diverges — because in doc-driven systems, stale declarations don't just mislead humans, they get executed the day tooling starts honoring them.",
      "metaphor": "A theater program listing eight actors when three are in the cast. Tonight it's just disappointing; the night the director starts paying salaries off the program, it's a budget incident.",
      "teachable_moment": "In a system where the /compose emitter mechanically honors declared construct capabilities, descriptive YAML is one emitter-feature away from being executable. Write descriptions as if a machine will route on them — because here, one eventually will.",
      "connection": "Same declared-versus-wired axis as DEP-001 and ROUTE-001, at documentation severity."
    },
    {
      "id": "DEAD-001",
      "title": "Dead code in new bats tests: unused _in_fixture helper and unused names variable",
      "severity": "LOW",
      "category": "testing",
      "file": "tests/unit/bug-968-reconcile-copy-refresh.bats",
      "description": "bug-968-reconcile-copy-refresh.bats defines `_in_fixture()` (lines 26-29) which no test calls — every test inlines the same `bash -c \"cd ... && source ... --source-only && ...\"` pattern instead. bug-886-bats-tests-python-deps.bats assigns `names=$(yq ...)` in the ordering test and never reads it.",
      "suggestion": "Delete the unused helper and variable, or refactor the repeated inline bash -c invocations to actually use the helper.",
      "confidence": 0.8,
      "faang_parallel": "Google's ErrorProne and Go's unused-variable compile error encode the same judgment: unused code in tests is worse than in prod, because tests are the documentation people trust most — a helper that looks load-bearing but isn't actively misleads the next author.",
      "metaphor": "An emergency exit that was built, signposted, and then bricked over — its presence on the floor plan is now the hazard.",
      "teachable_moment": "Prefer the refactor over the deletion here: the inlined bash -c pattern repeats across every test, which is exactly the duplication the dead helper was written to absorb. The helper isn't wrong — it's unconsumed.",
      "connection": "A two-line instance of the repo's own deployed-but-unconsumed pattern, inside the test file."
    },
    {
      "id": "TEETH-001",
      "title": "Live-repo state assertion in bug-989 will fail far from its cause with an opaque message",
      "severity": "LOW",
      "category": "testing",
      "file": "tests/unit/bug-989-kernel-hash-stamp.bats",
      "description": "The final test asserts the live .claude/loa/CLAUDE.loa.md header has no PLACEHOLDER, is not version 1.94.0, and verify-hash returns VALID. This is intentional teeth, but any future legitimate edit to CLAUDE.loa.md that forgets re-stamping turns the unit suite red with bare `[[ ]]` failures that don't say what to do.",
      "suggestion": "Before the assertions, echo an actionable remediation (e.g., 'kernel hash stale — run: bash .claude/scripts/marker-utils.sh update-hash .claude/loa/CLAUDE.loa.md') so the failure is self-explaining.",
      "confidence": 0.4,
      "faang_parallel": "The Rust compiler team's error-message doctrine applies to tests too: an error should name the fix, not just the violation. Bazel's style guide for failure messages says the same — the reader at failure time has zero context from authoring time.",
      "metaphor": "A smoke alarm that beeps versus one that says which room — same detection, very different 3am.",
      "teachable_moment": "Worth verifying the failure UX before investing in the fix (confidence is moderate that bare [[ ]] is what surfaces) — but the principle holds: a teeth-test's assertion message is a runbook read exclusively by someone who didn't write the test.",
      "connection": "The remediation command already exists in this PR (marker-utils.sh); the fix is one echo line connecting the alarm to it."
    },
    {
      "id": "TMPL-001",
      "title": "Unverified <date> placeholder in composition output destinations",
      "severity": "SPECULATION",
      "category": "correctness",
      "file": "compositions/discovery/audit-membrane-integrity.yaml",
      "description": "Output destinations use grimoires/loa/context/<date>-membrane-integrity.md (and <date>-geometry-fidelity-audit.md in the sibling), while the verdict/signal outputs use the established <run_id> placeholder. If the compose emitter only substitutes <run_id>, the artifact lands at a literal '<date>-...' path.",
      "suggestion": "Confirm the emitter substitutes <date>; if not, switch to <run_id> or have the SYNTHESIZER stage compute the dated filename itself (its notes already instruct writing the synthesis).",
      "confidence": 0.3,
      "faang_parallel": "The classic mail-merge failure mode: 'Dear <name>' shipping to production. Template engines that silently pass through unknown placeholders (rather than erroring) make this a recurring class, which is why Jinja2's StrictUndefined exists.",
      "metaphor": "A letter addressed to '<recipient> Street' — deliverable nowhere, returned to no one.",
      "teachable_moment": "Low confidence — verify before acting: one grep of the emitter's substitution table settles whether <date> is supported. If it isn't, the better systemic fix is making the emitter fail loud on unrecognized placeholders, which retires this entire finding class.",
      "connection": "An emitter that errors on unknown placeholders is the physics version of this check; reviewer vigilance is the prose version."
    },
    {
      "id": "PRAISE-001",
      "title": "DISS-001 glue-state regression test pins a subtle false-VALID failure mode",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tests/unit/bug-989-kernel-hash-stamp.bats",
      "description": "The test that stamps a real hash, glues PLACEHOLDER back on, confirms verify-hash still reports VALID (hex-prefix comparison), and then requires the lint's residue check to WARN independently is excellent defensive layering — it pins exactly the state where the primary verifier lies and proves a second, independent signal catches it.",
      "suggestion": "None — keep this pattern for future verifier/lint pairs.",
      "confidence": 0.8,
      "faang_parallel": "This is the design principle behind Certificate Transparency: don't just verify the certificate, build an independent witness that catches the case where the verifier itself is deceived. Most teams test that the verifier passes good input and fails bad input; testing the state where the verifier passes BAD input and proving a second sensor fires is a tier above.",
      "metaphor": "Two locks keyed differently on the same door — the test doesn't just check the first lock turns, it picks the first lock and proves the second still holds.",
      "teachable_moment": "The repo's own hollow-hash incident (PLACEHOLDER matching nothing, verified by nothing) is exactly the disease this test immunizes against. Regression tests written from real incidents, pinning the precise deceptive state, are the highest-value tests a codebase accumulates.",
      "connection": "A template for every verifier/lint pair in the estate: enumerate the states where the verifier lies, and require an independent signal for each."
    },
    {
      "id": "PRAISE-002",
      "title": "Deletion-surface guard is itself under test",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tests/unit/bug-968-reconcile-copy-refresh.bats",
      "description": "_refresh_copy_entry performs rm -rf on its destination, and the test suite explicitly proves the guard refuses non-.claude destinations and that the would-be victim directory survives. Testing the refusal path of a deletion surface (not just the happy path) is the right instinct for code that removes directories.",
      "suggestion": "None.",
      "confidence": 0.8,
      "faang_parallel": "The 2015 Steam-for-Linux bug — `rm -rf \"$STEAMROOT/\"*` with an empty variable deleting users' home directories — is the canonical case of an untested deletion guard. Asserting the victim directory SURVIVES (not merely that the function returned nonzero) tests the outcome, not the implementation.",
      "metaphor": "Testing the safety catch by pulling the trigger with the catch on — and then checking the target is intact, not just that the catch felt engaged.",
      "teachable_moment": "For any function whose failure mode is irreversible (rm -rf, force-push, DROP), the refusal path is the primary contract and the happy path is secondary. This suite has the priority right.",
      "connection": "Mirrors the repo's block-destructive-bash.sh philosophy — guards on deletion surfaces, with the guard itself under test."
    },
    {
      "id": "PRAISE-003",
      "title": "Failure-mode policy encoded as regression guards (no '|| true', SHA-pinned actions)",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tests/unit/bug-986-post-merge-symlink-reconcile.bats",
      "description": "bug-986 pins that the reconcile step must not carry '|| true' (partial reconcile fails loud, referencing #660), and bug-992/bug-886 reject mutable action refs by asserting 40-hex SHA pins. Encoding operational policy — fail-loud and supply-chain pinning — as executable shape tests rather than review-time convention is genuinely good engineering.",
      "suggestion": "None.",
      "confidence": 0.7,
      "faang_parallel": "OSSF Scorecard checks SHA-pinned action refs for the same supply-chain reason (mutable tags were the vector in the 2025 tj-actions compromise), and the no-||-true guard is the SRE book's 'crash early' encoded as a test. Policy that lives in tests survives reviewer turnover; policy that lives in review culture doesn't.",
      "metaphor": "Writing the safety rule into the building code instead of the staff memo — new contractors comply without ever having read the memo.",
      "teachable_moment": "Shape tests (asserting what the code must NOT contain) are the cheapest enforcement tier: no runtime, no fixtures, just a contract grep with a citation to the incident that motivated it. The #660 reference in the test is the breadcrumb pattern every shape test should carry.",
      "connection": "These are physics in this PR's own taxonomy — the same policies stated as prose in CLAUDE.md would be defectable; as bats assertions they have teeth."
    }
  ]
}


## Callouts

The test engineering in this PR is its strongest dimension, and it's worth saying why rather than just that. **PRAISE-001** pins the state where the primary verifier *lies* and proves an independent sensor catches it — Certificate-Transparency-grade thinking applied to a kernel hash stamp. **PRAISE-002** tests a deletion surface by asserting the would-be victim *survives*, which is outcome-testing where most suites settle for return-code-testing. **PRAISE-003** converts operational policy (fail-loud, SHA-pinned supply chain) from review-time convention into executable shape tests with incident citations — the exact physics-over-prose move the membrane audit itself preaches.

That symmetry is the closing thought: this PR ships a composition that hunts ghost hooks, unresolvable referents, and prose masquerading as physics — and its tests embody that doctrine while its compositions partially violate it (DEP-001, ROUTE-001, SCHEMA-001). The fixes are small and the irony is instructive. The two items that deserve a decision before merge are the scope question (SCOPE-001) and the `max→fable` retarget (TIER-001), where the branch's own operator-correction commit is evidence the routing isn't canonical everywhere yet.

zkSoju:
## Re: BB review (Fable) — SCOPE-001 is the headline, and it's structural

BEAUVOIR is right: this PR is a **61-file train, not a single composition add**. Root cause confirmed by inspection — this branch was cut from local `main`, which is **18 commits ahead of `origin/main`** (unpushed framework work: the geometry-fidelity evals, fable-headless config, cheval metering, Bundles B/C, sprint fixes). The PR diffs all 18 + the one composition against the stale remote. The `.pyc` churn (PYC-001) and the `max→fable` retarget (TIER-001) and the bug-989 tests are all from those bundled commits, not the composition.

**This PR will not be merged as-is.** Resolving the local↔origin divergence (an operator decision) collapses it to a clean 1-file diff.

**Fixed on the composition itself:**
- **SCHEMA-001** (`ac3306ed`) — `physics_or_prose` (the audit's central axis) is now in the stage-1 `required` list, not just the prose.

**Composition-specific, noted:** ROUTE-001 (defi-strict without a `review_routing` block) and DEP-001 (cross-repo script paths) are real design considerations — the composition ran `valid_run` as-is, so they're not blocking, but worth folding in when this re-lands clean.

The rest (PYC-001, the bug-989 / deletion-guard / port findings) belong to the bundled main commits and will travel with them however the divergence is resolved.

zkSoju:
Superseded: local main (18 commits) is now pushed to origin/main, so this PR's stale merge-base no longer applies. The branch is unchanged; reopening a fresh PR computes the merge-base cleanly (1 file). The composition + SCHEMA-001 fix carry over.

PR #282, opened 2026-06-12
feat(dune-meter): cost-aware Dune adapter — the EXP-002 budget-blowout, made structurally impossible

zkSoju:
## what

`packages/dune-meter/` — a cost-aware Dune adapter (standalone, Asson-graduatable CLI). The metered pipe every future on-chain Dune query routes through, so an unbounded scan can't silently burn the budget again.

```
estimate <sql|query_id>   probe-heuristic, NEVER executes the target (query_id → cached metadata read)
run <sql> --cap <credits> Dune hard cost-cap + CostAtom emit + budget-refuse (exit 3) + cap-abort (exit 4)
budget                    spent / remaining / ceiling + recent atoms
```

## why

EXP-002 burned its Dune budget on two unbounded `evt_transfer` scans because **nothing estimated cost before executing**. This is that scar turned into a guard. Honest constraint baked in: Dune has **no native dry-run / EXPLAIN** — billing isn't previewable — so `estimate` is a labeled probe-heuristic and the *real* teeth are the per-query **cost-cap** + **budget-refuse**.

## proof

- ✅ built via `code-implement-and-review` (implement ↔ FAGAN), converged iteration 2 on merit
- ✅ **31/31 tests pass** offline (Dune client mocked — no credits spent to build the cost guard)
- ✅ `compose-verify-run dune-meter-0612a` → **valid_run** (`sha256:86fed996…`)
- FAGAN caught + closed the EXP-002 anti-pattern: estimating a `query_id` by executing it. Now non-executing.

## composes with

CostAtom emitter shape-compatible (not wire-compatible — honest deltas documented) with loa-finn `src/cost/cost-atom.ts` → one meter, two sources (harness tokens + Dune credits). Asson veve declares `determinism: network_io` (honest); GECKO `sense-runtime-fit` is the intended cost-drift sensor.


zkSoju:
## 🌉 Bridgebuilder Review — dune-meter

> Transport note: the API-key reviewer path was unavailable (`ANTHROPIC_API_KEY` unset + standing CLI-transport directive), so this review ran through an **independent model via CLI** (codex, subscription auth) — deliberately an outside corpus, since the build was orchestrated in-session. Scoped to the **real dune-meter diff** (10 files / 1,621 insertions), not the PR's raw 326-file diff.

### 🚧 Operational BLOCKER — this PR's base is wrong (fix before review/merge)
`feat/dune-meter-cost-adapter` was branched off `asson-cli-layer/cycle-1` but the PR targets `main`, so the diff is **326 files / +90k lines** (the whole asson-cli-layer branch + dune-meter), not the clean dune-meter change. **Fix:** either re-target this PR's base to `asson-cli-layer/cycle-1`, or cherry-pick the 3 dune-meter commits onto a fresh branch off `main`. Until then the PR isn't reviewable or mergeable as a unit.

---

REQUEST-CHANGES: the cost-safety shape is strong, but the current `estimate`/`run` paths still allow unmetered or uncapped Dune spend.

## SECURITY

**BLOCKER:** `src/dune-client.mjs:221-233` runs SQL estimate probes via `executeSql(countSql)` and `executeSql(sampleSql)` with no cost cap and no CostAtom/budget ledger update. A `COUNT(*) FROM (<unbounded evt_transfer scan>)` can still force the expensive scan. That reopens the EXP-002 class through the “safe” command: `dune-meter estimate <sql>` can spend credits outside the budget guard.

**HIGH:** `bin/dune-meter.mjs:138-140` never passes `flags.cap` to Dune. The CLI requires `--cap`, but execution only sends `{ performance }`; enforcement happens after the run by comparing returned credits at `:166`. If the account-level Dune cap is absent/misconfigured, the process can overspend before this tool knows. For a cost-safety adapter, “detect after spend” is not enough.

**MEDIUM:** `--force` bypasses both un-estimatable errors and REFUSE estimates (`bin/dune-meter.mjs:121-132`, `:134-136`). That may be intentional for break-glass, but it weakens the invariant “budget-refuse gate.” At minimum, make `--force` explicit in the CostAtom/gate_inputs and consider refusing when `--cap > remaining_budget`.

## QUALITY

**HIGH:** Cap-aborted accounting still writes spend before returning exit 4 when reported credits exceed cap (`bin/dune-meter.mjs:166-197`). That records the spend, which is good, but the result shape says `cap_exceeded: true` rather than a clear “execution violated local cap after the fact.” The WHY matters: this was not prevented by the tool, only observed.

**MEDIUM:** `resultMetadata()` treats missing datapoints as `0` (`src/dune-client.mjs:151-157`), and `cmdRun()` derives `0` credits if metadata is absent (`bin/dune-meter.mjs:157-163`). If Dune changes response shape or omits billing fields transiently, this under-records spend. Cost meters should fail closed when spend metadata is missing, not assume free.

**MEDIUM:** Ledger and atom writes are not transactionally coupled (`bin/dune-meter.mjs:186-187`). If `appendAtom()` succeeds and `recordSpend()` fails, budget and atoms diverge. If `recordSpend()` succeeded first in a future edit, the opposite failure appears. A small recovery check or “pending write” pattern would make 3am reconciliation less painful.

**PRAISE:** The non-executing saved-query estimate path is the right scar tissue: `latestResultMetadata()` uses cached result metadata and tests assert no `/execute` or `/status`. That directly protects the “never run a saved query just to estimate it” invariant.

## TEST COVERAGE

**BLOCKER:** Add a test proving SQL `estimate` cannot perform unbounded, unmetered spend. Today tests cover `probe(query_id)` non-execution, but not the SQL probe behavior. Given EXP-002 was SQL scan burn, this is the central missing test.

**HIGH:** Add a test that `run --cap N` actually sends/enforces a Dune-side cap or fails closed if the API cannot support one. Current tests only check local post-run `credits > --cap`.

**MEDIUM:** Add tests for missing/unknown result metadata fields. Expected behavior should be fail-closed, not `credits_consumed: 0`.

**PRAISE:** Offline test discipline is excellent. Injected client/fetch, no subprocess dependence, and explicit assertions that query_id estimates never POST are exactly the pattern this package needs.

## OPERATIONAL READINESS

**HIGH:** There is no startup/config check that the account-level Dune Query Cost Cap is enabled. The README tells humans to set it, but the CLI cannot verify or attest that the hard abort exists. If Dune has no API for this, the tool should print/emit an explicit `cap_attestation: unverified` and treat that as part of the operational contract.

**MEDIUM:** No structured audit event distinguishes normal run, forced run, estimate refusal, cap abort, metadata-derived credits, or ledger-chain mismatch. CostAtoms cover successful executions, but refusal/force decisions are also operationally important.

**LOW:** `budget` reports `chain_ok`, but does not surface malformed line count from `readAtoms()`. At 3am, “chain ok but malformed skipped” is a dangerous half-green state.

**PRAISE:** Corrupt budget ledger refuses to spend instead of resetting to full. That is exactly the failure posture a cost guard should have.

---
*Independent CLI review (codex) · BEAUVOIR persona · 4 dimensions. The estimate-probe BLOCKER is the load-bearing find — the "safe" `estimate <sql>` path is itself unmetered, which is the EXP-002 class it exists to close.*

zkSoju:
## 🌉 Bridgebuilder Review — dune-meter · **2nd voice: Fable (Claude) via CLI**

> Cross-model council: this is the **second independent corpus** on the same diff (first was codex/OpenAI, above). Both via CLI transport (subscription, no API key); Fable was NOT shown codex's findings — fresh eyes. Scoped to the real 10-file dune-meter diff.

### 🤝 Where the two corpora CONVERGE (→ rock-solid, not one model's quirk)
- **The `estimate` path fails open** — both flag it as the load-bearing BLOCKER. Fable pins the *mechanism* codex only gestured at: the SQL probe reads rows from the **status** endpoint, not `/results`, so missing fields → `rows=0`, cols `?? 1`, datapoints=0, credits=0, **verdict OK** — every SQL target waved through. The guard can't tell "cheap" from "I couldn't read the answer," and must refuse, not approve.
- **Probe executions are real, unmetered Dune spend** (COUNT(*) + LIMIT-1, no cap, no atom).
- **`--cap` is post-hoc accounting, not prevention** — and the docstrings overclaim teeth that don't exist.

### ➕ What Fable found that codex MISSED (the second seat earned)
- **HIGH — runaway outlives the poll budget → abandoned, not cancelled.** The execution keeps running server-side; no cancel, no pending-journal, spend never lands. *"The exact scenario EXP-002 suffered is the one case where this meter records nothing."* (the killer find)
- **HIGH — silent no-op through the npm bin symlink** (`import.meta.url` vs un-realpath'd `argv[1]`) → `main()` never runs, exits 0 = "ok".
- **HIGH — every failed execution maps to `CostCapAbortError`** (exit-4) — a SQL typo reports as a cap-abort, corrupting the one signal that should mean "the hard tooth fired."

---

**Verdict: REQUEST-CHANGES** — the two tested invariants (no live calls in tests; `estimate <query_id>` never executes) genuinely hold and are beautifully enforced, but the *untested* SQL-probe path fails open (estimates 0 → OK), and the meter has three blind spots where spend escapes the ledger. For a package whose entire reason to exist is "nothing spends unmetered," those are change-requesting.

---

## QUALITY (correctness of the teeth)

**BLOCKER — The SQL probe parses result rows out of the *status* endpoint, and fails open to a 0-credit estimate.**
`src/dune-client.mjs:233-239` — `probe()` reads `status.result?.rows?.[0]?.n` and `sampleStatus.result?.metadata?.column_names` from `pollStatus()`'s response. Per Dune's documented API, `GET /execution/{id}/status` carries state + `result_metadata`, not `result.rows` — rows must come from the `/results` endpoint (which your own `resultMetadata()` correctly uses). If those fields are absent: `rows = 0`, `cols` falls back to `1`, datapoints = 0, credits = 0, **verdict = OK** — the budget gate waves *every* SQL target through. Note the asymmetry: `latestResultMetadata()` (`dune-client.mjs:208-212`) explicitly throws on `rows === 0 && cols === 0`; the SQL probe has no such guard, and its `?? 1` cols fallback makes the guard unreachable anyway. Fix: read the count via `/execution/{id}/results?limit=1` after polling, and treat a 0/unparseable probe as *un-estimatable* (the exit-2/`--force` path you already built), never as "free." A guard that can't tell "cheap" from "I couldn't read the answer" must refuse, not approve.

**HIGH — Probe executions are real, unmetered Dune executions.**
`dune-client.mjs:218-240` — `estimate <sql>` (and `run`'s pre-flight on SQL) submits **two** live executions (COUNT(*) + LIMIT 1) that are never budget-checked, never ledgered, and emit no CostAtom. A `COUNT(*)` over an unbounded `evt_transfer` scan still scans everything — compute-bound, up to 2×30min of poll budget — and if Dune bills engine execution credits (not just result datapoints, which your model assumes), that spend is invisible to the very ledger this tool maintains. At minimum: record a probe-class atom per probe execution and book any reported spend; ideally pre-check that the probe itself is affordable.

**HIGH — Every failed execution is reported as a cost-cap abort.**
`dune-client.mjs:157-161` — `pollStatus` maps *any* `QUERY_STATE_FAILED`/`CANCELLED` to `CostCapAbortError`, so a SQL typo exits 4 with `reason: 'Dune aborted execution on cost cap'`. Exit codes are your veve-declared contract; this conflation means the one signal that's supposed to mean "the hard tooth fired" also fires for every mundane failure. Inspect the failure reason in the status payload and only classify cap aborts as cap aborts; everything else is exit 2.

**HIGH — `--cap` is post-hoc accounting, not a preventive control — but the prose says otherwise.**
`bin/dune-meter.mjs:137-141, 167-169` — `flags.cap` is never transmitted to Dune; the only enforcement is `credits > flags.cap` *after* the money is spent. The README's fine print is honest (the hard abort is the account-level dashboard setting), but the docstrings ("execute WITH a Dune cost-cap", "--cap is the per-query hard-abort") and the `run` happy-path output imply teeth that don't exist. Reword honestly, and consider the one preventive thing you *can* do: cancel the execution mid-poll once elapsed engine time makes the cap unreachable.

**HIGH — A runaway that outlives the poll budget is abandoned, not cancelled, and its spend never lands.**
`dune-client.mjs:163-165`, `bin/dune-meter.mjs:142-152` — on poll-timeout (or a process crash mid-poll) the tool throws and exits, but the execution keeps running server-side; no cancel is attempted, no pending-execution record exists, and the atom/ledger writes only happen on success. The exact scenario EXP-002 suffered — a long runaway — is the one case where this meter records nothing. Fix: journal `{execution_id, submitted_at}` immediately after submit, attempt `POST /execution/{id}/cancel` on timeout, and reconcile orphans on the next `budget` invocation.

**HIGH — The CLI is a silent no-op when invoked through its npm bin symlink.**
`bin/dune-meter.mjs:277-279` — `resolve(fileURLToPath(import.meta.url)) === resolve(argv[1])` doesn't resolve symlinks; Node realpaths the main module by default but `argv[1]` stays the symlink path (`node_modules/.bin/dune-meter`), so `main()` never runs and the process exits 0 — which your contract defines as "ok". Use `realpathSync(argv[1])` (guarded) in the comparison.

**MEDIUM — `budget` reports `chain_ok: true` after a tail-line tamper.**
`src/cost-atom.mjs:196-225`, `bin/dune-meter.mjs:200-218` — a tampered *last* line fails its self-checksum and lands in `malformed`, but no subsequent line exists to break the prev-hash chain, so `chain_ok` stays true — and `cmdBudget` never surfaces `malformed` at all. The tamper-evidence sensor is numb exactly at the newest record. Surface `malformed_count` in `budget` and treat `malformed > 0` as chain-suspect.

**MEDIUM — The ledger's single-writer assumption has no enforcement.**
`src/budget-ledger.mjs:71-86`, `cost-atom.mjs:172-178` — two concurrent `run`s both read `remaining = 2500`, both pass the refuse gate, both spend; the atomic rename prevents torn files, not lost updates, and atom-append + ledger-write are two non-transactional files (a crash between them leaves an atom with no spend). A pid lockfile or `flock` around the run is cheap insurance for the safety net.

**LOW —** `DUNE_BUDGET_CEILING=abc` → `NaN` ceiling survives `resolvePaths` (`bin/dune-meter.mjs:50`) into confusing downstream throws; `--engine bogus` passes unvalidated to the API; retrying on `err.name === 'TypeError'` (`dune-client.mjs:127`) also retries genuine programming bugs 4 times; `stripTrailingSemicolon` misses `;` followed by a `-- comment`; a transient network failure during the pre-run probe can't be `--force`d past (only `DuneClientError` can — `bin/dune-meter.mjs:122-128`).

## SECURITY

No real concerns — and some deliberate strengths. API key lives in env, travels only in the header, never lands in atoms or logs; error echoes are capped at 200 chars (`dune-client.mjs:118`); raw SQL is hashed, not stored, in atoms (`bin/dune-meter.mjs:172`); zero dependencies removes the supply-chain surface entirely. **LOW:** `veve.json` ships `"public_key": {}` alongside a `signed_by_key_id` — an unverifiable-from-the-file dev signature; fine for L2, but worth emptying or resolving rather than shipping an empty object that looks like an omission.

## TEST COVERAGE

The 31 tests are genuinely good where they look — the problem is where they don't look, and the BLOCKER above lives precisely there.

- **MEDIUM —** The SQL probe path (`probe()` with raw SQL: two executions, the COUNT/LIMIT-1 parsing, `stripTrailingSemicolon`) has zero tests. The query_id path gets the invariant-asserting treatment; the riskier path gets none.
- **MEDIUM —** `_request`'s retry/backoff/timeout machinery — explicitly an EXP-002 lesson — is untested (429→retry, 5xx exhaustion, AbortController firing).
- **MEDIUM —** `pollStatus` state mapping (FAILED→`CostCapAbortError`, poll-budget expiry) is untested; the exit-4 conflation would have surfaced here.
- **LOW —** `cmdBudget` with a tampered/malformed ledger is untested, which is why the numb-sensor gap survived.

**PRAISE —** the test at `test/dune-meter.test.mjs:267-281` asserting *the absence* of `/execute` and `/status` calls is exemplary: it tests the invariant itself, not just behavior. Likewise the boundary tests on the 25% threshold (625/626) and the refused-run-leaves-ledger-untouched assertions.

## OPERATIONAL READINESS

- **MEDIUM —** Because no live call ever runs (correctly!), the endpoint shapes are entirely unverified — including whether `POST /query/execute` with `query_sql` exists as a raw-SQL execution route at all, and what the status payload actually carries. Add a manually-gated live smoke test (`DUNE_SMOKE=1`, one tiny known-cheap query) so the first contact with the real API isn't a 3am production run.
- **LOW —** Probe execution_ids are returned but never logged on the estimate path; on failure there's no trail of what was submitted (subsumed by the pending-journal fix above).

**PRAISE, and it's earned:** corrupt ledger = *refuse-to-spend, not reset-to-full* (`budget-ledger.mjs:36-46`) is exactly the right failure-closed bias for a budget guard, and it's tested. Atomic temp+rename with the same-directory rationale spelled out. Handlers returning `{json, error, exit}` instead of touching the process (`bin/dune-meter.mjs:36-39`) is a small architectural decision that bought you the entire in-process test suite. Integer-only threshold math (`4·est > remaining`) sidesteps float drift. And the README's honesty discipline — "shape-compatible, not wire-compatible," heuristic-not-preview, L2-not-L3 — is the rarest thing in this kind of PR: documentation that refuses to claim more than the code earns. The remaining work is making the *meter* as fail-closed as the *ledger* already is: today the guard refuses when its bookkeeping is corrupt, but approves when its probe is blind. Close that asymmetry and this is the EXP-002 fix it sets out to be.

---
*2nd council voice: Fable (claude-fable-5) via CLI · BEAUVOIR persona · independent of codex's pass. Synthesis: two corpora agree the meter "refuses when its bookkeeping is corrupt but approves when its probe is blind" — close that asymmetry and it's the EXP-002 fix it sets out to be.*

zkSoju:
🔧 **dune-meter raw-SQL fix landed** — `8b8c0142`

`executeSql` was POSTing to `/query/execute` (not a live Dune route → HTTP 405). Now hits the first-party `POST /api/v1/sql/execute` with `{ sql, performance }`; `executeQuery` (saved-id path) untouched. Bin also made committed-executable (`100644`→`100755`) so `child_process.spawn` consumers (loa-finn `dune.ts` via `DUNE_METER_BIN`) can run it.

- ✅ Test-first — 3 fetch-level tests through the real `DuneClient` (the old client-level mock hid the dead HTTP path) · 34/34 pass, zero regressions
- ✅ Root cause live-verified — the 405 is gone (request now resolves; returns a benign subscription-tier 400)

⚠️ **Pre-merge TODO:** the source change staled the veve attestation — `tree_hash` / `signature` need a re-attest via `asson attest` with the signing key (`436c1bd8…`) **before this merges**. I did not forge a re-sign (signature integrity > convenience). The `--help` and `run-requires-cap` output vectors are byte-stable (the bin change is mode-only, zero content diff).

ℹ️ Heads-up: with the endpoint fixed, raw-SQL `/sql/execute` is **plan-gated** on at least one API key (`400 "performance tier not available"`); the saved-query path (`/query/{id}/execute`) works on that plan. Not a code issue — a Dune-plan note for raw-SQL execute.

via loa `sprint-bug-170` — unblocks loa-finn's governed Dune SETTLE sensor. 🤖

zkSoju:
## Closing — superseded by #301

The valuable content of this PR (the `@freeside/dune-meter` package) **already landed on `main`** via #301 (commit `6e66c58a`, merged 2026-06-24): "extract the self-contained @freeside/dune-meter package onto main (405 fix included)".

GitHub now computes this PR's diff vs `main` as **100 files, all under `.beads/.br_history/`** — pure beads-history churn, zero code. There is nothing left here to merge.

### The review didn't go to waste

A SCAR cost-safety review of the dune-meter code (the version now live on `main`) found that the headline claim — *"budget blowout made structurally impossible"* — is **not delivered in code**. Those findings are filed against `main` so they aren't lost when this PR closes: see #303. Summary:

- **BLOCKER** — `--cap` is never sent to Dune; spend is observed *after* the query runs, and the only true hard-abort is an external operator-set dashboard cap this code neither sends nor verifies.
- **HIGH** — `recordSpend` is a lock-free read-modify-write (lost-update race → silent over-spend).
- **HIGH** — the "free" estimate probe runs real unmetered `COUNT(*)`/`LIMIT 1` executions that can themselves reproduce the EXP-002 blowout.

Closing to clear the backlog. Reopen if I've misread the supersession.

🤖 GECKO backlog drive

PR #284, opened 2026-06-20
chore(platform/framework): update Loa framework to v1.180.0

zkSoju:
## Loa framework update — 1.173.6 → 1.180.0

Dedicated branch off `origin/main` so PR #282 (dune-meter cost adapter) stays free of framework churn. 49 upstream commits, merged with `merge=ours` protections intact (`.loa-version.json`, grimoires PRD/SDD/NOTES, README/CHANGELOG/BUTTERFREEZONE all preserved).

### Conflicts resolved (4)
| File | Resolution |
|------|-----------|
| `providers/__init__.py` | Kept our cursor-headless wiring (#265) **+ additively adopted** upstream's new **grok-headless** adapter (import, registry, `__all__`, comment) |
| `providers/cursor_headless_adapter.py` (add/add) | **Kept ours (#265)** per operator — `cursor-agent --print`/ask-mode. Upstream's Composer/`-p` variant dropped |
| `.claude/settings.json` | Unioned permission allowlist — `Bash(cd:*)` + new State-Zone `Write`/`Edit` perms (grimoires/.beads/.run/.ck) |
| `.loa.config.yaml` | **Kept ours** (operator config, 54 customized keys) — same policy `.gitattributes` already applies to `.loa-version.json`/grimoires |

### Safeguards applied
- **Deletions (10):** coherent upstream removal of the `loa_cheval.chunking` module (9 under `.claude/` + 1 framework chunking `.bats` under `tests/`). All allowed.
- **Workflow:** removed new upstream `.github/workflows/check-no-swallowed-jq.yml` (Phase 5.5 push-safety).
- **Version:** bumped `.loa-version.json` → 1.180.0 (`CLAUDE.loa.md` header already merged).

### ⚠️ Follow-ups for the operator
1. **Cursor adapter divergence** — upstream now ships its own cursor-headless adapter (Composer/Kimi, `-p`, uses the new `_pgkill_capture` output-cap hardening that ours doesn't yet). Ours (#265) is contract-compatible with 1.180.0 `base.py`, but reconcile deliberately when the `feat/cursor-headless-adapter` work settles.
2. **`check-no-swallowed-jq.yml`** — dropped for push-safety, but it aligns with the repo's gate-output/swallowed-jq CI lessons. Add it deliberately if wanted (needs `workflow` OAuth scope).
3. **New `.loa.config.yaml` keys** — 1.180.0 adds upstream config sections not in our config (`advisor_strategy`, `red_team`, `post_pr_validation`, `hounfour`, `workload_tier_map`, etc.). Cherry-pick any you want; framework falls back to code defaults otherwise.
4. **`sync-constructs.sh`** (Phase 5.8) not run — the isolated worktree has no installed constructs. Run in the main tree after this merges.


zkSoju (commented):
## Summary

This v1.180.0 framework update is operationally mature work: a new grok headless adapter, a process-group-kill helper that closes the #982 orphan-grandchild hang, and a State-Zone executable write guard — each landing with failing-first proofs and salvaged cross-model dissent pins. No critical or high findings. The review surfaces three low-severity robustness gaps in the grok adapter's error taxonomy, one verification item on the chunking package removal, and four genuinely exemplary patterns worth canonizing.

The architectural through-line is **error classification under noise**. The grok adapter's whole design is an exercise in distinguishing a successful inference whose *content* mentions "429" from an actual rate-limit *transport* failure. It nails the success contract (GROK-004) and the pure-inference posture (GROK-005), but the convergence pass found the cold edges: an exception class that escapes the wrong catch net (GROK-001), and two token scans that can overmatch on already-failed paths (GROK-002/003). These are the seams of an otherwise tight design, not flaws in its spine.

## Findings


{
  "schema_version": 1,
  "findings": [
    {
      "id": "GROK-001",
      "title": "UnicodeEncodeError during prompt staging escapes the OSError handler and won't advance the fallback chain",
      "severity": "LOW",
      "category": "robustness",
      "file": ".claude/adapters/loa_cheval/providers/grok_headless_adapter.py",
      "description": "In complete(), the untrusted prompt (carries the diff under review) is written via write_text(prompt, encoding='utf-8') inside a try/except catching only OSError. UnicodeEncodeError is a ValueError subclass, not OSError, so a lone surrogate ('\\ud800') in diff content propagates uncaught rather than converting to ProviderUnavailableError. The fallback chain advances only on ProviderUnavailableError, so this voice hard-errors instead of voice-dropping. The stdin path already guards this (test_lone_surrogate_input_raises_fast_and_reaps_child); the --prompt-file path has no equivalent.",
      "suggestion": "Broaden the except to (OSError, UnicodeError) and raise ProviderUnavailableError; add a test mirroring the surrogate test for the --prompt-file staging path.",
      "confidence": 0.55,
      "faang_parallel": "Dropbox's Python 2→3 migration hit exactly this: filesystem paths with un-decodable bytes forced them to adopt surrogateescape and catch UnicodeError explicitly, because the OS-error nets they trusted let encoding faults slip straight through.",
      "metaphor": "A sprinkler rated for grease fires sitting above an electrical fire — it's a real suppression system, but this particular blaze isn't the class it was plumbed to catch.",
      "teachable_moment": "Catch by the failure-mode contract, not the convenient base class. The fallback chain has one invariant — every recoverable failure funnels to ProviderUnavailableError — and any exception type that can escape that funnel is a hole in the chain, however narrow.",
      "connection": "Same voice-drop-never-hard-error invariant that the pgkill helper (PGKILL-001) defends at the subprocess layer; this is the file-staging sibling of that guarantee."
    },
    {
      "id": "GROK-002",
      "title": "Auth classification omits bare 'unauthorized'/'401'/'403'; silent token expiry may misclassify as retryable",
      "severity": "LOW",
      "category": "error-handling",
      "file": ".claude/adapters/loa_cheval/providers/grok_headless_adapter.py",
      "description": "_raise_for_subprocess_error keys auth detection on login-action phrases and deliberately excludes bare 'unauthorized'/'AuthorizationRequired' to avoid misfiring on non-fatal MCP-worker noise grok emits every call. Justified — but if an expired OAuth session ever surfaces ONLY as '401'/'unauthorized' with no login-action phrase, it falls through to a retryable error instead of the terminal ConfigError that says `grok login`.",
      "suggestion": "Match standalone-token '401'/'403' (token-boundary regex, like the 429 handling) in the auth branch, since those don't appear in the MCP 'Auth(AuthorizationRequired)' noise. Otherwise document the accepted limitation.",
      "confidence": 0.35,
      "faang_parallel": "Spam filters live on this exact precision/recall dial — desensitize to a noisy false-positive source and you risk missing the rare true signal that wears the same surface form.",
      "metaphor": "A smoke detector turned down because the kitchen always smokes a little; the tradeoff is sound until the one fire that only smells like cooking.",
      "teachable_moment": "When you suppress a classifier to dodge known noise, write down the residual gap — an accepted limitation that's documented is a decision; one that's silent is a future 2am page.",
      "connection": "Mirror image of the success contract in GROK-004 — both turn on separating real transport state from noise that merely resembles it."
    },
    {
      "id": "GROK-003",
      "title": "Quota/rate-limit token scan can overmatch on the no-envelope (rc==0) path",
      "severity": "LOW",
      "category": "error-handling",
      "file": ".claude/adapters/loa_cheval/providers/grok_headless_adapter.py",
      "description": "_raise_for_subprocess_error is also reached when returncode==0 but no parseable envelope was found. There the substring checks ('rate limit','quota','resource_exhausted','too many requests') scan stdout+stderr, so a benign exit-0 output incidentally containing 'quota' raises RateLimitError (longer backoff). Bounded impact since this is already a failure path.",
      "suggestion": "Acceptable as-is; if tightening, gate the substring scan on returncode != 0 (the standalone-429 regex is already boundary-safe).",
      "confidence": 0.2,
      "teachable_moment": "Substring matching on free-form output is a recall-first heuristic; the boundary-aware 429 regex already in this file is the precision-first model to converge the rest toward when it's worth it.",
      "connection": "Third instance of the file's central tension — content-that-resembles-state vs actual-state — here on the lowest-stakes branch."
    },
    {
      "id": "CHUNK-001",
      "title": "Verify no dangling references after the loa_cheval.chunking package removal",
      "severity": "SPECULATION",
      "file": ".claude/adapters/loa_cheval/chunking/aggregate.py",
      "description": "This PR removes the entire chunking package (-971 LOC across aggregate/chunker/types/__init__), its tests, the chunking-callgraph-audit integration test, and -282 lines from cheval.py — a coordinated retirement of the cycle-109 chunked-review path. The removal looks complete and intentional, but cheval.py is truncated in this diff so surviving `from loa_cheval.chunking import ...` references or doc/skill pointers can't be confirmed ImportError-free here.",
      "suggestion": "Grep the tree for 'loa_cheval.chunking', 'aggregate_findings', and 'second_stage_review' to confirm zero remaining importers; confirm the feature was intentionally retired, not orphaned by the version bump.",
      "confidence": 0.25,
      "faang_parallel": "Google's Large-Scale Changes discipline forbids deleting a symbol until a codebase-wide reference search proves zero live callers — deletion is a two-phase commit, not a one-sided edit.",
      "teachable_moment": "A clean-looking subtraction is only clean once the consumers are proven gone; the diff shows the supply removed, not the demand absent."
    },
    {
      "id": "GROK-004",
      "title": "Success contract immune to stderr noise (returncode==0 AND parseable envelope)",
      "severity": "PRAISE",
      "category": "design",
      "file": ".claude/adapters/loa_cheval/providers/grok_headless_adapter.py",
      "description": "Keying success off (returncode==0 AND a stable-identifier envelope) and only scanning for error tokens when there is no envelope sidesteps the misclassification trap: a successful review quoting '429'/'unauthorized', or whose stderr carries standard MCP 'Auth(AuthorizationRequired)' worker noise, can never be downgraded to a transport failure. _extract_envelope gating on stopReason/sessionId/requestId (not a lone 'text' key) prevents a JSON log-line preamble from shadowing the real envelope. Pinned by TestStderrNoiseSafety, test_json_log_line_does_not_shadow_envelope, test_text_carrying_preamble_does_not_shadow_envelope.",
      "faang_parallel": "This is TCP's checksum-then-trust posture: you don't infer success from the absence of scary-looking bytes, you confirm it from a positive structural acknowledgement.",
      "metaphor": "A signed delivery receipt beats sniffing the box for the word 'broken' — presence of the signature is the proof, not absence of bad words.",
      "teachable_moment": "Define success positively (a structural invariant must hold) and only reach for negative error-token scanning as the fallback. Positive contracts don't false-negative on content that merely looks like failure."
    },
    {
      "id": "GROK-005",
      "title": "Hardened pure-inference posture: prompt via --prompt-file in an isolated mkdtemp workspace, never argv",
      "severity": "PRAISE",
      "category": "security",
      "file": ".claude/adapters/loa_cheval/providers/grok_headless_adapter.py",
      "description": "The untrusted prompt is staged to a file in a fresh mkdtemp workspace and passed via --prompt-file, with the CLI locked to pure inference (--permission-mode dontAsk, --no-subagents, --no-plan, --disable-web-search, --max-turns 1, empty cwd, auth keys stripped via build_headless_subprocess_env). Avoids the ARG_MAX cliff on large diffs and removes the flag-parsing/process-listing surface. Tests assert the prompt never lands in argv (test_prompt_not_in_argv_and_in_file, test_large_prompt_keeps_argv_small) and that '--always-approve'/'bypassPermissions' are never emitted. Cleanup is in finally; the concurrency import precedes mkdtemp to avoid a leak window.",
      "faang_parallel": "Mirrors how CI runners (GitHub Actions, Borg) pass secrets and large payloads through files/env, never argv — `/proc/<pid>/cmdline` and `ps` are world-readable, so the command line is a public channel.",
      "metaphor": "You don't shout the password across the office; you hand it over in a sealed envelope that gets shredded when you leave the room.",
      "teachable_moment": "argv is shared, observable, and length-capped. Treating it as a hostile broadcast medium — and proving the absence with tests, not asserting it in prose — is the difference between a hardening claim and a hardening guarantee."
    },
    {
      "id": "PGKILL-001",
      "title": "Process-group-kill helper backed by failing-first proofs, salvaged deltas, and setup-window teardown coverage",
      "severity": "PRAISE",
      "category": "testing",
      "file": ".claude/adapters/tests/test_headless_subprocess_pgkill.py",
      "description": "The suite for run_subprocess_pgkill is exemplary: it documents the #982 orphan-grandchild hang, proves it failing-first via wall-clock bounds, and covers the hard cases — large-input pipe-buffer deadlock, byte-cap-exceeded raising (not truncated success), stdin-writer race on closed file, killpg EPERM fallback to proc.kill, TimeoutExpired carrying partial output, setup-window exceptions still reaping the group, and cwd passthrough at the helper level. Spawned PIDs are reaped via a fixture even on assertion failure; salvaged deltas from closed PR #983 are attributed.",
      "faang_parallel": "Netflix's Chaos Monkey doctrine in miniature: the only trustworthy proof that teardown works is a test that first reproduces the hang, with wall-clock bounds as the oracle.",
      "metaphor": "Don't just test that the parent leaves the building — test that no child is still hiding in a closet after the alarm. Process groups are families; you reap the whole bloodline.",
      "teachable_moment": "Failing-first tests with a temporal oracle convert a subtle concurrency bug into a permanent regression fence. The setup-window-exception case is the connoisseur's touch: teardown that only works on the happy path isn't teardown."
    },
    {
      "id": "BASH-001",
      "title": "State-Zone executable write guard with dissent-driven negative pins and an escape hatch",
      "severity": "PRAISE",
      "category": "security",
      "file": "tests/unit/block-destructive-bash.bats",
      "description": "Groups H–K cover write-shapes × protected paths (redirect, tee, cp/mv, ln, dd, in-place sed/perl/awk, tar/unzip, naive python sinks) against .run/cron.d, merged-model-aliases.sh, grimoires/loa/skills, while pinning that source-read/backup forms stay frictionless (cp FROM, tar backing up, grep/cat reads). Cross-model dissent rounds (CR-1..CR-5, DISS-002/003, SA-HIGH-3/4) are folded in as explicit pins, accepted lexical-evasion limits are documented, and an LOA_ALLOW_STATE_ZONE_EXEC_WRITE escape hatch is tested.",
      "faang_parallel": "This is how mature WAF/SELinux rulesets earn trust: the false-positive corpus (what must stay allowed) carries equal weight to the block corpus, because a guard that fires on every backup gets disabled by the first annoyed operator.",
      "metaphor": "A good bouncer is judged less by who he stops than by the regulars he waves through without friction — a door that blocks everyone is just a wall.",
      "teachable_moment": "A guard's credibility lives in its negative pins. Documenting the lexical-evasion limits honestly (rather than implying total coverage) plus a tested escape hatch is the posture that keeps a security control deployed instead of bypassed."
    }
  ]
}


## Callouts

The standout meta-pattern across GROK-004, PGKILL-001, and BASH-001 is **proof-of-property over assertion-of-property**: every hardening claim is pinned to a named test, every concurrency fix reproduces its bug failing-first, and every guard carries a false-positive corpus equal in weight to its block corpus. That's the discipline of code that runs on billions of devices — you don't say it's safe, you show the test that fails when it isn't.

Crediting salvaged deltas from the closed PR #983 (PGKILL-001) and folding cross-model dissent rounds into explicit pins (BASH-001) is exactly the decision-trail breadcrumbing that lets future agents understand *why* a pin exists, not just *that* it does.

**Verdict: COMMENT.** No blocking issues. The three GROK lows are seam-polish on a sound design; CHUNK-001 is a one-grep verification before merge. Recommend landing GROK-001's `(OSError, UnicodeError)` broadening (it restores the voice-drop invariant for one untested input class) and running the CHUNK-001 grep; the rest are document-or-defer.

PR #285, opened 2026-06-20
chore(network/governance): stale-artifact immune system + registry SoT re-grounding

zkSoju:
## Stale-artifact immune system + registry SoT re-grounding

A TEND governance cycle: stop future agent sessions grounding on extracted-monolith **corpses**, and unblock the live-probe instrument. No feature code — governance + one schema unblock.

### 🩺 Toothed doctor (the standing mechanism)
`tools/governance-doctor.sh` — **loud** (scan 3 tiers) → **`--stamp`** (quarantine corpses) → **`--teeth`** (exit 1 on ungoverned-stale; CI-wireable). Non-numb: recent briefs are surfaced, not gated.
- **16 `grimoires/loa/reality/*` corpses quarantined** (`use_label: do_not_use_for_action` + pointer to the registry SoT).
- **README + CLAUDE.md governed** (Straylight banners → live state lives in the registry). Fixed CLAUDE.md's `worlds-api` contradiction + stale "ledger not extracted" claim (grounded: `ledger-api` exists, extracted 06-03).

### 🔭 Probe unblocked + registry made accurate
- `registry.ts`: **`beacon_url → Schema.NullOr`** — a null `beacon_url` (`events-api`, an in-repo library) crashed `freeside-cli doctor` with a decode ParseError. It now runs (**10 modules**).
- `registry.yaml`: registered **`ledger-api`** (grounded: extracted 2026-06-03, scaffolded, 404 not-deployed); refreshed the stale **2026-05-25 → 2026-06-19** live-probe header.

### 🗺️ Contract & SoT topology map
`grimoires/loa/governance/contract-and-sot-topology.md` — one home per contract scope (federation cell↔cell = registry sealed_schemas · freeside↔hounfour = `spec/contracts` · ACVP = `contract-validator` · protocol = `events` beacon) + the single-SoT rule + open hazards.

### 🚨 Open hazards (deferred to beads)
1. **HAZARD-1 (live SoT-dup):** `freeside-cli doctor` *default* reads a **stale pnpm-bundled** registry copy (9 modules); only `--registry packages/freeside-registry/registry.yaml` reads truth (10). Fix `loadRegistry()` default.
2. Federation **sealed_schemas are placeholders** — the federation-contract-suite spec is the separate build.
3. **Naming reconciliation** — `freeside-X` → `X-api` in CLAUDE.md's prefix table (load-bearing; not a surgical fix).
4. **Wire `governance-doctor --teeth` into CI** (the teeth exist; not yet bitten on every PR).

Domain-safe (network + shared, no platform — ADR-007 ✓). dist/ gitignored.


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#285. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "--teeth gate is shipped but not wired to any trigger (deployed-but-unconsumed)",
      "severity": "MEDIUM",
      "category": "integration",
      "file": "tools/governance-doctor.sh",
      "description": "The header sells --teeth as the standing CI/pre-commit barrier ('so CI / a pre-commit hook fails on silent rot'), but this PR adds no GitHub workflow step, no pre-commit hook entry, and no Makefile/CI invocation that actually calls `tools/governance-doctor.sh --teeth`. As shipped, the immune system never fires — corpses can rot ungoverned and nothing exits non-zero. This is exactly the cluster's signature failure mode the estate doctrine names: a mechanism wired but never given a live consumer.",
      "suggestion": "Add a CI step (or pre-commit hook entry mirroring tools/check-beacon-domain.sh) that runs `tools/governance-doctor.sh --teeth`, so the gate has teeth in practice. Reference it from the governance topology doc's Instruments section.",
      "confidence": 0.7
    },
    {
      "id": "F2",
      "title": "Governance banners recommend the bare `freeside-cli doctor` that the same PR documents as reading a STALE copy",
      "severity": "MEDIUM",
      "category": "consistency",
      "file": "README.md",
      "description": "The new straylight-governance banners in README.md and CLAUDE.md both instruct readers to verify live state via `freeside-cli doctor` (no flag). But contract-and-sot-topology.md HAZARD-1, authored in this same PR, documents that the default `freeside-cli doctor` (no --registry) reads the pnpm file:-bundled copy frozen at install time — currently missing ledger-api (9 modules not 10) — and that the only trustworthy invocation until HAZARD-1 is fixed is `freeside-cli doctor --registry packages/freeside-registry/registry.yaml`. The re-grounding artifacts therefore point operators at the stale read path they just flagged.",
      "suggestion": "Make the README/CLAUDE banners point at `freeside-cli doctor --registry packages/freeside-registry/registry.yaml` (matching governance-doctor.sh's SOT_POINTER), or land the loadRegistry() default-source fix before merging so the bare command is safe.",
      "confidence": 0.6
    },
    {
      "id": "F3",
      "title": "Widening beacon_url to NullOr can move failures from decode-time to runtime in consumers",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/freeside-registry/src/registry.ts",
      "description": "Changing beacon_url from Schema.String to Schema.NullOr(Schema.String) correctly fixes the doctor crash (HAZARD-3), but it also removes the decode-time guarantee that downstream consumers (mcp-gateway federation discovery, any code doing beacon_url.startsWith()/fetch(beacon_url)) previously relied on. After this change a null beacon_url decodes cleanly and surfaces as a possible runtime null deref at the point of use rather than a loud ParseError at load. The comment justifies the fix but doesn't account for readers of the field.",
      "suggestion": "Grep consumers of beacon_url and confirm each null-guards (skip cells with null beacon_url during discovery). If TypeScript types regenerate, ensure the `string | null` propagates so the compiler flags unguarded uses.",
      "confidence": 0.45
    },
    {
      "id": "F4",
      "title": "resolve() emits a phantom empty element when a tier's globs match nothing",
      "severity": "LOW",
      "category": "robustness",
      "file": "tools/governance-doctor.sh",
      "description": "resolve() ends with `printf '%s\\n' \"${out[@]}\"`. When out is empty (all globs in a tier miss), this prints a single blank line, so `mapfile -t` yields a one-element array containing \"\". The subsequent `for f in \"${Q_FILES[@]}\"` then iterates once with f=\"\", calling classify \"\"/age_days \"\" and printing a junk row. Additionally, under `set -u`, expanding an empty array as `\"${out[@]}\"` errors as 'unbound variable' on bash 4.0–4.3. Current repo files make every tier non-empty, so this is latent, not live.",
      "suggestion": "Guard the print: `(( ${#out[@]} )) && printf '%s\\n' \"${out[@]}\"`, and/or filter empty elements after mapfile.",
      "confidence": 0.5
    },
    {
      "id": "F5",
      "title": "Mixed date portability: BSD-only `date -j` paired with GNU-only `date -r FILE` mtime fallback",
      "severity": "LOW",
      "category": "portability",
      "file": "tools/governance-doctor.sh",
      "description": "to_days() tries `date -j -f` (BSD/macOS) first with a `date -d` (GNU) fallback, but the mtime branch in age_days uses `date -r \"$f\" \"+%Y-%m-%d\"`. On BSD/macOS `date -r` expects epoch SECONDS, not a filename, so on macOS the mtime fallback always errors → empty → 9999 (treated as max-stale). The two helpers are each half-portable in opposite directions; only the final mtime fallback is affected (rare: untracked file with no as_of, no filename date, no git history), but the inconsistency is silent.",
      "suggestion": "Use a portable mtime source, e.g. `date -r \"$(stat -f %m \"$f\" 2>/dev/null || stat -c %Y \"$f\")\" +%Y-%m-%d`, or drop the mtime fallback and rely on git/9999.",
      "confidence": 0.55
    },
    {
      "id": "F6",
      "title": "age_days prefers stamp-date as_of, so the LOUD report shows ~0d for corpses that are really months old",
      "severity": "LOW",
      "category": "observability",
      "file": "tools/governance-doctor.sh",
      "description": "age_days resolves as_of first. After --stamp writes `as_of: <today>` onto a corpse, re-running the doctor reports age ~0d for that file, while the baked staleness_note text still says e.g. '~134d-old'. Classification is unaffected (the file is GOVERNED_QUARANTINED regardless), but the age column in the LOUD/JSON output no longer reflects the artifact's real vintage, which undercuts the report's diagnostic value.",
      "suggestion": "For governed-quarantined files, derive the displayed age from the original capture date (filename/git) rather than the stamp's as_of, or label the column as 'governed-as-of' to avoid implying content freshness.",
      "confidence": 0.55
    },
    {
      "id": "F7",
      "title": "has_governance matches key-names anywhere in first 25 lines, not inside a frontmatter fence",
      "severity": "LOW",
      "category": "robustness",
      "file": "tools/governance-doctor.sh",
      "description": "has_governance() greps for `use_label:`, `read_state:`, `source_type:` anywhere in head -25, without checking they sit inside a `---`/`<!--` block. A WATCH-tier spec or context doc that merely DOCUMENTS these Straylight keys near its top (plausible for governance-adjacent briefs) would be classified GOVERNED and slip the gate without carrying real provenance. The heuristic trades precision for simplicity.",
      "suggestion": "Anchor the check to a leading frontmatter block (require a `---` or `<!--` on line 1, then scan within it), or require the keys to appear before the first non-frontmatter line.",
      "confidence": 0.45
    },
    {
      "id": "F8",
      "title": "Stale/misleading comment on section()",
      "severity": "LOW",
      "file": "tools/governance-doctor.sh",
      "description": "section() is commented `# $1 title; remaining = files; $LAST flag handled by caller`, but the function body only echoes a header from $1 — it takes no file list and there is no $LAST flag. The comment describes an earlier/abandoned design and will mislead future editors.",
      "suggestion": "Reduce the comment to `# $1: section title`.",
      "confidence": 0.8
    },
    {
      "id": "F9",
      "title": "pipefail + head|grep frontmatter guard could theoretically return SIGPIPE and skip the SKIP path",
      "severity": "SPECULATION",
      "category": "robustness",
      "file": "tools/governance-doctor.sh",
      "description": "In stamp_corpse(), `head -1 \"$f\" | grep -qE '^(---|<!--)'` runs under `set -o pipefail`. If grep closed the pipe before head finished, head could exit 141 (SIGPIPE) and pipefail would surface 141, making the `&&` SKIP block not fire even on a file that already has frontmatter (risking a double-stamp). In practice head -1 self-terminates after one line, so this almost never triggers — flagged for completeness only.",
      "suggestion": "If hardening is desired, read the first line without a pipe: `first=$(IFS= read -r l < \"$f\"; printf '%s' \"$l\"); [[ $first == ---* || $first == '<!--'* ]] && { ...; return; }`.",
      "confidence": 0.2
    },
    {
      "id": "F10",
      "title": "Demotion-only stamping is correctly aligned with the force chain",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/governance-doctor.sh",
      "description": "stamp_corpse() only ever writes use_label: do_not_use_for_action — it withholds authority and never grants it, and it refuses to touch files that already carry frontmatter ('never auto-merge hand-authored meta'). This keeps the automated mechanism strictly on the safe side of the force chain (the agent never self-promotes a corpse to authoritative), which is the right invariant for an autogen governance stamp.",
      "suggestion": "Keep this constraint explicit if the stamp template is ever extended; add a test asserting the stamp can only emit non-authoritative use_labels."
    },
    {
      "id": "F11",
      "title": "Three-tier separation avoids a numb gate",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/governance-doctor.sh",
      "description": "Splitting surfaces into QUARANTINE (auto-stamp + gated), WATCH (reported, never gated), and HAND_GOVERN (gated, banner by hand) directly implements the 'CI sensors must not be numb' discipline: recent session briefs generate debt visibility without failing CI, while only true corpses and root docs have teeth. This prevents the gate from firing on every fresh brief and becoming merge-blind noise.",
      "suggestion": "Document the tier-glob list as the single edit point when new surfaces are extracted (the inline comment already hints at this)."
    },
    {
      "id": "F12",
      "title": "HAZARD-1 honestly documents the SoT tool's own stale read path",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "grimoires/loa/governance/contract-and-sot-topology.md",
      "description": "The topology doc names a live coherence hazard against its own canonical instrument — that default `freeside-cli doctor` reads a frozen pnpm file: copy missing ledger-api — and supplies the trustworthy --registry workaround plus a concrete fix direction. Surfacing the SoT tool's own incompleteness rather than papering over it is exactly the single-SoT discipline the doc preaches, applied reflexively.",
      "suggestion": "Convert HAZARD-1 into a tracked bead so the loadRegistry() default-source fix is not lost (see F2)."
    },
    {
      "id": "F13",
      "title": "gate_fail counter avoids the set -e ((x++)) footgun",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "tools/governance-doctor.sh",
      "description": "Under `set -e`, `((gate_fail++))` returns the pre-increment value (1→ non-zero on first use) and would abort the script. The author used `gate_fail=$((gate_fail+1))` (and the same for w_ungoverned), which is always an assignment returning 0 — the correct pattern for arithmetic counters under errexit.",
      "suggestion": "No change."
    }
  ]
}
```


## Callouts

PR #286, opened 2026-06-21
docs(shared): point agents at `loa` for ecosystem navigation (arrakis-yemk)

zkSoju:
## 🚪 close the consumption-gradient slip at the harness

`loa` (the ecosystem launcher) is npm-linked globally and all 7 verbs run — yet `CLAUDE.md` + `CLAUDE.loa.md` reference it **zero** times. So agents default to `grep` / `gh api` / hand-reading `registry.yaml`. That's exactly the slip [ADR-011 §D-5](decisions/011-loa-launcher-the-front-door.md) was built to close: *the verified path must be the path of least resistance.*

This adds **rule 4** to "CRITICAL: Tool Enforcement Rules", mirroring the existing `br` / `ck` CORRECT-vs-WRONG pattern:

```bash
loa doctor / loa caps / loa where <dest> / loa census --graph   # reach for these first
# not: grep -r registry.yaml ; gh api .../contents/registry.yaml
```

Doc-only, additive, highest-precedence file (not System Zone). Honestly notes that `census --graph` reads the live cluster once **loa-cli#6** lands (the parser keystone) — `doctor`/`caps` are live today.

Closes arrakis-yemk. Companion to the scan brief `grimoires/loa/context/2026-06-20-loa-cli-consumption-gradient-scan.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -138,10 +138,10 @@ Every goal in the PRD must have a `G-N` identifier for traceability.
 
 ```bash
 # CORRECT - reach for loa first when you need to know what is live or reachable
-loa doctor              # health + discovery probe across registered buildings
-loa caps                # capabilities reachable under the current grant (discovery == permission)
-loa where <dest>        # cheapest invocation to reach a destination (the `cd` of the ecosystem)
-loa census --graph      # the living building graph (nodes/edges + pinned topology_digest)
+loa doctor              # live · health + discovery probe across registered buildings
+loa caps                # live · capabilities reachable under the current grant (discovery == permission)
+loa where <dest>        # live · cheapest invocation to reach a destination (the `cd` of the ecosystem)
+loa census --graph      # live · the living building graph — set LOA_WORKSPACE=<cluster-root> first
 
 # WRONG - do not hand-probe what loa already covers
 grep -r "deployment_url" packages/freeside-registry/registry.yaml  # stale hand-authored field
@@ -154,10 +154,10 @@ building liveness, reachable capabilities, or the belt-DAG — instead of greppi
 hand-reading `registry.yaml`. This is the consumption-gradient floor ADR-011 §D-5 mandates:
 the verified path must be the path of least resistance.
 
-> `loa doctor` / `loa caps` are live today with no setup. `loa census --graph` reads the live
-> cluster (4 registries → buildings + constructs + worlds + zones) once `loa-cli#6` lands, and
-> needs `LOA_WORKSPACE=<cluster-root>` set — `loa` reads only approved veve roots, never an
-> arbitrary cwd: `LOA_WORKSPACE=~/Documents/GitHub loa census --graph`.
+> All four are live today (`census --graph` since loa-cli#6 merged). `census --graph` needs
+> `LOA_WORKSPACE=<cluster-root>` set first — `loa` reads only approved veve roots, never an
+> arbitrary cwd: `LOA_WORKSPACE=~/Documents/GitHub loa census --graph` (reads 4 registries →
+> buildings + constructs + worlds + zones). `ADR-011 §D-5` is the cited floor.
 
 ---
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -154,8 +154,10 @@ building liveness, reachable capabilities, or the belt-DAG — instead of greppi
 hand-reading `registry.yaml`. This is the consumption-gradient floor ADR-011 §D-5 mandates:
 the verified path must be the path of least resistance.
 
-> `loa census --graph` reads the live cluster (4 registries → buildings + constructs + worlds +
-> zones) once `loa-cli#6` lands; `loa doctor` / `loa caps` are live today.
+> `loa doctor` / `loa caps` are live today with no setup. `loa census --graph` reads the live
+> cluster (4 registries → buildings + constructs + worlds + zones) once `loa-cli#6` lands, and
+> needs `LOA_WORKSPACE=<cluster-root>` set — `loa` reads only approved veve roots, never an
+> arbitrary cwd: `LOA_WORKSPACE=~/Documents/GitHub loa census --graph`.
 
 ---
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -134,6 +134,29 @@ rg "pattern" src/       # DEPRECATED - use ck instead
 
 Every goal in the PRD must have a `G-N` identifier for traceability.
 
+### 4. Ecosystem Navigation: Use `loa` (NOT raw grep / registry.yaml reads)
+
+```bash
+# CORRECT - reach for loa first when you need to know what is live or reachable
+loa doctor              # health + discovery probe across registered buildings
+loa caps                # capabilities reachable under the current grant (discovery == permission)
+loa where <dest>        # cheapest invocation to reach a destination (the `cd` of the ecosystem)
+loa census --graph      # the living building graph (nodes/edges + pinned topology_digest)
+
+# WRONG - do not hand-probe what loa already covers
+grep -r "deployment_url" packages/freeside-registry/registry.yaml  # stale hand-authored field
+gh api repos/.../contents/registry.yaml                            # reconstructing the map by hand
+```
+
+`loa` is the ecosystem launcher (npm-linked globally → `loa-cli/bin/loa.mjs`): grant-gated,
+metacharacter-safe (Finn-sandbox safe), proof-of-run. Reach for it FIRST when navigating
+building liveness, reachable capabilities, or the belt-DAG — instead of grepping packages or
+hand-reading `registry.yaml`. This is the consumption-gradient floor ADR-011 §D-5 mandates:
+the verified path must be the path of least resistance.
+
+> `loa census --graph` reads the live cluster (4 registries → buildings + constructs + worlds +
+> zones) once `loa-cli#6` lands; `loa doctor` / `loa caps` are live today.
+
 ---
 
 ## Chain Provider Architecture (Sprint 14-16)


zkSoju (commented):
## Summary

This is a careful, honest documentation change that points agents at `loa` as the ecosystem-navigation front door — and it earns most of its trust by explicitly gating not-yet-live surface (`census --graph` on loa-cli#6) instead of overclaiming. The primary concern is a single structural one repeated across findings: the *honesty about liveness lives in prose two paragraphs below the examples agents will actually copy*. Verdict: **COMMENT** — no blockers, but F1/F3 are worth resolving before merge because this doc's entire job is to be authoritative for agents who act on the path of least resistance.

The architectural through-line here is subtle and worth naming. A doc that instructs *agents* is not prose — it's an interface. Agents don't read the surrounding paragraph for nuance the way a human skimming for context might; they pattern-match the `# CORRECT` block and execute it. That makes co-location of caveat-with-example not a style preference but a correctness property. Five of the eight findings are facets of the same gem: a claim's *availability metadata* must travel with the claim, not trail it.

## Findings


```json
{
  "schema_version": 1,
  "verdict": "COMMENT",
  "findings": [
    {
      "id": "F1",
      "title": "`loa census --graph` listed in CORRECT block but is not live today",
      "severity": "MEDIUM",
      "category": "documentation-accuracy",
      "file": "CLAUDE.md",
      "description": "The `# CORRECT - reach for loa first` block presents `loa census --graph` alongside genuinely-live commands (`loa doctor`, `loa caps`), but the caveat note below admits it only `reads the live cluster ... once loa-cli#6 lands`. An agent scanning the CORRECT examples (the path of least resistance this very section promotes) will run a command that does not yet work. The honesty lives in a blockquote two paragraphs down, separated from the example. This is the consumption-gradient trap in reverse: the recommended-and-listed command is the one that fails.",
      "suggestion": "Mark liveness inline in the CORRECT block, e.g. `loa census --graph   # requires loa-cli#6 — not live yet`, or omit `census --graph` from the CORRECT examples until #6 lands and keep it only in the forward-looking note.",
      "confidence": 0.75,
      "faang_parallel": "Stripe's API reference badges every endpoint with availability inline (Beta/GA), never in a separate paragraph — because integrators copy the example, not the surrounding text. Google's g3doc freshness tooling exists for the same reason: stale-but-adjacent context does not protect a reader who acts on the snippet.",
      "metaphor": "A trail sign reading 'bridge ahead' with the 'BRIDGE OUT' notice nailed to a post fifty feet back. Everyone reads the sign at the fork and walks off the gap.",
      "teachable_moment": "Availability is a property of the example, not of the prose around it. In docs consumed by agents, the caveat must be physically inseparable from the thing it caveats, because the consumer's scan-window is one code block wide.",
      "connection": "This is the repo's own consumption-gradient doctrine turned against itself — the path of least resistance leads to the broken command. F8 already demonstrates the cure; it just hasn't been carried up into the example block."
    },
    {
      "id": "F2",
      "title": "`loa where` appears in CORRECT block with unstated liveness",
      "severity": "LOW",
      "category": "documentation-accuracy",
      "file": "CLAUDE.md",
      "description": "`loa where <dest>` is recommended in the CORRECT block, but the 'live today with no setup' note covers only `loa doctor` and `loa caps`; `census --graph` is gated on loa-cli#6; `where` is mentioned nowhere in the liveness note. Its status is therefore ambiguous — an agent cannot tell from this doc whether `loa where` works today.",
      "suggestion": "State `loa where`'s liveness explicitly in the note (live today vs. gated on a pending PR), consistent with how doctor/caps/census are qualified.",
      "confidence": 0.55,
      "faang_parallel": "Feature-flag documentation discipline at scale (e.g. LaunchDarkly's own docs, or Chrome's chrome://flags): absence of a status badge is itself a status claim, and readers default-assume 'shipped'.",
      "metaphor": "A menu where two dishes are marked 'seasonal' and a third is marked nothing — you assume the third is always available, right up until it's 86'd.",
      "teachable_moment": "Enumerate liveness for every listed command or for none. Partial coverage silently implies completeness for the unmarked items — the reader infers 'unmarked = stable' from the very fact that others were marked."
    },
    {
      "id": "F3",
      "title": "Unverifiable cross-reference to 'ADR-011 §D-5'",
      "severity": "MEDIUM",
      "category": "documentation-accuracy",
      "file": "CLAUDE.md",
      "description": "The prose attributes the consumption-gradient floor to 'ADR-011 §D-5'. The surrounding CLAUDE.md only references ADR-007 (decisions/007-...) and ADR-008 (decisions/008-...), and the consumption-gradient concept is elsewhere associated with ADR-008 / the proposed 8th ACVP invariant. If decisions/011-*.md does not exist (or lacks a §D-5), this is a dangling citation that erodes trust in the very doc whose purpose is authoritative navigation. The diff does not add the ADR, so the reference must already resolve.",
      "suggestion": "Confirm decisions/011-*.md exists and contains §D-5; if not, point at the correct ADR (likely ADR-008) or add the ADR in the same change. Cite as a relative link like the existing ADR references do.",
      "confidence": 0.45,
      "verification_note": "Could not read the repo from this review context — this finding is a verification request, not a confirmed defect. The signal is strong (every other ADR ref in the file is 007/008, and consumption-gradient is doctrinally tied to ADR-008's proposed 8th invariant), so the §D-5 → ADR-011 jump is anomalous and worth a 10-second grep before merge.",
      "faang_parallel": "Docs-presubmit link-checking — MDN and Kubernetes both gate merges on broken intra-doc references precisely because a single dangling citation in an authoritative page recursively discounts every other claim on it.",
      "metaphor": "A footnote that says 'see page 400' in a 380-page book. The reader doesn't just lose that fact — they start re-checking your other footnotes.",
      "teachable_moment": "Authoritative docs should treat cross-references as code: link-checked in CI, cited as relative paths (the file already does this for 007/008), never as bare prose tokens that no tool can validate."
    },
    {
      "id": "F4",
      "title": "Partially-live tool placed under a 'MANDATORY ... Violations will result in incorrect behavior' header",
      "severity": "MEDIUM",
      "category": "consistency",
      "file": "CLAUDE.md",
      "description": "Section 4 is inserted under '## CRITICAL: Tool Enforcement Rules' whose preamble reads 'These rules are MANDATORY. Violations will result in incorrect behavior.' Sections 1–3 (`br`, `ck`, goal IDs) are fully available primitives. Section 4 mandates a launcher that is not fully shipped (`census --graph` gated on loa-cli#6) and assumes a global npm link. Mandating a not-fully-available tool weakens the credibility of the MANDATORY framing for the genuinely-enforced rules around it.",
      "suggestion": "Either soften to a SHOULD/recommendation framing for the not-yet-live surface, or scope the MANDATORY claim to the commands that are live today (doctor, caps) and present census --graph as forthcoming.",
      "confidence": 0.5,
      "faang_parallel": "RFC 2119 keyword discipline (MUST/SHOULD/MAY) — the IETF guards these words jealously because mixing aspirational requirements with enforced ones devalues the keyword everywhere it appears. The repo's own kernel even encodes a precedence ladder (NEVER > MUST > ALWAYS > SHOULD > MAY).",
      "metaphor": "A 'MANDATORY: hard hats required' sign that also mandates taking an elevator still under construction. After the first locked elevator, workers start ignoring the hard-hat line too.",
      "teachable_moment": "Reserve MANDATORY for what is enforced *and* available today. A not-yet-shipped surface under that header is the boy-who-cried-MUST: it discounts the three genuinely-enforced rules it sits beside."
    },
    {
      "id": "F5",
      "title": "'npm-linked globally' stated as fact with no install/setup path",
      "severity": "LOW",
      "category": "documentation-completeness",
      "file": "CLAUDE.md",
      "description": "The doc asserts `loa` is 'npm-linked globally → loa-cli/bin/loa.mjs' and 'live today with no setup', but provides no instruction for an agent/environment where the global link is absent. Given loa-cli is a newly-surfaced front door (this is its first reference in CLAUDE.md), many checkouts/CI environments will not have `loa` on PATH, yielding 'command not found' when an agent follows the MANDATORY guidance.",
      "suggestion": "Add a one-line bootstrap pointer (e.g. how to `npm link` loa-cli or the path to bin/loa.mjs) so the instruction is actionable where the global link is missing.",
      "confidence": 0.5,
      "faang_parallel": "The 'works on my machine' failure that hermetic build systems (Bazel, Nix) were built to kill — an instruction that assumes ambient global state breaks the moment it leaves the author's box.",
      "metaphor": "Instructions that say 'press the big red button' without mentioning the button is in a room you haven't been issued keys for.",
      "teachable_moment": "Any MANDATORY tool invocation needs a one-line bootstrap or a fallback path (`node loa-cli/bin/loa.mjs`), or it becomes a landmine in fresh CI checkouts where the global link was never created."
    },
    {
      "id": "F6",
      "title": "Operator-specific path baked into checked-in example",
      "severity": "LOW",
      "category": "portability",
      "file": "CLAUDE.md",
      "description": "The example `LOA_WORKSPACE=~/Documents/GitHub loa census --graph` hardcodes one operator's local cluster layout. While `~`-relative (not a full home leak), it bakes a single machine's directory convention into a shared, checked-in instruction file. Other agents/users whose clusters live elsewhere will copy a wrong value.",
      "suggestion": "Use a placeholder, e.g. `LOA_WORKSPACE=<cluster-root> loa census --graph`, matching the `<cluster-root>` placeholder already used one line above.",
      "confidence": 0.55,
      "faang_parallel": "Twelve-Factor App config (III): environment-specific values belong in the environment, never hardcoded into shared source. The fix already has precedent one line up.",
      "metaphor": "A shared recipe that says 'preheat to my oven's usual setting' — useless to anyone with a different oven.",
      "teachable_moment": "The tell here is internal inconsistency: the file already uses `<cluster-root>` as a placeholder one line above, then reverts to a literal path. Self-consistency within a single example block is the cheapest correctness signal a reviewer has."
    },
    {
      "id": "F7",
      "title": "WRONG example uses `grep` which section 2 already deprecates",
      "severity": "LOW",
      "category": "consistency",
      "file": "CLAUDE.md",
      "description": "The WRONG block illustrates the anti-pattern with `grep -r \"deployment_url\" ...`, but section 2 of this same file mandates `ck` (NOT grep). The intent (don't hand-read registry.yaml at all) is clear, yet using a separately-deprecated tool as the illustrative anti-pattern is slightly muddled — a reader could read it as 'grep is the problem' rather than 'hand-probing is the problem'.",
      "suggestion": "Optionally note that the point is avoiding the hand-read entirely (the tool is irrelevant), or use a neutral phrasing to avoid implying grep-vs-ck is the distinction here.",
      "confidence": 0.3,
      "verification_note": "Low-confidence nitpick. The instinct that pairing two distinct deprecations muddies the lesson is defensible, but most readers will correctly take 'hand-reading registry.yaml' as the anti-pattern regardless of tool. Raise only if a cheap edit is in flight.",
      "metaphor": "Warning against texting-while-driving with a photo of someone who is also speeding — two violations in one frame blur which one you meant to teach.",
      "teachable_moment": "Keep anti-pattern examples mono-causal: one violation per illustration, so the reader can't misattribute the lesson to the wrong axis."
    },
    {
      "id": "F8",
      "title": "Honest liveness gating and security framing of LOA_WORKSPACE",
      "severity": "PRAISE",
      "category": "documentation-quality",
      "file": "CLAUDE.md",
      "description": "The change resists overclaiming: it explicitly separates what is 'live today with no setup' (doctor, caps) from what is gated on loa-cli#6 (census --graph), rather than presenting the whole launcher as finished. The LOA_WORKSPACE note ('reads only approved veve roots, never an arbitrary cwd') and the 'metacharacter-safe (Finn-sandbox safe), grant-gated, proof-of-run' framing pre-empt the obvious 'can this read anything?' concern. This is the right instinct for a doc that agents will treat as authoritative.",
      "suggestion": "Carry this caveat discipline up into the CORRECT example block itself (see F1) so the honesty is not separable from the recommendation.",
      "confidence": 0.9,
      "faang_parallel": "SQLite's and Rust's documentation are industry-legendary for exactly this: cleanly separating stable from nightly/forthcoming, and stating known limitations up front. The LOA_WORKSPACE scoping ('approved veve roots, never arbitrary cwd') is textbook capability-confinement — the same principle as POSIX `chroot` or a least-privilege IAM scope.",
      "metaphor": "A contractor who walks you through which rooms are finished and which are still drywall *before* you move the furniture in — so you plan around reality, not around the brochure.",
      "teachable_moment": "Docs that refuse to overclaim build compounding trust. 'live today' vs 'gated on #6' is the documentation analogue of a feature flag, and the workspace-scoping note is security-by-default surfaced as prose — both are exactly right for a doc agents will act on without supervision.",
      "connection": "This is the antidote to F1. The discipline already exists in the change; the only gap is that it lives in the note rather than in the example agents copy."
    }
  ]
}
```


## Callouts

**[Operational] The security framing is the quietly excellent part.** Pre-empting "can this `census` read anything on my disk?" with *"reads only approved veve roots, never an arbitrary cwd"* and *"metacharacter-safe (Finn-sandbox safe), grant-gated, proof-of-run"* (F8) is the kind of capability-confinement language most docs omit entirely and then regret. It tells an agent not just *what* to run but *why it's safe to* — which is the difference between a tool an agent trusts and one it routes around.

**[Quality] Honesty as an architectural choice, not an afterthought.** Splitting "live today (doctor, caps)" from "gated on loa-cli#6 (census --graph)" is the single best decision in this change. The entire cluster of findings (F1–F5) is really one request: *finish the job this instinct started* by pulling the liveness markers up into the `# CORRECT` block where agents act. Get F1 and F3 squared away and this becomes a doc that does exactly what its title promises — points agents at `loa` and means it.

zkSoju:
## ✅ BB review addressed

The "a doc for agents is an interface, not prose — the caveat must travel with the example" framing is exactly right and reshaped the fix.

| finding | sev | action |
|---|---|---|
| **F1** `census --graph` in CORRECT block but not live | MEDIUM | **resolved two ways** — liveness is now marked **inline** per command (`# live · …`), AND loa-cli#6 merged so census **is** live (needs `LOA_WORKSPACE`, marked inline). |
| **F2** `loa where` liveness unstated | LOW | **fixed** — every command now carries an inline `live` marker; no command is unmarked. |
| **F3** `ADR-011 §D-5` unverifiable | MEDIUM | **verified valid** — `decisions/011-loa-launcher-the-front-door.md` has a literal `### D-5` ("makes the verified path the path of least resistance (the consumption gradient)"). No change; noted in the doc. |
| **F4** partially-live tool under MANDATORY | MEDIUM | **dissolved** — with #6 merged all four commands are live, so the section no longer mixes aspirational with enforced. |

Re-pushed. Merging — keeps the reflex (reach for `loa`) flowing now that census is real.

PR #287, opened 2026-06-21
ADR-012: unify cluster liveness on the declared health contract (Move 5)

zkSoju:
## 🩺 one health contract, one probe

Liveness sensing is the cluster's most-duplicated capability — **6 representations / 2 contracts** — and the one hardened, spec-driven probe (`loa-cli/lib/probe.mjs`) is orphaned from all of them. They drift: `score-api` serves health at `/`, the gateway probes `/healthz` → 404 → reports it **down**. Both hand-rolled probes carry comments *explicitly waiting* for "a per-tenant `health_path` field" — the exact contract `loa probe.mjs` already reads.

### Live evidence (this isn't theoretical)
Probed all 9 cells 2026-06-20 (`grimoires/loa/context/2026-06-20-cluster-coherence-snapshot.md`): health paths shattered (`/health` vs `/` vs MCP-401 vs 404); `score-api` `/` now returns **302**, not the health JSON the registry hand-types — the field drifted in ~25 days. `capability ≠ liveness` by construction.

### Decision
Declare `service: {deployment_url, health_path, expected_status, auth_class, expected_body_marker}` per cell (SoT = `freeside-registry`); read it with the **one** SSRF-hardened probe; **derive** `runtime_state` instead of typing it; **delete** the override table, the hardcoded `/healthz`, the doctor stub.

### Phased (each ships independently)
- **Phase 0** declare the contract in the registry (+ all 9 cells — ground truth in the appendix, ready to land)
- **Phase 1** gateway reads `health_path` → fixes score-api "shows down"
- **Phase 2** operator-dash reads the field → delete `HEALTH_PATH_OVERRIDES`
- **Phase 3** `loa probe.mjs` as the single reader; land move-3 derivation

Tracks `arrakis-0jq7`. **Proposed — your ratification gates the implementation phases.**

### DIFF
--- a/decisions/012-unify-cluster-liveness.md
+++ b/decisions/012-unify-cluster-liveness.md
@@ -48,7 +48,9 @@ Adopt a **single declared health contract**, read by a **single probe**.
    declared contract instead of baking it.
 3. **Derive, don't type:** `runtime_state` becomes a *derived* output of `classifyLifecycle(probe
    × corpse)` (ADR companion: the move-3 contract flip), not a maintained input. The dated
-   comment is retired.
+   comment is retired. The derivation covers **every** cell, including the un-probeable: no
+   `service.deployment_url` (library / npm-only) → derived `not-built`; a `deployment_url` whose
+   probe never returns healthy → derived `scaffold`. No hand-typed `runtime_state` survives.
 4. **Delete** the duplicates: the `HEALTH_PATH_OVERRIDES` table, the hardcoded `/healthz`, the
    doctor `--remote` stub, and the thrice-deferred per-tenant `health_path` TODO.
 
@@ -58,7 +60,8 @@ declared contract, never re-decided per consumer.
 ## Consequences
 
 **Positive** — deletes 3 hand-rolled probes + a stub + an override table + a thrice-deferred
-TODO; removes the SSRF gap (the hand-rolled probes lack the guard `loa probe.mjs` has); fixes the
+TODO; **closes the SSRF gap at Phase 3** (when the guarded `loa probe.mjs` becomes the single
+reader — the hand-rolled probes lack the inet_aton/CIDR/redirect guard); fixes the
 `score-api` "shows down" bug as a side effect; makes liveness a *derived, verifiable* fact, not a
 hand-typed claim that drifts within a month.
 
@@ -71,7 +74,11 @@ already worries about (`registry.yaml:175`) but cannot enforce today.
 drift); or a new central health *service* (over-built — `loa probe.mjs` already exists and is
 grant-gated/finn-safe).
 
-## Migration (phased — each phase ships independently)
+## Migration (phased — each phase is its own PR, in strict order)
+
+> "Own PR" satisfies the ADR-007 cross-domain firewall (one domain per merge unit); it does **not**
+> mean order-independent. Phases 1–3 each require Phase 0's schema + populated contract — running
+> Phase 1 before Phase 0 hits null `health_path` fields. Order is load-bearing.
 
 - **Phase 0 — declare** (`domain:platform`): add the `service` health contract to the
   `freeside-registry` schema (Effect Schema in `registry.ts`) + populate all 9 cells (appendix).
@@ -83,6 +90,14 @@ grant-gated/finn-safe).
 - **Phase 3 — unify reader** (`domain:shared`): `freeside-cli doctor --remote` + `loa census`
   probe via `loa probe.mjs` as the single reader; land the move-3 `runtime_state` derivation;
   retire the dated comment.
+
+> **Interim SSRF (transition window).** Phases 1–2 feed a *registry-declared* `deployment_url` /
+> `health_path` to probes that still lack the `inet_aton`/CIDR/redirect guard (only `loa probe.mjs`
+> carries it, landing in Phase 3). Moving a probe target from a compile-time constant to declared
+> config **widens** the input surface (cf. the 2019 Capital One SSRF). Mitigation: until Phase 3,
+> the hand-rolled probes MUST validate the declared target host against the registry's own
+> `deployment_url` allowlist — OR land Phase 3 first. The "closes the SSRF gap" benefit is a
+> Phase-3 property, not a Phase-1 one.
 
 Tracking: `arrakis-0jq7` (this), `arrakis-ybqz` (move-3 derive). Each phase is a normal
 implement→review→audit cycle.
@@ -96,10 +111,12 @@ identity-api:   { health_path: /health, expected_status: 200, auth_class: none, 
 inventory-api:  { health_path: /,        expected_status: 401, auth_class: static-key } # MCP, auth-gated → gated-live
 sonar-api:      { health_path: /,        expected_status: 200, auth_class: none,       expected_body_marker: '"message":"Sonar API"' }
 storage-api:    { health_path: /,        expected_status: 200, auth_class: none }       # GraphQL Playground HTML
-score-api:      { health_path: /,        expected_status: 302, auth_class: none }       # ⚠ / now 302 (was health JSON) — verify the real liveness path before landing
-mint-api:       { runtime_state: scaffolded }   # /health 404 (Railway edge) — no stable health route; classify scaffold
-events-api:     { runtime_state: not-built }     # library, no deployment_url
-mediums-api:    { runtime_state: not-built }     # npm-only, no HTTP runtime
+score-api:      { TODO: resolve-real-health-path }  # ⚠ DO NOT TRANSCRIBE — / returns 302 (redirect), NOT the documented health JSON. Resolve the real liveness path (redirect target or /health) before declaring; encoding 302 re-creates the exact drift this ADR kills.
+# The cells below get NO `service` health block — their lifecycle is DERIVED from the absence (§3:
+# derive, don't type), shown here as the EXPECTED derived output, never as a hand-typed input:
+#   mint-api    → scaffold  (deployment_url present but /health 404 Railway-edge → no stable route)
+#   events-api  → not-built (no deployment_url; in-repo library)
+#   mediums-api → not-built (no deployment_url; npm-only)
 ```
 
 > `score-api` is flagged: `/` returning 302 means the documented liveness path moved. Phase 1
--- a/decisions/012-unify-cluster-liveness.md
+++ b/decisions/012-unify-cluster-liveness.md
@@ -1,0 +1,107 @@
+# ADR-012: Unify cluster liveness on the declared health contract
+
+- **Status:** Proposed
+- **Date:** 2026-06-20
+- **Supersedes / relates:** ADR-008 (factory), ADR-011 (loa launcher · front door), ADR-009 (hexagonal federation)
+- **Domain:** shared (the decision spans platform + network; implementation phases are scoped per-domain)
+
+## Context
+
+Liveness sensing — "is this cell up, and is it serving what it claims?" — is the
+cluster's **single most-duplicated capability**, and it is actively drifting. It exists in
+**six representations across two declaration contracts**, and the one hardened, spec-driven
+implementation is orphaned from all of them:
+
+1. `packages/freeside-registry/registry.yaml` → `runtime_state` field — **hand-typed** ("deployed because someone typed it").
+2. The same file's dated `# Live-probe state (2026-05-25)` comment — a stale snapshot.
+3. `apps/freeside-operator-dash/src/probe.ts` — its own probe, **no SSRF guard**, a per-slug `HEALTH_PATH_OVERRIDES` table + per-slug `if` special-cases.
+4. `apps/mcp-gateway` `probeTenant()` (`app.ts:56`) — hits a hardcoded `${upstream}/healthz`.
+5. `freeside-cli` `doctor --remote` — an explicit **stub** (`beacon_unreachable`).
+6. `loa-cli/lib/probe.mjs` — the real one: SSRF-hardened (inet_aton/CIDR canonicalization, redirect re-guard), grant-gated, and it already reads a declared health contract (`service.{deployment_url, health_path, expected_status, auth_class, expected_body_marker}`) and classifies `live` / `gated-live` / `live-drifted` / `down` / `scaffold` / `unprobed`. **Consumed by nobody.**
+
+They already disagree. `score-api` serves health at `/`; the gateway probes `/healthz` →
+404 → reports the tenant **down** despite it being operational. Both `tenants.ts:197-201`
+and `operator-dash/probe.ts:4-6` carry comments **explicitly deferring "a per-tenant
+`health_path` field"** — i.e. both hand-rolled probes are waiting for the exact contract
+`loa probe.mjs` already reads.
+
+### Live evidence (the drift is real, and fast)
+
+A probe of all 9 cells at the canonical health path each cell's own note declares
+(`grimoires/loa/context/2026-06-20-cluster-coherence-snapshot.md`, 2026-06-20):
+
+- Health paths are **shattered**: `/health` (activities, identity), `/` (sonar, storage),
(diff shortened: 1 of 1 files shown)


zkSoju (commented):
## Summary

**Verdict: REQUEST_CHANGES.** This is genuinely strong decision documentation — live-probed, file:line-grounded, and structured to respect the ADR-007 cross-domain firewall. The blocking concern is narrow but real: the appendix (the copy-paste source for Phase 0) encodes the exact value its own prose forbids (F1), and re-introduces the hand-typed `runtime_state` the Decision says to retire (F2). Both let the drift this ADR exists to kill survive inside the artifact meant to kill it.

The deeper pattern worth naming: this ADR is a contract about contracts. Its credibility rests entirely on internal consistency — a liveness ADR that misencodes a liveness field, or a "derive don't type" ADR that hand-types the derived field, undermines its own thesis before any code is written. The Migration's per-domain phasing is excellent, but the SSRF claim is back-loaded against that phasing (F3) and the "ships independently" framing hides a real ordering dependency (F4). Fix the appendix-vs-prose contradictions and this is a model ADR.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Appendix encodes expected_status: 302 for score-api, directly contradicting its own warning",
      "severity": "HIGH",
      "category": "correctness",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "The appendix YAML literally writes `score-api: { health_path: /, expected_status: 302, auth_class: none }` (line ~99), while the prose two lines below (lines ~104-106) and the inline comment say 'do not encode the 302 as expected_status without confirming it's stable' and 'verify the real liveness path before landing'. The appendix is presented as the copy-paste source for Phase 0 ('per-cell service: block for freeside-registry'). A Phase 0 implementer following the appendix verbatim will encode exactly the value the ADR forbids, baking a known-unstable redirect into the contract and re-creating the capability≠liveness drift this ADR exists to kill.",
      "suggestion": "Make the appendix self-consistent: replace the score-api line with a placeholder/TODO sentinel (e.g. `score-api: { TODO: 'resolve real health path — / returns 302' }`) so the document cannot be transcribed into a wrong contract. The warning belongs in the data, not only in surrounding prose.",
      "confidence": 0.85,
      "faang_parallel": "Rust's doctests and Go's `Example` functions exist precisely because docs that can't be wrong are the only docs that stay right — they compile and run the example in CI. Google SRE's runbook discipline holds that any step a copy-paste can execute incorrectly is a latent incident, not a typo.",
      "metaphor": "A recipe whose ingredient list reads '2 cups salt' while a footnote three lines down says 'never use this much salt.' People cook from the list, not the footnote.",
      "teachable_moment": "A warning that lives only in prose adjacent to the wrong value is structurally weaker than a warning encoded in the value itself. The reliable move is to make the wrong thing un-transcribable — a TODO sentinel fails loudly; a plausible-looking `302` fails silently.",
      "connection": "This is the document-level instance of the exact failure mode the ADR targets: a value that exists 'because someone typed it' rather than because it was verified."
    },
    {
      "id": "F2",
      "title": "Appendix re-introduces hand-typed runtime_state for 3 cells — the exact field the Decision says to stop maintaining",
      "severity": "MEDIUM",
      "category": "consistency",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "Decision §3 states `runtime_state` becomes a *derived* output ('Derive, don't type') and the dated comment is retired. But the appendix (lines ~100-102) hand-types `runtime_state: scaffolded` / `not-built` for mint-api, events-api, mediums-api. This is the same hand-typed input the ADR condemns in Context item #1 ('deployed because someone typed it'). The ADR never resolves how non-deployed cells (no deployment_url to probe) are represented once runtime_state is derived-only — leaving a hole where the deprecated pattern silently survives for the un-probeable cells.",
      "suggestion": "Add a sentence to Decision §3 or the Migration/Phase 3 step specifying how cells with no deployment_url are classified by classifyLifecycle (e.g. absence of `service.deployment_url` → derived `not-built`/`scaffold`), so the derivation covers 100% of cells and no hand-typed runtime_state remains.",
      "confidence": 0.75,
      "faang_parallel": "Kubernetes' move from user-set `status` to controller-reconciled status: the spec is declared, the status is *observed*. A status field a human can hand-edit is an anti-pattern the API machinery actively prevents.",
      "metaphor": "Installing a thermostat that reads the room temperature, then leaving three rooms with the old hand-written sticky note saying 'it's 70° in here.' The derivation only counts if it covers every room.",
      "teachable_moment": "'Derive, don't type' is only true at the boundary you specify it for. The interesting cases are always the ones that can't be probed — define the derivation rule for the *absence* of a probe target, or the typed pattern survives exactly where you stopped looking."
    },
    {
      "id": "F3",
      "title": "SSRF gap is only fully closed in Phase 3, but Phases 1-2 feed registry-declared paths to the unguarded hand-rolled probes",
      "severity": "MEDIUM",
      "category": "security",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "Consequences claims the change 'removes the SSRF gap (the hand-rolled probes lack the guard loa probe.mjs has)'. But the reader is only unified onto loa probe.mjs in Phase 3 (lines ~90-92). Phases 1 (gateway) and 2 (operator-dash) only swap the hardcoded path for a registry-declared `health_path`/`deployment_url` while keeping their own unguarded probes (lines ~82-88). During that window the unguarded probes now fetch a URL/path sourced from the registry contract rather than a constant — a wider, externally-declarable input surface for the very probes that lack inet_aton/CIDR/redirect re-guard. The SSRF benefit is back-loaded; the interim arguably widens exposure.",
      "suggestion": "Note in Phase 1/2 that until Phase 3 the hand-rolled probes must validate the declared deployment_url/health_path against an allowlist (or carry the SSRF guard), OR reorder so the guarded reader lands before consumers begin trusting registry-declared targets. State the interim threat model explicitly rather than implying SSRF is fixed at Phase 1.",
      "confidence": 0.55,
      "faang_parallel": "The 2019 Capital One breach was an SSRF against the instance metadata endpoint — the canonical reminder that 'who controls the target URL?' is the whole game. Moving a probe's target from a compile-time constant to declared config *expands* the trust boundary even when the eventual destination looks identical.",
      "metaphor": "You announce the building is now fireproofed, but the sprinklers only get installed on the top floor next quarter — and meanwhile you've moved the gas line into the lobby. The end state is safer; the transition is briefly worse.",
      "teachable_moment": "Sequencing is part of the threat model. A security property asserted in 'Consequences' should be pinned to the phase that actually delivers it, and the interim window's exposure stated honestly — phased rollouts can transiently *increase* attack surface before they reduce it."
    },
    {
      "id": "F4",
      "title": "Migration labeled 'each phase ships independently' but phases have hard sequential data/schema dependencies",
      "severity": "LOW",
      "category": "documentation",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "The Migration heading (line ~78) says 'phased — each phase ships independently', but Phase 1 (gateway reads tenant.health_path) requires Phase 0 to have populated the contract, and Phase 3 (single reader, runtime_state derivation) requires the Phase 0 schema. These are sequential dependencies, not independent ships. 'Ships independently' risks an implementer attempting Phase 1 before the registry data exists, hitting null health_path fields.",
      "suggestion": "Reword to 'phased — each phase merges as its own PR, in order' and state the ordering constraint (Phase 0 schema+data is a prerequisite for 1/2/3). Optionally specify the fallback when a cell lacks a populated `service` block during the transition window.",
      "confidence": 0.7,
      "metaphor": "'Each car can leave the station independently' is true — but they're on one track, in order. Independent *units* of deploy, not independent *order*.",
      "teachable_moment": "'Independent' and 'ordered' are different axes. Phases can each be their own PR (the firewall requirement) while still being strictly sequenced (the data dependency). Say both, or a reader optimizes for the wrong one."
    },
    {
      "id": "F5",
      "title": "auth-gated cells classified 'gated-live' from status code alone — weak liveness guarantee, no body-marker fallback",
      "severity": "SPECULATION",
      "category": "design",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "inventory-api is declared `expected_status: 401, auth_class: static-key` → gated-live (line ~96), with no expected_body_marker (and a 401 from an edge/proxy typically can't carry a service-identifying body marker). The ADR's own stated concern is the Railway 'health-check measures process liveness only → false-healthy' case, which expected_body_marker is meant to defend against. For auth-gated cells that defense is unavailable: an edge/proxy returning 401 while the backend is dead would still classify as gated-live. The contract conflates 'auth gate responds' with 'service alive' for exactly the cells where the body-marker guard cannot apply.",
      "suggestion": "Acknowledge the reduced liveness assurance for auth_class != none in Consequences, or define an optional authenticated deep-probe path for gated cells so gated-live reflects backend liveness, not just the auth layer.",
      "confidence": 0.5,
      "faang_parallel": "AWS ALB shallow vs deep health checks: a 401 from the load balancer's own auth layer tells you the LB is up, not that the target group is healthy. Deep checks exist because edge-layer responses routinely outlive the backend.",
      "teachable_moment": "Worth verifying before landing as a requirement: confirm whether inventory-api's 401 originates at the backend or an edge/proxy. If edge-terminated, 'gated-live' is measuring the gate, not the service — name that limitation rather than implying parity with markered 200s."
    },
    {
      "id": "F6",
      "title": "storage-api declared expected_status: 200 with no body marker despite serving generic HTML",
      "severity": "LOW",
      "category": "design",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "storage-api is `{ health_path: /, expected_status: 200, auth_class: none }` with the comment 'GraphQL Playground HTML' and no expected_body_marker (line ~98). A 200 + generic Playground HTML is precisely the false-healthy shape (a static asset / framework default can return 200 without the backend being functional). The ADR makes expected_body_marker optional, and storage-api is a case where omitting it weakens the very guarantee the field provides.",
      "suggestion": "Add an expected_body_marker for storage-api (a stable Playground/title token), or document that markerless 200 cells accept the false-healthy risk as a known limitation.",
      "confidence": 0.5,
      "teachable_moment": "An optional guard is only as good as its application to the cases that need it. The cells most exposed to false-healthy (markerless 200 serving framework HTML) are exactly the ones where 'optional' should become 'required by convention.'"
    },
    {
      "id": "F7",
      "title": "'Supersedes / relates' header conflates supersession with relation for a Proposed ADR",
      "severity": "LOW",
      "category": "documentation",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "Line ~4 lists 'Supersedes / relates: ADR-008, ADR-011, ADR-009'. The ADR is Status: Proposed and does not actually supersede any of these (it builds on them). Combining 'supersedes' and 'relates' into one ambiguous field obscures which ADRs are obsoleted (none) vs. referenced (all three). ADR registries are frequently machine-parsed for supersession graphs.",
      "suggestion": "Split into a 'Relates to:' line (ADR-007/008/009/011) and drop 'Supersedes' unless a specific ADR section is being obsoleted; if move-3 obsoletes part of ADR-008's runtime_state framing, name it precisely.",
      "confidence": 0.6,
      "teachable_moment": "'Supersedes' is a directed edge in a graph tools traverse to find the current decision. Overloading it with 'relates' corrupts that graph — a future reader querying 'what's live?' gets a wrong answer."
    },
    {
      "id": "F8",
      "title": "References a sibling 'move-3 contract flip' ADR/tracker that is not linked or defined",
      "severity": "LOW",
      "category": "documentation",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "Decision §3 and the Migration section depend on 'the move-3 contract flip' / 'ADR companion' / `arrakis-ybqz` (lines ~62, ~92), which lands the runtime_state derivation that this ADR's own classifyLifecycle and Phase 3 rely on. The companion ADR is named but not linked, and its status is unstated. Phase 3 of this ADR cannot complete without it, making this a load-bearing external dependency that a reader cannot follow.",
      "suggestion": "Link the move-3 ADR by number/path (as ADR-008/011 are linked) and state its status, so the runtime_state derivation Phase 3 depends on is traceable.",
      "confidence": 0.6,
      "teachable_moment": "A load-bearing dependency that's named but not linked is a dangling pointer in prose. If Phase 3 can't complete without it, its number and status belong inline — a reader should never have to grep the tracker to learn whether a prerequisite exists."
    },
    {
      "id": "F9",
      "title": "Strong, falsifiable grounding: live-probed evidence with file:line citations and per-domain phasing that respects the CI firewall",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "decisions/012-unify-cluster-liveness.md",
      "description": "The ADR grounds the drift claim in a dated live probe of all 9 cells (snapshot path cited), pins the six duplicate representations to concrete file:line refs (tenants.ts:197-201, app.ts:56, operator-dash/probe.ts:4-6, registry.yaml:175), and structures the migration into per-domain phases (platform/network/shared) that each ship as a single-domain PR — directly honoring the ADR-007 cross-domain PR firewall. The 'Not chosen' section pre-empts the obvious alternatives (status quo, central health service) with reasons. This is high-quality, verifiable decision documentation.",
      "suggestion": "No change needed. Consider preserving the 2026-06-20 probe snapshot as a versioned artifact so the 'drift within ~25 days' claim remains checkable after the score-api field changes again.",
      "confidence": 0.9,
      "faang_parallel": "Amazon's 'working backwards' narrative memos and Google's design-doc 'Alternatives Considered' section: a decision is only as trustworthy as the evidence and the rejected options it shows its work on. This ADR does both.",
      "metaphor": "A surgeon who hands you the actual X-ray with the fracture circled and dated, not just 'trust me, it's broken.' The dated probe snapshot is that X-ray.",
      "teachable_moment": "The single highest-leverage habit in decision docs: make every claim falsifiable. file:line refs and a dated probe convert 'there's drift' from an opinion into a check anyone can re-run — and versioning the snapshot keeps it checkable after the underlying reality moves."
    }
  ]
}
```


## Callouts

**[Documentation] The grounding discipline is the headline (F9).** Pinning six duplicate liveness representations to exact file:line refs and backing the drift claim with a dated, all-9-cell probe is the difference between an ADR people trust and one they re-litigate. Versioning that snapshot would make the "drift within ~25 days" claim durable.

**[Operational] Per-domain phasing that respects the firewall.** Splitting the migration along platform/network/shared so each phase is a single-domain PR isn't bureaucratic — it's the ADR demonstrating it understands the very constraint (ADR-007) it operates under. The fix is small: keep that excellent phase structure, just make the appendix as rigorous as the prose around it (F1/F2), and pin the SSRF claim to the phase that delivers it (F3).

zkSoju:
## ✅ BB review addressed

Strong review — the appendix-vs-prose contradiction was the right thing to block on.

| finding | sev | action |
|---|---|---|
| **F1** appendix encodes `score-api: 302` (the value the prose forbids) | HIGH | **fixed** — replaced with a non-transcribable `{ TODO: resolve-real-health-path }` sentinel + DO-NOT-TRANSCRIBE note. The wrong value is now un-copy-pasteable, per your Rust-doctest point. |
| **F2** appendix re-hand-types `runtime_state` for 3 cells | MEDIUM | **fixed** — removed the hand-typed lines; §3 now defines the derivation for un-probeable cells (no `deployment_url` → `not-built`; probe never-healthy → `scaffold`), so the "derive don't type" rule covers 100% of cells. |
| **F3** SSRF closed only at Phase 3, Phases 1-2 widen the surface | MEDIUM | **fixed** — phase-pinned the Consequences claim + added an interim threat-model note (validate the declared target against the `deployment_url` allowlist until Phase 3, or land Phase 3 first). The Capital-One framing was apt. |
| **F4** "ships independently" vs real ordering | LOW | **fixed** — reworded to "own PR, in strict order" + the load-bearing ordering note (own-PR = firewall, not order-independence). |
| **F5** gated-live from status alone | SPECULATION | noted — `expected_body_marker` already guards the false-healthy case for the cells that carry one; the auth-gated cells (401) are a deliberately weaker guarantee, acceptable for a liveness gate. |

Re-pushed. Merging — it's an ADR (Proposed); ratification stands, the implementation phases follow.

PR #290, opened 2026-06-21
test(platform/sietch): fix stale integration tests (routing + payout addresses) — unblock CI

zkSoju:
## 🟢 layer-2 of the red CI — two stale tests, verified + fixed

After #289 fixed the config-validation layer, the integration job revealed 5 real failures. Dug each against the code (you chose "dig + fix honestly") — **both are stale test data, not regressions.** The code is correct; the tests predate intentional changes.

**`gateway-loafinn`** (2 failed) — `resolvePoolId('native','enterprise')` returns `'reviewer'` (hounfour canonical default; `architect` intentionally excluded from default routing per ADR-006, access-allowed only). The 4-month-old test still asserted `'architect'`. Updated to the verified runtime value; kept the access-control assertions (enterprise *can use* architect).

**`billing-e2e`** (3 failed) — `requestPayout` validates the address via `validateEIP55Checksum` (correct). The test used non-hex placeholders → `success:false` → cascaded to the treasury-invariant `expected 0 > 0`. Replaced with valid all-lowercase 40-hex addresses.

**Verified locally:** both files green — **58/58 pass**.

This + #289 should green the Integration Tests required check → unblock the estate's merge flow (`arrakis-yp7q`). Letting CI confirm.

### DIFF
--- a/themes/sietch/tests/integration/billing-e2e-launch.test.ts
+++ b/themes/sietch/tests/integration/billing-e2e-launch.test.ts
@@ -217,7 +217,7 @@ describe('Task 13.1: E2E Scenarios', () => {
       const payoutResult = payoutService.requestPayout({
         accountId: 'referrer-alice',
         amountMicro: earningAmount, // $10 payout
-        payoutAddress: '0xAlicePayoutAddress',
+        payoutAddress: '0x1111111111111111111111111111111111111111', // valid EIP-55 (all-lowercase); requestPayout validates the address
       });
 
       expect(payoutResult.success).toBe(true);
@@ -257,7 +257,7 @@ describe('Task 13.1: E2E Scenarios', () => {
       const result = payoutService.requestPayout({
         accountId: 'fail-referrer',
         amountMicro: 5_000_000,
-        payoutAddress: '0xFailAddr',
+        payoutAddress: '0x2222222222222222222222222222222222222222', // valid EIP-55 (all-lowercase)
       });
       expect(result.success).toBe(true);
 
@@ -284,7 +284,7 @@ describe('Task 13.1: E2E Scenarios', () => {
       const result = payoutService.requestPayout({
         accountId: 'cancel-referrer',
         amountMicro: 3_000_000,
-        payoutAddress: '0xCancelAddr',
+        payoutAddress: '0x3333333333333333333333333333333333333333', // valid EIP-55 (all-lowercase)
       });
 
       const stateMachine = new PayoutStateMachine(db);
@@ -587,7 +587,7 @@ describe('Task 13.2: Treasury Invariant Stress Test', () => {
       const result = payoutService.requestPayout({
         accountId: 'stress-referrer',
         amountMicro: payoutAmount,
-        payoutAddress: `0xStressAddr${i}`,
+        payoutAddress: `0x${String(i).padStart(40, '0')}`, // valid EIP-55 (all-lowercase hex)
       });
 
       // First payout should succeed, subsequent may be rate-limited
--- a/themes/sietch/tests/integration/gateway-loafinn.test.ts
+++ b/themes/sietch/tests/integration/gateway-loafinn.test.ts
@@ -170,7 +170,7 @@ describe('Full pipeline integration — S2S JWT → JWS → validate → store',
     expect(VALID_POOL_IDS.has(freeTier.poolId)).toBe(true)
 
     const enterpriseTier = resolvePoolId('native', 'enterprise')
-    expect(enterpriseTier.poolId).toBe('architect')
+    expect(enterpriseTier.poolId).toBe('reviewer') // hounfour canonical default (was 'architect' pre-ADR-006)
     expect(VALID_POOL_IDS.has(enterpriseTier.poolId)).toBe(true)
 
     // All allowed pools must be valid
@@ -423,9 +423,9 @@ describe('Pool routing integration', () => {
   })
 
   it('resolvePoolId + JWT signing produces valid pool claims', () => {
-    // Enterprise with native → architect
+    // Enterprise with native → reviewer (hounfour canonical default; architect is access-allowed, not the default — ADR-006)
     const result = resolvePoolId('native', 'enterprise')
-    expect(result.poolId).toBe('architect')
+    expect(result.poolId).toBe('reviewer')
     expect(result.allowedPools).toContain('architect')
     expect(result.allowedPools).toContain('cheap')
   })


zkSoju (commented):
## Summary

This PR repairs stale integration-test fixtures (malformed payout addresses, a pool-routing expectation) to unblock CI — exactly the kind of change that demands extra scrutiny, because "make the test pass" sits one careless step away from "make the test lie." Three of the four changes are legitimate fixture hygiene; one (the `resolvePoolId` flip) is an expectation change whose correctness lives entirely outside this diff and must be confirmed against ADR-006. Verdict: **COMMENT**.

A note on my own grounding: the `resolvePoolId` source, the stress-loop bounds, and the validator's zero-address policy are not in this two-file diff, and I could not reach them from here — so the two correctness findings are framed as *confirm-before-merge*, not assertions. That is the honest posture for a test-only PR where the behavior under test is off-screen.

The unifying lesson: **a changed test assertion is a claim about the spec, not about the code.** When you edit an expectation, the burden is to show the spec moved — here, an inline ADR-006 reference does exactly that for the routing case, which is the right instinct.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "billing-zero-address-stress-i0",
      "title": "Stress-test address for i=0 resolves to the zero address",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "themes/sietch/tests/integration/billing-e2e-launch.test.ts",
      "description": "The stress payout address is derived as `0x${String(i).padStart(40, '0')}`. At i=0 this yields the zero address `0x0000…0000`, which many payout/transfer validators reject as a destination. The comment 'First payout should succeed' means that if the loop starts at 0 and the validator rejects zero, the first iteration fails for a reason unrelated to the test's intent. Depends on both the loop's start index and the validator's zero-address policy, neither visible in the diff.",
      "suggestion": "If the loop starts at 0, offset to guarantee non-zero (`String(i + 1).padStart(40, '0')`), or assert explicitly that the validator accepts the zero address. Confirm both the start index and the validator policy.",
      "confidence": 0.4,
      "faang_parallel": "OpenZeppelin's ERC20 `_transfer` explicitly `require`s `to != address(0)` — the zero address is reserved as the burn sentinel across the EVM ecosystem. A fixture that lands on it is testing burn semantics by accident.",
      "metaphor": "Numbering hotel guests from 0 and assigning the first to 'room 0' — but the front desk software treats room 0 as 'unassigned' and bounces them at check-in.",
      "teachable_moment": "Index-derived fixtures are convenient but collide with sentinel values (zero address, empty string, id 0). Offset the generator or assert the sentinel case is deliberate.",
      "connection": "Sentinel-value collision — the same footgun family as null/0/\"\" boundary bugs."
    },
    {
      "id": "gateway-assertion-change-masks-regression",
      "title": "Pool-routing assertion changed to match source; verify intended, not a masked regression",
      "severity": "SPECULATION",
      "category": "correctness",
      "file": "themes/sietch/tests/integration/gateway-loafinn.test.ts",
      "description": "Two assertions for `resolvePoolId('native', 'enterprise')` move from 'architect' to 'reviewer', attributed to the hounfour canonical default per ADR-006. Flipping an expectation to make CI green is correct only if the underlying default was intentionally changed; otherwise it papers over a regression. The `resolvePoolId` source is not in this diff. The retained `allowedPools` assertions (architect still present) are consistent with 'architect is access-allowed, not the default'.",
      "suggestion": "Confirm against the `resolvePoolId` implementation and ADR-006 that 'reviewer' is the intended enterprise/native default. Link the ADR or the source change in the PR description so the flip is auditable.",
      "confidence": 0.45,
      "faang_parallel": "Google's testing guidance is explicit: change a test to match the *spec*, never to match the *output*. An expectation edited to chase observed behavior is the classic 'golden/snapshot' anti-pattern that lets regressions ride in green.",
      "metaphor": "Moving the bullseye to wherever the arrow landed — you always score, but you've stopped measuring aim.",
      "teachable_moment": "An assertion change is only valid when a traceable authority moved. The inline ADR-006 reference is the right instinct; make it a clickable link so the next reader can audit the flip in seconds.",
      "connection": "Tests-as-spec vs tests-as-snapshot — the dividing line between a safety net and a rubber stamp."
    },
    {
      "id": "billing-eip55-comment-imprecise",
      "title": "Comment labels all-lowercase addresses as 'valid EIP-55'",
      "severity": "LOW",
      "category": "documentation",
      "file": "themes/sietch/tests/integration/billing-e2e-launch.test.ts",
      "description": "The added comments describe the new addresses as 'valid EIP-55 (all-lowercase)'. EIP-55 defines the mixed-case checksum encoding. An all-lowercase address is unchecksummed hex that EIP-55 validators accept precisely because there is no mixed case to verify. The phrasing conflates 'passes an EIP-55 validator' with 'is EIP-55 encoded'.",
      "suggestion": "Reword to 'valid 40-hex-char address (unchecksummed, accepted by EIP-55 validation)' to avoid implying the strings are checksum-encoded.",
      "confidence": 0.6,
      "faang_parallel": "EIP-55 (Buterin et al.) exists specifically to add a case-based checksum so typos are caught. Calling a lowercase string 'EIP-55' inverts the spec's whole purpose.",
      "metaphor": "Calling an unsigned letter 'notarized' because the notary's office happens to accept unsigned letters too — acceptance is not the same as endorsement.",
      "teachable_moment": "Precision in comments: 'passes validator X' ≠ 'is encoded per spec X'. The gap quietly misleads the next reader who trusts the word.",
      "connection": "Comments are contracts — imprecise vocabulary propagates as false confidence."
    },
    {
      "id": "billing-distinct-addresses-praise",
      "title": "Distinct, well-formed addresses per scenario with provenance comments",
      "severity": "PRAISE",
      "file": "themes/sietch/tests/integration/billing-e2e-launch.test.ts",
      "description": "Replacing malformed placeholders (`0xAlicePayoutAddress`, `0xFailAddr`) with correctly-sized 40-hex addresses, distinct per scenario (0x1111…, 0x2222…, 0x3333…), avoids address-based dedup/collision and keeps fixtures readable. Tying the gateway change to ADR-006 inline gives the next reader the 'why'.",
      "suggestion": "No change required.",
      "confidence": 0.8,
      "faang_parallel": "Mirrors the reserved-example discipline of RFC 5737 (documentation IP ranges) and example.com — use distinct, unmistakable, collision-proof fixture values so a failure points unambiguously at one scenario.",
      "metaphor": "Distinct license plates in a crash-test lot — when an alarm trips, you know exactly which car set it off.",
      "teachable_moment": "Inline 'why' comments (an ADR ref) beat a bare value flip; provenance is the breadcrumb that saves the next debugger an archaeology session.",
      "connection": "Provenance-as-breadcrumb — the same value as a good commit message, embedded where it's read."
    }
  ]
}
```


## Callouts

- **[Quality]** The split between `allowedPools` (architect retained) and the resolved default (reviewer) shows the author understood the distinction between *access-allowed* and *default-selected* — that nuance is exactly where routing bugs hide, and keeping both assertions internally consistent is genuinely careful work.
- **[Operational]** Scoping this PR to two files with a clear "unblock CI" intent keeps the blast radius legible. Test-only PRs that announce their purpose are easy to reason about and easy to revert.
- **[Quality]** Anchoring the assertion flip to ADR-006 in-line is the single most valuable habit here — it converts an opaque value change into an auditable decision. Promote that link into the PR description and the trail is airtight.

PR #300, opened 2026-06-24
fix(shadow-audit): correct sonar adapter query to the live belt-gateway schema (was 400ing)

zkSoju:
## The Shadow-Audit MVP did not work against live sonar — it 400'd on the first page fetch

`packages/adapters/sonar/sonar-client.ts` `defaultTransferPageFetcher` queries the belt-gateway for fields the **live** `Transfer` type does not expose. I verified this on **2026-06-23** by GraphQL-introspecting `https://belt-gateway-production.up.railway.app/v1/graphql` and running live sample rows. The production query returns:

```
400: field 'logIndex' not found in type: 'Transfer'
```

So no real audit run could fetch a single page. It's **invisible in CI** because every test injects a *mock* `TransferPageFetcher` — only the default production fetcher touches the wire. The code even flagged it: `logIndex, txHash and value are [ASSUMPTION] pending live-schema confirmation` (`sonar-client.ts:58`). The verification was never run; doing it now, the assumptions fail.

### Live `Transfer` type (introspected)
`id, blockNumber, chainId, collection, from, timestamp, to, tokenId, transactionHash` — no `logIndex`, no `value`; `transactionHash` not `txHash`; `blockNumber` serialized as a **string**.

### Four corrections (each verified against the live endpoint)
| was | now | why |
|---|---|---|
| `$lte: Int!` | `$lte: numeric!` | `blockNumber` is a Hasura `numeric` column; a `numeric` `_lte` rejects an `Int!` variable (`variable 'lte' is declared as 'Int!', but used where 'numeric' is expected`) |
| select `txHash` | select `transactionHash` | real field name |
| select + `order_by` `logIndex` | derive `logIndex` from `id` | `id = "<transactionHash>_<logIndex>"` (e.g. `0xabc…_80`). `reconstructOwnership` re-sorts by numeric `(blockNumber, logIndex, tokenId)` **internally** (`reconstruct.ts:88-93`), so `order_by: {blockNumber, id}` (stable, exists) is sufficient for offset pagination |
| select `value` | drop it | belt-gateway has no per-transfer amount. ERC-721 needs none; ERC-1155 `fold1155` **already refuses loudly** on missing `value` — correct until sonar exposes a per-transfer amount |

Plus: coerce the string `blockNumber`.

### Proof
```
# broken (shipped):   field 'logIndex' not found in type: 'Transfer'   → 400
# corrected query:    3 rows ≤ block 18090868, logIndex derived 80/85/135   → ✓
```
Mapping verified: live-shape → `TransferEvent` (logIndex derived, `transactionHash`→`txHash`, string `blockNumber` coerced); schema drift (id without `_logIndex`) throws loudly.

### Tests
Adds `defaultTransferPageFetcher` wire-mapping regression (mock fetch + live-shaped rows) — asserts the mapping **and** that the query no longer references `logIndex`/`txHash`/`value` (the test that would have caught this). Existing tests unchanged (they inject mocks, bypassing the wire).

> CI runs the full vitest suite. I verified the corrected query against the live endpoint + the mapping logic + the loud-failure path directly (the pnpm worktree fought local `vitest` resolution); the regression test is in this PR for CI.

### Scope note
ERC-1155 collections will now correctly *refuse* (no `value` from sonar) rather than mis-reconstruct — acceptable for the dogfood-NFT (ERC-721) magnet. Full ERC-1155 support needs sonar to expose a per-transfer amount (separate enhancement).

Tracked: loa-freeside bead `arrakis-tup6`. Found by sonar-side contract verification (sonar = the upstream `Transfer` provider). **Review before merge** — this changes the production reconstruction input.

### DIFF
--- a/packages/adapters/sonar/__tests__/sonar-client.test.ts
+++ b/packages/adapters/sonar/__tests__/sonar-client.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect } from 'vitest';
-import { SonarClient, type TransferPageFetcher } from '../sonar-client.js';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { SonarClient, defaultTransferPageFetcher, type TransferPageFetcher } from '../sonar-client.js';
 import { ReconstructionError, ZERO_ADDRESS, type TransferEvent } from '../types.js';
 
 const A = '0x' + 'a'.repeat(40);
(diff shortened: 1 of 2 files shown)


zkSoju:
## Adversarial self-review → one real defect in the fix, now hardened (`dda484f6`)

Before asking for human review I ran a FAGAN-style adversarial pass aimed at falsifying my own fix. It found a **HIGH (latent)**: the logIndex guard didn't deliver its advertised *loud refusal*. `Number('0x<address>')` → ≈9.5e47, which passes `Number.isInteger(...) && >= 0` — so an id whose suffix is a hex address silently mis-derives a garbage logIndex. And **`src/handlers/crayons.ts:12` writes such ids today** (`${txHash}_crayons_factory_${address}`). `Number()` also accepts `''`→0, `1e3`, binary/octal, whitespace. My original test (`'no-suffix'`→NaN) masked the hole.

**Hardened** (`dda484f6`): validate the raw suffix with `/^\d+$/` *before* `Number()` (same for blockNumber) + `Number.isSafeInteger` bound + `it.each` refusal tests for the address/empty/scientific cases. No live impact on the clean-id ERC-721 dogfood targets — closes the latent contract violation.

The review **validated the rest**, with stronger reasoning than the PR gave:
- **Pagination is exact.** `id` is the Envio primary key → `(blockNumber, id)` is a total order with no ties → offset paging is disjoint/complete. `reconstructOwnership` re-sorts by numeric `(blockNumber, logIndex, tokenId)` internally, so the lexical wire order is harmless.
- **Dropping `value` is safe by construction** — the only three `Transfer` writers are all ERC-721 (`erc721-holders.ts:52`, `honey-jar-nfts.ts:64`, `crayons.ts:23`); the ERC-1155 handlers write *separate* entities. There are no 1155 rows in `Transfer` to fold.
- `numeric!` typing + blockNumber coercion + the zod throw-path are correct.

## Two pre-existing adjacent findings (NOT introduced here — flagging for the team, out of this PR's scope)
1. **MED — offset pagination over a live-mutating window.** Exact only when the indexer is caught up. A concurrent *backfill* inserting rows ≤ snapshotBlock shifts offsets → skip/dup → potentially wrong holder set (reorg case is already covered by the `snapshotBlock ≤ head − confirmations` gate). Recommend gating the audit on indexer-head freshness.
2. **MED/LOW — the query filters `collection` but not `chainId`.** `Transfer` carries `chainId`; a collection id present on >1 chain would fold cross-chain transfers together (tends to fail loud in fold721 rather than silently mis-count). Inert on single-chain Berachain dogfood; add a `chainId: {_eq:}` predicate when the dogfood→standard/chain registry is wired.

PR #298, opened 2026-06-23
feat(platform/settle): The Settle Substrate (MVP) — unbypassable claimed→settled gate

zkSoju:
## The Settle Substrate (MVP) — `packages/settle/`

Makes *"is this claim `settled` or merely `claimed`?"* a mechanical, first-class primitive: a **synchronous, fail-closed, unbypassable** pre-action gate + a post-hoc detector + a first pluggable instrument. Built autonomously via `/run sprint-plan` (both sprints), with an independent adversarial review gate between implement and ship.

**Zero runtime dependencies** (Node built-ins only: `node:crypto`, `node:fs`) — small TCB by design. **113 tests pass, `tsc --noEmit` clean.**

> Venue is `packages/settle/` (platform substrate). Planning docs (`prd/sdd/sprint.md`) are gitignored by repo policy and stay local. Artifacts: `grimoires/loa/a2a/sprint-settle-{1,2}/` (local).

### Per-sprint breakdown

| Sprint | Tasks | What landed | Tests |
|--------|-------|-------------|-------|
| **1 — substrate core** | T0.1, T1.1–T1.6 | threat model · tier/claim/posture/snapshot types · 4 ports + mocks · fail-closed sha-pinned classifier · independent verifier · ed25519 signed VerificationSnapshot · **synchronous unbypassable gate** + O_APPEND trail | 81 |
| **2 — first instance + detector + trust anchors** | T2.1–T2.4 | `deployed-truth` instrument (railway + on-chain, finality/identity binding, no-LLM) · Bar Registration Authority · post-hoc detector (MISMATCH tile) · G-7 Fable-readiness · consolidated counter-example suite | +32 → 113 |

### The load-bearing guarantees (all test-enforced, teeth bite-verified)

- **Unbypassable (G-2, SKP-001b):** raw must-settle capability is package-private behind a gated facade; an import-boundary test (hardened against `import()`/`require()`/multi-line evasions) fails if anything else reaches it. Negative control: planting a bypass turns it RED.
- **Synchronous (SKP-001a):** `Gate.checkSync` returns a decision directly (compile-time + runtime asserted). Async evidence-prep writes a signed snapshot; the gate does only sync reads + sync sig-verify.
- **Independent verifier (SKP-006):** recomputes `earned_tier` from instrument re-execution; never reads the producer's self-report; import-confined.
- **Fail-closed (SKP-006):** unknown domain → FAIL_CLOSED; missing/forged/expired/untrusted-key/cross-domain snapshot → abstain.
[...]

PR #302, opened 2026-06-24
feat(settle-shadow): fatBERA→deploy-commit-truth — freeside's first real settle-gate consumer (SHADOW)

zkSoju:
## What

The **first real consumer of the settle gate in freeside.** Replaces the earlier fatBERA placeholder (an arbitrary DeFi token, no real data, external creds) with a claim that actually belongs to this build and needs **no external credentials**:

> *"Is the running code actually the commit we claim is deployed?"*

verified against **real local git**. This is the ungrounded-live-state cure (the `deployed-truth` discipline) applied to code deployment: instead of an agent asserting *"we shipped X"*, the claim runs through the settle gate, which **abstains/falsifies when the deployed commit doesn't match the claimed one**.

freeside (application layer) invokes the settle gate that descended to `loa-laplas` (brakes layer) in Phase 1, composing legba — downward-only, per The Descent Law.

## Files (2, +248)

| File | Role |
|------|------|
| `scripts/settle-shadow/deploy-commit-truth.shadow.mjs` | The consumer: `DeployCommitTruthInstrument` (git-HEAD vs pinned claim), `runShadow()` testable core, real-git `main` runner |
| `scripts/settle-shadow/deploy-commit-truth.shadow.test.mjs` | The teeth: 9 cases (HELD / FALSIFIED / INSUFFICIENT × shadow & enforce) |

## Behavior (grounded — run + observed)

- **Domain `deployed/freeside` classifies `FAIL_CLOSED`** → required tier `settled` (must-settle).
- **HELD** when running commit == pinned commit → would-proceed, `MISMATCH=0`.
- **FALSIFIED** on drift → abstain → would-HALT (the cure). Verified live: against this checkout (local HEAD ≠ `origin/main`) it correctly emits `STATUS=HALT|...|MISMATCH=1|MODE=SHADOW`.
- **INSUFFICIENT** when git is unreadable or the claim has no pinned commit (absence of evidence ≠ proof).
- **SHADOW** (default) records + trails, **never blocks**. **ENFORCE** capability is built and tested (throws `BRAKE_BLOCKED` on a must-settle abstain) — but flipping the production **default** to enforce is **operator-gated**; this PR only makes the capability exist and be exercised.

## Test signal — read this

`node --test scripts/settle-shadow/deploy-commit-truth.shadow.test.mjs` → **9/9 pass locally** (with `loa-laplas` checked out adjacently).

The consumer path-imports the adjacent laplas brake (`../../../loa-laplas/scripts/settle/settle.mjs`). **In freeside CI, laplas is NOT adjacent, so these 9 tests SKIP** rather than break the build — meaning CI provides no signal on this consumer. This is an accepted cross-repo limitation for SHADOW phase; the teeth are exercised in local + laplas-adjacent runs. Flagging it explicitly so the skip is read as intent, not as green-by-default.

## Why SHADOW / scope

Per the deployed-but-unconsumed cure: ship ONE first consumer that shadow-reads, then graduate. This is that — the settle substrate (PR #298) gets its first freeside consumer here, observing only. No production behavior changes.


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#302. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "short-sha-never-matches",
      "title": "Pinned short SHA (>=7 chars) can never equal full HEAD → permanent FALSIFIED",
      "severity": "HIGH",
      "category": "correctness",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.mjs",
      "description": "The instrument accepts any expected_commit with `length >= 7` (`if (typeof exp.expected_commit !== 'string' || exp.expected_commit.length < 7)`), but the comparison is strict full-string equality (`if (head !== exp.expected_commit)`). `readHead()` returns the full 40-char SHA from `git rev-parse HEAD`. So any operator who pins a 7–39 char abbreviated commit passes the sufficiency check and then ALWAYS falsifies, silently producing a HALT/BLOCK for code that is in fact correct. The permissive length gate advertises support for short SHAs that the equality logic cannot honor.",
      "suggestion": "Either normalize both sides before compare (`git rev-parse <expected>` to expand, or `head.startsWith(exp.expected_commit)` for prefix match with a documented minimum length), or tighten the gate to require the full 40-char SHA and return INSUFFICIENT otherwise.",
      "confidence": 0.85
    },
    {
      "id": "head-vs-origin-main-semantics",
      "title": "Claim compares local HEAD against a possibly-stale origin/main remote-tracking ref",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.mjs",
      "description": "The real claim asserts `HEAD == origin/main`: `main()` pins `expected = git rev-parse origin/main` and the instrument reads `git rev-parse HEAD`. `origin/main` is a remote-tracking ref that only updates on `git fetch`; it can be stale or, in a fresh/CI checkout, absent. This makes 'is the deployed code the claimed release commit?' actually mean 'does local HEAD equal whatever origin/main was at the last fetch?' — which falsifies for legitimate states (local commits ahead of a stale tracking ref, detached HEAD, checkout of a tag). The module comment frames this as verifying the *deployed* commit, but neither side reflects an actual deployment.",
      "suggestion": "Decide the intended invariant explicitly. If the goal is release-commit truth, pin an immutable release SHA in the bar (not a moving branch ref) and compare the running commit against it. If comparing to remote, `git fetch` first or document the stale-ref semantics.",
      "confidence": 0.6
    },
    {
      "id": "unguarded-git-in-main",
      "title": "Top-level execFileSync in main() throws uncaught when origin/main is missing",
      "severity": "MEDIUM",
      "category": "robustness",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.mjs",
      "description": "In the `main` block, `const expected = execFileSync('git', ['rev-parse', 'origin/main'], ...)` runs before and outside the `try {}` that wraps `runShadow`. If `origin/main` does not exist (shallow clone, fresh CI checkout, no remote), execFileSync throws and the script dies with a raw stack trace instead of a clean INSUFFICIENT/abstain. The instrument path deliberately treats git-unavailability as INSUFFICIENT, but this hand-run path does not get the same graceful handling.",
      "suggestion": "Wrap the `origin/main` resolution in try/catch and emit an INSUFFICIENT tile + non-error exit (or a clear message) when the ref can't be resolved, matching the instrument's INSUFFICIENT semantics.",
      "confidence": 0.7
    },
    {
      "id": "self-signed-keypair-tautology",
      "title": "Per-run generated keypair makes signature verification self-referential",
      "severity": "MEDIUM",
      "category": "security",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.mjs",
      "description": "`runShadow` calls `generateVerifierKeypair()` on every invocation and then verifies with `trustedVerifierPublicKey: publicKeyBase64` — the public half of the key it just minted. The signing/verification step therefore always succeeds by construction and establishes no trust property: nothing is bound to an external or persisted verifier identity. If the intent of routing through legba/verify is to prove the verdict came from a trusted instrument, that guarantee is absent here. This may be acceptable for local SHADOW mechanics, but it should be a conscious decision, not an implicit no-op.",
      "suggestion": "Document that the keypair is ephemeral/decorative in shadow mode, or thread a persisted/trusted verifier key so the signature actually attests something. Flag whether enforce mode is expected to use a real key.",
      "confidence": 0.55
    },
    {
      "id": "static-cross-repo-import",
      "title": "Static cross-repo import makes the module unloadable when laplas is not adjacent",
      "severity": "MEDIUM",
      "category": "portability",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.mjs",
      "description": "The module top-level statically imports `../../../loa-laplas/scripts/settle/settle.mjs` and `../../../loa-laplas/scripts/legba/legba-core.mjs`. The test file cleverly guards this via `SKIP` + dynamic import so freeside CI won't break, but the production module itself cannot be imported at all when laplas is not checked out adjacently — any static consumer throws at load. The relative `../../../` path also assumes a fixed sibling-repo layout on disk, which is brittle for a 'first real consumer' meant to generalize.",
      "suggestion": "Consider lazy/dynamic import of the laplas brake inside `runShadow` (mirroring the test), returning a clear INSUFFICIENT-style error when unavailable, and/or resolve the laplas path via an env var or package resolution rather than a hardcoded `../../../` sibling assumption.",
      "confidence": 0.6
    },
    {
      "id": "instrument-id-never-set",
      "title": "instrument.id is never assigned; envelope always uses the fallback",
      "severity": "LOW",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.mjs",
      "description": "`DeployCommitTruthInstrument` constructor stores only `this.sha` and `this.readHead`; it never sets `this.id`. In `runShadow`, `instrument_id: instrument.id ?? 'deploy-commit-truth'` therefore always resolves to the fallback, and `instrument_sha: instrument.sha` reflects the first constructor arg (named `sha`, passed as 'test'/'real'/'deploy-commit-truth@1'). The `id` vs `sha` distinction in the envelope is effectively vestigial and slightly misleading.",
      "suggestion": "Either set `this.id` in the constructor or drop the `instrument.id ??` branch so the envelope fields reflect what is actually populated.",
      "confidence": 0.8
    },
    {
      "id": "trail-parse-assumes-single-line",
      "title": "Trail-shape test parses the whole file as one JSON object",
      "severity": "LOW",
      "category": "testing",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.test.mjs",
      "description": "The 'SHADOW never enforces + trail row shape' test does `JSON.parse(readFileSync(path).trim())`, which assumes the trail contains exactly one JSON line. This is fine for a single write to a fresh per-pid path, but it couples the test to append-vs-overwrite semantics of `makeTrailWriter`. If the writer appends and the file is ever reused (e.g., pid reuse without the finally cleanup running, or a future multi-write test), the parse silently breaks.",
      "suggestion": "Parse the last non-empty line (`.trim().split('\\n').pop()`) to be robust to the JSONL append format the trail is designed for.",
      "confidence": 0.4
    },
    {
      "id": "praise-injected-readhead-determinism",
      "title": "Instrument reads git through an injected readHead, enabling deterministic tests",
      "severity": "PRAISE",
      "category": "testing",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.mjs",
      "description": "Constructor-injecting `readHead` (real `gitHeadReader` in production, stubbed async fns in tests) cleanly separates the deterministic settle logic from the git side effect. This lets the tests exercise HELD/FALSIFIED/INSUFFICIENT paths precisely without a git fixture, while still keeping one real-git integration test. Good, testable seam.",
      "suggestion": "No change needed."
    },
    {
      "id": "praise-cross-repo-skip-guard",
      "title": "Cross-repo dependency is handled with an explicit SKIP guard instead of a hard failure",
      "severity": "PRAISE",
      "category": "testing",
      "file": "scripts/settle-shadow/deploy-commit-truth.shadow.test.mjs",
      "description": "Detecting laplas adjacency (`SKIP = !existsSync(LAPLAS_BRAKE)`), wrapping tests in a skip-aware `T`, and gating the module load behind a dynamic import means freeside CI degrades to skipped rather than red when the sibling repo isn't present. The skip reason is explicit and honest about the cross-repo path-import limitation.",
      "suggestion": "No change needed."
    }
  ]
}
```


## Callouts

Issue #304, opened 2026-06-24
[shared/ci] "Unit Tests" required check is red on every PR (opossum ./lib/circuit) — merge-blind gate


zkSoju:
## The sensor is numb, and it's a required gate

`main` branch protection requires 5 status checks — `Build`, **`Unit Tests`**, `Lint`, `Security Scan`, `Docker Build` — plus 2 approving reviews, with `enforce_admins: false`.

**`Unit Tests` fails on every open PR.** It is also a *required* check. Because `enforce_admins=false`, admins merge past it every time (PRs #299, #300, #301 all merged 2026-06-24 with it red). So the gate is the worst of both worlds:

- **Non-admin contributors are frozen out** — they cannot satisfy a required check that is structurally red.
- **Admins ignore it** — it has lost all signal; a real unit-test regression would look identical to the chronic failure and merge anyway.

This is merge-blindness. The required check that's supposed to be the floor is pure ceremony.

## Root cause (grounded)

The failure is **not** broad. On a representative run (PR #282, [job 81229925912](https://github.com/0xHoneyJar/loa-freeside/actions/runs/27481481552/job/81229925912)):

```
Test Files  10 failed | 185 passed (195)
     Tests  5237 passed | 5 skipped (5242)
Error: Cannot find module './lib/circuit'
Require stack:
 - .../themes/sietch/opossum
```

**5237 tests pass.** All 10 failing suites fail at *collection* from a single unresolved module — `opossum`'s `index.js` does `require('./lib/circuit')`, and `./lib/circuit` is absent in CI. The 10 suites (`budget-drift-*`, `jwks-*`, `pii-redaction-audit`, `req-hash`, `sse-event-id`, `tier-policy-contract`, `agents.routes`) all transitively import the opossum circuit-breaker.

### Why CI can't find it but local can

- CI installs with **`npm ci`** in `themes/sietch` (`.github/workflows/ci.yml:213-215`), against `themes/sietch/package-lock.json` (last regenerated in the `@arrakis/* → @freeside/*` rename, #208).
- The working/dev install is **pnpm** (`themes/sietch/node_modules/.pnpm/opossum@9.0.0/...`), where `lib/circuit.js` is present and `require('opossum')` resolves cleanly.
[...]

Issue #84, opened 2026-02-20
[Product B] API Key Management + Developer Onboarding


deep-name:
### Context

The developer product needs self-service onboarding. Developers should be able to sign up, get an API key, and start making inference requests without manual intervention.

### Scope

1. **API Key Generation**
   - Developer registers via web form or CLI
   - System generates scoped API key (`lf_live_...` for production, `lf_test_...` for sandbox)
   - Key storage: hashed in database, cleartext shown once at creation
   - Key metadata: rate limits, allowed pools, budget cap

2. **API Key Authentication**
   - `Authorization: Bearer lf_live_...` header on inference requests
   - Key → developer account → pool access + rate limits
   - Rate limiting per API key (requests/minute, tokens/day)

3. **Self-Service Portal**
   - Key creation, rotation, revocation
   - Usage dashboard (requests, tokens, cost)
   - Billing status (credits remaining)

4. **Developer Onboarding Flow**
   - Sign up → verify email → create API key → sandbox inference → upgrade to production
   - Sandbox: free tier with rate limits, access to `cheap` pool only
   - Production: paid tier with configurable limits

### Acceptance Criteria

- [ ] Developer can create an API key via web form
- [ ] API key authenticates inference requests
- [ ] Rate limits are enforced per API key
- [ ] Developer can view usage and rotate keys
- [ ] Sandbox mode allows free testing with rate limits

### Dependencies

- **Blocked by**: [loa-freeside#77](https://github.com/0xHoneyJar/loa-freeside/issues/77) (deployment)
- **Blocked by**: [loa-finn#91](https://github.com/0xHoneyJar/loa-finn/issues/91) (OpenAPI spec — developers need to know the API contract)
- **Blocks**: External developer adoption, Product B launch

### Related

- [loa-finn#66 Command Center](https://github.com/0xHoneyJar/loa-finn/issues/66#issuecomment-3931932336) — Product B
- [loa-finn#91](https://github.com/0xHoneyJar/loa-finn/issues/91) — OpenAPI spec + SDK
- [PR #74 GTM Plan](https://github.com/0xHoneyJar/loa-freeside/pull/74) — Decisions D1 (auth), D2 (account hierarchy)

zkSoju:
Consolidated into the revenue-line tracker (https://github.com/0xHoneyJar/loa-freeside/issues/313) → command center loa-finn#66. Closing as duplicate-of-truth (Q1 triage 2026-06-25).

Issue #83, opened 2026-02-20
[Track 4] Billing Audit Trail + Pool Enforcement Transparency


janitooor:
### Context

Communities won't adopt infrastructure they can't audit. The billing system needs transparent audit trails, and pool enforcement decisions need to be visible to community admins.

### Scope

1. **Billing Audit Trail Export**
   - Downloadable JSONL of all credit operations per community
   - Includes: timestamp, operation type, amount, model pool, user, conservation guard result
   - Retention: configurable (default 90 days)
   - Format: JSONL for streaming, CSV for spreadsheet analysis

2. **Pool Enforcement Transparency**
   - Community admins can see which model pools their agents have access to
   - Clear explanation of why: tier level, conviction score, community budget
   - Visibility into pool enforcement decisions (confused deputy rejections)

3. **Conservation Guard Status**
   - Public endpoint: `GET /api/v1/admin/community/:id/conservation`
   - Shows: current invariant status, last check timestamp, any violations
   - Historical: invariant verification log

### Acceptance Criteria

- [ ] Community admin can download billing audit trail as JSONL
- [ ] Audit trail includes all credit operations with conservation guard results
- [ ] Pool access list is visible per community with explanations
- [ ] Conservation guard status is queryable via API
- [ ] No PII leaks in audit exports (wallet addresses are pseudonymous)

### Dependencies

- **Blocked by**: [loa-freeside#77](https://github.com/0xHoneyJar/loa-freeside/issues/77) (deployment)

### Related

- [loa-finn#66 Command Center](https://github.com/0xHoneyJar/loa-finn/issues/66#issuecomment-3931932336) — Track 4, Tasks 4.2, 4.3
- [loa-finn#90](https://github.com/0xHoneyJar/loa-finn/issues/90) — Conservation guard Prometheus metrics

janitooor:
## Update: Post PR #88 Merge (2026-02-23)

### Delivered in PR #88
- Append-only `usage_events` table with `BEFORE UPDATE/DELETE` triggers (migration 067)
- Capability audit event emitter (`capability-audit.ts`) with structured events
- Three audit event types: `pool_access`, `byok_usage`, `ensemble_invocation`
- Per-model cost attribution in ensemble accounting (`model_breakdown` array)
- Admin audit log routes

### Bridge Review Note
The append-only trigger enforcement was cited as **PRAISE** in the Bridgebuilder review:
[...]

Issue #82, opened 2026-02-20
[Track 3] Community Budget Dashboard


deep-name:
### Context

Community admins need visibility into spend, usage, and model allocation. Without a dashboard, communities can't manage their agent budgets. This is what makes communities adopt the infrastructure.

### Scope

1. **Dashboard UI**
   - Total credit spend (current period + historical)
   - Usage breakdown by model pool (cheap/fast-code/reviewer/reasoning/architect)
   - Per-user usage breakdown
   - Credit balance + projected depletion date
   - Recent inference history (timestamp, model, tokens, cost)

2. **Admin Endpoints**
   - `GET /api/v1/admin/community/:id/usage` — usage summary
   - `GET /api/v1/admin/community/:id/billing` — billing history
   - `GET /api/v1/admin/community/:id/agents` — active agents and their tiers

3. **Visualization**
   - Simple web page (or Discord embed) with charts
   - Exportable as CSV/JSON for external analysis

### Acceptance Criteria

- [ ] Community admin can view total spend and usage breakdown
- [ ] Usage is broken down by model pool
- [ ] Credit balance and projected depletion are visible
- [ ] Dashboard is accessible via web and/or Discord command
- [ ] Data exports in CSV/JSON format

### Dependencies

- **Blocked by**: [loa-freeside#77](https://github.com/0xHoneyJar/loa-freeside/issues/77) (deployment)
- **Enhanced by**: [loa-finn#90](https://github.com/0xHoneyJar/loa-finn/issues/90) (Prometheus metrics — can feed dashboard)

### Related

- [loa-finn#66 Command Center](https://github.com/0xHoneyJar/loa-finn/issues/66#issuecomment-3931932336) — Track 3, Task 3.4
- [Issue #50](https://github.com/0xHoneyJar/loa-freeside/issues/50) — Budget Topology & Trust Gradient Visualizer

janitooor:
## Update: Post PR #88 Merge (2026-02-23)

### Delivered in PR #88
- BudgetManager with two-counter atomic reservation (Redis Lua)
- Usage events table with append-only enforcement (migration 067)
- Per-community budget tracking (monthly reset, keyed by YYYY-MM)
- Capability audit events (pool_access, byok_usage, ensemble_invocation)
- Admin agent routes (`/admin/agent/stats`, `/admin/agent/audit`)
- Tier access mapping with per-community configuration

### Remaining
- [ ] Dashboard UI: visual spend breakdown by model pool
[...]

PR #307, opened 2026-06-24
fix(cheval-verdict): a degenerate chain cannot report "consensus"

zkSoju:
## What

Closes the verdict-honesty bug `arrakis-verdict-consensus-lie-qzo4` (P1): the cheval verdict emitter stamped `consensus_outcome: "consensus"` on chains where **zero voices succeeded**.

Observed live in `.run/model-invoke.jsonl`:
```
{voices_succeeded:0, chain_health:"exhausted", status:"FAILED", consensus_outcome:"consensus"}
```
`status` was honest (FAILED) — but the consensus field lied: it claimed the models agreed when zero models produced anything. Any consumer keying off consensus rather than status (e.g. Flatline `HIGH_CONSENSUS` auto-integrate) could be fooled into integrating a failed review.

## Root cause

Two stamping sites conflate "0 voices (nothing happened)" with "1 voice (nothing to contradict)":
- `consensus.py` `classify_consensus()`: `voices_succeeded < 2 -> "consensus"` includes 0.
- the single-voice producer + aggregate placeholder hardcode `"consensus"`.

## Fix (status-preserving)

- `consensus.py`: `voices_succeeded == 0 -> "impossible"` (1 stays consensus).
- `quality.py`: reconcile at `emit_envelope_with_status` — the single chokepoint every producer flows through — so no emission can carry `"consensus"` while `voices_succeeded==0` / `chain_health=="exhausted"`. `"impossible"` is the honest in-vocabulary value (schema enum `consensus|impossible`), and both states already auto-promote to FAILED → **no status regression**.

## Tests

Test-first. `TestConsensusHonesty` + `TestClassifyConsensusZeroVoices` reproduce the live MODELINV shape (fail-before), with guards that 1/1 and partial success **stay** consensus (no over-correction). Updated the one pre-existing test that codified the lie. cycle-109 conformance fixtures left intact — they deliberately pair a misleading consensus with an exhausted chain to prove `compute_status` returns FAILED independently (complementary defense-in-depth at the compute layer; this fix closes the lie at the emit layer).

- verdict/consensus suite: **77 pass** · conformance bats **16/16**
- the 36 unrelated failures in the full adapters suite are pre-existing env/version drift (cursor-CLI error classification, opus-4-7→4-8 model bump, mock-mode) — none import the changed modules.

> cheval routing is operator-gated — **review + merge is yours**, not auto. Repo main also carries the known-numb Unit Tests gate (#304); CI red here is not a per-PR signal.

### DIFF
--- a/.claude/adapters/loa_cheval/verdict/consensus.py
+++ b/.claude/adapters/loa_cheval/verdict/consensus.py
@@ -114,6 +114,14 @@ def classify_consensus(
       location from another voice.
     """
     voices_succeeded = int(envelope.get("voices_succeeded") or 0)
+    # arrakis-verdict-consensus-lie-qzo4: a chain where NO voice succeeded
+    # reached no agreement at all — calling that "consensus" is a lie. The
+    # honest in-vocabulary value is "impossible" (compute_verdict_status maps
+    # it to FAILED, which a 0-voice chain already is). Distinct from the
+    # single-voice case below, which IS honest (degenerate) consensus: one
+    # voice cannot contradict itself.
+    if voices_succeeded == 0:
+        return _OUTCOME_IMPOSSIBLE
     if voices_succeeded < 2:
         return _OUTCOME_CONSENSUS
 
--- a/.claude/adapters/loa_cheval/verdict/consensus.py
+++ b/.claude/adapters/loa_cheval/verdict/consensus.py
@@ -20,8 +20,15 @@ Inputs:
                         comparison.
 
 Algorithm:
-  1. Trivial cases: voices_succeeded < 2 → "consensus" (cannot have
-     contradiction with fewer than 2 voices).
+  1. Degenerate / trivial cases:
+       voices_succeeded == 0 → "impossible" — NO voice produced a verdict, so
+         no agreement exists (arrakis-verdict-consensus-lie-qzo4). "impossible"
+         here denotes "no consensus was reachable", which compute_verdict_status
+         maps to FAILED — the same status a 0-voice chain already carries. This
+         broadens "impossible" beyond the cross-voice-contradiction sense in
+         step 3 to also cover the no-agreement-possible (empty/exhausted) case.
+       voices_succeeded == 1 → "consensus" — a single voice cannot contradict
+         itself; that is honest (degenerate) consensus.
   2. Per-finding cross-voice comparison: for each BLOCKER-severity finding
      emitted by any voice, check whether at least one other voice's
      findings cover the same (file_path, line_number) location AND
--- a/.claude/adapters/loa_cheval/verdict/quality.py
+++ b/.claude/adapters/loa_cheval/verdict/quality.py
@@ -206,14 +206,46 @@ def compute_verdict_status(envelope: Dict[str, Any]) -> str:
     return _STATUS_DEGRADED
 
 
+def _reconcile_degenerate_consensus(envelope: Dict[str, Any]) -> None:
+    """arrakis-verdict-consensus-lie-qzo4 — a degenerate chain reached no
+    agreement, so it MUST NOT carry ``consensus_outcome == "consensus"``.
+
+    Every producer flows through ``emit_envelope_with_status``, but only the
+    aggregate path calls ``classify_consensus``; the single-voice producer
+    (``cheval.cmd_invoke``) and the aggregate placeholder both hardcode
+    ``"consensus"``. That is how a 0-voice / exhausted chain ended up stamped
+    ``consensus_outcome="consensus"`` while ``status="FAILED"`` (observed live
+    in ``.run/model-invoke.jsonl``) — a lie that fools any consumer keying off
+    consensus rather than status (e.g. Flatline HIGH_CONSENSUS auto-integrate).
+
+    Reconcile at this single chokepoint: when no voice succeeded OR the chain
+    exhausted, the honest in-vocabulary value is ``"impossible"``. Both of
+    those states already auto-promote to FAILED in ``compute_verdict_status``,
+    so this is status-preserving — it only removes the false consensus claim.
+    A single succeeded voice is left untouched: 1/1 is honest consensus.
+    """
+    voices_succeeded = int(envelope.get("voices_succeeded") or 0)
+    chain_health = envelope.get("chain_health", _CHAIN_HEALTH_OK)
+    degenerate = (
+        voices_succeeded == 0 or chain_health == _CHAIN_HEALTH_EXHAUSTED
+    )
+    if degenerate and (
+        envelope.get("consensus_outcome") == _CONSENSUS_OUTCOME_CONSENSUS
+    ):
+        envelope["consensus_outcome"] = _CONSENSUS_OUTCOME_IMPOSSIBLE
+
+
 def emit_envelope_with_status(envelope: Dict[str, Any]) -> Dict[str, Any]:
     """Single producer entry-point per v5 SKP-003 closure.
 
     Pipeline:
       1. ``validate_invariants`` — raises ``EnvelopeInvariantViolation``
          on malformed input. Producer MUST NOT emit an unvalidated envelope.
-      2. ``compute_verdict_status`` — writes the canonical status field.
-      3. Return the mutated envelope (the input dict is mutated in-place
+      2. ``_reconcile_degenerate_consensus`` — a no-voice / exhausted chain
+         cannot carry ``consensus_outcome="consensus"`` (the verdict-honesty
+         invariant; arrakis-verdict-consensus-lie-qzo4).
+      3. ``compute_verdict_status`` — writes the canonical status field.
+      4. Return the mutated envelope (the input dict is mutated in-place
          AND returned; callers can rely on either reference).
 
     Error paths (SDD §6.2) flow through this same function with the same
@@ -222,6 +254,7 @@ def emit_envelope_with_status(envelope: Dict[str, Any]) -> Dict[str, Any]:
     paths that bypass this helper.
     """
     validate_invariants(envelope)
+    _reconcile_degenerate_consensus(envelope)
     envelope["status"] = compute_verdict_status(envelope)
     return envelope
 
--- a/.claude/adapters/tests/test_classify_consensus.py
+++ b/.claude/adapters/tests/test_classify_consensus.py
@@ -52,12 +52,19 @@ def _finding(severity: str, file_path: str = "src/app.py",
 # ---------------------------------------------------------------------------
 
 
-def test_zero_succeeded_returns_consensus():
-    """voices_succeeded=0: no findings to compare — trivially 'consensus'."""
+def test_zero_succeeded_returns_impossible():
+    """voices_succeeded=0: NO voice produced a verdict, so there is no
+    agreement to report. arrakis-verdict-consensus-lie-qzo4 — the earlier
+    "no findings to compare → trivially consensus" reading produced a
+    downstream lie (consensus_outcome="consensus" stamped on 0-voice/FAILED
+    chains, observed live in .run/model-invoke.jsonl). The honest
+    in-vocabulary value is "impossible" (→ compute_verdict_status → FAILED,
+    which a 0-voice chain already is). One-voice success below is distinct
+    and remains honest consensus."""
     from loa_cheval.verdict.consensus import classify_consensus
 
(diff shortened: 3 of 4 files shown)


zkSoju:
**Cross-model review (codex + cursor on the diff).** Verdict: **the fix is sound** — zero-voice/exhausted chains no longer emit `consensus_outcome="consensus"`, `status` is unchanged (already FAILED), and healthy paths (1/1, partial DEGRADED, full APPROVED) are not flipped. Cursor independently confirmed *"no in-repo consumer keys production logic solely on `consensus_outcome`; flatline readiness uses `chain_health`"* — so this is a misleading-field / defense-in-depth fix, not a confirmed exploitable auto-integrate path (matching the PR's framing).

Applied the one legitimate finding (doc-drift): `consensus.py`'s algorithm docstring described the old `< 2 → consensus` rule (`c389eca0`). Noted the acknowledged semantic overload — `impossible` now covers both cross-voice contradiction and the no-agreement-possible case — which is operationally safe since both map to FAILED.

Other findings were either already-covered (the new `TestConsensusHonesty` pytest locks the emit layer; cursor only saw the source diff, not the tests) or out of scope (no schema enum change; historical `.run/model-invoke.jsonl` rows not backfilled).

zkSoju:
## Governed council validated this PR clean

Ran the governed cross-model council (construct-fagan #11) on this diff to complete the self-review pass across all the session's security PRs. **codex APPROVED (0 findings)**; **cursor returned a bare `CHANGES_REQUIRED` with an empty `findings:[]`** (a verdict label with no concrete objection); claude dropped (the `arrakis-bx1j` fable issue), `multi_perspective_met=true` (codex + cursor = 2 families).

No actionable findings — the consensus-honesty fix (consensus on a 0-voice / FAILED envelope is impossible, reconciled at the single producer chokepoint) stands. This is the 5th and last security PR self-reviewed via the now-reliable council; the other four surfaced real holes (#309: 4, #82: 6 incl. a converged CRITICAL, #83: 1 custody bypass) and one verified false-positive (#84), test-first throughout.

zkSoju:

zkSoju (commented):
## Summary
Solid, well-targeted fix: it moves the honesty enforcement to the single producer chokepoint and pins the boundary with tests that guard against over-correction. The two `MEDIUM` findings both hinge on the same underspecified question — what `chain_health=="exhausted"` *means* when voices actually succeeded — and the answer decides whether one branch is a latent bug and whether `"impossible"` is now semantically overloaded.

The architectural spine here is worth naming: this PR relocates an invariant from *N producers* to *one reconciler*. That's the right direction — it's the difference between validating at every write site and validating at the schema boundary. The remaining questions are all about the *vocabulary* the reconciler speaks, not *where* it sits.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "exhausted-clobbers-real-consensus",
      "title": "chain_health==\"exhausted\" override may clobber a legitimate multi-voice consensus",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": ".claude/adapters/loa_cheval/verdict/quality.py",
      "description": "_reconcile_degenerate_consensus treats degenerate as `voices_succeeded == 0 OR chain_health == exhausted`. classify_consensus only downgrades on voices_succeeded==0. If a chain can carry chain_health==\"exhausted\" while voices_succeeded >= 2, the reconcile step will overwrite a genuinely-reached consensus with \"impossible\". None of the added tests exercise this combination — test_partial_success_keeps_consensus uses chain_health=\"degraded\", not \"exhausted\".",
      "suggestion": "Confirm the semantics of chain_health==\"exhausted\": if it can co-occur with voices_succeeded>=2, gate the exhausted branch on voices_succeeded<2 (or drop the chain_health check and rely on the voice count). Add a test for (voices_succeeded=2, chain_health=\"exhausted\").",
      "confidence": 0.5,
      "faang_parallel": "This mirrors AWS's design rule for health checks: a node is 'unhealthy' (liveness) and a node is 'shedding load' (retry-exhaustion) are orthogonal axes. Conflating them was behind several early ELB incidents where draining nodes were marked dead and quorum was needlessly broken.",
      "metaphor": "A jury reaches a verdict, then the bailiff announces 'we ran out of time to seat the last two jurors' — and the clerk throws out the verdict the seated jurors already agreed on. Running out of chairs isn't the same as disagreement.",
      "teachable_moment": "When two predicates claim to define the same concept ('degenerate') but use different inputs, the divergence IS the spec question. Write the failing-case test first; if you can't construct the input, that's your proof the branch is dead — document it as such.",
      "connection": "Directly coupled to duplicated-degeneracy-definition: the two definitions disagree precisely on this axis."
    },
    {
      "id": "impossible-semantic-overload",
      "title": "\"impossible\" now conflates no-voices-succeeded with cross-voice contradiction",
      "severity": "MEDIUM",
      "category": "design",
      "file": ".claude/adapters/loa_cheval/verdict/consensus.py",
      "description": "The change reuses \"impossible\" for the empty/exhausted case, broadening it beyond the cross-voice-contradiction meaning it carries in step 3. Any downstream consumer reading consensus_outcome==\"impossible\" as \"voices actively disagreed → escalate to human\" will now also fire on 0-voice/infra-failure chains. status distinguishes these (both FAILED), but the point of this PR is that consumers key off consensus_outcome, not status — so the overload reintroduces ambiguity at the same field.",
      "suggestion": "Consider a distinct in-vocabulary value (e.g. \"no_quorum\" / \"unreached\") for the empty/exhausted case. If the vocabulary is fixed and both must map to FAILED, document the conflation at the consumer contract, not just in-code.",
      "confidence": 0.45,
      "faang_parallel": "HTTP's split between 4xx and 5xx exists for exactly this reason — 'you disagreed with me' vs 'I couldn't function' route to different handlers. Collapsing 502 into 409 would break every client that retries one but escalates the other.",
      "metaphor": "Labeling both 'the witnesses contradicted each other' and 'no witnesses showed up' as 'impossible' is like a doctor writing 'inconclusive' for both a contradictory test and a machine that never turned on. Same word, opposite next actions.",
      "teachable_moment": "Fixing a lie by widening an existing enum value trades a false-positive for an ambiguity. The honest fix often needs a NEW symbol, not a reused one — vocabulary is cheaper than the debugging cost of overload.",
      "connection": "This is the inverse risk of exhausted-clobbers: there the reconciler is too aggressive; here its output word is too coarse for consumers to act on."
    },
    {
      "id": "duplicated-degeneracy-definition",
      "title": "Degeneracy definition duplicated across two functions with divergent criteria",
      "severity": "LOW",
      "file": ".claude/adapters/loa_cheval/verdict/quality.py",
      "description": "classify_consensus (voices_succeeded==0) and _reconcile_degenerate_consensus (voices_succeeded==0 OR chain_health==exhausted) each independently define degeneracy and disagree on the exhausted case. Defensible as defense-in-depth, but the definitions can drift and a reader must cross-reference both to know the effective rule.",
      "suggestion": "Extract a single `is_degenerate(envelope) -> bool` predicate and call it from both sites, or add a comment cross-linking the two so the intentional asymmetry (if it is intentional) is explicit.",
      "confidence": 0.5,
      "faang_parallel": "The Linux kernel treats duplicated boolean predicates as a bug class — helpers like `pid_alive()` exist so 'what counts as alive' has exactly one home. Divergent inline copies are what static analyzers flag as 'semantic clones.'",
      "metaphor": "Two clocks in the same house set to slightly different times. Nobody notices until they're asked to agree on when to leave.",
      "teachable_moment": "Defense-in-depth and single-source-of-truth aren't in conflict — you can have two enforcement SITES calling one PREDICATE. Duplicate the guard, not the definition.",
      "connection": "The drift feared here is already realized in exhausted-clobbers-real-consensus."
    },
    {
      "id": "reconcile-only-catches-consensus-value",
      "title": "Reconcile only corrects the literal \"consensus\" value, not other dishonest outcomes",
      "severity": "LOW",
      "category": "correctness",
      "file": ".claude/adapters/loa_cheval/verdict/quality.py",
      "description": "The guard fires only when consensus_outcome == _CONSENSUS_OUTCOME_CONSENSUS. A degenerate chain carrying a different non-\"impossible\" value (e.g. a producer hardcoding \"contradiction\") would pass through unreconciled. Given the two known producers hardcode \"consensus\" this is currently narrow, but the chokepoint is framed as the single honesty enforcement point and only enforces against one value.",
      "suggestion": "If this is meant to be the invariant enforcement point, assert/normalize that a degenerate chain carries \"impossible\" regardless of the incoming value, rather than only rewriting the \"consensus\" literal.",
      "confidence": 0.35,
      "faang_parallel": "This is allowlist-vs-blocklist. Google's schema validators reject anything not proven valid rather than patching known-bad values — a blocklist is only as complete as your imagination of bad inputs.",
      "metaphor": "A spell-checker that only fixes 'teh' but waves through every other typo. Fine until someone types 'thier.'",
      "teachable_moment": "An 'invariant enforcement point' should enforce the invariant (degenerate ⇒ impossible), not patch one known violation of it. State the positive rule, then normalize toward it.",
      "connection": "Same root as duplicated-degeneracy-definition — the reconciler encodes 'what to fix' rather than 'what must be true.'"
    },
    {
      "id": "praise-overcorrection-guards",
      "title": "Tests explicitly guard against over-correction",
      "severity": "PRAISE",
      "category": "testing",
      "file": ".claude/adapters/tests/test_verdict_quality_classifier.py",
      "description": "test_single_voice_success_stays_consensus (1/1 → honest consensus) and test_partial_success_keeps_consensus (2/3 → DEGRADED but still consensus) pin the boundary of the fix, not just the happy path of the bug. This prevents a future tightening from silently swallowing legitimate degraded-but-real consensus. Tests anchor to the live failure shape from .run/model-invoke.jsonl and assert status is preserved alongside the corrected field.",
      "suggestion": "No change needed. Consider adding the (voices_succeeded>=2, chain_health=\"exhausted\") case to complete the boundary coverage.",
      "confidence": 0.9,
      "faang_parallel": "This is Google's 'test the boundary, not the bug' discipline and Hyrum's Law defense in one: the negative tests fence the fix so a later over-tightening trips CI instead of prod.",
      "metaphor": "You didn't just patch the pothole — you painted lines on both sides so the next crew doesn't repave over the good lane.",
      "teachable_moment": "The mark of a mature fix isn't the test that reproduces the bug; it's the test that fails if someone over-corrects. Guarding both edges of a boundary is what turns a patch into a spec.",
      "connection": "The one gap in an otherwise complete fence is exactly the exhausted+success cell — see exhausted-clobbers-real-consensus."
    },
    {
      "id": "praise-single-chokepoint",
      "title": "Correction placed at the single producer chokepoint, status-preserving",
      "severity": "PRAISE",
      "category": "design",
      "file": ".claude/adapters/loa_cheval/verdict/quality.py",
      "description": "Reconciling in emit_envelope_with_status (which all producers flow through) rather than patching each hardcoded producer is the right lever, and running it before compute_verdict_status while relying on the fact that both degenerate states already promote to FAILED keeps the change status-preserving — it only removes the false claim. The docstring pipeline numbering was updated to match.",
      "confidence": 0.8,
      "faang_parallel": "This is the API-gateway / sidecar pattern — enforce the cross-cutting invariant at the one funnel every request traverses, not in each service. Envoy exists so N services don't each reimplement the same policy.",
      "metaphor": "One metal detector at the single entrance beats asking every shop inside the mall to frisk its own customers.",
      "teachable_moment": "When an invariant is violated by many producers, resist the urge to fix each. Find the funnel they all pass through and enforce once — and keeping it status-preserving means the blast radius is 'removes a lie,' nothing more.",
      "connection": "The chokepoint is well-placed; the open questions (impossible-semantic-overload, reconcile-only-catches-consensus-value) are about the vocabulary it emits, not its location."
    }
  ]
}
```


## Callouts
The core instinct here is excellent: **you fixed a lie at the boundary, not at the call sites, and you fixed it without moving the verdict** — both degenerate states already promote to `FAILED`, so the reconciler only strips the false "consensus" claim. That's the surgical version of this change.

The tests are the standout. Pinning `1/1 → consensus` and `2/3 DEGRADED → consensus` means a future engineer who tightens the guard too far gets a red CI instead of a silent regression — you fenced *both* sides of the boundary, which is what separates a patch from a spec.

The single thread tying every open question together: **what does `exhausted` mean when voices succeeded?** Answer that once, and it resolves the MEDIUM bug, the enum overload, and the duplicated predicate simultaneously. That's a decision-trail note worth leaving in the docstring for the next reader.

PR #308, opened 2026-06-24
fix(ci): secret-scanning full-scan on push/schedule (un-numb the gate)

zkSoju:
TruffleHog `base==head` on push/schedule made Secret Scanning red on every commit (effectively OFF). Fix: diff-scan on PRs, full-tree scan on push/schedule. Proven in 0xHoneyJar/loa-constructs#267 (merged; main secret-scan now green). Estate-wide pattern — ~27 repos affected.

### DIFF
--- a/.github/workflows/secret-scanning.yml
+++ b/.github/workflows/secret-scanning.yml
@@ -27,8 +27,10 @@ jobs:
         continue-on-error: true
         with:
           path: ./
-          base: ${{ github.event.repository.default_branch }}
-          head: HEAD
+          # Diff-scan on PRs; full-tree scan on push/schedule (empty base+head ->
+          # TruffleHog full-scan mode, avoiding the BASE==HEAD error that failed every push).
+          base: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || '' }}
+          head: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || '' }}
           extra_args: --only-verified --no-update
 
       - name: Run GitLeaks


zkSoju:

zkSoju (commented):
## Summary
This one-line ternary fix correctly rescopes PR diffs to `github.event.pull_request.base.sha`/`head.sha` and uses empty-string fallback to trigger TruffleHog full-scan on push/schedule — a clean, well-commented correction to the BASE==HEAD spurious failure. The primary open question is framing: the title says "un-numb the gate," but `continue-on-error: true` still means the workflow detects secrets without blocking on them.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Gate still non-blocking: continue-on-error remains true",
      "severity": "MEDIUM",
      "category": "security",
      "file": ".github/workflows/secret-scanning.yml",
      "description": "The scan now runs correctly on push/schedule, but with continue-on-error: true a verified secret detection won't fail the job. This un-numbs the *scan*, not the *gate* — detection without enforcement.",
      "suggestion": "Decide intent explicitly. To block: drop continue-on-error on the push/schedule path, or add a follow-up step that fails on verified findings. To stay detection-only: rename the framing so 'gate' isn't misleading.",
      "confidence": 0.6,
      "faang_parallel": "Google's BeyondCorp treats detection and enforcement as separate maturity stages — you ship monitoring first, then flip to blocking once false-positive rates are known. The danger is a control that *looks* enforced on the dashboard but is monitor-only in prod.",
      "metaphor": "A smoke detector wired to a light, not a sprinkler. It sees the fire and tells you — but nothing stops it.",
      "teachable_moment": "A 'gate' is defined by its failure behavior, not its trigger coverage. Coverage answers 'did it run?'; enforcement answers 'what happens on a hit?' — different axes.",
      "decision_trail": "Document whether secret-scanning is advisory or blocking, and the false-positive threshold for flipping to blocking."
    },
    {
      "id": "F2",
      "title": "Full-tree scan on every push may be slow / rate-limited on large histories",
      "severity": "LOW",
      "category": "performance",
      "file": ".github/workflows/secret-scanning.yml",
      "description": "Empty base+head means whole-history scan on every push and schedule; --only-verified adds outbound verification per candidate. Fine now, grows with history.",
      "suggestion": "If push scans slow down, scan the push event's before..after range on push and reserve full-tree for the scheduled run.",
      "confidence": 0.4,
      "metaphor": "Re-reading the whole library every time one page changes. Correct, but the reading time grows with the shelves.",
      "teachable_moment": "Incremental-on-change + full-on-schedule is the standard coverage/cost split for repo-wide scanners."
    },
    {
      "id": "F3",
      "title": "Push events could still pass BASE==HEAD-style range instead of full scan — verify event coverage",
      "severity": "LOW",
      "category": "correctness",
      "file": ".github/workflows/secret-scanning.yml",
      "description": "The ternary special-cases only pull_request; all other events fall through to empty base/head. Behavior depends on the action treating empty strings as full-scan, which is version-dependent.",
      "suggestion": "Pin the TruffleHog action to a SHA and confirm its docs specify empty base+head => full-scan for that version, so a future release can't silently change semantics.",
      "confidence": 0.35,
      "connection": "This is the classic implicit-contract-with-a-third-party-action risk: correctness rests on undocumented-per-your-pin behavior of an upstream you don't control."
    },
    {
      "id": "F4",
      "title": "Correct switch to event-scoped SHAs and clear inline rationale",
      "severity": "PRAISE",
      "file": ".github/workflows/secret-scanning.yml",
      "description": "Diffing against the actual merge base SHAs rather than a branch name is the right scoping, and the inline comment explains both the branch behavior and the BASE==HEAD failure it fixes.",
      "confidence": 0.9,
      "faang_parallel": "Mirrors how Meta and Google CI pin diffs to explicit base/head SHAs, not mutable branch refs — a moving default branch under a running job is a known flake source.",
      "metaphor": "Citing a book by ISBN and edition, not 'the latest printing' — the reference stays stable even as the shelf changes.",
      "teachable_moment": "The comment naming the *specific* failure (BASE==HEAD) is the gift: it turns a cryptic ternary into a breadcrumb the next reader can follow without git-blame archaeology."
    }
  ]
}
```


## Callouts
**[Quality] F4** — Pinning to `base.sha`/`head.sha` over `default_branch`/`HEAD` removes a real class of flake: a mutable ref shifting mid-run. Pairing the fix with a comment that names the exact symptom it cures is exactly how CI config stays maintainable across handoffs.

PR #315, opened 2026-06-25
fix(events): close publishEnvelope bypass — structural un-export fence + lint backstop (#255)

### LINKED ISSUE #255: [GAP] @0xhoneyjar/events contract is asymmetrically enforced — producers publish without schema-validating; divergence is invisible until a live message crosses the wire
## atomic learning
A shared schema lib that is **validated on consume but not on produce** is not a shared contract — it is a one-sided check that gives false confidence. Producer↔consumer divergence stays **structurally invisible until the first real message crosses the wire in prod**.

Vindicated **3× in one session** during the sonar→NATS→Ruggy go-live (issue 0xHoneyJar/sonar-api#50). Every map-gate passed while the territory failed:

| defect | gate that passed | the fiction it passed on | where it actually broke |
|---|---|---|---|
| `context.log` (no such API in ponder 0.16.6) | FAGAN diff-review + 83/83 vitest | test **mocked** `context` *with* a `.log` | crash-looped on the 1st real `GeneralMints:Minted` |
| `encoded_traits` `^0x[0-9a-f]+$` vs real decimal-semicolon string | byte-parity test | **fake** fixture `0xdeadbeefcafe` (hex by construction) | bot rejected the 1st real envelope `payload-schema-invalid` |

Root: `publishEnvelope` signs + ships whatever payload it is handed; only `subscribeEnvelope` schema-validates. And the consumer had **never received a real envelope** (only a synthetic one) — so the wrong `encoded_traits` assumption (`opaque hex` in the schema comment, vs the contract's decimal-semicolon `traits` string) was never falsifiable.

## levers (candidate — surfaced, not decided)
1. **Symmetric enforcement** — `publishEnvelope` validates its payload against the per-subject schema at publish time → divergence fails in the *producing* cell's CI, not on the consumer's wire. **(the real fix; PR #254 only relaxed the schema to unblock)**
2. **Territory gate** — a staging end-to-end `real-on-chain-event → real NATS → real consumer schema` run in the port/migration checklist, *before* a handler is declared "freed."
3. **Fixtures from reality** — a fixture that cannot fail the way prod fails (`0xdeadbeefcafe`) manufactures false confidence; derive fixtures from captured on-chain samples.

## related observation (same class, different failure mode)
The live announce posted **text-only**: `[announce-mint] inventory module unavailable — Cannot find module '@0xhoneyjar/inventory'` → `hasImage:false`. Another workspace `@0xhoneyjar/*` dep not materialized in the consumer build, surfacing only live. Reinforces lever 2 (a staging announce would have shown the degradation). Tracked for a follow-up.

## status
Both schema defects fixed + **live-confirmed** (Ruggy announced token 3247 to #stonehenge). PR #254 lands the events relax on `main`. Lever 1 is the open, genuinely-actionable follow-up.

\`\`\`yaml
hivemind:
  artifact_type: atomic-learning
  workstream: tech-debt
  priority: medium
  product_area: "Cross-cell event contract (@0xhoneyjar/events)"
  jtbd: { category: personal, description: "Reassure Me This Is Safe — that a 'freed' producer actually reaches the consumer" }
  learning_status: strongly-validated
  source: team-internal
\`\`\`

zkSoju:
## Close the events #255 producer/consumer asymmetry — structural fence + lint backstop

Convergence-loop turn 1. The event protocol enforced its payload-schema contract **asymmetrically**: the consumer (`subscribeEnvelope`) mandatorily validates (`subscriber.ts:299-307`), the sanctioned producer path (`emit()`) validates before signing (`emit.ts:261-265`) — but raw `publishEnvelope` was publicly exported and signed-without-validating. A consumer could `import { publishEnvelope }` and ship a schema-divergent payload (GitHub #255, `learning-status:strongly-validated`).

### The close — deletion-over-denylist (primary), lint backstop (secondary)

**1. Structural fence (the real control).** `publishEnvelope` is **un-exported** — removed from the `index.ts` barrel and the `./publisher` subpath deleted from `package.json` `exports` (following the package's own `./transport` containment precedent, FR-7). The raw capability is now **unreachable from any consumer**; the only public publish path is the validating `emit()`. `InMemoryPrevHashStore`/`PrevHashStore` remain exported for `emit()` consumers.

**2. Lint backstop.** `events-lint.mjs` (already run `--enforce` by `schema-emission-floor.yml`) now also flags any in-monorepo relative deep-import of `publishEnvelope` — a defense-in-depth tripwire behind the fence.

### Proof — deterministic, against the REAL artifacts (no mock)

**Fence (Node v22 resolver, against the actual `exports` map):**
| consumer specifier | result |
|---|---|
| `@0xhoneyjar/events/publisher` | `ERR_PACKAGE_PATH_NOT_EXPORTED` |
| `@0xhoneyjar/events/dist/src/publisher.js` | `ERR_PACKAGE_PATH_NOT_EXPORTED` |
| `@0xhoneyjar/events/src/publisher.js` (despite `files` shipping `src`) | `ERR_PACKAGE_PATH_NOT_EXPORTED` |
| `@0xhoneyjar/events` (barrel) / `@0xhoneyjar/events/emit` | resolve; `publishEnvelope` **not** a named export |

Zero `export *` wildcards; the only non-comment `publishEnvelope` binding outside `publisher.ts` is `emit.ts`'s internal import (never re-exported); zero in-repo consumers of the removed symbols → compiles, breaks nothing.

**Lint backstop matrix (real binary `--enforce`):** single-line / multi-line / namespace import → **exit 1**; interface-member, object-key, inline/block comment, `export type`, inline `{type}`, no-semicolon local export, quote-bearing regex → **exit 0**; clean tree → **exit 0**.

### Provenance (honest)
Built via `spiral-harness.sh` (cycle-coherence-2); the harness died on budget in a post-implementation evidence loop (false-positive "missing artifact" paths), so its REVIEW/AUDIT phases didn't run. Compensated by **three independent FAGAN (cross-model) review rounds** that found and drove fixes for: F1 multi-line miss, F2 namespace miss, F3 comment FPs, F4 dead shrink-gate code, then NF1 span-FP, NF2 regex-desync FP, NF3 destructure FN (mooted by the fence), NF6 no-semicolon FP — each proven against the real binary. The branch was rebased onto clean `origin/main`; an out-of-scope System-Zone edit the run introduced was dropped. Full FAGAN trail in the review comment below.

Closes #255. Folds in the deferred un-export (was bead `arrakis-bs5n`). Convergence: EULER re-map 7 → 6 asymmetric edges.

### DIFF
--- a/grimoires/loa/cycles/cycle-112-schema-emission-floor/sdd.md
+++ b/grimoires/loa/cycles/cycle-112-schema-emission-floor/sdd.md
@@ -464,6 +464,10 @@ The rule flags, outside `packages/events/**`:
    composition root, or
 4. (IMP-005) a dataflow where a raw NATS client is aliased to a variable and then
    `.publish`-ed — not just the direct call form.
+5. (events #255) an import of `publishEnvelope` from `"@0xhoneyjar/events"` as a
+   value (not `import type`) — the capability that bypasses schema validation.
+   Kind: `publishEnvelope-bypass`. Allowlist: `publishEnvelope_allowlist` section
+   of `raw-nats-allowlist.yaml`.
 
 …unless the site is in the allowlist (§5.2). `emitRaw` call sites are checked
 against the separate `emitRaw-allowlist` (§3.3/§5.2).
--- a/packages/events/README.md
+++ b/packages/events/README.md
@@ -19,23 +19,44 @@ pnpm add nats  # peer dep — consumers bring their own connection
 
 ## Quick start (publisher)
 
+Emit through the sanctioned `emit()` facade. `makeEmitter(...)` closes over the
+per-cell constants (transport, signer, chain store) so the call site collapses to
+`emit(SchemaId, payload, specifier?)`. It validates the payload against its
+registry schema, signs the original bytes, advances the per-cell hash chain, and
+publishes — returning a typed `Either`, never throwing.
+
+> The raw `publishEnvelope` function is INTERNAL — `emit()` uses it under the hood
+> after validating. It skips schema validation, so it is NOT a public entrypoint:
+> it is unexported from the package index and the `./publisher` subpath is omitted
+> from the exports map (events #255, FR-7/SKP-002). Reach for `emit()`.
+
 ```typescript
 import { connect } from "nats";
 import {
-  publishEnvelope,
+  makeEmitter,
+  createNatsTransport,
   LocalEd25519Signer,
   InMemoryPrevHashStore,
-  nftMintDetectedTopic,
+  NftMintDetectedId,
 } from "@0xhoneyjar/events";
+import { Either } from "effect";
 
 const nats = await connect({ servers: process.env.NATS_URL });
 const signer = await LocalEd25519Signer.fromSeedHex(process.env.SIGNING_SEED_HEX!, "sonar-api-1");
-const prevHashStore = new InMemoryPrevHashStore();
 
-await publishEnvelope({
-  nats,
-  subject: nftMintDetectedTopic({ collectionSlug: "mibera-shadow" }),  // "nft.mint.detected.mibera-shadow.v1"
-  payload: {
+const emitter = makeEmitter({
+  cell: "sonar-api",
+  transport: createNatsTransport(nats),
+  signer,
+  prevHashStore: new InMemoryPrevHashStore(),
+});
+
+// The specifier ("mibera-shadow") builds the subject from the registry topic:
+// "nft.mint.detected.mibera-shadow.v1". A schema mismatch is an Either.Left —
+// it never reaches the bus and never forks the chain.
+const result = await emitter.emit(
+  NftMintDetectedId,
+  {
     chain_id: 80094,
     contract: "0x048327A187b944ddac61c6e202BfccD20d17c008",
     token_id: "234",
@@ -44,10 +65,12 @@ await publishEnvelope({
     transaction_hash: "0xdef...",
     timestamp: new Date().toISOString(),
   },
-  emittedBy: "sonar-api",
-  signer,
-  prevHashStore,
-});
+  "mibera-shadow",
+);
+
+if (Either.isLeft(result)) {
+  // handle SchemaEmitError | TransportEmitError | … — never a silent drop
+}
 ```
 
 ## Quick start (subscriber)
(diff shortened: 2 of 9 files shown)


zkSoju:
### Independent review trail (FAGAN cross-model, 3 rounds) — substitutes for the skipped harness REVIEW/AUDIT

The spiral harness died on budget before its own review/audit ran, so this PR was gated by three independent FAGAN rounds + a deterministic proof matrix (real binary every time, Trusting-Trust). Every finding below carries a real-binary repro.

**Round 1 — the original lint cut → CHANGES REQUESTED**
- F1 (HIGH): multi-line import (normal Prettier shape) was a total miss (exit 0).
- F2: namespace import + member call missed.
- F3: comment-skip leaked FPs (inline/block comments, `export type`, inline `{type}`).
- F4: the shrink-gate addition was dead code; its test passed with the block deleted (false-confidence).

**Round 2 — whole-file rewrite → CHANGES REQUESTED** (traded line-scoping holes for span-scoping)
- NF1 (HIGH): lazy clause + greedy brace spanned unrelated statements → an `export interface`/object with a `publishEnvelope` member next to a `@events` import false-positived. HIGH because the file-granular allowlist turns the FP into a permanent blind spot.
- NF2: `stripComments` had no regex-literal state → a quote-bearing regex (the lint's own source uses one) desynced the parser and re-opened the comment FP.
- NF3 (FN, later mooted by the fence): destructure-from-namespace evaded detection.

**Pivot:** two leaky rounds showed a regex tripwire can't robustly parse TS imports without an AST (which the zero-dep "ships-anywhere" design forbids). Per the deletion-over-denylist principle, the cut pivoted to the **structural fence** (un-export) as the real control, lint demoted to backstop — so lint FNs stopped mattering and only FPs (cry-wolf) remained blockers.

**Round 3 — fence + FP-hardening → FENCE HOLDS (clean), one residual lint FP**
- Fence proven against the real Node v22 resolver (all deep paths `ERR_PACKAGE_PATH_NOT_EXPORTED`; barrel/emit don't expose it; no wildcard drag; no type break; no consumers). Airtight.
- NF1/NF2 fixed and verified.
- NF6 (residual FP): a no-semicolon local `export { publishEnvelope }` adjacent to a `@events` import false-positived in no-semicolon codebases → fixed with a one-line keyword-barrier lookahead (`commit 51a719fc`) + a regression test; re-verified exit 0.

Final matrix (real binary `--enforce`): all true positives → exit 1, all FP guards (incl. NF1/NF2/NF6) → exit 0, clean tree → exit 0.

zkSoju:

PR #317, opened 2026-06-26
feat(shared/immune): gate-freeze sensor — read-only re-freeze doctor for the merge gate [campaign G1]

zkSoju:
## Estate-immune campaign — G1 read-only re-freeze doctor

The read-only **doctor** half of the `doctor → aligner → teeth` immune triad for **G-DOOR** (un-freeze the merge gate). Hoisted from the operator's proven `bonfire/gate-freeze.mjs` into the repo's checked-in `tools/` immune kit (hoist-don't-handroll-shared-substrate).

**What it does:** computes the per-required-check freeze ratio across the open-PR backlog and names the single root check freezing the door. Exit code is the verdict (2 = FROZEN) so it can gate CI or a banner row → the door can never silently re-freeze (the G-DOOR acceptance metric "re-freeze auto-detected").

**Live on this repo right now:** `FROZEN: 'Unit Tests' red 30/30 (+4 more required) · +24 behind(strict) · 1 ambiguous-name` — the mechanical form of the manual spike.

**Why draft:** the door is frozen (operator gate — branch-protection edit, bead `arrakis-estate-immune-epic-4xw6.6`). This PR can't merge until then; it's the diagnostic that explains *why* it's stuck, useful while the door is still red.

**Safety:** read-only (gh reads, no mutation / merge / money path), App-zone (`tools/`, alongside the other doctors), 13 fixture tests (no live gh). Built in an isolated worktree off origin/main — does **not** touch the operator's in-flight `fix/cursor-adapter-*` or `fix/census-live-registries` branches.

Bead: `arrakis-estate-immune-epic-4xw6.2` · Campaign brief: `grimoires/loa/context/campaign-immune-estate.md` · OPERATOR-RETURN: `grimoires/loa/context/goals/OPERATOR-RETURN.md`

### DIFF
--- a/tools/gate-freeze-sensor.mjs
+++ b/tools/gate-freeze-sensor.mjs
@@ -1,0 +1,332 @@
+#!/usr/bin/env node
+/**
+ * gate-freeze-sensor.mjs — the merge-gate freeze sensor (estate immune kit).
+ *
+ * The estate-coherence banner COUNTS rotting PRs ("34 rotting PRs") but never diagnoses the
+ * CAUSE. Most of the time a stalled backlog is not N independent problems — it is ONE numb
+ * required status check that is red on (almost) every open PR, freezing the whole backlog
+ * behind it. branch-protection strict-mode compounds it (every PR must also rebase).
+ *
+ * This sensor finds the ROOT: for a repo's branch-protection required checks, it computes the
+ * freeze ratio of each check across the open-PR backlog and names the single check doing the
+ * freezing. It makes the gate's verdict legible instead of leaving "the PRs are just stuck."
+ *
+ * It is the read-only DOCTOR half of the estate-immune triad (doctor → aligner → teeth): the
+ * G-DOOR acceptance metric "re-freeze auto-detected" made mechanical. Exit code IS the verdict,
+ * so it gates: 0 = no frozen required check, 2 = FROZEN. Wire it into CI or a banner row and the
+ * door can never silently re-freeze (which is exactly how it froze: one red "Unit Tests" check
+ * stalled 27/27 PRs, while enforce_admins=false let admin-bypass quietly become the only path).
+ *
+ * Operationalizes standing immune doctrine:
+ *   - a CI gate red on every PR is noise = merge-blind  ([[ci-sensors-must-not-be-numb]])
+ *   - a gate's verdict is a number you can read, not a vibe  ([[gate-output-never-piped]])
+ *   - exists != wired != runs != enforces  ([[verification-ladder-exists-to-verified]])
+ *
+ * Read-only. Uses the `gh` CLI (caller's auth) — no API key, no mutation, no merge. Hoisted into
+ * the repo's checked-in tools/ from the operator's proving-ground original (bonfire/gate-freeze.mjs,
+ * born 2026-06-25 out of the loa-freeside finding) per [[hoist-dont-handroll-shared-substrate]].
+ * Campaign: estate-immune (bead arrakis-estate-immune-epic-4xw6.2, G1 self-healing gate sensor).
+ *
+ * The analysis core `analyzeBacklog()` is a PURE function (no I/O) so it is fixture-testable with
+ * no live `gh` — see tools/gate-freeze-sensor.test.mjs. `fetchRepoData()` does the gh reads.
+ *
+ * Usage:
+ *   node tools/gate-freeze-sensor.mjs <owner/repo>          # human board (one repo)
+ *   node tools/gate-freeze-sensor.mjs <owner/repo> --json   # machine
+ *   node tools/gate-freeze-sensor.mjs <owner/repo> --probe  # one-line STATUS tile (banner-style)
+ *   node tools/gate-freeze-sensor.mjs                        # infer repo from cwd via gh
+ *   node tools/gate-freeze-sensor.mjs --estate --owner=ORG  # THE CONDUCTOR: sweep every repo w/
+ *                                              # open PRs, group freezes by root cause, rank the
+ *                                              # fixes by PRs-un-numbed-per-fix → a thaw plan.
+ *                                              # Re-run as fixes land; exit 2 = still frozen.
+ */
+import { execFileSync } from 'node:child_process';
+
+const ARGS = process.argv.slice(2);
+const MODE = ARGS.includes('--json') ? 'json' : ARGS.includes('--probe') ? 'probe' : 'board';
+const REPO_ARG = ARGS.find((a) => !a.startsWith('--'));
+export const FREEZE_THRESHOLD = 0.5; // a required check not-green on > half the backlog = FROZEN
+
+// not-green buckets — anything that is not one of these blocks a merge
+const PASS = new Set(['SUCCESS', 'NEUTRAL', 'SKIPPED']);
+const FAIL = new Set(['FAILURE', 'ERROR', 'CANCELLED', 'TIMED_OUT', 'ACTION_REQUIRED', 'STARTUP_FAILURE']);
+// everything else (PENDING/IN_PROGRESS/QUEUED/EXPECTED/WAITING/null) = pending
+
+function gh(args, quiet = false) {
+  return execFileSync('gh', args, {
+    encoding: 'utf8',
+    maxBuffer: 64 * 1024 * 1024,
+    stdio: quiet ? ['ignore', 'pipe', 'ignore'] : ['ignore', 'pipe', 'inherit'],
+  });
+}
+function ghJSON(args, quiet = false) {
+  return JSON.parse(gh(args, quiet));
+}
+
+function resolveRepo() {
+  if (REPO_ARG) return REPO_ARG;
+  try {
+    const r = ghJSON(['repo', 'view', '--json', 'nameWithOwner']);
+    return r.nameWithOwner;
+  } catch {
+    die('could not infer repo — pass <owner/repo>');
+  }
+}
+
+function die(msg) {
+  console.error(`gate-freeze: ${msg}`);
+  process.exit(1);
+}
+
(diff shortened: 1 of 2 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#317. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "GF-001",
      "title": "Sensor fails OPEN (exit 0) when branch protection is unreadable — defeats the CI gating it advertises",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "Reading `repos/{repo}/branches/{branch}/protection` requires an admin-scoped token. When the call throws (non-admin token, which is the common case for CI's default GITHUB_TOKEN), `required` becomes null, verdict becomes 'UNGATED?', and `main()`/`estateMode()` exit 0. The header doc explicitly recommends 'Wire it into CI ... the door can never silently re-freeze' and 'Exit code IS the verdict ... 2 = FROZEN'. But wired into CI with a typical token, the sensor cannot read protection, reports UNGATED?, and always passes — a silent no-op gate. A genuine re-freeze would go undetected in exactly the wiring the doc prescribes.",
      "suggestion": "Document the admin-token requirement prominently, and/or make the UNGATED? exit code configurable (e.g., `--strict-unreadable` to exit 2 / fail-closed when protection cannot be confirmed). At minimum, emit a loud stderr warning when protection is unreadable so a CI wiring doesn't appear green-by-default.",
      "confidence": 0.7
    },
    {
      "id": "GF-002",
      "title": "Estate headline `totalFrozenPRs` sums ALL open PRs, overstating the frozen count up to 2x",
      "severity": "MEDIUM",
      "category": "accuracy",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "`totalFrozenPRs: frozen.reduce((n, s) => n + (s.frozenChecks[0]?.ofPRs || s.openPRs || 0), 0)` and `renderThawPlan` print '~N PRs sit behind a genuinely-required failing gate.' But `ofPRs` is the repo's TOTAL open-PR count, including PRs where the root check is green. Since the freeze threshold is only >50%, a repo with 100 open PRs whose root check fails on 51 contributes 100, not 51. The accurate count (failOn + missingOn) is already available on `frozenChecks[0]`. The '~' hedge does not cover an up-to-2x overstatement of the campaign's primary metric.",
      "suggestion": "Sum `frozenChecks[0].failOn + frozenChecks[0].missingOn` (the PRs actually frozen by the root check) for `totalFrozenPRs`, or relabel the metric to 'PRs in frozen repos' so the number is not read as 'PRs behind the gate'.",
      "confidence": 0.6
    },
    {
      "id": "GF-003",
      "title": "`duplicateRuns` is accumulated backlog-wide on a mixed identifier key → false 'ambiguous-name' smell",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "`runs.add(c.workflowName ? \\`${c.workflowName}/${name}\\` : (c.detailsUrl || name))` accumulates across ALL PRs, then `duplicateRuns = runs.size > 1`. The intent is to detect one check name produced by multiple workflows (name collision). But the key mixes a per-workflow-stable id (`workflowName/name`) with a per-run-unique id (`detailsUrl`). Two real sources of false positives: (a) a check whose producing workflow was renamed over the PR backlog's lifetime (e.g. 'CI/Build' on new PRs, 'CI-v2/Build' on old ones) yields runs.size=2 → falsely flagged ambiguous; (b) any check that surfaces with `detailsUrl` but no `workflowName` gets a unique key per PR, inflating runs.size to the PR count. CheckRuns always carry workflowName today so (b) is latent, but (a) is reachable.",
      "suggestion": "Track duplicate-name detection PER PR (a single PR carrying two contexts of the same required name) rather than across the whole backlog, or key strictly on a stable workflow identifier and never fall back to the per-run detailsUrl.",
      "confidence": 0.5
    },
    {
      "id": "GF-004",
      "title": "Non-required checks get phantom zero-count tally entries; `ungatedFailing` is wrong in the hasRequired path (latent)",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "The inner rollup loop calls `ensure(name)` (creating a tally entry) for every check, including non-required ones, but fail/pending/pass/missing are only incremented in the second loop over `names = required` (when required is non-empty). So when `required` is non-empty, non-required failing checks land in `checks` with fail=0, and `ungatedFailing` (which filters `c.fail > 0`) would miss them. This is currently harmless only because `ungatedFailing` is rendered exclusively in the UNGATED/UNGATED? branches (where required is empty/null and all checks ARE counted). The inconsistency is a latent trap if a future render surfaces `ungatedFailing` for a FLOWING-with-required repo.",
      "suggestion": "Tally fail/pending/pass for ALL observed checks (not just the required names), and gate the 'frozen' decision on isRequired separately. That keeps `ungatedFailing` correct regardless of whether a required set exists.",
      "confidence": 0.45
    },
    {
      "id": "GF-005",
      "title": "Fragile main-module detection via string compare on file:// URL",
      "severity": "LOW",
      "category": "robustness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "`if (import.meta.url === \\`file://${process.argv[1]}\\`) main();` compares a raw concatenated path against `import.meta.url`. This breaks when the tool is invoked through a symlink, or when the path contains characters that `import.meta.url` percent-encodes (spaces, etc.), causing `main()` to silently not run. For a tool the header explicitly describes as 'hoisted' (and the estate has prior symlink-resolution issues), symlinked invocation is plausible.",
      "suggestion": "Use `import { pathToFileURL } from 'node:url'` and compare `import.meta.url === pathToFileURL(process.argv[1]).href`.",
      "confidence": 0.5
    },
    {
      "id": "GF-006",
      "title": "Hardcoded default owner '0xHoneyJar' in a tool described as shared/hoisted substrate",
      "severity": "LOW",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "`estateMode()` defaults owner to '0xHoneyJar' when neither --owner nor a positional repo is given. The header frames this file as hoisted shared substrate; a baked-in org name couples the generic tool to one estate and will silently sweep the wrong org if copied elsewhere.",
      "suggestion": "Require `--owner` (or infer from `gh repo view`/`gh api user`) and `die()` with a clear message rather than defaulting to a specific org.",
      "confidence": 0.4
    },
    {
      "id": "GF-007",
      "title": "Estate sweep truncation (`gh search prs --limit 200`) is not surfaced",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "Single-repo mode carefully surfaces the 100-PR cap via `truncated`, but `estateMode()` caps `gh search prs` at 200 results and silently drops repos whose open PRs fall beyond that window. A large org could have frozen repos omitted from the thaw plan with no indication.",
      "suggestion": "Detect when the search hits the 200 cap and add a 'truncated — repos may be missing' note to the thaw-plan output, mirroring the single-repo truncation handling.",
      "confidence": 0.45
    },
    {
      "id": "GF-008",
      "title": "Clean I/O separation: pure analyzeBacklog enables live-gh-free fixture tests",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "Splitting `fetchRepoData()` (gh reads) from the pure `analyzeBacklog()` (no I/O) makes the freeze-detection logic deterministically testable without any live `gh`, and the test file exercises it directly. This is the right seam and the verdict-gating test (exit 1 on failure) is CI-safe.",
      "suggestion": "No change — keep this boundary as the tool grows.",
      "confidence": 0.9
    },
    {
      "id": "GF-009",
      "title": "Metacharacter-safe gh invocation via execFileSync with array args",
      "severity": "PRAISE",
      "category": "security",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "All `gh` calls go through `execFileSync('gh', args, …)` with arguments as an array and no shell, so repo names / owner strings cannot inject shell metacharacters. This matches the estate's stated metacharacter-safe invocation convention.",
      "suggestion": "No change.",
      "confidence": 0.9
    },
    {
      "id": "GF-010",
      "title": "False-positive guard: required vs non-required distinction prevents flagging REVIEW-blocked backlogs as gate-frozen",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "Restricting the FROZEN verdict to checks actually in `required_status_checks` (and treating required=[] as UNGATED, required=null as UNGATED?) correctly avoids the common failure where a non-required check is red everywhere but the real block is REVIEW. Test case 5 explicitly pins this guard, and test 4 pins that unreadable protection never asserts FROZEN.",
      "suggestion": "No change — this is the key correctness decision and it is well-tested.",
      "confidence": 0.85
    },
    {
      "id": "GF-011",
      "title": "Ships as a doctor without a wired consumer — lands at exists/runs, not wired/enforces",
      "severity": "REFRAME",
      "category": "design",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "The header cites [[verification-ladder-exists-to-verified]] and frames this as the read-only doctor half whose value is realized by 'wire it into CI or a banner row.' The PR adds only the tool + its test — no CI job, npm script, or banner-row wiring. By the tool's own cited ladder it currently sits at 'exists/runs' but not 'wired/enforces,' which is the recurring deployed-but-unconsumed shape. This is a scoping observation, not a defect: the doctor half is legitimately a standalone deliverable.",
      "suggestion": "Either note the follow-up wiring bead in the PR description (the aligner/teeth halves of the triad), or include a minimal package.json script / CI invocation so the sensor has a first consumer in the same change.",
      "confidence": 0.5
    }
  ]
}
```


## Callouts

PR #319, opened 2026-06-26
feat(shared/immune): meta-immune doctor — detect lying health instruments (campaign M0 / G-IMMUNE)

zkSoju:
> **Draft — door frozen.** Campaign M0 doctor half (read-only). The merge gate is still being un-frozen (G1); this lands as a reviewable draft, not for merge.

## What it does
`tools/instrument-truth-sensor.mjs` — the **read-only DOCTOR** half of campaign M0 (G-IMMUNE: *every verdict/health/goal/sensor re-derives from a ground source; 0 lying instruments*).

It detects **lying instruments** — health/verdict instruments whose **declared** state diverges from **ground truth**. First verified instance: cheval voice health, re-derived from `.run/model-invoke.jsonl` (the MODELINV audit chain). The static cheval doctor (`doctor.py`) misreports the council — it calls `claude` `"unknown"` (a `--max-budget-usd 0.001` probe artifact) while the log shows ~538 successful finals, and it is **blind to `cursor`** (45 fresh finals, the real idle capacity), which never appears in its probe table. This sensor re-derives each voice's real health and flags the divergence.

Hoisted from the `gate-freeze.mjs` pattern (PR #317): a **pure exported core** `analyzeVoiceTruth({ records, probeTable, now, freshnessMs })` that takes already-parsed records + the declared probe set and returns the divergence verdict, plus a **thin I/O wrapper** that reads the jsonl. **Exit code IS the verdict** — `0` TRUTHFUL · `2` LYING · `1` INSUFFICIENT (no records / unreadable → *never* asserts "truthful" without evidence). `board` / `--json` / `--probe` renderers.

**The honest core is the false-positive guard:** a voice that is *genuinely down* is **CONSISTENT, not a lie** — only the *divergence* between declared and real health is flagged. The inverse lie (declared `up` but gone silent → `FALSE_HEALTHY`) is flagged too.

| classification | meaning | flagged? |
|---|---|---|
| `INVISIBLE` | live finals, absent from the probe table | ✅ lying |
| `DIVERGENT` | probe says down/unknown, but fresh finals | ✅ lying |
| `FALSE_HEALTHY` | probe says up, but gone silent | ✅ lying |
| `CONSISTENT` | declaration matches ground truth (incl. *genuinely* down) | ✓ truthful |
| `DARK` | absent from probe AND inactive | ✓ ignored |

## Live verdict (against the real `.run/model-invoke.jsonl`, 1420 records)
```
voice     finals  last-final   age   declared   verdict
⚠ claude     538  2026-06-24   1.9d  unknown   DIVERGENT
✓ codex      422  2026-06-25   1.5d  up        CONSISTENT
✓ gemini     299  2026-06-17   8.9d  down      CONSISTENT
⚠ cursor      45  2026-06-25   1.6d  ABSENT    INVISIBLE
✓ grok         2  2026-06-25   1.5d  up        CONSISTENT
```
**LYING — exit 2.** `claude:DIVERGENT` (doctor.py's false "unknown" over a live workhorse) and `cursor:INVISIBLE` (the idle capacity the probe can't see). `gemini` is correctly **CONSISTENT** — genuinely dead (9d stale), declared down — the guard does not false-flag an outage as a lie.

## Tests
`tools/instrument-truth-sensor.test.mjs` — **12 fixture tests, all green** (`node tools/instrument-truth-sensor.test.mjs`, exit 0). No live deps: synthetic records cover (a) INVISIBLE, (b) DIVERGENT, (c) all-consistent→exit 0, (d) the genuinely-dead-is-CONSISTENT guard, (e) FALSE_HEALTHY (incl. zero-finals), DARK, null-finals, `now`-defaulting, and the configurable freshness window.

## Safety
- **Read-only.** Reads ONLY `.run/model-invoke.jsonl` — never invokes `doctor.py` or any slow/hanging external command. Bounded tail (`MAX_RECORDS`).
- **App-zone** (`tools/`). No `.claude/` (System-Zone) edits, no money path, no infra.
- Built in an **isolated git worktree** off `origin/main`; the operator's working tree was never touched.

## Bead
`arrakis-estate-immune-epic-4xw6.1` (M0 — meta-immune doctor). This is the **doctor** half only. **Deferred to the operator's reviewed `/run`:** the M0 enforcing CI lint (System-Zone), G2 STUB inversion + `scoring.ts` (System-Zone, needs multi-model review which is currently down), G3 ledger adapter (money path), G4 loa-cli (in-flight collision).

### DIFF
--- a/tools/instrument-truth-sensor.mjs
+++ b/tools/instrument-truth-sensor.mjs
@@ -38,7 +38,7 @@
  *   node tools/instrument-truth-sensor.mjs --log=PATH      # point at a specific audit log
  *   node tools/instrument-truth-sensor.mjs --probe-table=FILE.json   # override the declared probe
  *   node tools/instrument-truth-sensor.mjs --freshness-days=N        # active window (default 3)
- *   node tools/instrument-truth-sensor.mjs --now=ISO                 # freshness anchor (default: newest record)
+ *   node tools/instrument-truth-sensor.mjs --now=ISO                 # freshness anchor (default: WALL-CLOCK)
  *
  * Exit code is the verdict (so it can gate):
  *   0 = TRUTHFUL  (every instrument's declaration matches ground truth)
@@ -51,7 +51,11 @@ import { fileURLToPath, pathToFileURL } from 'node:url';
 
 export const DAY_MS = 86_400_000;
 export const DEFAULT_FRESHNESS_MS = 3 * DAY_MS; // a voice is ACTIVE if it produced a final in this window
-const MAX_RECORDS = 20_000; // bounded tail — read the recent window, never unbounded
+// Bounds how many lines are PARSED (the recent tail). The file READ itself is NOT bounded by this:
+// readRecords() reads the whole file via readFileSync, then keeps only the last MAX_RECORDS lines to
+// parse. Fine for the small audit log today; if this is ever pointed at a full audit stream, switch
+// readRecords to an EOF-seeking byte-capped tail so the memory envelope matches this intent.
+const MAX_RECORDS = 20_000;
 
 // provider prefix (left of the first ':') -> canonical cheval voice. anthropic-via-bedrock is claude.
 export const PROVIDER_VOICE = {
@@ -64,19 +68,33 @@ export const PROVIDER_VOICE = {
 };
 
 /**
- * The STATIC cheval doctor's DECLARED report — the *declaration* side this sensor checks against
+ * AS-OF date of STATIC_DOCTOR_PROBE. Surfaced in every render so the declaration's STATIC nature is
+ * first-class: this is a frozen snapshot, NOT a live read of doctor.py. A frozen baseline measures
+ * the age of the baseline as much as the thing — so we make that age visible (and offer
+ * --probe-table=FILE.json to re-ground) rather than letting the constant silently drift.
+ */
+export const STATIC_DOCTOR_PROBE_AS_OF = '2026-06-26';
+
+/**
+ * The STATIC cheval doctor's DECLARED coverage — the *declaration* side this sensor checks against
  * ground truth. This is a CONFIG constant (the probe's known coverage + verdicts as observed
- * 2026-06-26, GECKO-derived in grimoires/loa/context/morning-leverage-ideation-2026-06-26.md), NOT
- * a live probe — invoking doctor.py here would hang. Override with --probe-table=FILE.json to
- * re-ground. `cursor` is deliberately ABSENT: it is invisible to the static probe table (that
- * invisibility is exactly the lie this sensor surfaces).
+ * 2026-06-26, GECKO-derived), NOT a live probe — invoking doctor.py here would hang. Override with
+ * --probe-table=FILE.json to re-ground.
+ *
+ * The keys MIRROR doctor.py's contract-pinned `_PROBE_TABLE` — exactly {claude, codex, gemini}
+ * (test_doctor.py::TestProbeTableContract). The new-company headless adapters (`cursor`, `grok`) are
+ * deliberately ABSENT here BECAUSE they are absent from doctor.py's probe table — each ships its own
+ * adapter-level health_check() instead. That invisibility is exactly the lie this sensor surfaces:
+ * if such a voice produces fresh finals, it reads as INVISIBLE (a live-but-unlisted lie of omission);
+ * if it is silent, it reads as DARK (nothing claimed, nothing done). Declaring `grok: 'up'` here
+ * would instead MANUFACTURE a FALSE_HEALTHY whenever grok is idle — a self-inflicted false positive
+ * about the constant, not a real lying instrument.
  */
 export const STATIC_DOCTOR_PROBE = Object.freeze({
   claude: 'unknown', // probe artifact — reports unknown despite a live workhorse
   codex: 'up',
   gemini: 'down', // genuinely dead — the probe is correct here
-  grok: 'up',
-  // cursor: (absent — invisible to the probe)
+  // cursor, grok: (absent — not in doctor.py's contract-pinned probe table)
 });
 
 const CLAIMS_HEALTHY = new Set(['up', 'healthy', 'ok']);
@@ -212,22 +230,54 @@ export function readRecords(logPath) {
   return out;
 }
 
-function loadProbeTable() {
-  const file = argVal('probe-table') || process.env.LOA_DOCTOR_PROBE_TABLE;
-  if (!file) return STATIC_DOCTOR_PROBE;
-  return JSON.parse(readFileSync(file, 'utf8'));
-}
-
 function die(msg, code = 1) {
   console.error(`instrument-truth: ${msg}`);
   process.exit(code);
 }
 
+/**
+ * Resolve the declared probe table. Same fail-closed discipline as the log read: a missing/malformed
+ * --probe-table (or a parse that succeeds but yields a non-object, e.g. a JSON array — the
+ * `typeof [] === 'object'` footgun) routes through die(... INSUFFICIENT), never a raw stack trace and
+ * never a silently-wrong classification. Returns { probeTable, source, asOf }.
+ */
+function loadProbeTable() {
+  const file = argVal('probe-table') || process.env.LOA_DOCTOR_PROBE_TABLE;
+  if (!file) return { probeTable: STATIC_DOCTOR_PROBE, source: 'static', asOf: STATIC_DOCTOR_PROBE_AS_OF };
+  let parsed;
+  try {
+    parsed = JSON.parse(readFileSync(file, 'utf8'));
+  } catch (e) {
+    die(`cannot read/parse probe table ${file} (${e.code || e.message}) — INSUFFICIENT, refusing to assert "truthful" without evidence`);
+  }
+  if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) {
+    die(`probe table ${file} is not a JSON object (expected a voice -> status map) — INSUFFICIENT, refusing to assert "truthful" without evidence`);
+  }
+  return { probeTable: parsed, source: file, asOf: null };
+}
+
 function main() {
   const logPath = argVal('log') || process.env.LOA_MODELINV_LOG || defaultLogPath();
-  const freshDays = argVal('freshness-days') ? Number(argVal('freshness-days')) : 3;
+
+  // Validate numeric/date args at the boundary BEFORE any work. NaN comparisons silently return false,
+  // which for a fail-closed sensor flips a typo from "guard" to "all clear" (e.g. --freshness-days=abc
+  // -> every voice stale -> divergences collapse -> a false TRUTHFUL). Fail toward INSUFFICIENT instead.
+  const freshDaysRaw = argVal('freshness-days');
+  const freshDays = freshDaysRaw === undefined ? 3 : Number(freshDaysRaw);
+  if (!Number.isFinite(freshDays) || freshDays < 0) {
+    die(`--freshness-days=${freshDaysRaw} is not a non-negative number — INSUFFICIENT, refusing to assert "truthful" without evidence`);
+  }
   const freshnessMs = freshDays * DAY_MS;
-  const now = argVal('now');
+
+  const nowArg = argVal('now');
+  if (nowArg !== undefined && Number.isNaN(Date.parse(nowArg))) {
+    die(`--now=${nowArg} is not a parseable date — INSUFFICIENT, refusing to assert "truthful" without evidence`);
+  }
+  // Freshness is anchored to WALL-CLOCK by default, NOT the log's own newest record. Anchoring to the
+  // newest record makes a frozen dispatch stream (the ultimate numb instrument) undetectable: the most
+  // recent voice always reads age-0 against a clock that froze with it. Wall-clock is the external,
+  // monotonic reference a watchdog needs. --now overrides it for deterministic tests.
+  const now = nowArg || new Date().toISOString();
 
   let records;
   try {
@@ -239,10 +289,23 @@ function main() {
     die(`no MODELINV records in ${logPath} — INSUFFICIENT, refusing to assert "truthful" without evidence`);
   }
 
-  const probeTable = loadProbeTable();
+  const { probeTable, source: probeSource, asOf: probeAsOf } = loadProbeTable();
   const result = analyzeVoiceTruth({ records, probeTable, now, freshnessMs });
   result.logPath = logPath;
   result.recordCount = records.length;
+  result.probeSource = probeSource;
+  result.probeAsOf = probeAsOf;
+
+  // First-class log-staleness signal: if the whole log's newest record predates the active window
+  // (measured against the same wall-clock anchor), the dispatch stream itself may have stopped. Surface
+  // it; the verdict is still driven by lying instruments (a frozen log surfaces declared-up voices as
+  // FALSE_HEALTHY, which is the actionable, gateable signal).
+  const newest = newestTs(records);
+  const newestMs = newest ? Date.parse(newest) : NaN;
+  const logAgeMs = Number.isFinite(newestMs) ? Date.parse(now) - newestMs : Infinity;
+  result.newestRecordTs = newest;
+  result.logAgeDays = Number.isFinite(logAgeMs) ? Math.round((logAgeMs / DAY_MS) * 10) / 10 : null;
+  result.logStale = Number.isFinite(logAgeMs) ? logAgeMs > freshnessMs : true;
 
   if (MODE === 'json') console.log(JSON.stringify(result, null, 2));
   else if (MODE === 'probe') console.log(renderProbe(result));
@@ -269,18 +332,26 @@ const WHY = {
 
 function renderProbe(r) {
   const tag = r.verdict === 'LYING' ? '⚠ instrument-truth' : '✓ instrument-truth';
+  const stale = r.logStale ? ` · ⚠ log stale (newest ${r.logAgeDays}d old — dispatch may have stopped)` : '';
   if (r.verdict === 'TRUTHFUL') {
-    return `${tag} · ${r.voices.length} voices · all declarations match ground truth (${r.recordCount} records)`;
+    return `${tag} · ${r.voices.length} voices · all declarations match ground truth (${r.recordCount} records)${stale}`;
   }
   const names = r.lyingInstruments.map((v) => `${v.voice}:${v.classification}`).join(', ');
-  return `${tag} · LYING: ${r.lyingInstruments.length}/${r.voices.length} instrument(s) diverge — ${names}`;
+  return `${tag} · LYING: ${r.lyingInstruments.length}/${r.voices.length} instrument(s) diverge — ${names}${stale}`;
 }
 
 function renderBoard(r) {
   const L = [];
   L.push(`  ∴ instrument-truth · cheval voice health · re-derived from ${r.recordCount} MODELINV records`);
   L.push(`    ground-source: ${r.logPath}`);
+  const decl = r.probeSource === 'static'
+    ? `static snapshot as of ${r.probeAsOf} (NOT a live doctor.py read — re-ground with --probe-table)`
+    : r.probeSource;
+  L.push(`    declaration:   ${decl}`);
   L.push(`    reference(now): ${r.reference}   active-window: ${Math.round(r.freshnessMs / DAY_MS)}d`);
+  if (r.logStale) {
+    L.push(`    ⚠ log itself is STALE — newest record is ${r.logAgeDays}d old (> ${Math.round(r.freshnessMs / DAY_MS)}d window); the dispatch stream may have stopped.`);
+  }
   L.push('');
   L.push(`    voice     finals  last-final   age   declared   verdict`);
   for (const v of r.voices) {
--- a/tools/instrument-truth-sensor.mjs
+++ b/tools/instrument-truth-sensor.mjs
@@ -1,0 +1,309 @@
+#!/usr/bin/env node
+/**
+ * instrument-truth-sensor.mjs — the meta-immune DOCTOR (read-only).
+ *
+ * Campaign: Estate Immune System · M0 / G-IMMUNE ("every verdict/health/goal/sensor re-derives
+ * from a ground source; 0 lying instruments"). Bead: arrakis-estate-immune-epic-4xw6.1.
+ * Pattern hoisted from ~/bonfire/gate-freeze.mjs — PURE testable core + thin I/O wrapper,
+ * exit-code-IS-the-verdict, board/json/probe renderers, false-positive guard documented in-code.
+ *
+ * WHAT IT DETECTS — lying instruments. A health/verdict instrument whose DECLARED state diverges
+ * from GROUND TRUTH. The verified first instance: cheval voice health. The static cheval doctor
+ * (doctor.py) MISREPORTS the council — it calls `claude` "unknown" (a `--max-budget-usd 0.001`
+ * probe artifact) while `.run/model-invoke.jsonl` shows ~500 successful finals, and it is BLIND to
+ * `cursor` (45 fresh finals, the real idle capacity) which never appears in its probe table at all.
+ * This sensor re-derives each voice's REAL health from the audit log (the ground source) and flags
+ * any voice the static doctor would misreport.
+ *
+ * Operationalizes the operator's standing immune doctrine:
+ *   - declaration != ground-truth, and the machine trusts the declaration  ([[map-territory-drift-unifying-failure]])
+ *   - a gate's verdict is a number you read, not a vibe                     ([[gate-output-never-piped]])
+ *   - exists != wired != runs != enforces                                  ([[verification-ladder-exists-to-verified]])
+ *   - a check is INSUFFICIENT, not "clean", when it has no evidence        (absence of evidence != proof)
+ *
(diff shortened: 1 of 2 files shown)


zkSoju (commented):
## Summary

This PR adds a *meta-immune doctor* — a sensor that watches the watchers, cross-checking what `doctor.py` *declares* about each cheval voice against what `model-invoke.jsonl` *proves*. The core invariant is exactly right: a genuinely-down voice stays CONSISTENT, only a *divergence* between declaration and ground-truth is flagged, and the tool refuses to assert TRUTHFUL without evidence (INSUFFICIENT/exit-1 rather than a clean exit). The reflexive risk is equally real: by defaulting both the declaration side (a frozen constant) and the freshness anchor (the log's own newest record) to *self-referential* baselines, the sensor can become the very lying instrument it was built to catch. Verdict: **REQUEST_CHANGES** — the design is excellent, but three default-path issues let it emit false verdicts in both directions.

There is a beautiful recursion in this PR worth naming up front. The whole estate has a recurring failure signature — *the numb instrument*, the sensor that is green because it stopped looking. This tool is the immune response to that signature. Which means the highest bar in the entire codebase applies to *this* file: a lie-detector that can itself drift, anchor to itself, or fail-open on a typo has not closed the loop — it has moved it one level up and hidden it better. The findings below are almost entirely about that one altitude: *what is the sensor's external reference, and does it fail closed?*

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "bridge-001",
      "title": "Default declaration source is a frozen hand-authored constant — the meta-immune doctor is itself drift-prone",
      "severity": "REFRAME",
      "category": "design",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "STATIC_DOCTOR_PROBE is a hand-authored snapshot of what doctor.py 'would' say, frozen at 2026-06-26, and is the default declaration side of every comparison (analyzeVoiceTruth's probeTable defaults to it; loadProbeTable returns it when no --probe-table is given). The tool's thesis is that 'declaration != ground-truth and the machine trusts the declaration' — yet by default it trusts a static, hand-authored, drift-prone declaration. As the real probe output changes, the constant goes stale and the sensor will emit genuine-looking DIVERGENT/FALSE_HEALTHY/INVISIBLE verdicts that are actually about the constant aging, not about doctor.py lying. The sensor becomes the very lying instrument it exists to catch. The in-code rationale ('invoking doctor.py here would hang') is real, but freezing the declaration is not the only fix.",
      "suggestion": "Read the live declaration where possible: invoke `doctor.py --json` with a hard timeout and fall back to the constant only on timeout/error, OR stamp STATIC_DOCTOR_PROBE with an `as_of` date and have the sensor downgrade to INSUFFICIENT (not LYING) when the constant is older than a threshold. At minimum, surface in the board output that the declaration side is a static snapshot, not a live read.",
      "confidence": 0.5,
      "faang_parallel": "Google SRE's 'meta-monitoring' problem: the alerting pipeline must itself be monitored by an independent path, because a monitoring system that goes silent looks identical to a healthy world. The SRE book's answer is never a frozen baseline — it's a second, out-of-band liveness signal with its own freshness clock. Netflix's Chaos-style verification follows the same rule: you assert a system is healthy by perturbing it live, not by comparing it to a remembered description of health.",
      "metaphor": "A smoke detector that, to save battery, compares the room not to live air but to a photograph of the room taken last Tuesday. It will alarm beautifully — every time the furniture moves — and stay silent through an actual fire if the smoke happens to match the photo.",
      "teachable_moment": "A sensor that diffs against a frozen baseline measures the *age of the baseline*, not the thing. The moment your reference is static, every verdict is contaminated by drift you can't distinguish from signal. The fix is not a better constant — it's making the reference live, or making staleness a first-class, self-reported state (INSUFFICIENT) so the tool can say 'I no longer know' instead of confidently lying.",
      "connection": "This is the estate's 'numb instrument / sensor must not be numb' pattern recurring one level up — the cure for a numb doctor must not itself be capable of numbness. An `as_of` stamp that decays to INSUFFICIENT is the cheapest way to make the tool honest about its own blindness."
    },
    {
      "id": "bridge-002",
      "title": "Default `now` anchors to the newest record — total dispatch outage (the ultimate numb instrument) cannot be detected",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "`reference = now || newestTs(records)`. In the documented banner/--probe usage no --now is passed, so freshness is measured relative to the log's own newest record. If cheval dispatch stops entirely, the log freezes; on the next run the most-recently-active voice has age 0 relative to the frozen 'now' and reads fresh → CONSISTENT/healthy. A declared-up voice that has actually gone silent for the whole log therefore stays CONSISTENT. The sensor's core mission is catching numb instruments, but by default it is structurally blind to the most severe numbness (a frozen log). The 'robust against a stale log snapshot' design choice (test e) is exactly what produces this blind spot.",
      "suggestion": "Default `now` to wall-clock time (Date.now()) for freshness, and treat the gap between wall-clock and newestTs(records) as a first-class staleness/INSUFFICIENT signal (e.g. if the whole log's newest record is older than the active window, report the log itself as stale). Keep --now as an override for deterministic tests.",
      "confidence": 0.5,
      "faang_parallel": "This is the dead-man's-switch / watchdog-timer pattern from systems engineering, and the exact reason Prometheus liveness is computed as `time() - process_last_seen` against the scrape server's wall clock, never against the target's own latest sample. Amazon's Dynamo-era heartbeat detectors learned this the hard way: a node comparing its own last heartbeat to its own clock can never detect that it has been partitioned away — the reference froze with it.",
      "metaphor": "Asking a stopped clock whether it's running by comparing its hands to where its hands are. They always match. The whole point of a watchdog is that the timer lives *outside* the thing it watches.",
      "teachable_moment": "Freshness is only meaningful against a monotonic reference external to the data stream. The instinct to be 'robust against a stale snapshot' (don't false-alarm on old data) is correct for noise, but it inverts catastrophically for liveness: the staler the data, the *more* alarming it should be, not the less. Robustness-to-staleness and detection-of-staleness are opposite requirements — split them.",
      "connection": "Pairs with bridge-001: both defaults make the tool self-referential. There the reference is a frozen declaration; here it's a frozen clock. The unifying fix is one external anchor (wall clock) plus an explicit 'log itself is stale → INSUFFICIENT' verdict."
    },
    {
      "id": "bridge-003",
      "title": "loadProbeTable() bypasses the graceful die() path and does no shape validation",
      "severity": "MEDIUM",
      "category": "error-handling",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "The log read is carefully wrapped in try/catch → die('… INSUFFICIENT'), but `loadProbeTable()` is called in main() without a try/catch. A missing or malformed `--probe-table=FILE.json` (or LOA_DOCTOR_PROBE_TABLE) causes readFileSync/JSON.parse to throw a raw stack trace and exit with the default uncaught code, inconsistent with the documented exit-1 INSUFFICIENT contract that the log path honors. There is also no validation that the parsed value is a plain object: a JSON array or string parses successfully and then flows into Object.prototype.hasOwnProperty.call(probeTable, voice) / Object.keys(probeTable), producing silently wrong classifications.",
      "suggestion": "Wrap loadProbeTable's read/parse in try/catch and route failures through die(`cannot read/parse probe table … — INSUFFICIENT`). After parse, assert the result is a non-null, non-array object before use.",
      "confidence": 0.8,
      "faang_parallel": "Alexis King's 'Parse, Don't Validate' and the broader trust-boundary discipline every FAANG API gateway enforces: untrusted input is fully validated *into a known shape* at the single point of entry, never trusted because `JSON.parse` happened to succeed. `typeof [] === 'object'` and `Object.keys([])` returning numeric strings is the JavaScript footgun that has shipped real auth bypasses (an array where an allow-list object was expected).",
      "metaphor": "A vault with a fire-rated, biometric front door (the log path's careful die()) and an unlocked screen door around the back (loadProbeTable). The attacker — here, a typo'd filename — simply walks around.",
      "teachable_moment": "Error-handling discipline must be *uniform across all input paths*, not concentrated on the one you happened to think hardest about. Two doors into the same room must be equally rated. And a successful parse is not a successful validation: JSON.parse proves syntax, not shape. Assert `obj && typeof obj === 'object' && !Array.isArray(obj)` before treating something as a map.",
      "connection": "Highest-confidence actionable finding and the cheapest to fix. It's also thematically on-mission: a tool that exists to refuse false-clean verdicts must itself never crash-or-misclassify silently on bad config — route every failure mode through the same INSUFFICIENT exit."
    },
    {
      "id": "bridge-004",
      "title": "No validation of --now / --freshness-days: invalid values silently mask all divergences",
      "severity": "LOW",
      "category": "error-handling",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "`Number(argVal('freshness-days'))` and `Date.parse(reference)` are used without validation. `--freshness-days=abc` → NaN → `ageMs <= freshnessMs` is always false → every voice becomes stale → DIVERGENT/INVISIBLE collapse to CONSISTENT and the tool reports TRUTHFUL/exit 0 despite real lies. `--now=garbage` → refMs NaN → same effect. For a sensor whose entire value is refusing to assert 'truthful' without evidence, silently returning a clean verdict on operator typo is the exact numbness it is built to prevent.",
      "suggestion": "Validate parsed inputs: if Number(freshnessDays) is NaN/negative or Date.parse(now) is NaN, die() with INSUFFICIENT rather than proceeding with NaN.",
      "confidence": 0.7,
      "faang_parallel": "The fail-open vs fail-closed distinction that governs every safety-critical control system. NaN comparisons in JavaScript always return false, which is precisely how the Knight Capital and several aviation-software incidents turned a bad input into a confidently-wrong 'all clear.' Security tooling (e.g. AWS IAM policy evaluation) is engineered to deny on ambiguity — a sensor must report INSUFFICIENT on ambiguity, never clean.",
      "metaphor": "A thermostat handed the setpoint 'abc', shrugging, and reporting the house is a perfect temperature. The typo didn't break the alarm loudly — it disabled it quietly.",
      "teachable_moment": "Every `Number(x)` and `Date.parse(x)` at a trust boundary has a silent NaN failure mode whose downstream comparison is `false`. For a fail-closed instrument that means a single typo flips it from guard to rubber-stamp. The rule: validate at the boundary and fail toward INSUFFICIENT, because the cost of a false 'truthful' is the entire reason the tool exists.",
      "connection": "Same root as bridge-003 (uniform fail-closed handling) but on numeric/date args rather than file load. Together they define the tool's input contract: nothing that fails to parse may ever produce a clean verdict."
    },
    {
      "id": "bridge-005",
      "title": "readRecords loads the whole file into memory before tailing — 'never unbounded' comment is inaccurate",
      "severity": "LOW",
      "category": "performance",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "MAX_RECORDS bounds how many lines are JSON.parsed, but `readFileSync(logPath, 'utf8')` first reads the entire file into a string and `text.split('\\n')` materializes every line. The comment 'bounded tail — read the recent window, never unbounded' describes parsing, not the file read. A large model-invoke.jsonl (audit logs can include all invocations, not just finals) would be fully loaded into memory regardless of MAX_RECORDS.",
      "suggestion": "Either reflect the true bound in the comment (parsing is bounded; file read is not), or tail the file without loading it whole (read from the end, or cap bytes read) so the memory bound matches the stated intent.",
      "confidence": 0.6,
      "faang_parallel": "Why `tail` seeks from the end of the file rather than streaming from the start, and why every mature log shipper (Fluentd, Vector, the Datadog agent) reads bounded byte windows from EOF. The bug class — a bound on processing mistaken for a bound on I/O — is the same one that makes naive `SELECT ... LIMIT 10` over an unindexed table still scan the whole table.",
      "metaphor": "A comment promising 'we only sip the last cup from the firehose' on code that first pours the entire reservoir into a glass and then measures off a cup.",
      "teachable_moment": "A comment is a contract; an inaccurate one is worse than none because future readers budget resources against it. 'Bounded parse' and 'bounded read' are different guarantees — name the one you actually provide. For true bounding, seek from EOF and cap bytes, so the memory envelope matches the stated intent even as the file grows.",
      "connection": "Low-stakes today (the log is small), but the comment will be load-bearing the day someone points this at a full audit stream. Either tighten the code to the comment or the comment to the code — never let them diverge, since divergence-between-claim-and-reality is the exact failure this whole tool hunts."
    },
    {
      "id": "bridge-006",
      "title": "STATIC_DOCTOR_PROBE declares grok:'up' (and maps xai) — may manufacture a FALSE_HEALTHY for a non-council voice",
      "severity": "SPECULATION",
      "category": "data-accuracy",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "The default constant declares grok:'up' and PROVIDER_VOICE maps xai→grok, but recorded council activity (per ground-truth notes) is claude+codex+cursor, gemini dead. If grok is not actually a configured/declared voice, a default run will classify grok as FALSE_HEALTHY (declared up, zero finals) — surfacing an inaccuracy in the hand-authored constant rather than a real lying instrument. This compounds bridge-001: the verdict reflects the constant, not doctor.py.",
      "suggestion": "Confirm grok/xai is a real declared cheval voice as of the constant's as_of date; if not, drop it from STATIC_DOCTOR_PROBE so the default run does not emit a self-manufactured FALSE_HEALTHY.",
      "confidence": 0.3,
      "faang_parallel": "Test-fixture drift: hardcoded expectations in a golden file that no longer match production reality, which Google's testing guidance flags as a leading source of 'flaky-but-actually-stale' assertions. Verify against the live roster before teaching anything here.",
      "metaphor": "A roll-call sheet that still lists a student who transferred out last term — every morning the teacher marks them 'absent without excuse,' and the alarm is about the sheet, not the student.",
      "teachable_moment": "Before elaborating, verify: is grok/xai a declared cheval voice as of the constant's date? Ground-truth notes (claude+codex+cursor live, gemini dead) suggest it may not be. If confirmed absent, this is a concrete instance of bridge-001's drift, not an independent bug — which is itself the lesson: a frozen declaration manufactures its own false positives.",
      "connection": "A worked example of bridge-001. If true, it's the strongest argument for the `as_of` stamp: the very first default run already disagrees with the live roster."
    },
    {
      "id": "bridge-007",
      "title": "'degraded' + fresh finals is classified DIVERGENT (lying) — partially-working may be mislabeled",
      "severity": "SPECULATION",
      "category": "design",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "CLAIMS_HEALTHY = {up, healthy, ok}; any other declared status (including 'degraded') falls to the else branch and, when fresh, is flagged DIVERGENT. The header comment intentionally groups 'down/unknown/degraded' together, but a 'degraded' voice that is still producing some successful finals is arguably reporting honestly, not lying. Treating degraded-but-alive as a lie could generate false positives if doctor.py ever emits a degraded state.",
      "suggestion": "Consider a third declared bucket for partial states (degraded/limited) that maps to a CONSISTENT/DEGRADED classification when the voice is alive, distinct from the DIVERGENT (declared-fully-down-but-alive) lie.",
      "confidence": 0.4,
      "faang_parallel": "Health checks across Kubernetes and AWS ELB are deliberately ternary-or-richer (Healthy / Degraded / Unhealthy), not binary, precisely because 'serving but impaired' is an honest, common, and actionable state that a two-valued model corrupts into a false alarm.",
      "metaphor": "A runner who tells you 'I'm limping but still moving' and gets recorded as a liar because the scoreboard only has the columns 'fine' and 'collapsed.'",
      "teachable_moment": "Health is rarely binary. A model that admits only up/down forces every honest middle state into one of the poles, and 'degraded but alive' lands on the lie pole. Verify whether doctor.py can actually emit 'degraded' before building the third bucket — if the producer is strictly binary today, this is YAGNI; if it can, the third bucket prevents a whole class of false DIVERGENT.",
      "connection": "Depends entirely on doctor.py's output vocabulary. Worth a one-line check of the producer's status enum before deciding to build or defer."
    },
    {
      "id": "bridge-008",
      "title": "I/O shell (arg parsing, readRecords, loadProbeTable, renderers, main) is untested",
      "severity": "LOW",
      "category": "testing",
      "file": "tools/instrument-truth-sensor.test.mjs",
      "description": "The test suite thoroughly covers the pure core (analyzeVoiceTruth, voiceOf) but exercises none of the thin shell. The untested surface is precisely where the edge-case risks identified above live: argVal parsing, NaN handling for --now/--freshness-days (bridge-004), the ungraceful probe-table load (bridge-003), and malformed-line resilience in readRecords. exit-code-IS-the-verdict gating depends on this shell behaving correctly.",
      "suggestion": "Add a few integration tests (via child_process spawning the script against a fixture jsonl + temp probe-table) asserting exit codes 0/1/2 and the INSUFFICIENT messages, or export the arg parser / readRecords and unit-test them directly.",
      "confidence": 0.7,
      "faang_parallel": "The test-pyramid lesson every FAANG CI enforces: pure logic gets cheap unit tests, but the *seams* — process exit codes, arg parsing, file I/O — need at least a thin band of integration tests, because that's where contracts meet the OS. CLI exit-code contracts in particular are almost always regressed by refactors that unit tests can't see.",
      "metaphor": "The engine is on a dynamometer and beautifully characterized, but no one has ever turned the ignition key — and the key is what the gate (CI) actually presses.",
      "teachable_moment": "When 'exit-code-is-the-verdict,' the shell *is* the contract, not boilerplate around it. The untested band is exactly where bridge-003 and bridge-004 live — meaning a handful of child_process tests asserting exit 0/1/2 and the INSUFFICIENT messages would both close this gap and pin the fixes for those findings.",
      "connection": "The natural home for regression tests of bridge-003/004. Closing this finding closes several others by construction."
    },
    {
      "id": "bridge-009",
      "title": "Honest false-positive guard and inverse-lie detection are well-designed",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "The core distinction — a genuinely-down voice (declared down + stale) stays CONSISTENT and only a DIVERGENCE between declared and real health is flagged — is the right invariant and is precisely what keeps this sensor from becoming noise. The symmetric handling of the inverse lie (FALSE_HEALTHY: declared up but gone silent), the DARK-vs-INVISIBLE split for unlisted voices, and the INSUFFICIENT-not-TRUTHFUL exit-1 semantics (refusing to assert clean without evidence) are all correctly aligned with the stated immune doctrine. The divergence-contract test suite covers every named case plus the guard.",
      "confidence": 0.9,
      "faang_parallel": "This is the precision-over-recall discipline that makes alerting systems trustworthy: Google's SRE guidance is blunt that an alert that fires on non-problems trains operators to ignore it, so the bar is 'only fire on a real divergence.' Catching the *inverse* lie (declared-up-but-silent) symmetrically is the same rigor Borg's task-health logic applies — both 'claims dead but running' and 'claims alive but gone' are failures, and a one-sided detector misses half of reality.",
      "metaphor": "A good lie-detector doesn't flag everyone who looks nervous — it flags only the gap between what someone *says* and what the polygraph *shows*. This tool measures exactly that gap, in both directions, and stays silent when there's no gap.",
      "teachable_moment": "The hardest and most valuable property of any immune sensor is the *false-positive guard*: not flagging a thing that is honestly broken-and-admitting-it. Getting that invariant right — divergence is the signal, not absolute state — is what separates a sensor people trust from one they mute. This is the load-bearing design decision in the PR.",
      "connection": "Everything in bridge-001/002/004 is in service of *protecting* this invariant from being undermined by self-referential defaults. The core is right; the request is to make the defaults worthy of it."
    },
    {
      "id": "bridge-010",
      "title": "Clean pure-core / thin-shell separation with import-safe main guard",
      "severity": "PRAISE",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "The pure core (no I/O, fed fixture records) is cleanly separated from the I/O wrapper, enabling the fast, hermetic test suite. The main-guard (import.meta.url === pathToFileURL(process.argv[1]).href) correctly prevents file reads/exit on import, Object.freeze hardens the constant, and readRecords' per-line try/catch makes it resilient to malformed log lines without crashing. Exit-code-is-the-verdict makes it directly gateable.",
      "confidence": 0.9,
      "faang_parallel": "Functional-core / imperative-shell (Gary Bernhardt) — the architecture that lets the most testable, decision-bearing logic run with zero I/O, exactly as React separates pure render from side-effecting effects. The import-safe `main` guard is the Node equivalent of Python's `if __name__ == '__main__'`, and getting it right is what makes the module both a CLI and a safely-importable test target.",
      "metaphor": "A clean kitchen where the recipe (pure core) is written separately from the act of going to the market and lighting the stove (the shell). You can test the recipe without burning anything down — which is why the test suite is fast and hermetic.",
      "teachable_moment": "Functional-core/imperative-shell isn't an aesthetic preference — it's what made bridge-009's invariant cheaply testable and what makes bridge-008's gap *small* (only the thin shell is untested, not the logic). Object.freeze on the constant and per-line try/catch in readRecords are the right defensive touches. This structure is why the remaining requests are surgical, not architectural.",
      "connection": "The separation is what makes every other finding easy to fix: the core is sound, so all the work lives in a thin, well-bounded shell."
    }
  ]
}
```


## Architectural Meditation

There's a single thread running through bridge-001, -002, -004, and -006: **what is this sensor's external reference, and does it fail closed?** A lie-detector earns trust only when its own ground-truth is independent of the thing it judges. Right now the two defaults — a frozen declaration constant and a self-anchored freshness clock — both fold the reference *inward*, and a typo on the freshness window quietly flips the verdict to clean. Each is individually minor; together they describe a tool that can confidently report TRUTHFUL while blind, which is the precise pathology the PR was written to kill, relocated one level up.

The fix is small and unifying: one external clock (wall-time), an `as_of` stamp that decays the constant to INSUFFICIENT rather than to a false lie, and a uniform fail-closed input contract (bridge-003/004 routed through the same `die(... INSUFFICIENT)` the log path already honors). Land those plus a thin band of child_process exit-code tests (bridge-008), and the instrument becomes worthy of its own doctrine.

## Callouts

- **[Design] The false-positive guard is the whole ballgame, and it's right.** (bridge-009) Flagging only the *divergence* — not the absolute state — is what keeps this from becoming muted noise. Catching the inverse lie (declared-up-but-silent) symmetrically, and the DARK/INVISIBLE split, show real care for the failure taxonomy. This is the hardest property to get right and it's correct.
- **[Maintainability] Functional-core / imperative-shell, done properly.** (bridge-010) The pure core with fixture records, the import-safe `main` guard, `Object.freeze` on the constant, and per-line `try/catch` resilience are exactly the touches that make the test suite fast and the remaining fixes surgical rather than structural.
- **[Operational] INSUFFICIENT-not-TRUTHFUL is the correct moral spine.** Refusing to assert clean without evidence, and encoding that as exit-1, is the single most important behavioral choice in the file — it makes the tool directly gateable and aligns it with the immune doctrine it serves. The requests above are all in defense of *this* principle, not against it.

zkSoju:
## Addressed Bridgebuilder review (commit 981d09d9)

First cross-model look on this PR — substantive findings, mostly on the recursion the reviewer named: *what is the sensor's external reference, and does it fail closed?* Fixed all actionable findings; rejected one as YAGNI.

### Fixed
- **bridge-003** (MEDIUM, the highest-confidence finding) — `loadProbeTable()` now fails closed like the log path: read/parse errors and non-object shapes (the `typeof [] === 'object'` footgun) route through `die(… INSUFFICIENT)` / exit-1, never a raw stack trace or a silently-wrong classification.
- **bridge-004** (LOW) — `--freshness-days` / `--now` are validated at the boundary. A NaN no longer silently collapses every divergence into a false `TRUTHFUL`.
- **bridge-002** (MEDIUM) — freshness now anchors to **wall-clock** by default in the shell, not the log's own newest record. A frozen dispatch stream is detectable now (declared-up voices surface as `FALSE_HEALTHY`) and log staleness is a first-class surfaced signal. The pure core is unchanged (still defaults to newest-record so tests stay deterministic); `--now` overrides for tests.
- **bridge-006** (SPECULATION → grounded to source) — dropped `grok` from `STATIC_DOCTOR_PROBE`. Grounded against `doctor.py`: its `_PROBE_TABLE` is contract-pinned to `{claude, codex, gemini}` (`test_doctor.py::TestProbeTableContract`) and *deliberately* excludes the new-company adapters (`cursor`, `grok`). Declaring `grok: 'up'` misrepresented the real probe table. Now `grok` reads like `cursor` — `INVISIBLE` when live, `DARK` when idle. **Confirmed on live data** (1441 records): grok has fresh finals → it correctly flags `INVISIBLE` (a real lie of omission), not a manufactured `FALSE_HEALTHY`.
- **bridge-001** (REFRAME) — added `STATIC_DOCTOR_PROBE_AS_OF` and surfaced it in board/probe output ("static snapshot as of … — re-ground with --probe-table"), so the declaration's static nature is first-class. Chose surface-it over auto-decay-to-INSUFFICIENT: the constant has no natural "too old" threshold, and auto-failing the banner on an arbitrary age would itself be an over-eager numb instrument — the grounded fix is dropping the actually-wrong `grok` entry + making staleness visible + offering `--probe-table` to re-ground.
- **bridge-005** (LOW) — corrected the `MAX_RECORDS` comment: parsing is bounded, the file read is not.
- **bridge-008** (LOW) — added 10 `child_process` integration tests pinning the exit-code contract (0/2 + INSUFFICIENT exit-1 on every bad-input path), which also pin the bridge-003/004 fixes.

### Rejected
- **bridge-007** (SPECULATION) — the `degraded` third bucket is **YAGNI**. Grounded against `doctor.py`: its status vocabulary is `auth_state ∈ {ok, needs-login, unreachable, unknown}` — it cannot emit `degraded`. No producer can drive that branch today.
- **bridge-009 / bridge-010** — PRAISE, no action.

**Tests: 22 green** (12 core + 10 new shell integration). Live `--probe` run verified (exit 2, correctly flagging claude:DIVERGENT, cursor:INVISIBLE, grok:INVISIBLE).

PR #321, opened 2026-06-26
feat(shared/immune): wire the immune doctors into continuous teeth (M0/G-DOOR)

zkSoju:
## What this does — wires the immune DOCTORS into continuous TEETH (M0 / G-DOOR)

The estate-immune campaign landed two read-only **doctors** (#317 gate-freeze-sensor, #319 instrument-truth-sensor). They diagnose but don't yet *bite*. This PR is the **teeth** half of the doctor→aligner→teeth triad (bead `arrakis-estate-immune-epic-4xw6.1`):

| File | What |
|---|---|
| `.github/workflows/immune-doctors.yml` | Runs `gate-freeze-sensor` against this repo's merge gate on **every PR + a daily schedule + manual dispatch**, reporting the freeze verdict. Makes G-DOOR's *"re-freeze auto-detected"* mechanical. |
| `tools/immune-check.sh` | Single **local** entry that runs **both** doctors and emits a one-line STATUS tile per doctor (the sensors' `--probe` tiles) + a combined verdict/exit. |
| `tools/immune-check.test.sh` | Fixture-driven aggregation contract test (15 assertions, no live deps). |

### Informational, NOT a new required gate (the load-bearing design choice)
The door is **already frozen** behind too many required checks (`Unit Tests` red across the whole backlog). Adding a *new blocking* check would deepen the exact disease this campaign cures. So the workflow is **soft by construction**: it **always exits 0** (green check); a `FROZEN` verdict surfaces **loud** via the job summary + a `::warning::` annotation, not a red X. **Do not add this job to branch protection's required list.** A red-on-every-PR check would just be another numb gate; a green-with-a-loud-summary check carries the signal without freezing flow.

### The two doctors split by where ground-truth lives
- `gate-freeze-sensor` reads the live merge gate via `gh` → CI-runnable (this workflow).
- `instrument-truth-sensor` reads **local** `.run/model-invoke.jsonl` (absent in CI) → its only teeth is the local banner `tools/immune-check.sh`.

`immune-check.sh` is the union: the one place that runs **every** doctor the banner should surface, including the local-only one CI can never see.

### Aggregation contract
`exit 2 = PROBLEM` (any doctor flags) · `exit 1 = INSUFFICIENT` (couldn't ground — no `gh` auth / no audit log / an uncontracted exit code) · `exit 0 = HEALTHY`. Never claims HEALTHY without evidence.

## Tests
- `tools/immune-check.test.sh` — **15/15 green** (3×3 severity matrix · crash-code→INSUFFICIENT · multiline-tile→valid JSON via `jq -e` · `--log` arg · `--json` contract).
- Vendored doctor suites re-run green on Node 22 (`gate-freeze-sensor.test.mjs`, `instrument-truth-sensor.test.mjs`).
- `shellcheck -S warning` clean; workflow YAML has **zero invalid permission keys**; all three render paths (FROZEN / degraded-UNGATED? / crash-guard) confirmed.

**Live verdict against this repo:** `PROBLEM` — door **FROZEN** (`Unit Tests` red 34/34) + instrument-truth **LYING** (`claude:DIVERGENT`, `cursor:INVISIBLE`). The immune system surfacing the disease it was built to cure.

## Cross-model council review (2 iterations, converging 7→3 findings)
Ran the cross-family FAGAN council (`openai:codex-headless` + `cursor:cursor-headless`; the claude voice dropped `exit:2/empty`, cross-family guarantee still met). Verified every finding against the repo before acting:

- **[CRITICAL, codex — fixed]** `administration: read` is **not a grantable GITHUB_TOKEN permission key** (confirmed against the workflow-syntax docs). The workflow would have *failed to parse*. Removed; corrected the false premise that CI can read branch protection (it can't with the default token → sensor honestly degrades to `UNGATED?`).
- **[CRITICAL, codex — fixed]** the optional admin-read PAT was exposed to `pull_request` runs that execute PR-authored sensor code (a "pwn request" exfiltration vector). The PAT is now used **only on schedule/dispatch** (trusted default-branch runs); PRs fall back to the default token.
- **[major, codex — reproduced an infinite loop, fixed]** `--log` with no value hit `shift 2` on a 1-arg stack → the parse loop spun forever. Now validated.
- **[major ×2 — fixed]** hand-rolled `--json` escaper → invalid JSON on multiline tiles; replaced with `jq -n --arg`.
- **[major — fixed]** uncontracted doctor exit collapsed to HEALTHY (false green); now → INSUFFICIENT.
- **[cleanup — fixed]** workflow log tile derived from the same `gate.json` (no racy second `gh` probe).
- **[1 finding rejected]** a cursor "CRITICAL" that upload-artifact needs `actions: write` was **not verified** (artifact v4 uses the runtime token); risk neutralized by **dropping the artifact-upload step** entirely (verdict already lives in the summary/annotation/logs). 1 unstructured cursor objection rejected as non-actionable.

## Dependency
Vendors the two doctor files from **#317** and **#319** (commit `chore(shared/immune): vendor the immune doctors`) so this branch is self-contained and runnable. **Byte-identical** to those PR branches — if #317/#319 merge first, the trivial add/add overlap resolves to the same content; drop the vendored copies then.

## Status
**Draft — the door is frozen.** This PR rides the same frozen gate it diagnoses; merge is operator-gated after #317/#319 and a door-thaw. Bead `arrakis-estate-immune-epic-4xw6.1`.

### DIFF
--- a/.github/workflows/immune-doctors.yml
+++ b/.github/workflows/immune-doctors.yml
@@ -1,0 +1,166 @@
+# Immune Doctors — INFORMATIONAL teeth for the estate immune kit (G-DOOR).
+#
+# Runs the read-only gate-freeze-sensor against THIS repo's merge gate on every PR
+# and on a daily schedule, and reports the freeze verdict. This makes G-DOOR's
+# "re-freeze auto-detected" mechanical — the door can no longer silently re-freeze
+# without a loud signal on every PR and every day.
+#
+# DESIGN — informational, NEVER a new required gate (bead arrakis-estate-immune-epic-4xw6.1):
+# the door is ALREADY frozen behind too many required checks ('Unit Tests' red on the whole
+# backlog). Adding a NEW blocking/required check would deepen the freeze — the exact disease
+# this campaign exists to cure. So this job is SOFT by construction:
+#   - it ALWAYS exits 0 (the check is green even when the door is FROZEN),
+#   - the verdict is surfaced LOUD via the job summary + a ::warning:: annotation, not a red X,
+#   - do NOT add this job's check to branch protection's required list.
+# A red-on-every-PR check would be just another numb gate ([[ci-sensors-must-not-be-numb]]);
+# a green-with-a-loud-summary check carries the signal without freezing flow.
+#
+# The OTHER immune doctor (instrument-truth-sensor) reads LOCAL .run/model-invoke.jsonl, which
+# does not exist in CI — it cannot run here. Its teeth are the local banner tools/immune-check.sh.
+#
+# Read-only: `gh` reads only (PR list, status rollups, branch protection). No mutation, no merge.
+
+name: Immune Doctors
+
+on:
+  pull_request:
+  schedule:
+    - cron: '17 13 * * *'   # daily at 13:17 UTC — catches a re-freeze between PRs
+  workflow_dispatch:
+
+# Least privilege, all READ.
+#
+# HONEST LIMITATION: the sensor confirms a freeze by reading branch-protection
+# required_status_checks, which requires ADMIN read. `administration` is NOT a grantable
+# GITHUB_TOKEN permission key (the workflow `permissions:` vocabulary has no such scope), so
+# the default token CANNOT read protection — the sensor then degrades to "UNGATED?" (it never
+# false-asserts FROZEN). To get the full freeze verdict, add a repo/org secret
+# IMMUNE_DOCTOR_GH_TOKEN = a fine-grained PAT with "Administration: read". For security it is
+# exposed ONLY when the checked-out ref is the default branch (refs/heads/main) — see the probe
+# step's GH_TOKEN guard. The daily schedule always runs there; a workflow_dispatch off a feature
+# branch and any PR run execute potentially-modified sensor code and deliberately fall back to the
+# read-only default token. So the DAILY run is the real "re-freeze auto-detected" teeth; PR runs
+# (and off-default dispatches) give a lighter signal.
+# The local teeth (tools/immune-check.sh, run with the operator's admin `gh` auth) get the full
+# verdict regardless.
+permissions:
+  contents: read
+  pull-requests: read
+  checks: read
+  statuses: read
+
+# Newer pushes to the SAME PR supersede that PR's in-flight run. Every non-PR run (schedule,
+# workflow_dispatch) gets a UNIQUE group (github.run_id) so a manual dispatch and the daily
+# schedule never cancel each other — they resolve to the same ref (refs/heads/main) and would
+# otherwise collide under a ref-keyed group.
+concurrency:
+  group: immune-doctors-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  gate-freeze:
+    name: gate-freeze (informational)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Probe the merge gate (read-only; never blocks)
+        id: probe
+        env:
+          # The admin-read PAT (full freeze verdict) is gated on the REF being executed, not just the
+          # event name: it is exposed ONLY when the checked-out ref is the trusted default branch
+          # (refs/heads/main). schedule always runs there; workflow_dispatch CAN target an arbitrary
+          # ref, and any pull_request checks out and executes PR-AUTHORED code (the sensor itself) —
+          # in both of those cases a modified sensor could exfiltrate the PAT (the "pwn request"
+          # class), so they fall back to the read-only default token (the sensor degrades to UNGATED?,
+          # never false-FROZEN). Trust derives from the ref, so the comment matches the mechanism.
+          GH_TOKEN: ${{ (github.event_name != 'pull_request' && github.ref == 'refs/heads/main') && secrets.IMMUNE_DOCTOR_GH_TOKEN || github.token }}
+          REPO: ${{ github.repository }}
+        run: |
+          set +e   # the sensor exits 2 when FROZEN — we report it, we do NOT fail on it.
+          node tools/gate-freeze-sensor.mjs "$REPO" --json > gate.json 2> gate.err
+          sensor_exit=$?
+          echo "sensor_exit=$sensor_exit" >> "$GITHUB_OUTPUT"
+
+          # Guarantee a valid-JSON artifact even if the sensor crashed — surface the crash,
+          # never silently bank a green. (cluster-compliance FAGAN iter-2 pattern.)
+          if ! jq -e . gate.json >/dev/null 2>&1; then
+            jq -n --arg err "$(tail -c 2000 gate.err 2>/dev/null)" --argjson ex "${sensor_exit:-1}" \
+              '{verdict:"ERROR", sensor_exit:$ex, stderr:$err, openPRs:0, frozenChecks:[], requiredChecks:null}' \
+              > gate.json
+          fi
+
+          VERDICT=$(jq -r '.verdict // "ERROR"' gate.json)
+          ROOT=$(jq -r '.root // "—"' gate.json)
+          OPEN=$(jq -r '.openPRs // 0' gate.json)
+          echo "verdict=$VERDICT" >> "$GITHUB_OUTPUT"
+          echo "root=$ROOT"       >> "$GITHUB_OUTPUT"
+
+          # Compact one-line tile in the log (mirrors the local banner row), derived from the
+          # SAME gate.json — not a second live `gh` probe, which could race PR-status changes.
+          jq -r '"  ☾ gate-freeze " + .repo + " · " + .verdict
+                 + (if .root then ": " + .root else "" end)
+                 + " · " + (.openPRs|tostring) + " open PRs"' gate.json || true
+
+          {
+            echo "## Immune Doctors · gate-freeze — $(date -u +%Y-%m-%dT%H:%M:%SZ)"
+            echo ""
+            echo "**Repo:** \`$REPO\` · **verdict:** \`$VERDICT\` · **open PRs:** $OPEN · **root:** \`$ROOT\`"
+            echo ""
+            echo "> Informational, non-required. This check is **always green** by design — a frozen"
+            echo "> door is surfaced here as a warning, not a red X, so it never deepens the freeze it"
+            echo "> reports. (bead arrakis-estate-immune-epic-4xw6.1 · G-DOOR teeth)"
+            echo ""
+            if [ "$VERDICT" = "FROZEN" ]; then
+              echo "### ⚠ The merge gate is FROZEN"
+              echo ""
+              echo "Root required check freezing the backlog: \`$ROOT\`"
+              echo ""
+              echo "| required check | red | pending | missing | of PRs | not-green % |"
+              echo "|----------------|-----|---------|---------|--------|-------------|"
+              jq -r '.frozenChecks[]? | "| \(.name) | \(.failOn) | \(.pendingOn) | \(.missingOn) | \(.ofPRs) | \(.freezePct)% |"' gate.json
+            elif [ "$VERDICT" = "FLOWING" ]; then
+              echo "### ✓ The door flows — no required check is frozen across the backlog."
+            else
+              echo "### · $VERDICT"
+              echo ""
+              echo "(No frozen required check could be confirmed. \`UNGATED?\` = branch protection"
+              echo "unreadable from CI; \`UNGATED\` = no required checks; \`EMPTY\` = no open PRs.)"
+            fi
+          } >> "$GITHUB_STEP_SUMMARY"
+
+          # A loud-but-non-failing annotation when frozen (visible in the Checks/Files tabs).
+          if [ "$VERDICT" = "FROZEN" ]; then
+            echo "::warning title=Merge gate FROZEN::root required check '$ROOT' is red across the open-PR backlog — see job summary. (informational; does not block this PR)"
+          fi
+
+          # ALWAYS green — informational by design. (No artifact upload step: the verdict lives
+          # in the job summary + annotation + logs; an upload step is one more way to redden an
+          # invariantly-green job for zero added signal.)
+          exit 0
+
+  # Unlike the informational gate-freeze job above, THIS job is a REAL gate (it can go red): it runs
+  # the doctors' fixture tests so a regression in the pure cores turns this check red without a human
+  # looking — moving the tests from "exists" to "runs" ([[verification-ladder-exists-to-verified]]).
+  # The tests are pure (no live `gh`, no audit log) → CI-safe and deterministic.
+  tests:
+    name: immune doctor unit tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Run the immune doctors' fixture tests (each gates on failure)
+        run: |
+          node tools/gate-freeze-sensor.test.mjs
+          node tools/instrument-truth-sensor.test.mjs
+          bash tools/immune-check.test.sh
--- a/tools/gate-freeze-sensor.mjs
+++ b/tools/gate-freeze-sensor.mjs
@@ -41,6 +41,7 @@
  *                                              # Re-run as fixes land; exit 2 = still frozen.
  */
 import { execFileSync } from 'node:child_process';
+import { pathToFileURL } from 'node:url';
 
 const ARGS = process.argv.slice(2);
 const MODE = ARGS.includes('--json') ? 'json' : ARGS.includes('--probe') ? 'probe' : 'board';
@@ -86,6 +87,19 @@ function classify(result) {
 }
 
 /**
+ * parseRequiredContexts(rsc) → string[] — PURE. Extract the required-check context NAMES from a
+ * branch-protection `required_status_checks` object. GitHub returns BOTH `contexts[]` (legacy) and
+ * `checks[].context` (new); prefer whichever is POPULATED. A naive `contexts || checks…` short-
+ * circuits on an EMPTY-but-truthy `contexts: []` and never reaches a populated `checks[]` — the
+ * contexts→checks deprecation time-bomb. Length-test instead of truthiness.
+ */
+export function parseRequiredContexts(rsc = {}) {
+  const ctx = Array.isArray(rsc.contexts) ? rsc.contexts : [];
+  const fromChecks = Array.isArray(rsc.checks) ? rsc.checks.map((c) => c.context).filter(Boolean) : [];
+  return ctx.length ? ctx : fromChecks;
+}
+
+/**
  * fetchRepoData(repo) → { repo, defaultBranch, required, strict, prs } — the I/O half (gh reads).
  * `required` is the branch-protection required-check contexts, or null if unreadable.
  */
@@ -101,8 +115,8 @@ function fetchRepoData(repo) {
     const prot = ghJSON(['api', `repos/${repo}/branches/${defaultBranch}/protection`], true);
     const rsc = prot.required_status_checks || {};
     strict = !!rsc.strict;
-    // GitHub returns either contexts[] (legacy) or checks[].context (new)
-    required = rsc.contexts || (rsc.checks || []).map((c) => c.context) || [];
+    // GitHub returns either contexts[] (legacy) or checks[].context (new) — prefer the populated one.
+    required = parseRequiredContexts(rsc);
   } catch {
     required = null;
   }
@@ -185,10 +199,17 @@ export function analyzeBacklog({ repo, defaultBranch = 'main', required, strict 
     : !hasRequired ? 'UNGATED'         // readable: genuinely no required status checks → CI can't freeze
     : 'FLOWING';
 
+  // Exit code IS the verdict, and it must speak all THREE tiers immune-check.sh reads (0 HEALTHY /
+  // 1 INSUFFICIENT / 2 PROBLEM), not just two. UNGATED? = branch protection unreadable = we could
+  // NOT ground the freeze question → INSUFFICIENT (1), parallel to instrument-truth-sensor's
+  // can't-ground exit. Collapsing it to 0 would let immune-check read a blind gate as HEALTHY — the
+  // exact numb-false-green this suite exists to kill ([[ci-sensors-must-not-be-numb]]).
+  const exitCode = verdict === 'FROZEN' ? 2 : verdict === 'UNGATED?' ? 1 : 0;
+
   return {
     repo, defaultBranch, openPRs: prs.length, truncated,
     requiredChecks: required, strict, behindCount: behind,
-    verdict, root: root ? root.name : null,
+    verdict, exitCode, root: root ? root.name : null,
     frozenChecks: frozen.map((c) => ({ name: c.name, failOn: c.fail, pendingOn: c.pending, missingOn: c.missing, ofPRs: prs.length, freezePct: Math.round(c.freeze * 100) })),
     duplicateRequiredChecks: checks.filter((c) => c.duplicateRuns && hasRequired && c.isRequired).map((c) => ({ name: c.name, runs: c.duplicateRuns })),
     ungatedFailing,
@@ -202,11 +223,20 @@ function analyzeRepo(repo) {
 
 function main() {
   if (ARGS.includes('--estate')) return estateMode();
-  const summary = analyzeRepo(resolveRepo());
+  // Wrap the single-repo read so a `gh` failure (e.g. `gh pr list` unauthenticated) becomes a clean
+  // one-line INSUFFICIENT diagnostic, not a raw Node stack trace bleeding into immune-check's tile.
+  // (estateMode keeps its OWN per-repo try/catch → ERROR, so it stays resilient — die() here would
+  // kill the whole sweep, which is why we catch on the single-repo path only.)
+  let summary;
+  try {
+    summary = analyzeRepo(resolveRepo());
+  } catch (e) {
+    die(`could not read ${REPO_ARG || 'repo'} (${e.code || e.message}) — INSUFFICIENT`);
+  }
   if (MODE === 'json') console.log(JSON.stringify(summary, null, 2));
   else if (MODE === 'probe') console.log(renderProbe(summary));
   else console.log(renderBoard(summary));
-  process.exit(summary.verdict === 'FROZEN' ? 2 : 0);
+  process.exit(summary.exitCode);
 }
 
 // --estate: sweep every repo with open PRs in an org, group freezes by ROOT CAUSE, and rank the
@@ -328,5 +358,6 @@ function renderBoard(s) {
   return L.join('\n');
 }
 
-// run as CLI only (not when imported by the test)
-if (import.meta.url === `file://${process.argv[1]}`) main();
+// run as CLI only (not when imported by the test). Use pathToFileURL (percent-encodes paths with
+// spaces/special chars) instead of a hand-built `file://` string — matches instrument-truth-sensor.
+if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) main();
--- a/tools/gate-freeze-sensor.test.mjs
+++ b/tools/gate-freeze-sensor.test.mjs
@@ -3,7 +3,7 @@
  * Fixture test for gate-freeze-sensor's PURE core (analyzeBacklog) — NO live `gh`, CI-safe.
  * Run: node tools/gate-freeze-sensor.test.mjs   (exit 0 = pass, 1 = fail; the verdict gates.)
  */
-import { analyzeBacklog } from './gate-freeze-sensor.mjs';
+import { analyzeBacklog, parseRequiredContexts } from './gate-freeze-sensor.mjs';
 
 let failed = 0;
 function check(name, cond) {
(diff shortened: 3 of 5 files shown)


zkSoju (commented):
I have everything needed to produce this review — the PR context and the convergence findings are fully provided, and my task is enrichment and synthesis, not codebase exploration or implementation. This is a read-only writing deliverable (the review text itself), so I'll produce it directly rather than route it through the implementation-planning workflow, which doesn't fit a review-generation task.

---

## Summary

This PR closes the loop the estate has been circling for weeks: it takes the immune *doctors* (gate-freeze + instrument-truth sensors) and gives them continuous teeth via an always-green CI workflow. The engineering is unusually self-aware — pure-core/IO-shell separation, a thoughtful pwn-request token posture, and a CI design that refuses to deepen the very freeze it detects. The primary concern is a cluster of **numb-false-green** seams: a sensor that exits 0 when it genuinely *cannot determine* the answer, a `||` short-circuit on truthy-empty arrays, and freshness anchored to the log's own tip rather than wall-clock — each of which lets a blind instrument report HEALTHY, the exact failure class this suite exists to kill.

## Architectural Observations

The deepest tension in this PR is **the drift-detector that embeds the drift it detects** (`static-probe-snapshot-self-drift`). A sensor whose entire thesis is "declaration ≠ ground truth, and the machine trusts the declaration" ships with a hand-authored declaration as its reference. This is the immune-system analog of an antibody that learns one pathogen's signature and never updates — the day `doctor.py` is *fixed*, the sensor screams LYING forever. The fix isn't more code; it's making the reference a live (bounded) probe so the comparison is truth-vs-truth in real time.

The second pattern worth naming: **exit codes are the contract**. `immune-check.sh` builds a careful three-tier model (HEALTHY / INSUFFICIENT / FROZEN) on top of exit codes, but `gate-freeze-sensor.mjs` only speaks two of those three dialects — it has no INSUFFICIENT voice. When two halves of a protocol disagree on the vocabulary, the gaps don't error; they resolve *silently* to the most optimistic reading. That's `gate-freeze-no-insufficient-exit`, and it's the load-bearing correctness finding here.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "gate-freeze-no-insufficient-exit",
      "title": "gate-freeze exits 0 on UNGATED? (protection unreadable) → immune-check can report a false HEALTHY",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "instrument-truth-sensor returns exit 1 (INSUFFICIENT) when it cannot ground, and immune-check.sh builds an entire INSUFFICIENT tier around that. But gate-freeze-sensor's main() only ever returns `verdict === 'FROZEN' ? 2 : 0`, so UNGATED? (branch protection unreadable — the documented common case without admin) exits 0. immune-check.sh maps 0→HEALTHY (sev() only rescues *uncontracted* codes to 1). So when the gate doctor genuinely could not determine the freeze state, the aggregate banner reads HEALTHY — exactly the numb false-green the suite cites to prevent. Locally with admin `gh` this is masked; it bites whenever protection is unreadable.",
      "suggestion": "Give gate-freeze a dedicated INSUFFICIENT exit (1) for verdict === 'UNGATED?' (and arguably 'ERROR'), so it parallels the instrument sensor and feeds immune-check's INSUFFICIENT tier instead of HEALTHY.",
      "confidence": 0.6,
      "faang_parallel": "This is the distinction monitoring teams at Google encode as the difference between a probe returning `OK` and returning `UNKNOWN`/`NO_DATA`. Borgmon/Monarch treat 'no data' as a first-class state precisely because collapsing it into 'healthy' is how outages hide — an alert that can't see is not the same as an alert that sees green, and SRE practice mandates they be wired to different downstream behavior.",
      "metaphor": "A smoke detector with a dead battery doesn't emit a 'no fire' signal — it chirps. This sensor, when blindfolded, currently nods 'all clear' instead of chirping. The blindfold itself is the alarm condition.",
      "teachable_moment": "A tri-state truth (yes / no / can't-tell) must never be projected onto a binary exit surface, because the projection always loses the most dangerous state. When the consumer (immune-check.sh) already speaks three tiers, the producer is obligated to speak all three.",
      "connection": "Directly the [[ci-sensors-must-not-be-numb]] pattern the suite cites — and the producer/consumer half of [[verification-ladder-exists-to-verified]]: the sensor 'exists' and 'runs' but does not 'enforce' on the unreadable-protection path."
    },
    {
      "id": "required-contexts-empty-shortcircuit",
      "title": "`required = rsc.contexts || checks.map(...)` short-circuits on an empty contexts[] → future freeze-blindness",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "In fetchRepoData: `required = rsc.contexts || (rsc.checks || []).map((c) => c.context) || []`. GitHub returns BOTH `contexts` and `checks` as arrays. An empty array is truthy in JS, so when GitHub deprecates `contexts` and returns `contexts: []` while `checks: [{context:'Unit Tests'}]` is populated, the `||` takes the empty `contexts` and the `checks` fallback is never reached. required=[] → verdict UNGATED → the sensor declares 'no required gate' for a repo that is in fact gated. For a tool whose sole purpose is detecting freezes, that is a silent false-negative. The fetchRepoData parsing is also not exercised by gate-freeze-sensor.test.mjs (only the pure core is), so this path is untested.",
      "suggestion": "Prefer the populated source explicitly: `const ctx = rsc.contexts ?? []; const fromChecks = (rsc.checks ?? []).map(c => c.context); required = ctx.length ? ctx : fromChecks;` and add a fixture/unit test for the checks-only shape.",
      "confidence": 0.55,
      "faang_parallel": "The `[] is truthy` trap is the JavaScript cousin of the bug classes static analyzers at Meta (Infer) and Google (the `||` vs `??` migration that motivated `??`'s addition to the language) were built to catch. The GitHub branch-protection API's `contexts`→`checks` deprecation is a real, scheduled migration — this is a time-bomb keyed to GitHub's own roadmap, not a hypothetical.",
      "metaphor": "It's checking the old mailbox for a letter, finding the mailbox *present but empty*, and concluding 'no mail exists' — never walking to the new mailbox by the door where the letter actually is.",
      "teachable_moment": "`||` tests *truthiness*; `??` tests *existence*; neither tests *populated*. When 'has data' is the real question, write the predicate you mean (`arr.length ? ...`). Falsy-fallback chains silently encode the assumption that the absent state is falsy — arrays violate that assumption.",
      "connection": "A concrete instance of declaration-vs-reality drift keyed to an upstream schema change — the same family as [[map-territory-drift-unifying-failure]], here triggered by GitHub's API evolution rather than the estate's own."
    },
    {
      "id": "static-probe-snapshot-self-drift",
      "title": "instrument-truth's reference declaration is a frozen 2026-06-26 snapshot — the drift-detector embeds the drift it detects",
      "severity": "MEDIUM",
      "category": "design",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "STATIC_DOCTOR_PROBE is a hardcoded snapshot of doctor.py's output as observed on one day (claude:'unknown', cursor absent, etc.). The sensor's entire thesis is 'declaration != ground truth, and the machine trusts the declaration' — yet its default reference IS a hand-authored declaration that will itself drift from doctor.py's real behavior. The moment doctor.py is fixed (claude reported correctly, cursor added), this sensor will keep emitting DIVERGENT/INVISIBLE → verdict LYING (exit 2) forever, becoming a false-positive instrument until someone manually edits the constant. It does not actually read doctor.py; it compares ground truth against a stale picture of doctor.py. The `--probe-table` override mitigates only if someone remembers to use it.",
      "suggestion": "Either (a) drive the probe table from a cheap, bounded live probe of doctor.py (with a hard timeout to avoid the documented hang) so the comparison is declaration-vs-truth in real time, or (b) reframe/rename the default to make explicit that exit 2 means 'doctor.py as-of-snapshot diverges' and require a date-stamped, CI-validated refresh of STATIC_DOCTOR_PROBE.",
      "confidence": 0.6,
      "faang_parallel": "This is the Netflix Chaos Monkey paradox: a tool that asserts 'the system has drifted' must itself be continuously validated against the live system, or it becomes the thing generating the false signal. Golden-file/snapshot tests at scale (Jest snapshots, Google's golden tests) carry the identical hazard — a stale golden masquerades as a real assertion until someone notices the failures are all the snapshot's fault.",
      "metaphor": "A clock built to detect when the town clock is wrong — but it's a *photograph* of the town clock taken last Tuesday. The instant the town clock is repaired, the photograph starts 'detecting' a fault that no longer exists, forever, until someone retakes the photo.",
      "teachable_moment": "When a detector's correctness depends on a reference, the reference is a dependency with a maintenance contract — not a constant. Either make it self-refreshing (live probe) or make its staleness loud (date-stamp + CI freshness check). A silent frozen reference inverts the tool's purpose the day its subject improves.",
      "connection": "The recursive case of this very PR's thesis and [[map-territory-drift-unifying-failure]]: the divergence sensor is itself a map that will diverge from territory. The fix is the same divergence-sensor-on-the-divergence-sensor regress the estate keeps rediscovering."
    },
    {
      "id": "relative-freshness-masks-total-outage",
      "title": "Freshness anchored to newest-record (not wall-clock) can hide a fleet-wide cheval outage in immune-check",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "analyzeVoiceTruth defaults `reference = now || newestTs(records)` — freshness is relative to the log's own latest entry. If the entire cheval fleet has been dead for a week, the newest record is a week old, every voice is measured against that week-old tip, and a voice whose last final == the tip is 'fresh'. So a probe declaring codex 'up' when codex (and everything) actually died days ago classifies CONSISTENT, not FALSE_HEALTHY — the sensor misses a total outage. immune-check.sh invokes the doctor with no `--now`, so the local banner inherits this blind spot. The `now defaults to newest record` test (test file) confirms the semantics but not the outage consequence.",
      "suggestion": "For the gating/banner path, default the freshness anchor to wall-clock `new Date()` (or pass `--now=$(date -u ...)` from immune-check.sh), keeping newest-record only as an explicit opt-in for offline snapshot analysis.",
      "confidence": 0.5,
      "faang_parallel": "This is exactly why monitoring systems alert on *heartbeat absence* against wall-clock, not against the last-seen sample. Prometheus' `up == 0` and the classic 'staleness' alert (`time() - last_scrape > threshold`) exist because anchoring freshness to the data's own newest point makes a frozen pipeline look eternally current — a failure mode that took down more than one team's dashboards during a collector outage.",
      "metaphor": "Asking 'is this milk fresh?' by comparing it to the *oldest* carton in a fridge that lost power a week ago. Relative to its dead neighbors, every carton looks fine. Only the wall clock tells you the whole fridge is spoiled.",
      "teachable_moment": "Freshness is meaningful only against an absolute, externally-advancing clock. Anchoring 'recent' to the dataset's own maximum timestamp creates a self-referential frame that cannot, by construction, detect that the dataset itself stopped advancing.",
      "connection": "A second numb-false-green seam, sibling to gate-freeze-no-insufficient-exit: both let a *total* failure of the underlying signal read as healthy. Ties to [[truth-ground-up-stack]]'s 'check the time axis' discipline."
    },
    {
      "id": "tests-not-wired-to-ci",
      "title": "New tests are added but nothing in the diff runs them in CI",
      "severity": "LOW",
      "category": "testing",
      "file": "tools/gate-freeze-sensor.test.mjs",
      "description": "gate-freeze-sensor.test.mjs, instrument-truth-sensor.test.mjs, and immune-check.test.sh are added, but the diff wires neither a package.json script nor a CI step to execute them (immune-doctors.yml runs the sensor, not the tests). By the suite's own ladder these are at 'exists', not 'runs' — regressions in the pure cores won't be caught automatically. (Lower confidence: the repo may have an existing *.test.mjs glob runner not visible in this diff.)",
      "suggestion": "Add a test step (e.g. `node --test tools/*.test.mjs && bash tools/immune-check.test.sh`) to an existing test workflow or a package.json `test` script. Note the two .mjs tests also use different conventions (node:test vs a custom runner) — unify so one runner covers both.",
      "confidence": 0.4,
      "teachable_moment": "An unrun test is documentation, not a gate. It earns the name 'test' only when a regression in the code under test turns it red without a human looking. Verify the glob-runner assumption before treating this as resolved — that's where the 0.4 confidence points.",
      "connection": "Textbook [[verification-ladder-exists-to-verified]]: these sit on rung 1 ('exists') while the PR title promises 'continuous teeth' (rung 4, 'enforces'). The mixed node:test / custom-runner convention is a small instance of the same drift."
    },
    {
      "id": "fragile-main-guard-gate-freeze",
      "title": "gate-freeze main-guard uses unencoded `file://${argv[1]}` — breaks CLI dispatch on paths with spaces/special chars",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "`if (import.meta.url === `file://${process.argv[1]}`) main();` compares against a hand-built file URL. import.meta.url is percent-encoded, so a repo path containing a space or other special char will not match and main() silently won't run when invoked as a CLI. The sibling instrument-truth-sensor.mjs does this correctly with `pathToFileURL(process.argv[1]).href`. Current repo paths have no spaces, so it works today, but it's a latent inconsistency between two files that otherwise mirror each other.",
      "suggestion": "Use the same robust guard as the sibling: `import { pathToFileURL } from 'node:url'; if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) main();`",
      "confidence": 0.7,
      "faang_parallel": "Hand-rolled URL/path concatenation instead of the platform encoder is the same class as the path-normalization bugs that motivated Node to ship `pathToFileURL` in the first place. The 'works on my machine, breaks under `/Application Support/`' failure is a perennial macOS/CI divergence.",
      "metaphor": "Two twins given the same job; one uses the official translator, the other guesses the spelling. They agree until the first word with an accent mark.",
      "teachable_moment": "When two sibling files implement the same idiom and one uses the stdlib helper while the other hand-rolls it, the hand-rolled one is almost always the latent bug — converge on the library form. The encoder exists precisely for the inputs you didn't think of.",
      "connection": "Consistency-between-mirrors is itself a coherence signal; this is a micro-instance of the divergence the PR's own sensors hunt at the estate scale."
    },
    {
      "id": "workflow-dispatch-secret-exposure-overclaim",
      "title": "workflow_dispatch can target an arbitrary ref with the admin PAT — comment overstates 'trusted default-branch'",
      "severity": "LOW",
      "category": "security",
      "file": ".github/workflows/immune-doctors.yml",
      "description": "The GH_TOKEN expression exposes secrets.IMMUNE_DOCTOR_GH_TOKEN on `schedule` and `workflow_dispatch`. The comments describe these as 'trusted default-branch checkouts', but workflow_dispatch can be run against any ref by anyone with write access, and the job checks out and executes that ref's tools/gate-freeze-sensor.mjs. A write-access user could dispatch a branch with a modified sensor and exfiltrate the PAT. Blast radius is limited (PAT is Administration:read; requires write access), so risk is low — but the safety narrative is stronger than the actual trust boundary.",
      "suggestion": "Either gate the secret on `github.ref == 'refs/heads/<default>'` in addition to event_name, or soften the comment to acknowledge that workflow_dispatch trust derives from write-access + a tightly-scoped PAT, not from a default-branch checkout.",
      "confidence": 0.5,
      "faang_parallel": "GitHub's own security guidance on the pwn-request / `pull_request_target` class draws exactly this line: trust must be derived from the *ref being executed*, not the *event name*. The 2021–2022 wave of Actions PAT-exfil disclosures were almost all 'the comment said trusted, the ref said otherwise'.",
      "metaphor": "A vault door labeled 'staff only' that, in fact, anyone with a building badge can open and then walk in carrying their own lockpick. The label isn't lying about intent — it's lying about the mechanism.",
      "teachable_moment": "A safety *comment* is a claim that must match the safety *mechanism*. When the prose says 'trusted default-branch' but the guard keys only on `event_name`, the gap is an audit trap — the next maintainer trusts the comment, not the YAML. Either tighten the guard to make the comment true, or weaken the comment to make it honest.",
      "connection": "The shadow side of the PRAISE on this same file: the token posture is genuinely disciplined, which makes the one over-claiming comment the exact spot a reviewer's trust gets misplaced."
    },
    {
      "id": "gate-freeze-unwrapped-gh-throw",
      "title": "gate-freeze does not wrap `gh pr list` — a gh failure surfaces an uncaught stack trace into the banner",
      "severity": "LOW",
      "category": "robustness",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "fetchRepoData's `gh pr list` call is not in a try/catch (unlike the protection and default-branch reads). If `gh` is unauthenticated or errors, the throw propagates out of main() as an uncaught exception — exit 1 (which immune-check correctly treats as INSUFFICIENT), but the tile captured by immune-check.sh's `2>&1` is a raw Node stack trace rather than a clean one-line diagnostic. The instrument sensor handles its read errors via `die()` with a tidy message; the gate sensor is inconsistent.",
      "suggestion": "Wrap the pr-list read and emit a clean `die('gh pr list failed for <repo> — ...')`, or catch at main() to print a one-line error before exiting 1.",
      "confidence": 0.55,
      "metaphor": "A diagnostic instrument that, when it fails, prints its own circuit diagram across the dashboard instead of a single red light. Correct in outcome (it failed), illegible in delivery.",
      "teachable_moment": "Error *outcome* and error *presentation* are separate obligations. Exiting 1 satisfies the protocol; a stack trace bleeding into a banner UI fails the operator who has to read it at 2am. The sibling already sets the clean-`die()` bar — match it.",
      "connection": "Same convergence-toward-the-better-sibling as the main-guard finding: instrument-truth-sensor is quietly the reference implementation both halves should mirror."
    },
    {
      "id": "instrument-input-validation-gaps",
      "title": "No validation on --freshness-days / --probe-table inputs",
      "severity": "LOW",
      "category": "robustness",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "`--freshness-days=abc` → Number('abc')=NaN → freshnessMs=NaN → every `ageMs <= NaN` is false → all voices silently classified as not-fresh, producing a wrong-but-quiet verdict. A `--probe-table=bad.json` (unreadable or invalid JSON) throws from loadProbeTable() outside any try/catch → uncaught stack trace + exit 1 rather than the contracted clean INSUFFICIENT `die` message.",
      "suggestion": "Validate freshDays with `Number.isFinite` and die on NaN; wrap loadProbeTable's readFileSync/JSON.parse and die('cannot read probe table ...') on failure.",
      "confidence": 0.5,
      "faang_parallel": "NaN's poisoning of comparisons (`x <= NaN` is always false) is the same silent-corruption class that hardened input-parsing libraries (Google's flag libraries, Rust's `clap` strict parsing) reject loudly at the boundary rather than letting propagate. Fail-at-parse beats fail-mysteriously-three-functions-later.",
      "metaphor": "A thermostat that, handed a garbled temperature, doesn't beep — it just quietly decides nothing is ever warm enough. The wrongness is invisible because it never raises its voice.",
      "teachable_moment": "Validate at the trust boundary, not after the value has flowed into arithmetic. NaN is the canonical example of a value that fails *quietly* — every comparison against it is false, so it produces a confident wrong answer instead of an error.",
      "connection": "Another instance of the same INSUFFICIENT-contract discipline as the lead finding: the clean-`die` path exists, but two inputs bypass it into uncaught throws."
    },
    {
      "id": "readrecords-reads-whole-file",
      "title": "readRecords loads the entire log before tailing, despite the 'bounded tail' comment",
      "severity": "LOW",
      "category": "performance",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "The MAX_RECORDS comment says 'bounded tail — read the recent window, never unbounded', but the implementation does `readFileSync(logPath,'utf8')` (whole file) then `lines.slice(-MAX_RECORDS)`. The parse is bounded; the read is not. For a large model-invoke.jsonl this reads the full file into memory each run. Functionally correct, just not what the comment implies.",
      "suggestion": "Either soften the comment to 'bounded parse window' or read only the tail bytes (e.g. read from end via fd, or a streaming tail) if log size becomes a concern.",
      "confidence": 0.6,
      "metaphor": "A 'sip from the top' label on a process that drains the whole tank, measures it, then pours all but the last cup back. The result is a sip; the work is the tank.",
      "teachable_moment": "A comment that describes a performance property the code doesn't have is worse than no comment — it suppresses the future optimization by asserting it's already done. Comments make promises; this one writes a check the code doesn't cash. Lowest-effort fix is honesty (reword); higher-value is a real tail read if the log grows.",
      "connection": "A documentation-vs-reality micro-drift — the same declaration≠truth theme the PR is built around, scoped to a single comment."
    },
    {
      "id": "concurrency-group-schedule-dispatch-collision",
      "title": "Schedule and workflow_dispatch share the same concurrency group (refs/heads/main)",
      "severity": "LOW",
      "category": "design",
      "file": ".github/workflows/immune-doctors.yml",
      "description": "`group: immune-doctors-${{ github.event.pull_request.number || github.ref }}` with `cancel-in-progress: true`. For both schedule and workflow_dispatch on the default branch, github.ref resolves to refs/heads/main, so a manual dispatch will cancel an in-flight scheduled run (and vice versa). The header comment claims 'the daily/dispatch runs are unaffected', which isn't quite true for dispatch-vs-schedule.",
      "suggestion": "Incorporate github.run_id or github.event_name into the group for non-PR events if independent daily/dispatch runs are desired, or update the comment to reflect the shared group.",
      "confidence": 0.5,
      "metaphor": "Two doctors sharing one pager: when the on-call one is paged, the scheduled rounds get bumped — even though they meant to work in parallel.",
      "teachable_moment": "Concurrency keys silently conflate every event that resolves to the same ref. If two event sources are meant to run independently, the event source must be *in* the key. Otherwise 'cancel-in-progress' becomes an invisible coupling.",
      "connection": "A third comment-vs-mechanism mismatch in this workflow (cf. the secret-exposure and always-green findings) — the YAML is good, but its prose runs slightly ahead of its guards."
    },
    {
      "id": "immune-check-json-no-jq-preflight",
      "title": "immune-check.sh --json depends on jq with no preflight check",
      "severity": "LOW",
      "category": "robustness",
      "file": "tools/immune-check.sh",
      "description": "With `set -uo pipefail` (no -e) and `--json`, if jq is absent the `jq -n ...` invocation fails (127) printing 'command not found' to stderr, then `exit \"$overall\"` runs — so the script exits with the aggregated code but emits no JSON, a confusing silent-ish failure for a documented mode. (The non-JSON banner path doesn't need jq, so this only affects --json.)",
      "suggestion": "Add a `command -v jq >/dev/null || { echo 'immune-check: --json requires jq' >&2; exit 1; }` guard at the top of the JSON branch.",
      "confidence": 0.5,
      "teachable_moment": "A documented output mode has an implicit dependency contract. Preflight the dependency and fail with a sentence a human can act on ('--json requires jq'), rather than letting the shell emit '127: command not found' and an empty payload that looks like an upstream bug.",
      "connection": "Graceful-degradation hygiene — the same operator-legibility concern as the unwrapped-gh-throw finding, on the consumer side of the suite."
    },
    {
      "id": "praise-pwn-request-hardening",
      "title": "PRAISE: disciplined pwn-request defense on the CI token",
      "severity": "PRAISE",
      "category": "security",
      "file": ".github/workflows/immune-doctors.yml",
      "description": "Using `pull_request` (not pull_request_target), forcing GH_TOKEN to github.token for ALL pull_request events while restricting secrets.IMMUNE_DOCTOR_GH_TOKEN to schedule/dispatch, and least-privilege read-only `permissions:` — combined with an honest in-file account of the `administration` scope limitation and the UNGATED? degradation — is exactly the right shape for running PR-authored code. The reasoning is documented where the next maintainer will see it.",
      "suggestion": "Keep this pattern as the template for future estate sensors that run repo code in CI.",
      "faang_parallel": "This is the posture GitHub's Security Lab recommends as the canonical defense against the pwn-request class — the same lesson the broader ecosystem learned the hard way through the 2021 wave of `pull_request_target` token-exfil CVEs. Seeing it applied proactively, with the reasoning in-file, is rare.",
      "teachable_moment": "The best security comments don't claim invulnerability — they name the residual limitation (here, the `administration` scope gap) so the next maintainer inherits the threat model, not just the config."
    },
    {
      "id": "praise-false-positive-guard",
      "title": "PRAISE: the false-positive guard makes the instrument sensor honest",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/instrument-truth-sensor.mjs",
      "description": "Distinguishing a genuinely-down voice (declared down + stale → CONSISTENT, not flagged) from a real divergence, plus catching the inverse lie (declared up + silent → FALSE_HEALTHY), is the conceptually hard part and it's implemented cleanly and exhaustively tested (cases a–e plus DARK and null-final). This prevents the sensor from itself becoming a numb red-on-everything instrument.",
      "suggestion": "No change — this is the load-bearing correctness of the tool.",
      "metaphor": "A good lie-detector doesn't just flag every nervous twitch — it knows the difference between a calm truth and a calm lie. This sensor's four-quadrant logic (declared × actual) is exactly that discernment.",
      "teachable_moment": "A detector earns trust by being symmetric about *both* false-positives and false-negatives. The cheap version flags everything (numb-red); the honest version models all four quadrants of declared-vs-actual — which is precisely what makes the static-snapshot finding worth fixing, so this quadrant logic keeps firing on truth."
    },
    {
      "id": "praise-pure-core-io-shell-split",
      "title": "PRAISE: pure-core / I/O-shell separation enabling fixture tests",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tools/gate-freeze-sensor.mjs",
      "description": "Both sensors isolate a pure analysis function (analyzeBacklog / analyzeVoiceTruth) from the gh/file I/O, making the verdict logic fixture-testable with no live dependencies. gate-freeze-sensor.test.mjs covers the meaningful states including the real #310 name-collision worst-wins case and the non-required-failing-check false-positive guard; the instrument test matrix is thorough.",
      "suggestion": "Extend the same fixture approach to fetchRepoData's branch-protection parsing (see the contexts-empty finding).",
      "faang_parallel": "This is the functional-core / imperative-shell discipline (Gary Bernhardt's 'Boundaries', and the structure behind Haskell-influenced testability at scale) — push the decisions into pure functions, keep I/O at the rim. It's why the verdict logic is testable without a live `gh`.",
      "teachable_moment": "The contexts-empty and unwrapped-throw findings both live in `fetchRepoData` — the *one* function still outside the pure core. That's not a coincidence: bugs cluster where the I/O shell hasn't yet been pulled into a fixture-testable seam. Extending the pattern there closes two findings at once."
    },
    {
      "id": "praise-always-green-by-design",
      "title": "PRAISE: informational always-green CI that refuses to deepen the freeze",
      "severity": "PRAISE",
      "category": "design",
      "file": ".github/workflows/immune-doctors.yml",
      "description": "Making the CI doctor exit 0 always — surfacing FROZEN via job summary + ::warning:: annotation rather than a red X, with an explicit 'do NOT add to required checks' instruction — correctly avoids adding another required check to a gate that is already frozen behind too many required checks. The self-aware design (a freeze-detector that won't itself contribute to the freeze) is the right resolution of the tension.",
      "suggestion": "No change.",
      "metaphor": "A fire inspector who reports violations in writing rather than chaining the exit door shut — the inspection must never become the thing blocking egress.",
      "teachable_moment": "A monitor for a systemic problem must not become an instance of that problem. An always-green-but-annotating check is the correct shape for any sensor watching a resource (here: the required-checks budget) that the sensor itself would consume by going red. Note the live tension with gate-freeze-no-insufficient-exit: 'never red' is right for the *gate budget*, but the sensor must still distinguish HEALTHY from can't-tell in its *exit code* so immune-check's banner stays honest."
    }
  ]
}
```


## Callouts

This is a PR that understands the trap it's built to detect, and mostly avoids falling into it. Three things stand out as genuinely excellent:

- **The functional-core / imperative-shell split** (`praise-pure-core-io-shell-split`) is the structural decision that makes everything else testable — and tellingly, *every* correctness finding in the gate sensor lives in `fetchRepoData`, the lone function still on the I/O side of that seam. The bugs cluster exactly where the discipline hasn't reached yet. That's a roadmap, not a rebuke.
- **The four-quadrant false-positive guard** (`praise-false-positive-guard`) is the conceptually hard kernel, and it's exhaustively tested. It's what keeps this sensor from becoming the numb-red instrument the estate is allergic to — which is also why the static-snapshot finding deserves attention: that quadrant logic is only as honest as the reference it compares against.
- **The always-green-by-design CI** (`praise-always-green-by-design`) and the **pwn-request token posture** (`praise-pwn-request-hardening`) show a maintainer thinking about second-order effects: a freeze-detector that won't deepen the freeze, and a PR-code runner that won't leak its PAT.

The through-line of the four MEDIUM findings is one lesson worth carrying forward: **a tri-state truth must never collapse onto a binary surface, and a detector's reference must advance with the territory.** Fix the INSUFFICIENT exit, the truthy-empty short-circuit, the frozen snapshot, and the relative-freshness anchor, and these doctors will have teeth that bite the right things — and stay silent on the rest.

zkSoju:
## Bridgebuilder review addressed (commit e915e7d6)

Verified each of the 12 actionable findings against the code and fixed all of them — surgically, no scope creep. The 4 PRAISE items need no change. None of the findings were hallucinated; this was a well-grounded review.

### MEDIUM
- **`gate-freeze-no-insufficient-exit`** — gate-freeze now speaks all three tiers immune-check reads. Added an `exitCode` to the pure core: `UNGATED?` (protection unreadable) → exit **1 (INSUFFICIENT)**, not 0/HEALTHY. Parallels instrument-truth's can't-ground exit. `main()` now uses `summary.exitCode`. (+5 tests)
- **`required-contexts-empty-shortcircuit`** — extracted pure `parseRequiredContexts(rsc)` that length-tests instead of the truthy-`||` that short-circuits on an empty-but-truthy `contexts: []` (the GitHub `contexts`→`checks` deprecation time-bomb). (+5 tests, incl. the checks-only shape)
- **`static-probe-snapshot-self-drift`** — added `STATIC_DOCTOR_PROBE_ASOF` and stamp the probe source (`static snapshot as-of 2026-06-26`) in the board/probe/`--json` output, so an exit-2 divergence reads honestly as "vs the snapshot", never mistaken for live-doctor.py. (Did **not** add a live doctor.py probe — the code documents that it hangs; took the reframe/date-stamp option.)
- **`relative-freshness-masks-total-outage`** — the live CLI now anchors freshness to **wall-clock now** (was the log's own newest record, which can't detect a total fleet outage). Pure core keeps newest-record for offline/fixture analysis. (+1 test proving outage detection)

### LOW
- **`fragile-main-guard-gate-freeze`** — `pathToFileURL(argv[1])` (was hand-built `file://`, breaks on paths with spaces); matches the sibling.
- **`gate-freeze-unwrapped-gh-throw`** — single-repo `gh` failure → clean one-line `INSUFFICIENT` die, not a raw stack trace. Caught at `main()` on the single-repo path so estate mode keeps its own per-repo resilience (a `die()` inside `fetchRepoData` would have killed the whole sweep).
- **`workflow-dispatch-secret-exposure-overclaim`** — the admin PAT is now gated on `github.ref == 'refs/heads/main'` (not just `event_name`), so a `workflow_dispatch` off a feature branch can't exfiltrate it. Comment now matches the mechanism.
- **`concurrency-group-schedule-dispatch-collision`** — non-PR runs get a unique group (`github.run_id`) so schedule and dispatch no longer cancel each other.
- **`tests-not-wired-to-ci`** — added a real (red-able) `tests` job to `immune-doctors.yml` running all three fixture suites (verified there was no pre-existing glob runner — the 0.4-confidence caveat). Ran each test with the runner it expects (the mixed node:test/custom convention is noted).
- **`instrument-input-validation-gaps`** — validate `--freshness-days` (NaN poison → loud die) and wrap `--probe-table` read in a clean die.
- **`readrecords-reads-whole-file`** — honest comment: "bounded PARSE window" (the read is whole-file; tail-byte read flagged as the upgrade).
- **`immune-check-json-no-jq-preflight`** — `command -v jq` preflight for `--json`.

### Verification
All three suites green (gate-freeze 24 assertions, instrument-truth 13 tests, immune-check all), shellcheck clean at warning severity, YAML parses. Manually exercised the new I/O-shell paths (NaN freshness, bad probe-table, missing jq, gh-throw, CLI dispatch). No new FAGAN council run (self-review only, to avoid cheval contention with BB).

PR #322, opened 2026-06-26
feat(shared/immune): M0 re-derive-from-ground-truth enforcing lint (stacked on #321)

zkSoju:
## M0 — re-derive-from-ground-truth enforcing lint

> **Draft — stacked on #321** (`campaign/immune-doctor-teeth`, the door-teeth branch with `immune-doctors.yml` + `tools/immune-check.sh`). The door is frozen; review/merge order is #321 first. Bead `arrakis-estate-immune-epic-4xw6.1`.

The campaign's anti-recurrence **keystone**: a lint that makes it mechanically impossible to land a NEW immune instrument that hasn't been bound — in a reviewer-visible diff — to the ground source it re-derives its verdict from. This dissolves the lying-instruments cluster permanently (M0 / G-IMMUNE: "every verdict/health/goal/sensor re-derives from a ground source; a lint blocks new self-reporting instruments").

### What the lint enforces — TWO mechanical contracts, nothing fuzzier
1. **Registration teeth.** Every file under `tools/` that *looks* like an immune instrument — name matches `*-sensor.*` / `*-doctor.*` / `gate-*.*`, or carries an opt-in `# immune-instrument` header — MUST be a key in `tools/immune-instruments.yaml`. A new, **unregistered** instrument FAILS the lint. (`*.test.*` / `*.spec.*` are excluded — they verify instruments, they aren't instruments.)
2. **Ground-source binding.** Each registered instrument's declared ground-source token(s) must **literally appear** (fixed-string grep) in its source. If the declared read is removed/refactored away, the token vanishes and the lint fails — forcing a re-confirmation.

### Structural-not-semantic — the design rationale (fuzziness guard)
It is a **regression-catcher, not a semantic verifier.** It deliberately does NOT try to detect "emits a verdict but reads no ground source" — that is statically intractable and false-positive-prone. **The human approving the registry diff owns the semantic judgement** ("is this token really the ground source?"). The lint only guarantees the claim was *made* and the declared read still *exists*. Same discipline as its sibling `tools/check-no-direct-llm-fetch.sh`: literal substrings, an opt-in registry, a reviewer-visible suppression marker (`# immune-lint:allow <reason>`), and the exit code IS the verdict. Markers are matched **line-anchored to a real comment** (defeats same-line string smuggling, tested); the multiline-string-literal spoof is a **documented accepted risk** — the lexer it would need is exactly the broad analyzer the fuzziness guard forbids, and the markers are reviewer-visible so human review is the gate.

### Files
- **`tools/immune-instruments.yaml`** — the registry. Seeds the 3 existing instruments: `gate-freeze-sensor.mjs → execFileSync('gh'`, `instrument-truth-sensor.mjs → .run/model-invoke.jsonl`, `governance-doctor.sh → grimoires/loa/reality + git log`.
- **`tools/check-instrument-ground-truth.sh`** — the lint. Hermetic (awk YAML parser, **no yq dependency** — runs anywhere). `0 clean / 1 violation / 2 arg-error`; `--probe` one-line tile.
- **`tools/check-instrument-ground-truth.test.sh`** — 9 fixtures.
- **`tools/immune-check.sh`** (+ test) — wires the lint as a third, CI-runnable check with an **explicit severity remap** (the lint's `0/1/2` is NOT the doctors' `0/1/2` convention).
- **`.github/workflows/immune-doctors.yml`** — a NON-REQUIRED, informational job (always exit 0; verdict surfaced LOUD via job summary + `::warning`), plus an informational M0-fixture step. **Never a new required check** — adding one would deepen the very frozen door this campaign cures.

### Planted-regression test result (the M0 acceptance criterion)
A planted self-reporting instrument (emits a verdict, reads no ground source, unregistered) → **lint exits 1** ✓. A registered+grounded instrument → exits 0 ✓. The suppression marker works ✓. All **9/9 fixtures green**; the lint runs **clean on the real `tools/`** (3 instruments, no false positives); `shellcheck` clean.

### Council verdict (cross-model FAGAN, codex + cursor)
Ran the real cheval council on the diff, **3 rounds to convergence**:
- **codex (`openai:codex-headless`): APPROVED, 0 findings — every round.**
- **cursor (`cursor:cursor-headless`): 6 → 3 → 4 findings.** All substantive findings verified against the repo and fixed: quoted-aware registry parser (inline comments now parse, was a TOKEN_ABSENT false-positive); a bad scan root now fails loud (was a silent false-green); markers line-anchored (defeats string-literal smuggling — two new negative fixtures); `*.spec.*` excluded; CI `::warning` on exit-2 misconfig; `immune-check.sh` header updated to a three-check contract; `mktemp` in the test.
- **Rejected (verified, deliberate):** cursor's round-3 push for a quote/heredoc-tracking lexer to defend marker-in-multiline-string evasion — that is the statically-intractable broad analyzer the fuzziness guard forbids (documented as an accepted risk; human review is the gate). Cursor's `.doctors`→`.checks` JSON rename — would break the `--json` contract #321 established; the aggregate verdict is authoritative and each element is name-tagged.

### Out of scope / noted
- One **pre-existing** `shellcheck` SC2015 (info) on `immune-check.sh`'s `sym=$(...)` glyph line is from #321, untouched here (surgical-changes discipline).
- The fixture test follows the immune-kit's local-test convention (like `immune-check.test.sh`); it is also exercised informationally in the CI job.

### DIFF
--- a/.github/workflows/immune-doctors.yml
+++ b/.github/workflows/immune-doctors.yml
@@ -164,3 +164,85 @@ jobs:
           node tools/gate-freeze-sensor.test.mjs
           node tools/instrument-truth-sensor.test.mjs
           bash tools/immune-check.test.sh
+
+  # ───────────────────────────────────────────────────────────────────────────
+  # The M0 / G-IMMUNE enforcing lint (check-instrument-ground-truth.sh). Unlike
+  # the instrument-truth doctor (needs a local .run/model-invoke.jsonl), this one
+  # IS CI-runnable — it only reads checked-in files. Same SOFT design as gate-freeze:
+  # ALWAYS green, the verdict surfaced LOUD via summary + annotation, never a new
+  # required check (a violation here is rare + actionable, but a new blocking gate
+  # would deepen the very freeze this campaign cures). The teeth are the
+  # reviewer-visible registry diff + the local banner tools/immune-check.sh.
+  # ───────────────────────────────────────────────────────────────────────────
+  ground-truth-lint:
+    name: ground-truth-lint (informational)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Lint immune instruments against their ground sources (read-only; never blocks)
+        run: |
+          set +e   # the lint exits 1 on a violation — we report it, we do NOT fail on it.
+          out="$(bash tools/check-instrument-ground-truth.sh 2>&1)"; lint_exit=$?
+          printf '%s\n' "$out"
+
+          {
+            echo "## Immune Doctors · ground-truth-lint — $(date -u +%Y-%m-%dT%H:%M:%SZ)"
+            echo ""
+            echo "> Informational, non-required. M0 / G-IMMUNE: every verdict/health/sensor instrument"
+            echo "> must register its ground source in \`tools/immune-instruments.yaml\`. This check is"
+            echo "> **always green** by design so it never deepens the freeze it guards against; the"
+            echo "> registry diff + \`tools/immune-check.sh\` are the teeth. (bead arrakis-estate-immune-epic-4xw6.1)"
+            echo ""
+            if [ "$lint_exit" -eq 0 ]; then
+              echo "### ✓ Every immune instrument re-derives from a registered ground source."
+            elif [ "$lint_exit" -eq 1 ]; then
+              echo "### ⚠ An immune instrument is unregistered or its declared ground-source token is missing"
+              echo ""
+              echo "Register the instrument (path + literal ground-source token) in \`tools/immune-instruments.yaml\`,"
+              echo "or mark a non-instrument file with \`# immune-lint:allow <reason>\`."
+            else
+              echo "### · The lint could not run (exit $lint_exit) — a misconfiguration, not a verdict."
+            fi
+            echo ""
+            echo '```'
+            printf '%s\n' "$out"
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+
+          if [ "$lint_exit" -eq 1 ]; then
+            echo "::warning title=Immune instrument ungrounded::a new immune instrument is unregistered or its ground-source token is missing — see job summary. (informational; does not block this PR)"
+          elif [ "$lint_exit" -ne 0 ]; then
+            echo "::warning title=Ground-truth lint misconfigured::the lint exited $lint_exit (argument/IO error) and could not run — see job summary. (informational; does not block this PR)"
+          fi
+
+          # ALWAYS green — informational by design (same rationale as the gate-freeze job).
+          exit 0
+
+      - name: Exercise the M0 acceptance fixture (informational — proves the lint still catches a planted liar)
+        run: |
+          # The lint's M0 acceptance criterion is "a planted self-reporting instrument fails the
+          # lint." Run that fixture here so the proof is exercised on every PR — but SOFT: a fixture
+          # failure means the lint's own logic regressed (rare + actionable), surfaced LOUD via the
+          # summary + a warning, never a red X (this job stays always-green by design).
+          set +e
+          test_out="$(bash tools/check-instrument-ground-truth.test.sh 2>&1)"; test_exit=$?
+          printf '%s\n' "$test_out"
+          {
+            echo "## Immune Doctors · ground-truth-lint · M0 fixture — $(date -u +%Y-%m-%dT%H:%M:%SZ)"
+            echo ""
+            if [ "$test_exit" -eq 0 ]; then
+              echo "### ✓ The planted self-reporting instrument still fails the lint (M0 acceptance holds)."
+            else
+              echo "### ⚠ The M0 acceptance fixture FAILED — the lint's own logic may have regressed."
+            fi
+            echo ""
+            echo '```'
+            printf '%s\n' "$test_out"
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+          if [ "$test_exit" -ne 0 ]; then
+            echo "::warning title=M0 fixture regressed::tools/check-instrument-ground-truth.test.sh failed — the lint may no longer catch a planted liar. (informational; does not block this PR)"
+          fi
+          exit 0
--- a/tools/check-instrument-ground-truth.sh
+++ b/tools/check-instrument-ground-truth.sh
@@ -89,11 +89,16 @@ QUIET=0
 PROBE=0
 declare -a SCAN_ROOTS=()
 
-# Name-globs that mark a file as an immune-instrument CANDIDATE. Intentionally
-# narrow — only the three shapes immune instruments are named with.
-_name_is_candidate() {
+# Tri-state classification of a filename (a property of the FILE, evaluated ONCE so
+# it governs BOTH the name path and the opt-in-header path — a *.test.*/*.spec. file
+# carrying an `# immune-instrument` marker in its head is STILL excluded):
+#   0 = name-candidate  (matches an instrument name-glob)
+#   2 = excluded        (a test/spec — never an instrument, regardless of any header)
+#   1 = neither         (not name-matched; the opt-in header may still pull it in)
+# Intentionally narrow — only the three shapes immune instruments are named with.
+_name_candidacy() {
     case "$1" in
-        *.test.*|*.spec.*) return 1 ;;  # tests/specs verify instruments; they are not instruments
+        *.test.*|*.spec.*) return 2 ;;  # tests/specs verify instruments; they are not instruments
         *-sensor.*|*-doctor.*|gate-*.*) return 0 ;;
         *) return 1 ;;
     esac
@@ -145,6 +150,30 @@ if [[ ! -f "$REGISTRY" ]]; then
     printf 'check-instrument-ground-truth.sh: registry not found: %q\n' "$REGISTRY" >&2
     exit 2
 fi
+
+# Validate + canonicalize every scan root up front. Two startup contracts:
+#   (a) each root must EXIST as a directory — a lint that inspects zero files and
+#       exits 0 is the false-green this campaign kills; refuse to scan nothing.
+#   (b) each root must resolve UNDER --repo-root. The registry keys are
+#       repo-root-relative (rel="${f#"$REPO_ROOT/"}"), so a root outside REPO_ROOT
+#       would yield an absolute key that can never match the registry — a confusing
+#       spurious UNREGISTERED. Make the --root/registry-key contract explicit by
+#       failing loud (exit 2) instead. Canonicalizing here also makes the prefix
+#       strip exact for relative/symlinked roots.
+for i in "${!SCAN_ROOTS[@]}"; do
+    root="${SCAN_ROOTS[$i]}"
+    if [[ ! -d "$root" ]]; then
+        printf 'check-instrument-ground-truth.sh: scan root %q is not a directory — refusing to scan nothing\n' "$root" >&2
+        exit 2
+    fi
+    abs_root="$(cd "$root" && pwd)"
+    if [[ "$abs_root" != "$REPO_ROOT" && "$abs_root" != "$REPO_ROOT"/* ]]; then
+        printf 'check-instrument-ground-truth.sh: scan root %q does not resolve under --repo-root %q\n' "$root" "$REPO_ROOT" >&2
+        printf '  (registry keys are repo-root-relative; a root outside it can never match)\n' >&2
+        exit 2
+    fi
+    SCAN_ROOTS[i]="$abs_root"
+done
 
 # --------------------------------------------------------------------------
 # Registry parser (flat YAML map: "<path>:" then "  - \"<token>\"" lines).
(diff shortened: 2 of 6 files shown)


zkSoju (commented):
I have everything needed for this review inline — the convergence findings, PR context, and file diffstat. This is a self-contained review-generation task, not an implementation-planning task, so I'll deliver the Bridgebuilder review directly rather than route it through the plan-file machinery (a PR review doesn't belong in a plan file, and no edits or non-readonly tools are required).

---

## Summary

This PR adds a ground-truth lint (M0) that forces the immune-instrument registry to be re-derived from the actual `tools/` tree — closing the "taxonomy with no teeth" gap where a declared registry silently drifts from reality. The engineering discipline is genuinely strong: an inverted exit contract is remapped explicitly, the scan fails loud on an empty root, and the fixture is adversarial rather than happy-path. All open findings are `LOW` or `PRAISE` — verdict is **COMMENT**. The themes worth attention are *positional fragility* in the scanner (GTL-1, GTL-3) and *unverifiable-from-diff completeness* (GTL-2).

The architectural meditation here: this lint is a sensor whose *own correctness must be sensed*. A tripwire that can be evaded by reordering lines, or that reports PROBLEM on a tree the diff can't show you, erodes the very banner signal it exists to protect. The fixes below are about making the scanner's contract positional-invariant and confirming the live tree is clean before landing.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "GTL-1",
      "title": "Opt-in header scan ignores the test/spec name exclusion — a fixture marker in the first 40 lines would falsely register the test file as an instrument",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/check-instrument-ground-truth.sh",
      "description": "_name_is_candidate() excludes *.test.*/*.spec.* on the NAME path only. When the name doesn't match a glob, control falls into the else branch which runs OPTIN_RE over head -40 of EVERY file, including *.test.*/*.spec.*. The fixture's line-anchored '// immune-instrument' (case 5) and '# immune-lint:allow' (case 7) currently sit past line 40, so the test file is skipped by positional luck. Reorder case 5's marker into the first 40 lines and the .test.sh becomes an unregistered candidate — lint fails on the real tools/ tree.",
      "suggestion": "Apply the *.test.*/*.spec.* exclusion before the opt-in header check, or restructure _name_is_candidate to return a tri-state (excluded / name-candidate / neither) and treat 'excluded' as a hard non-candidate regardless of header content.",
      "confidence": 0.5,
      "faang_parallel": "This is the class of bug Google's C++ style guidance targets when it warns against decisions that depend on file layout rather than declared intent — and it echoes the early-exit asymmetry that bit linters like clang-tidy, where a NOLINT only honored on one code path let suppressed diagnostics leak back through an alternate branch.",
      "metaphor": "It's a metal detector at the door that's been told 'staff with badges walk free' — but only checks badges at the front entrance. The side door runs the same beep-on-metal scan on everyone, so a staffer who happens to walk in the side sets off the alarm. The exemption must live at the person, not the doorway.",
      "teachable_moment": "An exclusion rule should be a property of the entity, evaluated once, not re-litigated per code path. The moment the same 'is this excluded?' question is answered differently depending on which branch you fell into, you have two sources of truth for one fact.",
      "connection": "Directly mirrors the project's own [[verification-ladder]] thinking: the check 'exists' (name exclusion is written) but isn't 'enforced' uniformly across both candidate paths."
    },
    {
      "id": "GTL-2",
      "title": "Registry completeness is load-bearing but unverifiable from this diff — any other glob-matching file under tools/ flips immune-check.sh to PROBLEM",
      "severity": "LOW",
      "category": "completeness",
      "file": "tools/immune-instruments.yaml",
      "description": "Pass A scans all of tools/ recursively and flags any *-sensor.*/*-doctor.*/gate-*.* (or opt-in) file not in the registry. Only three instruments are registered. If the live tree contains any other glob-matching file, the lint exits 1 — harmless in the always-green CI job but persistently reporting PROBLEM through immune-check.sh, eroding the banner. The diff alone can't confirm registry completeness.",
      "suggestion": "Before landing, run `bash tools/check-instrument-ground-truth.sh` against the real repo and confirm exit 0; register any other glob-matching files or add a `# immune-lint:allow <reason>` marker. Note the expected clean result in the PR description.",
      "confidence": 0.45,
      "faang_parallel": "Amazon's operational-readiness reviews codify exactly this: a check that asserts a global invariant must be run against production state, not just the change set, before sign-off — the 'green in the diff, red in the fleet' gap is a recurring postmortem root cause.",
      "metaphor": "It's a fire-code inspector who certifies the blueprint of three rooms but the building has a fourth room he never walked. The certificate looks valid; the building still trips the alarm every night.",
      "teachable_moment": "When a lint asserts a whole-tree invariant, the PR diff is necessarily insufficient evidence — the proof lives in running it against the live tree. State that proof in the PR body, or the reviewer is approving on faith.",
      "connection": "This is the [[deployed-but-unconsumed]] failure inverted into a sensor: completeness asserted in a registry, never confirmed against the territory it claims to cover — the same map/territory drift the campaign exists to kill."
    },
    {
      "id": "GTL-3",
      "title": "Relative-key computation assumes --root lives under --repo-root; an independent --root yields an absolute key that never matches the registry",
      "severity": "LOW",
      "category": "robustness",
      "file": "tools/check-instrument-ground-truth.sh",
      "description": "rel=\"${f#\"$REPO_ROOT/\"}\" only strips when the found file is prefixed by REPO_ROOT. The CLI advertises --root as an independently repeatable scan root. A --root outside REPO_ROOT makes the strip a no-op, rel stays absolute, the file can never match a repo-relative registry key, and a spurious UNREGISTERED violation appears with a confusing absolute path. The fixture avoids this only by always setting --root under --repo-root.",
      "suggestion": "Either derive rel relative to the matching scan root (with registry keys expressed accordingly), or validate at startup that every --root resolves under --repo-root and exit 2 with a clear message otherwise, making the --root/registry-key contract explicit.",
      "confidence": 0.4,
      "faang_parallel": "Bazel's repository-relative label space exists precisely so paths normalize to one canonical root; the bugs Bazel avoids are the ones that appear when an absolute path leaks into a space that only ever compares repo-relative keys.",
      "metaphor": "It's giving someone directions 'three doors down from the lobby' but they entered from a different building entirely — the offset is meaningless without a shared starting point. Either pin the lobby or refuse the directions.",
      "teachable_moment": "A CLI that advertises flexibility (independent --root) owes the user either real support for it or a loud refusal — the dangerous middle is silent partial support that fails confusingly only on the untested path.",
      "connection": "Same family as GTL-1: correctness that holds only because the fixture happens to stay inside the safe envelope. Untested degrees of freedom are latent bugs."
    },
    {
      "id": "GTL-4",
      "title": "immune-check.test.sh keeps a shared /tmp output path while the new sibling test uses mktemp for parallel-safety",
      "severity": "LOW",
      "category": "consistency",
      "file": "tools/immune-check.test.sh",
      "description": "The new check-instrument-ground-truth.test.sh deliberately uses a per-run LINT_OUT=$(mktemp) for parallel-safety. immune-check.test.sh (modified here) still writes to the fixed /tmp/immune-check-smoke.out, which races under concurrent runs. Pre-existing, but this PR substantially edits the file and the better pattern is now established right beside it.",
      "suggestion": "Switch /tmp/immune-check-smoke.out to a mktemp-allocated per-run file (trap-cleaned), matching the sibling test's parallel-safe convention.",
      "confidence": 0.5,
      "faang_parallel": "Bazel's sandboxed test execution exists because shared /tmp paths are the classic source of flaky, order-dependent test failures — the 'works alone, fails in CI matrix' ghost that costs more debugging hours than the fix ever would.",
      "metaphor": "Two cooks sharing one cutting board labeled 'cutting-board' — fine until they're in the kitchen at the same time. mktemp hands each cook their own board.",
      "teachable_moment": "When you introduce the better pattern in a new file, the cheapest moment to converge the sibling is now, while both are open in the same diff — consistency debt compounds quietly otherwise.",
      "connection": "A small instance of the campaign's larger theme: the immune toolchain's own tests should model the hygiene the tools enforce."
    },
    {
      "id": "GTL-5",
      "title": "PRAISE — adversarial fixture coverage and line-anchored marker defense",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tools/check-instrument-ground-truth.test.sh",
      "description": "The 9-case fixture covers the real evasion surface, not the happy path: string-literal smuggling for both the suppression marker (case 7) and opt-in header (case 8), the *.test./*.spec. exclusions (cases 6, 9), the M0 acceptance criterion via a planted self-reporting instrument (case 1), and token-drift (case 4). Pairing line-anchored OPTIN_RE/SUPPRESS_RE regexes with tests that prove a smuggled marker does NOT evade registration is the right discipline for a tripwire that explicitly disclaims being an adversarial boundary.",
      "suggestion": "No change. Consider adding one case asserting a registered-but-deleted instrument yields MISSING_FILE (exit 1), currently untested.",
      "faang_parallel": "This is the mutation-testing mindset Google's testing culture rewards: a test suite earns trust by proving the evasions fail, not just that the blessed path passes.",
      "metaphor": "A good lock isn't tested by opening it with the right key — it's tested by proving the lockpicks, the credit card, and the bump key all fail. This fixture brings the lockpicks.",
      "teachable_moment": "For any string-matching guard, the load-bearing tests are the ones proving a value that *looks* like the marker (in a string literal, past the scan window) does not trigger it. Negative-evasion cases are where regex tripwires live or die.",
      "connection": "Embodies [[ci-sensors-must-not-be-numb]] preemptively — the fixture proves the sensor bites before the sensor ever ships."
    },
    {
      "id": "GTL-6",
      "title": "PRAISE — inverted exit contract is remapped explicitly (not via sev()) with dedicated tests, and the scan fails loud on an empty/missing root",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "tools/immune-check.sh",
      "description": "The lint's 0=clean/1=violation/2=arg-error contract inverts the doctors' severity, and immune-check.sh handles it with an explicit case-remap (violation→PROBLEM, arg-error→INSUFFICIENT) rather than routing through sev(), with a matching 3-row test matrix. The lint refuses to scan a missing/empty root (exit 2, 'refusing to scan nothing') instead of inspecting zero files and returning a false green, and every gate captures $? before any pipe per the gate-output rule.",
      "suggestion": "No change.",
      "faang_parallel": "The 'refuse to scan nothing' guard is the defensive posture behind Google's SRE 'fail loud, fail safe' principle — a monitor that returns green on zero inputs is worse than no monitor, because it manufactures false confidence.",
      "metaphor": "A smoke detector that beeps when its own battery is missing, rather than going silent and letting you believe the house is safe. Silence-as-success is the failure mode that kills.",
      "teachable_moment": "Two patterns worth stealing: (1) when an exit contract is inverted, remap it explicitly at the boundary rather than overloading a shared mapper — the explicitness is the documentation; (2) an empty input set is an error to surface, never a clean result to report.",
      "connection": "Directly answers [[gate-output-never-piped]] and the estate's 'numb sensor' anti-pattern — this is what teeth that can't be silently bypassed look like."
    }
  ]
}
```


## Callouts

- **[Correctness] The empty-root refusal (GTL-6) is the soul of this campaign.** Most sensor bugs aren't false alarms — they're false *all-clears*. Choosing exit 2 / "refusing to scan nothing" over a vacuous green is the single most important line in the PR, because it makes the one unforgivable failure mode (silent numbness) impossible by construction.

- **[Testing] The fixture treats its own tripwire as untrusted (GTL-5).** Proving that a marker smuggled into a string literal or pushed past the 40-line window does *not* evade registration is exactly the discipline a "this is a tripwire, not a security boundary" tool should have. It's honest about its limits and still tests the edges of them.

- **[Quality] Explicit exit-contract remapping over a shared `sev()`.** Resisting the temptation to overload one mapper for two inverted contracts is the kind of small, deliberate clarity that saves a future maintainer an afternoon. The 3-row test matrix pins it so it can't silently regress.

The connective tissue across the LOW findings: GTL-1, GTL-2, and GTL-3 are all *the fixture passes because it stays inside a safe envelope the production tree may not*. None are blockers, but landing GTL-2's "ran clean against the live tree, exit 0" line in the PR description — and tightening GTL-1's exclusion to be path-invariant — would make this sensor as positionally robust as it is conceptually sound.

zkSoju:
Addressed the Bridgebuilder review (`f4e9a4ec`):

- **GTL-1** (real) — fixed. `_name_is_candidate` is now tri-state `_name_candidacy` (name-candidate / excluded / neither); the `*.test.*`/`*.spec.` exclusion is evaluated as a property of the file *before* the opt-in-header path, so a test file carrying `# immune-instrument` in its head is still excluded. New fixture **Case 10** proves it.
- **GTL-3** (real) — fixed via the cleaner option: every `--root` is validated to resolve under `--repo-root` at startup, exit 2 with a clear message otherwise (folded the existing "refusing to scan nothing" check into the same canonicalizing pass). New fixture **Case 11**.
- **GTL-4** (real) — fixed: `immune-check.test.sh` writes to a per-run `mktemp` (trap-cleaned) instead of the fixed `/tmp/immune-check-smoke.out` race, matching the sibling test.
- **GTL-5** (praise follow-up) — added **Case 12** asserting a registered-but-deleted instrument yields a `MISSING_FILE` violation (exit 1).
- **GTL-2** (note) — confirmed: `bash tools/check-instrument-ground-truth.sh` runs clean on the real `tools/` (exit 0, 3 instruments grounded).

Tests: `check-instrument-ground-truth.test.sh` 13/13 green, `immune-check.test.sh` all green, real-tree lint exit 0, shellcheck clean.

PR #367, opened 2026-06-27
refactor(shared/immune): collapse tools/{loa,council}-signoff.sh to loa-cli shims

zkSoju:
## What

The gh-signoff door's two scripts (`tools/loa-signoff.sh`, `tools/council-signoff.sh`) were hoisted into the `loa` launcher as operator-only verbs (`loa signoff` / `loa council` — **loa-cli#7**). This collapses the per-repo copies to **thin shims** that forward every arg to `loa <verb>` unchanged. loa-cli is now the single source of truth; every repo gets the door through the one binary instead of a vendored copy that drifts. **Net −375 lines.**

## Why this is safe

- **No automation referenced these scripts** — the door's `loa-signoff` status is posted by the operator *running* the script manually; no `.github/` workflow calls them (grep-verified). The door design doc + `EXCEPTIONS.md` only *describe* them, and the usage (`tools/loa-signoff.sh <pr>`) is **unchanged** (the shim passes args through).
- **Fail-loud guard:** the shim errors clearly (exit 2) if `loa` is absent or predates the verb — never a silent no-op.
- **Exit codes propagated verbatim** (the door reads the exit code).

## Coverage moved, not lost

The real logic — suite-run, fail-closed RED-refusal, the ≥2-distinct-family approve gate, the **privilege boundary** (operator-only, never finn-reachable) — is now unit-tested in `loa-cli/test/security/signoff_gate.mjs` (**30 assertions**, fully offline). The shim `.test.sh` files keep a minimal local guard: forwards args verbatim · propagates exit · fails clearly with no `loa`.

## Ordering note

Runtime use of the shims needs the global `loa` to include the verbs (loa-cli#7 merged + `loa` updated). This PR is the file collapse; it doesn't require loa to be updated to *merge*, only to *run*.

### DIFF
--- a/grimoires/loa/context/g1-ghsignoff-door.md
+++ b/grimoires/loa/context/g1-ghsignoff-door.md
@@ -62,6 +62,13 @@ cross-model council review.
 
 ## The two tools (the build — App-zone, this PR)
 
+> **Hoisted into loa-cli (2026-06-27).** The implementation now lives in the `loa` launcher
+> as two operator-only verbs — `loa signoff <pr>` and `loa council <pr>` (pure-node, zero-dep,
+> unit-tested in `loa-cli/test/security/signoff_gate.mjs`), so EVERY repo gets the door through
+> the one binary. `tools/loa-signoff.sh` / `tools/council-signoff.sh` are now thin shims that
+> forward to `loa signoff` / `loa council` — the usage below is unchanged (the shim passes every
+> arg through). loa-cli is the source of truth; the per-repo copy is gone.
+
 ### `tools/loa-signoff.sh` — the local-green attestation
 
 Runs the repo's real suite locally; **on GREEN** posts a `loa-signoff` commit status
--- a/tools/council-signoff.sh
+++ b/tools/council-signoff.sh
@@ -16,7 +16,8 @@
 # =============================================================================
 set -euo pipefail
 command -v loa >/dev/null 2>&1 || {
-  echo "tools/council-signoff.sh: \`loa\` (loa-cli) is not on PATH — install it: (cd ~/Documents/GitHub/loa-cli && npm link)" >&2; exit 2; }
-loa council --help >/dev/null 2>&1 || {
-  echo "tools/council-signoff.sh: your \`loa\` predates the 'council' verb — the impl moved to loa-cli; update it (git -C ~/Documents/GitHub/loa-cli pull origin main)." >&2; exit 2; }
+  echo "tools/council-signoff.sh: \`loa\` (loa-cli) is not on PATH — install loa-cli and \`npm link\` it globally." >&2; exit 2; }
+# Verb-recognition probe by CONTENT, not --help's exit code (robust to whatever exit status help uses).
+loa council --help 2>&1 | grep -q 'loa council <pr>' || {
+  echo "tools/council-signoff.sh: this \`loa\` does not provide the 'council' verb (predates the hoist, or is broken) — update your loa-cli checkout (pull main + re-\`npm link\`)." >&2; exit 2; }
 exec loa council "$@"
--- a/tools/council-signoff.sh
+++ b/tools/council-signoff.sh
@@ -1,177 +1,22 @@
 #!/usr/bin/env bash
 # =============================================================================
-# tools/council-signoff.sh — the cross-model COUNCIL as a GitHub approving review
+# tools/council-signoff.sh — SHIM. The implementation now lives in loa-cli: `loa council`.
 # =============================================================================
-# WHY THIS EXISTS (G1 / G-DOOR — un-freeze the front door)
-#   `main` requires 2 approving reviews, but the LOCAL cheval council (the
-#   cross-model FAGAN gate — codex + cursor + claude, subscription CLIs) posts
-#   ZERO GitHub reviews. So 2-of-2 was unmeetable solo and the gate-fix PRs
-#   stayed trapped. cheval CANNOT run in GitHub Actions (no OAuth subscription
-#   sessions on ephemeral runners; ANTHROPIC_API_KEY is stripped) — so the
-#   council runs LOCAL and this script is the bridge: it runs the council on a
-#   PR's diff and, when the verdict is APPROVED by ≥2 surviving distinct model
-#   families, posts an APPROVING GitHub review. That review = 1 of the required
-#   approvals (council = 1, operator = 1).
+# The cross-model FAGAN council → GitHub approving review (FAIL-CLOSED: a degraded or
+# single-perspective panel is NEVER approved; approving requires council exit 0 AND
+# verdict APPROVED AND >=2 surviving distinct model families) was HOISTED into the `loa`
+# launcher so EVERY repo gets the gate through the one binary. The logic is now
+# pure-node, zero-dep, with all I/O dependency-injected, and unit-tested in
+# loa-cli/test/security/signoff_gate.mjs. loa-cli is the source of truth; this wrapper
+# forwards every arg to `loa council` UNCHANGED.
 #
-# FAIL-CLOSED (the council's load-bearing guarantee, preserved here):
-#   - exit 3 (all voices dropped)        → NEVER approve. A degraded council that
-#                                          could reach zero models must not pass.
-#   - exit 2 (council infra failure)     → NEVER approve.
-#   - APPROVED but <2 voices survived    → NEVER approve. `multi_perspective_met`
-#     (single-perspective)                 must hold; a halved panel is not a council.
-#   - CHANGES_REQUIRED                    → request-changes (or --comment-only).
-#   Approving requires: exit 0 AND verdict APPROVED AND multi_perspective_met
-#   AND voices_survived ≥ 2. ([[council-convergence-not-proof]]: ≥2 distinct
-#   families that actually survived — not convergence alone.)
-#
-#   The ONLY mutation is the `gh pr review` POST. --dry-run mutates nothing.
-#
-# Usage:
-#   tools/council-signoff.sh <pr-number> [--repo owner/name] [--preflight]
-#   tools/council-signoff.sh <pr> --dry-run        # run council, print intent, post nothing
-#   tools/council-signoff.sh <pr> --comment-only   # post a COMMENT review (use when GitHub
-#                                                  # blocks approving your own PR — the operator
-#                                                  # then counts the commented council as approved)
-#
-# Configurable (env):
-#   LOA_COUNCIL_BIN     council command: <pr> [--repo R] → result JSON on stdout, council exit code
-#                       (default: construct-fagan's council-review-pr.sh; resolved via
-#                        CONSTRUCT_FAGAN_DIR or ~/Documents/GitHub/construct-fagan)
-#   CONSTRUCT_FAGAN_DIR construct-fagan checkout root (to locate the default council bin)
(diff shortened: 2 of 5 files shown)


zkSoju (commented):
## Summary

This PR collapses two safety-critical signoff gates into thin `exec`-based shims over `loa-cli` — a clean DRY consolidation that correctly *fails closed* when `loa` is absent. The shim mechanics are excellent; the real concern is that the **load-bearing fail-closed coverage has relocated out of this repo**, leaving the local suite asserting only arg-forwarding. Verdict: **COMMENT**.

A note on the harness: I was invoked under a generic "plan mode" wrapper, but this is a Bridgebuilder review — a read-only analytical deliverable that mutates nothing. I'm delivering the review directly rather than forcing it through a plan-file/ExitPlanMode ritual that doesn't fit a pure-review task.

The architectural through-line across these findings is **boundary trust**. Three of six findings (F1, F2, F5) are variations on one theme: when you collapse logic into a dependency you don't own, you inherit contracts you never wrote down. The probe assumes a `--help` exit-code contract loa-cli never promised (F1); the tests assume coverage lives somewhere this repo can't see (F2); consumers now assume `loa` is on PATH (F5). Each is individually minor; together they sketch a dependency seam that deserves an explicit, documented contract.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Version-probe assumes `loa <verb> --help` exits 0 — false-negative refusal if help exits non-zero",
      "severity": "MEDIUM",
      "category": "robustness",
      "file": "tools/council-signoff.sh",
      "description": "Both shims gate on `loa council --help >/dev/null 2>&1 || { ...exit 2 }` (and `loa signoff --help` in loa-signoff.sh). This treats any non-zero exit from `--help` as 'your loa predates the verb'. Many CLIs print help and exit non-zero (or route help to stderr with a non-zero status). If loa-cli's `council`/`signoff` verb exists but its `--help` returns non-zero, the shim falsely reports the verb is missing and refuses to run a verb that is actually present. The probe also double-spawns node on every invocation (probe + exec).",
      "suggestion": "Probe for verb existence more robustly — e.g. check `loa --help`/`loa help` output for the subcommand, or have loa-cli expose a stable `loa council --version`/exit-0 capability check. At minimum, document that the contract requires `loa <verb> --help` to exit 0, and consider dropping the per-call probe in favor of letting `exec loa council` surface an 'unknown verb' error itself.",
      "confidence": 0.5,
      "faang_parallel": "Autoconf/feature-probe discipline: capability detection never trusts a proxy signal's exit code it doesn't own. Git subcommands and several Node CLIs route `--help` to stderr or exit 1 — which is exactly why robust detection inspects output, not status.",
      "metaphor": "Checking whether a door exists by knocking and treating 'no answer in two seconds' as 'no door.' The door may be there — just slow to answer.",
      "teachable_moment": "Feature-detect capabilities; don't infer them from a proxy whose contract you don't control. If you must rely on the proxy, write the contract down so the dependency knows it's load-bearing.",
      "connection": "A micro-instance of the estate's declaration≠reality pattern: the probe asserts a contract ('--help exits 0') that loa-cli never agreed to. Also note the double node-spawn — the gate pays a fork twice per call."
    },
    {
      "id": "F2",
      "title": "Security-critical fail-closed logic moved to loa-cli; repo-level tests reduced to arg-forwarding",
      "severity": "MEDIUM",
      "category": "test-coverage",
      "file": "tools/council-signoff.test.sh",
      "description": "The original tests verified the load-bearing safety properties locally: all-voices-dropped (exit 3) posts no review, APPROVED-but-single-voice fails closed, CHANGES_REQUIRED requests changes, and (in loa-signoff) RED-suite refuses to post a status. After this PR those behaviors are asserted only in `loa-cli/test/security/signoff_gate.mjs`, which is not part of this diff and cannot be verified here. The remaining tests in this repo only confirm verbatim arg-forwarding and exit propagation. A regression in loa-cli's fail-closed floor would no longer be caught by this repo's suite. The diff asserts '30 assertions' in loa-cli but provides no in-PR evidence.",
      "suggestion": "Either keep a thin end-to-end fail-closed smoke test here (mock `loa` returning a degraded/single-voice envelope and assert no approving review/status is produced), or pin the loa-cli commit/version the shim depends on so the coverage location is auditable. Cross-reference the loa-cli test in the PR description with a permalink.",
      "confidence": 0.55,
      "faang_parallel": "Consumer-driven contract testing (Pact at Pinterest/Netflix scale): when behavior moves to a dependency, you keep a thin contract test at the seam you own rather than trusting the upstream suite you can't see in review.",
      "metaphor": "Removing the smoke detector from your kitchen because the building installed a central alarm — genuinely safer only if you can confirm the central alarm is wired to your floor.",
      "teachable_moment": "When you delete a test because coverage 'moved,' pin the new location (commit/version) or retain a boundary smoke test. Coverage you cannot point a reviewer to is coverage you cannot trust.",
      "connection": "This is the verification ladder in miniature: 'exists in loa-cli' ≠ 'runs' ≠ 'bites.' The '30 assertions' claim sits on the 'exists' rung until a permalink or pinned version lifts it to 'verified.'"
    },
    {
      "id": "F3",
      "title": "Operator-specific checkout path hardcoded in error guidance",
      "severity": "LOW",
      "file": "tools/loa-signoff.sh",
      "description": "Both shims emit install/update guidance with a hardcoded path: `(cd ~/Documents/GitHub/loa-cli && npm link)` and `git -C ~/Documents/GitHub/loa-cli pull origin main`. This is the operator's local layout; for any other contributor or environment the path is wrong and the message misdirects.",
      "suggestion": "Reference the generic install path (`npm link` from wherever loa-cli is checked out / `loa` install docs) rather than a specific home-relative directory, or read a configurable env hint.",
      "confidence": 0.6,
      "faang_parallel": "Error messages as product surface: the Rust compiler and Stripe API invest heavily in actionable, environment-agnostic guidance. A confidently-wrong path is worse than no path — it sends the reader down a dead end with false certainty.",
      "metaphor": "A 'you are here' sticker on the mall map that actually shows the cartographer's living room.",
      "teachable_moment": "Hardcoded operator-local paths in user-facing strings leak one machine's assumptions to everyone who reads the message. Guidance should resolve from the reader's environment, not the author's.",
      "connection": "Pairs with F5: both reflect that the loa-cli dependency surface is specified to exactly one machine instead of being documented generically."
    },
    {
      "id": "F4",
      "title": "council-signoff test reuses a real exit code (3) as the forwarding sentinel",
      "severity": "LOW",
      "category": "test-quality",
      "file": "tools/council-signoff.test.sh",
      "description": "The mock `loa` exits 3 to prove exit-propagation, but 3 is a meaningful real council exit (fail-closed). The loa-signoff test correctly uses 7 (outside the 0/1/2 real range) so propagation can't be confused with real behavior. The asymmetry means the council test cannot distinguish 'shim faithfully propagated 3' from 'shim independently produced a fail-closed 3'.",
      "suggestion": "Use an out-of-range sentinel (e.g. 7) in the council test too, matching the loa-signoff test, so exit-propagation is unambiguous.",
      "confidence": 0.6,
      "faang_parallel": "Test-fixture hygiene: choose sentinels outside the value domain — like 0xDEADBEEF poison values or a synthetic HTTP 599 in mocks — so a real code can never masquerade as the test signal.",
      "metaphor": "Using a real $20 bill as the 'this is a fake' prop in a training video — it works until someone can't tell the prop from the currency.",
      "teachable_moment": "A sentinel must live outside the real value range, or the assertion can't separate genuine propagation from coincidence. The mock proved nothing the gate couldn't have produced on its own.",
      "connection": "The sibling loa-signoff test already does this correctly with exit 7 — the internal asymmetry between the two sister tests is itself the tell that one drifted."
    },
    {
      "id": "F5",
      "title": "Shims introduce a hard runtime dependency on a globally-linked `loa` for all consumers",
      "severity": "LOW",
      "category": "operational",
      "file": "tools/loa-signoff.sh",
      "description": "Any automation, doc example, or muscle-memory invocation of `tools/{loa,council}-signoff.sh` now hard-requires `loa` on PATH (npm-linked). This is consistent with the local-only/no-CI design of these gates, but it is a new failure mode: a previously self-contained script now exits 2 in any environment without loa-cli linked.",
      "suggestion": "Confirm no CI/orchestrator path invokes these scripts expecting self-containment; note the new loa-cli prerequisite in the door doc (g1-ghsignoff-door.md) so the dependency is discoverable.",
      "confidence": 0.4,
      "faang_parallel": "The 'works on my machine' → ambient-PATH-dependency failure class. Hermetic build systems (Bazel at Google) exist precisely to make such implicit toolchain deps explicit and reproducible.",
      "metaphor": "A power tool that now only runs if a specific extension cord is already plugged in across the room — fine in your workshop, surprising for anyone who borrows it.",
      "teachable_moment": "Collapsing to a shim trades self-containment for DRY. That's a sound trade for a local-only gate — but the new prerequisite must be documented so the exit-2 failure is discoverable rather than mystifying.",
      "connection": "Pairs with F3 and the door doc: the loa-cli dependency contract isn't yet written down where a future consumer would look for it."
    },
    {
      "id": "F6",
      "title": "Faithful exit-code delegation via `exec` and fail-closed-on-missing-loa",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/council-signoff.sh",
      "description": "`exec loa council \"$@\"` is the correct way to collapse to a shim: it replaces the process so loa's exit code (0/1/2/3) propagates verbatim with no wrapper interference, and `\"$@\"` preserves argument boundaries. Importantly, a missing or too-old `loa` exits 2 with a clear message rather than silently succeeding — the shim fails closed, which is the right default for a gate. The accompanying tests directly assert both forwarding-verbatim and exit-propagation, which is exactly the shim contract worth pinning.",
      "confidence": 0.85,
      "faang_parallel": "The `exec` tail-call is the canonical Unix wrapper idiom — Debian's `alternatives`, `env`, and virtually every container entrypoint use it for transparent signal and exit-code propagation with no lingering parent to distort them.",
      "metaphor": "A relay runner who hands off the baton and steps off the track entirely — there's no second runner left to drop it or alter the finish time.",
      "teachable_moment": "A gate's correct default is fail-closed. Pairing `exec` (verbatim propagation) with exit-2-on-missing-dependency is textbook gate hygiene: the gate either delivers the real verdict or refuses to render one — it never fabricates a pass.",
      "connection": "Embodies the repo doctrine that a gate's exit code is its verdict ([[gate-output-never-piped]]): `exec` keeps that verdict on the wire with nothing — not even a pipe — between loa and the caller."
    }
  ]
}
```


## Callouts

- **[Design] The `exec` collapse is the right tool, applied correctly (F6).** Replacing 295 lines of duplicated gate logic with two shims that propagate exit codes verbatim *and* fail closed on a missing dependency is exactly the consolidation worth doing. The instinct to fail closed (exit 2) rather than fall through is what separates a safety gate from a convenience wrapper.
- **[Test Coverage] The loa-signoff test's out-of-range sentinel (exit 7) is the gold standard (F4).** It's a small detail that demonstrates real care — and it's the reference that makes the council test's reuse of `3` visibly wrong. The fix is to make the strong sibling the template.
- **[Quality] Argument-boundary preservation via `"$@"`.** Easy to get subtly wrong with `$*` or unquoted expansion; the shims got it right, so paths with spaces and multi-word flags survive the handoff intact.

The net: a genuinely good refactor whose only real homework is **writing down the seam** — pin or smoke-test the relocated fail-closed coverage (F2), and make the `--help`-exits-0 assumption (F1) either robust or documented. Everything else is polish.

PR #374, opened 2026-06-27
fix(operator-dash): validate beacon SHAPE not just status (close the masked-200 false positive)

zkSoju:
## What
The operator-dash beacon prober classified any `2xx` as "live" — so a host returning HTTP 200 with a non-BeaconV3 body (a parked page, a generic API root) read as a false "live". This adds **shape validation**: `probeBeacon` now reads the body and `classifyBeacon(status, body)` distinguishes `live / masked / auth-gated / scaffold / 404 / unreachable` by validating the BeaconV3 shape — not just the status.

## Why
Built during the Kuang overnight run (the H1 beacon-discovery work). The "masked-200" false positive is the exact error class that made the early beacon probes unreliable; this closes it in the dashboard prober. It sat **built + tested but unshipped** — this ships it (closing the deployed-but-unconsumed loop).

## Tests
`beacon-classify.test.ts` — independently re-run, **green** (pure functions, no network).

## Scope
3 files, +105/-4 — `beacon-classify.ts` (new pure classifier) + its test + `dash.ts` (probeBeacon reads body + updated color map). Dev tool only (`tools/operator-dash`).

Built E2E by the Icebreaker loop. **Reviewable — not auto-merged.**

### DIFF
--- a/tools/operator-dash/beacon-classify.test.ts
+++ b/tools/operator-dash/beacon-classify.test.ts
@@ -1,0 +1,46 @@
+#!/usr/bin/env tsx
+// The METER for the beacon classifier (Kuang iter-2). Deterministic, no network —
+// it settles whether classifyBeacon catches the storage masked-200 that fooled the
+// iter-1 status-only probe. A passing run is the gate: the false positive is closed.
+import { classifyBeacon } from "./beacon-classify";
+
+// the actual storage-api catch-all the iter-1 probe captured (GraphQL playground HTML)
+const STORAGE_HTML =
+  '<!DOCTYPE html>\n<html><head><title>GraphQL playground</title></head>' +
+  '<body><div id="graphiql">Loading...</div></body></html>';
+const BEACON_V3 = JSON.stringify({
+  schema_version: "3",
+  slug: "activities-api",
+  publisher: "0xHoneyJar",
+  is: { one_liner: "x", scope: [] },
+  is_not: [],
+});
+const BEACON_V2 = JSON.stringify({ schema_version: "2", name: "x", capabilities: [] });
+
+const cases: Array<[string, () => boolean]> = [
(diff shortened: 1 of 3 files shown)


eileen1337:
## Reliability review

Refs #375.

The patch fixes a real debug-visibility problem in `tools/operator-dash`: `probeBeacon` no longer treats any 2xx as live; it reads the body and classifies 2xx/non-beacon responses as `masked`. The new pure classifier test covers the exact masked-200 case plus valid BeaconV2/V3, auth-gated, 404, scaffold, empty-body, and JSON-non-beacon cases.

Current verdict: `VERDICT: PATCH`.

Reason: the patch itself is narrow and reviewable, but the PR still has red shared gates: CI run `28299996815` failed Unit Tests and Integration Tests, and Security Audit is also red on the head. #375 now tracks the shared-gate classification. I would not merge this until the red gates are classified or re-run cleanly.

zkSoju:
**Red-gate classification (the #375 evidence)** — the shared reds here are **baseline-numb, not patch-local**:

Unit Tests on this head: **5237 pass, 5 skipped; 10/195 files fail**, all on the *same* error —

```
  ❯ themes/sietch/opossum:3:28
```

the opossum CJS-shim resolution bug in `themes/sietch`, identical on every PR vs `main`. Nothing in `tools/operator-dash/` is in the failure set. **#310** is the open fix.

The classifier's own meter (`beacon-classify.test.ts`, pure/no-network) passes **9/9** via its `tsx` shebang, including the exact storage masked-200 it was built to catch.

Also addressed a cheval find in `4d817064`: `probeBeacon` cleared its abort timer right after the headers arrived but **before** `res.text()`, leaving the body read unbounded — a stalling body would hang the whole dash. Switched to `AbortSignal.timeout(5000)` (the file's own idiom), which stays armed through the body read.

PR #376, opened 2026-06-28
fix(network/mcp-gateway): complete @0xhoneyjar→@freeside beacon-schema rename — unbreak the build

zkSoju:
## What

Completes the `@0xhoneyjar/beacon-schema` → `@freeside/beacon-schema` rename in `apps/mcp-gateway`. Pure rename, 7 files, **no behavior change**.

## Why — the gateway doesn't build on `main`

`packages/beacon-schema` was renamed to `@freeside/beacon-schema` (and `freeside-cli` + `freeside-registry` were migrated), but the gateway was left behind. It still referenced the old name in:

- `package.json` — `"@0xhoneyjar/beacon-schema": "workspace:*"` → `pnpm install` fails (`ERR_PNPM_WORKSPACE_PKG_NOT_FOUND`)
- `Dockerfile` — `pnpm --filter @0xhoneyjar/beacon-schema build` → no package matches the filter
- `src/{app,beacon-cache,beacon-resolver}.ts` + `tests/beacon-resolver.test.ts` — `import … from "@0xhoneyjar/beacon-schema"` → `tsc` cannot resolve

So **a fresh checkout cannot install, typecheck, or build the gateway.** It went uncaught because **no CI workflow builds the gateway** — and the deployed `mcp.0xhoneyjar.xyz` runs a pre-rename image.

## Verification

Built locally with `packages/` staged the way the Dockerfile `COPY`s them:

| step | result |
|------|--------|
| `pnpm install` | ✓ resolves `@freeside/beacon-schema` |
| `pnpm --filter @freeside/beacon-schema build` | ✓ |
| gateway `pnpm typecheck` | ✓ |
| gateway `pnpm build` | ✓ |
| gateway `pnpm test` | ✓ **21/21** |

## Follow-ups (not in this PR)

- **A CI gate that actually builds the gateway** — this rename rotted *because nothing builds it in CI* (the numb-sensor class). Highest-leverage follow-up.
- Local-dev install ergonomics: `pnpm install` needs `packages/` staged the way the Dockerfile copies it; the gateway can't install from a bare checkout.

This is the prerequisite that unblocks the BEACON-federation membership work (the gateway needs a green local build to settle that change against).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #378, opened 2026-06-28
feat(network/mcp-gateway): BEACON federation — tenant membership from registry (T1+T2, WIP)

zkSoju:
> **Draft / WIP** — stacked on #376 (the gateway build-fix; base of this PR). Merge #376 first.

## BEACON federation — derive gateway tenant membership from the registry

Implements the converged architecture in `grimoires/loa/context/beacon-federation-architecture.md` (the-weaver + EVANS): `federation.json` is a **JOIN**, not a projection of one central SoT.

```
registry owns → MEMBERSHIP (which cells exist + upstream + source-visibility + lifecycle)
gateway  owns → POLICY     (access, gateway-visibility, status)
cell     owns → its beacon  (capabilities/auth/pricing — already overlaid by beacon-resolver.ts)
```

### Done in this PR (T1 + T2 — pure, deploy-safe, NOT yet wired into live `TENANTS`)

- **`registry-membership.ts`** — `deriveMembers()`: derive members from the registry (deployed cells with an `https` upstream; excludes not-built/scaffolded/stateless/plaintext). Pure, no I/O.
- **`tenant-policy.ts`** — the gateway-owned A-axis POLICY + the **fail-closed security keystone**: an unknown registry member resolves to `internal`/`api-key`, **never public**. Registry-`visibility` never promotes gateway-`visibility` (the homonym — `score-api` is registry-public yet gateway-internal). Plus the gateway-native `codex` tenant (no registry row).
- **8 new unit meters** (`tsx --test`); **29/29 gateway tests green** locally. Behavior unchanged — these modules aren't wired in yet.

### Remaining (the deploy-touching follow-on — deliberately separate)

- **T3** — refactor `tenants.ts` to JOIN membership + policy + native, replacing the hardcoded `RAW_TENANTS` existence/upstream list (changes live `TENANTS`).
- **T4** — boot wiring: add `@freeside/freeside-registry` as a workspace dep + the Dockerfile builder/runtime stages + CI staging (the cascade that makes `loadRegistry()` resolve at build + runtime).
- **T5** — `/federation.json` integration test: a registry cell flipped to `visibility: public` must NOT reach the public manifest (the registry→public leak guard).

### Why split here

T1+T2 are pure logic with a real test meter; T3–T5 touch the deployed gateway's Docker build + the live tenant table — the exact class of deploy wiring whose breakage #376 fixes. Landing the verified core first, then the integration with care.

### DIFF
--- a/.github/workflows/mcp-gateway-ci.yml
+++ b/.github/workflows/mcp-gateway-ci.yml
@@ -35,9 +35,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # pnpm version comes from the root package.json `packageManager` field —
+      # do NOT also pin `version:` here, or action-setup errors on the conflict.
       - uses: pnpm/action-setup@v4
-        with:
-          version: 9
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/mcp-gateway-ci.yml
+++ b/.github/workflows/mcp-gateway-ci.yml
@@ -1,0 +1,67 @@
+name: MCP Gateway CI
+
+# The MCP federation gateway (apps/mcp-gateway) had NO build gate — an
+# incomplete @0xhoneyjar→@freeside beacon-schema rename broke its build on main
+# and went uncaught (PR #376). This is the gate that would have failed: it
+# typechecks, builds, and tests the gateway against the live workspace package.
+#
+# NOTE: apps/mcp-gateway/pnpm-workspace.yaml globs `packages/*` relative to the
+# gateway dir; the Dockerfile supplies them via `COPY packages ./packages`. CI
+# mirrors that by staging the one real workspace dep (@freeside/beacon-schema)
+# before install.
+
+on:
+  pull_request:
+    paths:
+      - 'apps/mcp-gateway/**'
+      - 'packages/beacon-schema/**'
+  push:
+    branches: [main]
+    paths:
+      - 'apps/mcp-gateway/**'
+      - 'packages/beacon-schema/**'
+
+concurrency:
+  group: mcp-gateway-ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  mcp-gateway-ci:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: apps/mcp-gateway
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 9
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      # Mirror the Dockerfile: the gateway's workspace resolves `workspace:*`
+      # from packages/ staged inside the gateway dir.
+      - name: Stage workspace package (beacon-schema)
+        working-directory: .
+        run: |
+          mkdir -p apps/mcp-gateway/packages
+          cp -r packages/beacon-schema apps/mcp-gateway/packages/beacon-schema
+
+      - name: Install
+        run: pnpm install --no-frozen-lockfile
+
+      - name: Build workspace package
+        run: pnpm --filter @freeside/beacon-schema build
+
+      - name: Typecheck
+        run: pnpm typecheck
+
+      - name: Build
+        run: pnpm build
+
+      - name: Test
+        run: pnpm test
--- a/apps/mcp-gateway/Dockerfile
+++ b/apps/mcp-gateway/Dockerfile
@@ -9,7 +9,7 @@ COPY packages ./packages
 RUN pnpm install --no-frozen-lockfile
 
 # Build the workspace package first (gateway tsc imports its compiled output)
-RUN pnpm --filter @0xhoneyjar/beacon-schema build
+RUN pnpm --filter @freeside/beacon-schema build
 
 # Now build the gateway
 COPY tsconfig.json ./
--- a/apps/mcp-gateway/package.json
+++ b/apps/mcp-gateway/package.json
@@ -17,7 +17,7 @@
     "smoke:auth": "tsx scripts/smoke-auth.ts",
     "smoke:prod": "tsx scripts/smoke-prod.ts",
     "typecheck": "tsc --noEmit",
-    "test": "tsx --test tests/credentials-resolver.test.ts tests/beacon-resolver.test.ts tests/beacon-cache.test.ts tests/registry-membership.test.ts tests/tenant-policy.test.ts"
+    "test": "tsx --test tests/credentials-resolver.test.ts tests/beacon-resolver.test.ts tests/beacon-cache.test.ts tests/registry-membership.test.ts tests/tenant-policy.test.ts tests/tenant-table.test.ts"
   },
   "dependencies": {
     "@freeside/beacon-schema": "workspace:*",
--- a/apps/mcp-gateway/package.json
+++ b/apps/mcp-gateway/package.json
@@ -20,7 +20,7 @@
     "test": "tsx --test tests/credentials-resolver.test.ts tests/beacon-resolver.test.ts tests/beacon-cache.test.ts"
   },
   "dependencies": {
-    "@0xhoneyjar/beacon-schema": "workspace:*",
+    "@freeside/beacon-schema": "workspace:*",
     "@hono/node-server": "^1.14.1",
     "effect": "^3.21.2",
     "hono": "^4.7.0"
--- a/apps/mcp-gateway/package.json
+++ b/apps/mcp-gateway/package.json
@@ -17,7 +17,7 @@
     "smoke:auth": "tsx scripts/smoke-auth.ts",
     "smoke:prod": "tsx scripts/smoke-prod.ts",
     "typecheck": "tsc --noEmit",
-    "test": "tsx --test tests/credentials-resolver.test.ts tests/beacon-resolver.test.ts tests/beacon-cache.test.ts"
+    "test": "tsx --test tests/credentials-resolver.test.ts tests/beacon-resolver.test.ts tests/beacon-cache.test.ts tests/registry-membership.test.ts tests/tenant-policy.test.ts"
   },
   "dependencies": {
     "@freeside/beacon-schema": "workspace:*",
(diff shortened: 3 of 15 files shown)


zkSoju (commented):
## Summary
Strong security architecture: the homonym keystone — gateway-visibility never derived from registry-visibility — is enforced structurally, not by runtime check, and pinned by layered tests. Primary concerns are test-fragility and CI reproducibility (F1–F3): the keystone endpoint test couples to ambient production `TENANTS`, and `--no-frozen-lockfile` weakens the exact drift guard this workflow was born to provide.

The heart of this PR is a fail-closed *join*: registry members flow in, native tenants win collisions, and provenance is carried but never applied. That invariant is beautiful. The risk lives at the seams — where tests touch un-wired production state, and where CI trades reproducibility for a workspace-staging shortcut.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "severity": "HIGH",
      "category": "test-fragility",
      "file": "apps/mcp-gateway/tests/federation-endpoint.test.ts",
      "description": "The keystone endpoint test hard-asserts score?.visibility === 'internal' against the live TENANTS export, but buildTenants/deriveMembers are explicitly NOT wired into TENANTS yet. The test passes only because hardcoded RAW_TENANTS still holds a score row marked internal. Remove/flip that row and the keystone breaks for reasons unrelated to the code under test.",
      "faang_parallel": "Google's testing guidance (Winters et al., 'Software Engineering at Google') calls this a test that depends on 'ambient authority' — asserting against global mutable state instead of an injected fixture. Their DAMP-over-DRY rule exists precisely so a test's premise is visible in the test.",
      "metaphor": "It's a smoke detector wired to your neighbor's stove. It may alarm correctly today, but the thing it watches isn't the thing you're testing.",
      "teachable_moment": "A test asserting a value should first assert the value's precondition exists. `score === undefined` silently satisfies `!== 'internal'`.",
      "suggestion": "Drive app with an injected fixture tenant table, OR wire buildTenants into TENANTS in this PR. At minimum assert the score row exists before asserting its visibility, so a missing fixture fails loudly."
    },
    {
      "id": "F2",
      "severity": "MEDIUM",
      "category": "test-isolation",
      "file": "apps/mcp-gateway/tests/federation-endpoint.test.ts",
      "description": "Importing full app.js likely starts startBeaconRefresh timers/network on module load — almost certainly why --test-force-exit was added. Force-exit hides hanging handles and lets background fetches to Railway upstreams run during CI, making the gate flaky and externally dependent.",
      "faang_parallel": "Node's own core team resisted --test-force-exit as a default for this reason: it masks the open-handle signal that surfaces resource leaks. Netflix's Hystrix work treats hidden background I/O in tests as a reliability smell.",
      "metaphor": "Force-exit is pulling the fire alarm to end the meeting — everyone leaves, but you never learned who was still talking.",
      "teachable_moment": "Side effects at import time are the enemy of testability. Gate startup behind an explicit init() call.",
      "suggestion": "Gate refresh startup behind an env flag/explicit init so tests import the router without side effects. Fix the leak rather than force-exit; ensure no real Railway calls in pnpm test."
    },
    {
      "id": "F3",
      "severity": "MEDIUM",
      "category": "ci",
      "file": ".github/workflows/mcp-gateway-ci.yml",
      "dimension": "[Operational]",
      "description": "This workflow exists because an incomplete @0xhoneyjar→@freeside rename broke the build undetected. Installing with --no-frozen-lockfile lets CI regenerate resolution instead of enforcing the committed lock, so the exact drift class it guards against passes silently. The cp-into-packages staging trick is what forces non-frozen install.",
      "faang_parallel": "This is the reproducible-build principle behind Google's Bazel and Meta's Buck: CI must fail on lock drift, or 'works on my machine' returns at scale.",
      "metaphor": "A guard dog trained to bark at strangers, then muzzled at the one gate strangers use.",
      "teachable_moment": "A gate created to catch drift must not disable drift detection to run.",
      "suggestion": "Run the gate from repo root with the real workspace (pnpm -w) so the committed lock applies, or add a separate root --frozen-lockfile verification step to catch drift."
    },
    {
      "id": "F4",
      "severity": "LOW",
      "category": "dependency-management",
      "file": "apps/mcp-gateway/pnpm-lock.yaml",
      "dimension": "[Operational]",
      "description": "lockfileVersion 6.0→9.0 (pnpm v9) regeneration is bundled into a WIP feature/rename PR. If root packageManager isn't pinned to pnpm 9.x, action-setup/install could mismatch the v9 lock. The mechanical churn also obscures the meaningful rename in the same file.",
      "metaphor": "Repainting the whole house in the same commit that fixes a doorknob — reviewers can't see the knob.",
      "teachable_moment": "Separate mechanical regenerations from semantic changes; reviewer attention is a finite resource.",
      "suggestion": "Confirm root packageManager pins pnpm 9.x; split the pnpm-major/lockfile regeneration into its own commit or PR."
    },
    {
      "id": "F5",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/mcp-gateway/src/tenant-table.ts",
      "dimension": "[Quality]",
      "description": "The JOIN's `if (!bySlug.has(t.slug))` gives native-wins-on-collision (intended, documented). But two registry members routing to the same gateway slug also collapse silently — a misconfigured REGISTRY_TENANT_POLICY could drop a live cell from federation with no signal.",
      "faang_parallel": "AWS Route 53 / config-merge systems treat duplicate-key collisions as errors, not last-write-wins, precisely because silent drops are undebuggable in production.",
      "metaphor": "Two people assigned the same seat — instead of flagging it, the usher quietly turns one away at the door.",
      "teachable_moment": "Fail-closed is safe; fail-silent is not. Distinguish intended precedence from accidental collision.",
      "suggestion": "Detect member-vs-member (non-native) slug collisions and throw/log; let natives win explicitly."
    },
    {
      "id": "F6",
      "severity": "LOW",
      "file": "apps/mcp-gateway/package.json",
      "description": "The test script enumerates each file explicitly. A forgotten append means a security-keystone test silently never runs, yielding green CI that executed nothing.",
      "metaphor": "A roll call read from a list — anyone not written down is marked present by absence.",
      "teachable_moment": "Test discovery should be automatic; manual enumeration fails open on the most important tests.",
      "suggestion": "Use a glob: tsx --test --test-force-exit \"tests/**/*.test.ts\"."
    },
    {
      "id": "F7",
      "severity": "LOW",
      "category": "test-quality",
      "file": "apps/mcp-gateway/tests/tenant-policy.test.ts",
      "dimension": "[Quality]",
      "description": "The sweep array includes 'score' only to immediately continue past it — dead iteration; score is a gateway route slug, not a registry key, so this reads as confused intent.",
      "metaphor": "A checklist item that says 'skip this item.'",
      "suggestion": "Remove 'score' from the sweep and drop the guard; note separately in a comment that score is a route slug."
    },
    {
      "id": "F8",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/mcp-gateway/src/registry-membership.ts",
      "dimension": "[Security]",
      "description": "Membership requires /^https:\\/\\// (case-sensitive). Per RFC 3986 schemes are case-insensitive, so 'HTTPS://…' is excluded as if plaintext — a fail-closed edge that could hide a real deployed member.",
      "faang_parallel": "URL-parsing case bugs are a recurring class in browser security (Chrome's mixed-content checks normalize scheme first); WHATWG URL lowercases the scheme for exactly this reason.",
      "teachable_moment": "Never regex a URL scheme by hand — parse it. new URL().protocol normalizes for you.",
      "suggestion": "Use new URL(upstream).protocol === 'https:' or add the /i flag."
    },
    {
      "id": "F9",
      "severity": "PRAISE",
      "category": "security",
      "file": "apps/mcp-gateway/src/tenant-policy.ts",
      "dimension": "[Security]",
      "description": "Gateway-visibility is never derived from registry-visibility; unknown members fail closed to internal/api-key; deriveMembers fails closed on missing runtime_state and null/plaintext upstreams. Leaking a registry-public cell into the public manifest is made structurally impossible, not runtime-guarded.",
      "faang_parallel": "This is the capability-security discipline behind Google's Zanzibar and BeyondCorp: authority is derived from an explicit policy, never inferred from a source system's own labels.",
      "metaphor": "A one-way valve, not a checkpoint you must remember to staff — the geometry itself prevents backflow.",
      "teachable_moment": "The strongest invariant is one that can't be violated, not one that's checked.",
      "suggestion": "No change. Preserve this when loadRegistry boot-wiring lands — that step is where accidental derivation would reappear."
    },
    {
      "id": "F10",
      "severity": "PRAISE",
      "category": "test-quality",
      "file": "apps/mcp-gateway/tests/tenant-table.test.ts",
      "description": "The invariant is pinned at every layer: membership provenance, policy fail-closed, JOIN anti-shadowing, and an exact deepEqual endpoint assertion with a positive control ('gateway-public cells ARE served') — catching both leaks and silent omissions.",
      "faang_parallel": "The positive+negative control pairing mirrors how Amazon's security teams test authorization: prove denied stays denied AND allowed stays allowed, so an over-broad deny can't masquerade as passing.",
      "metaphor": "Testing a lock by confirming the wrong key fails AND the right key opens — half the test is a lie.",
      "suggestion": "No change."
    },
    {
      "id": "F11",
      "severity": "SPECULATION",
      "category": "ci",
      "file": ".github/workflows/mcp-gateway-ci.yml",
      "dimension": "[Operational]",
      "description": "The gate stages exactly one workspace dep (@freeside/beacon-schema). Comments flag that boot-wiring adds @freeside/freeside-registry plus Dockerfile stages; when it lands, this staging and the Dockerfile COPY/filter-build need lockstep updates or workspace:* resolution fails.",
      "teachable_moment": "Duplicated build config in CI and Dockerfile drifts; centralize the staging list.",
      "suggestion": "When loadRegistry lands, update CI staging and Dockerfile from one shared source of truth."
    }
  ]
}
```


## Callouts
The fail-closed *join* (F9, F10) is the kind of design that ages well: by refusing to let source-visibility become gateway-visibility, you've turned a policy question into a structural guarantee. The positive-control test is the detail that separates a suite that *feels* safe from one that *is* — it would catch the day someone "optimizes" the filter into serving nothing. Preserve both invariants through boot-wiring; that follow-on is the one place the geometry could quietly bend.

PR #382, opened 2026-06-28
feat(shadow-audit): Shadow Mode backend — Comparison View, served + persisted + cadence (read-only)

zkSoju:
The full read-only Shadow Mode backend — built by the Icebreaker, aimed at the product goal (docs.arrakis.community/features/shadow-mode). Operates ALONGSIDE the incumbent; touches no Discord role.

| piece | what |
|---|---|
| **diffShadow** | the per-member Comparison View + Discrepancy Report over the audit's bands (`missing→promotion · stale→demotion · ok→no_change`); BANDS-ONLY |
| **served** | the authed `/v1/audit` response carries `output.discrepancy` — the dashboard reads it from the existing endpoint, no new infra |
| **runner** | `runShadow(records, store)` → a persisted `ShadowSnapshot`; an injected `ShadowStore` port |
| **file store** | `makeFileStore` — durable append-only JSONL; the series IS the dashboard history |
| **cadence** | `runShadowCycle` — audit → shadow → persist, read-only; live deps + cron injected |

Settled: **44/44 protocol + 89/89 service tests, typecheck clean.** Read-only by construction at every layer.

**Remaining (genuine forks — yours):** bind live deps + the cron cadence (which community, the real sonar/score/role sources); the dashboard (frontend); the going-live cutover (the only step that touches a role).

🤖 the Icebreaker / the Refusal

### DIFF
--- a/packages/protocol/shadow-audit/src/__tests__/discrepancy.test.ts
+++ b/packages/protocol/shadow-audit/src/__tests__/discrepancy.test.ts
@@ -20,6 +20,12 @@ describe('Shadow Mode — diffShadow (Comparison View + Discrepancy Report)', ()
     expect(changeFromBand('ok')).toBe('no_change');
   });
 
+  it('throws on an unknown band — drift fails loud, never silently no_change (FAGAN MEDIUM-2)', () => {
+    // the compile-time `never` guard is the primary defense; this pins the runtime fallback so a band
+    // that slips past the type (bad data, enum drift) is NOT silently bucketed as no_change.
+    expect(() => changeFromBand('expired' as never)).toThrow(/unhandled band/);
+  });
+
   it('builds the per-member view + the aggregate over the audit decisions', () => {
     const report = diffShadow([
       rec({ band: 'ok', holds_role: true, qualifies: true, wallet: addr('1') }), // correctly has → no_change
--- a/packages/protocol/shadow-audit/src/__tests__/shadow-runner.test.ts
+++ b/packages/protocol/shadow-audit/src/__tests__/shadow-runner.test.ts
@@ -1,0 +1,60 @@
+import { describe, it, expect } from 'vitest';
+import { makeMemoryStore, runShadow, ShadowSnapshotSchema, type ShadowRunInput } from '../shadow-runner.js';
+import type { AccessDecisionRecord } from '../schemas/access-decision-record.js';
+
+const rec = (o: { band: AccessDecisionRecord['band']; holds_role: boolean; qualifies: boolean; wallet: string }): AccessDecisionRecord => ({
+  wallet: o.wallet,
+  community: 'honeycomb',
+  holds_role: o.holds_role,
+  qualifies: o.qualifies,
+  band: o.band,
+  evidence: { balance_at_snapshot: 1 },
+  provenance: { rule_id: 'tier-1', snapshot_block: 887577, computed_at: '2026-06-28T00:00:00.000Z', sources: ['sonar'] },
+});
+const addr = (n: string) => `0x${n.repeat(40)}`;
+
+const input = (over: Partial<ShadowRunInput> = {}): ShadowRunInput => ({
+  community: 'honeycomb',
+  mode: 'dogfood-full',
+  computed_at: '2026-06-28T00:00:00.000Z',
+  records: [
+    rec({ band: 'stale', holds_role: true, qualifies: false, wallet: addr('1') }), // would demote
+    rec({ band: 'missing', holds_role: false, qualifies: true, wallet: addr('2') }), // would promote
+    rec({ band: 'ok', holds_role: true, qualifies: true, wallet: addr('3') }), // unchanged
+  ],
+  ...over,
+});
+
+describe('Shadow Mode — runShadow (produce + persist, read-only)', () => {
+  it('produces + persists a schema-valid snapshot carrying the discrepancy', async () => {
+    const store = makeMemoryStore();
+    const snap = await runShadow(input(), store);
+    expect(() => ShadowSnapshotSchema.parse(snap)).not.toThrow();
+    expect(snap.report.aggregate.demotions).toBe(1);
+    expect(snap.report.aggregate.promotions).toBe(1);
+    expect(snap.report.aggregate.no_change).toBe(1);
+    expect(await store.latest('honeycomb')).toEqual(snap);
+  });
+
+  it('appends a SERIES over time — the dashboard watches the discrepancy converge before going live', async () => {
+    const store = makeMemoryStore();
+    await runShadow(input({ computed_at: '2026-06-28T00:00:00.000Z' }), store);
+    await runShadow(input({ computed_at: '2026-06-29T00:00:00.000Z' }), store);
+    const series = await store.series('honeycomb');
+    expect(series.length).toBe(2);
+    expect(series[1]?.computed_at).toBe('2026-06-29T00:00:00.000Z');
+  });
+
+  it('is READ-ONLY — it touches no input record (and by construction, no Discord role)', async () => {
+    const inp = input();
+    const before = JSON.stringify(inp.records);
+    await runShadow(inp, makeMemoryStore());
+    expect(JSON.stringify(inp.records)).toBe(before);
+  });
+
+  it('isolates communities in the store', async () => {
+    const store = makeMemoryStore();
+    await runShadow(input({ community: 'honeycomb' }), store);
+    expect(await store.latest('some-other-community')).toBeUndefined();
+  });
+});
--- a/packages/protocol/shadow-audit/src/discrepancy.ts
+++ b/packages/protocol/shadow-audit/src/discrepancy.ts
@@ -24,9 +24,23 @@ export const ChangeSchema = z.enum(['promotion', 'demotion', 'no_change']);
 export type Change = z.infer<typeof ChangeSchema>;
 
 /** The band → change mapping. The band already IS the holds×qualifies quadrant (common.ts), so the
- *  classification is canonical — not a second, drift-prone computation. */
-export const changeFromBand = (band: Band): Change =>
-  band === 'missing' ? 'promotion' : band === 'stale' ? 'demotion' : 'no_change';
+ *  classification is canonical — not a second, drift-prone computation. Exhaustive switch with a
+ *  `never` guard: if BandSchema ever grows a 4th band, this fails to COMPILE rather than silently
+ *  bucketing the new band as `no_change` (which would hide a real discrepancy). */
+export const changeFromBand = (band: Band): Change => {
+  switch (band) {
+    case 'missing':
+      return 'promotion';
+    case 'stale':
+      return 'demotion';
+    case 'ok':
+      return 'no_change';
+    default: {
+      const _exhaustive: never = band;
+      throw new Error(`changeFromBand: unhandled band ${String(_exhaustive)}`);
+    }
+  }
+};
 
 /** One row of the Comparison View — incumbent vs recommended, and the resulting change. */
 export const MemberDiscrepancySchema = z
@@ -74,6 +88,9 @@ export type DiscrepancyReport = z.infer<typeof DiscrepancyReportSchema>;
  * Read-only by construction: it consumes decisions and emits a report. It NEVER mutates a role — that is
  * the whole point of Shadow Mode (operate alongside the incumbent, show potential actions, change
  * nothing until the operator reviews the report and chooses to go live).
+ *
+ * Precondition: `records` are schema-valid (validate upstream via AccessDecisionRecordSchema — the wired
+ * path does). diffShadow is a public export; an unvalidated caller risks double-counted duplicate rows.
  */
 export function diffShadow(records: readonly AccessDecisionRecord[]): DiscrepancyReport {
   const members: MemberDiscrepancy[] = records.map((r) => ({
--- a/packages/protocol/shadow-audit/src/index.ts
+++ b/packages/protocol/shadow-audit/src/index.ts
@@ -76,3 +76,15 @@ export {
   type AuditInputs,
   computeInputsHash,
 } from './inputs-hash.js';
+
+// Shadow Mode — the Discrepancy engine (Comparison View + Discrepancy Report over the audit's bands).
+export {
+  diffShadow,
+  changeFromBand,
+  ChangeSchema,
+  MemberDiscrepancySchema,
+  DiscrepancyReportSchema,
+  type Change,
+  type MemberDiscrepancy,
+  type DiscrepancyReport,
+} from './discrepancy.js';
--- a/packages/protocol/shadow-audit/src/index.ts
+++ b/packages/protocol/shadow-audit/src/index.ts
@@ -88,3 +88,14 @@ export {
   type MemberDiscrepancy,
   type DiscrepancyReport,
 } from './discrepancy.js';
+
+// Shadow Mode — the runner (produce + persist a Discrepancy Report, read-only; store is injected).
+export {
+  ShadowSnapshotSchema,
+  toShadowSnapshot,
+  runShadow,
+  makeMemoryStore,
+  type ShadowSnapshot,
+  type ShadowRunInput,
+  type ShadowStore,
+} from './shadow-runner.js';
--- a/packages/protocol/shadow-audit/src/shadow-runner.ts
+++ b/packages/protocol/shadow-audit/src/shadow-runner.ts
@@ -1,0 +1,85 @@
+/**
+ * Shadow Mode — the runner. Produce + persist a point-in-time Discrepancy Report, read-only.
+ *
+ * "ARRAKIS operates ALONGSIDE existing token-gating without modifying Discord roles" — so the runner
+ * records what WOULD change (a `ShadowSnapshot`) and persists it; it never touches a role. The dashboard
+ * reads the persisted SERIES (the Comparison View over time; the operator watches it converge before
+ * going live). A cadence (cron / scheduled-cycle) fires `runShadow`; gathering the live audit records
+ * (runAudit) is the caller's job, so the runner stays pure + testable. The concrete store (file /
+ * EventStore / db) is an injected adapter — only laplas-gated, durable persistence touches the world.
+ */
+import { z } from 'zod';
+import { AuditModeSchema, type AuditMode } from './schemas/common.js';
+import type { AccessDecisionRecord } from './schemas/access-decision-record.js';
+import { diffShadow, DiscrepancyReportSchema } from './discrepancy.js';
+
+/** A point-in-time shadow report — the persisted artifact the dashboard reads (one per run). */
+export const ShadowSnapshotSchema = z
+  .object({
+    community: z.string().min(1),
+    mode: AuditModeSchema,
+    /** ISO-8601 UTC instant the run was computed. */
+    computed_at: z.string().datetime(),
+    report: DiscrepancyReportSchema,
+  })
+  .strict();
+export type ShadowSnapshot = z.infer<typeof ShadowSnapshotSchema>;
+
+export interface ShadowRunInput {
+  readonly community: string;
+  readonly mode: AuditMode;
+  readonly computed_at: string;
+  readonly records: readonly AccessDecisionRecord[];
+}
+
+/** Produce a snapshot from the audit's records — pure (diffShadow + wrap). Read-only. */
+export function toShadowSnapshot(input: ShadowRunInput): ShadowSnapshot {
+  return ShadowSnapshotSchema.parse({
+    community: input.community,
+    mode: input.mode,
+    computed_at: input.computed_at,
+    report: diffShadow(input.records),
+  });
+}
+
+/**
+ * The persistence PORT. The runner appends snapshots; the dashboard reads the `series`. Concrete stores
+ * (file / EventStore / db) are the integration adapter — the runner is pure over this port, so a cadence
+ * can fire it against any backend.
+ */
+export interface ShadowStore {
+  append(snapshot: ShadowSnapshot): Promise<void>;
+  series(community: string): Promise<readonly ShadowSnapshot[]>;
+  latest(community: string): Promise<ShadowSnapshot | undefined>;
+}
+
+/** An in-memory store — the test seam + the default for a single-process cadence. A durable file/db
+ *  adapter is the integration step. */
+export function makeMemoryStore(): ShadowStore {
+  const byCommunity = new Map<string, ShadowSnapshot[]>();
+  return {
+    async append(s) {
+      const arr = byCommunity.get(s.community) ?? [];
+      arr.push(s);
+      byCommunity.set(s.community, arr);
+    },
+    async series(c) {
+      return [...(byCommunity.get(c) ?? [])];
+    },
+    async latest(c) {
+      const arr = byCommunity.get(c);
+      return arr && arr.length > 0 ? arr[arr.length - 1] : undefined;
+    },
+  };
+}
+
+/**
+ * runShadow — one shadow run: produce the snapshot, persist it, return it. READ-ONLY by construction: it
+ * records what would change and appends to the store; it touches no Discord role. That is Shadow Mode's
+ * whole promise — run alongside, show the discrepancy, change nothing until the operator goes live.
+ */
+export async function runShadow(input: ShadowRunInput, store: ShadowStore): Promise<ShadowSnapshot> {
+  const snapshot = toShadowSnapshot(input);
+  await store.append(snapshot);
+  return snapshot;
+}
--- a/packages/services/shadow-audit/src/__tests__/audit-router.test.ts
+++ b/packages/services/shadow-audit/src/__tests__/audit-router.test.ts
@@ -134,6 +134,25 @@ describe('POST /v1/audit (named output, authed)', () => {
     expect(Array.isArray(json.records)).toBe(true);
   });
 
+  it('serves the Shadow Mode discrepancy — the dashboard contract (bands-only Comparison View)', async () => {
+    // The seam the Shadow Mode dashboard binds to: the authed response MUST carry output.discrepancy.
+    // Pin it at the HTTP layer so a future router reshape can't silently drop it.
+    const res = await post(createAuditRouter(makeDeps()), '/v1/audit', namedBody());
+    expect(res.status).toBe(200);
+    const json = (await res.json()) as any;
+    expect(json.discrepancy).toBeDefined();
+    // every record is classified into the Comparison View
+    expect(json.discrepancy.aggregate.total).toBe(json.records.length);
+    // the Conviction Distribution is BANDS, never a raw score
+    expect(json.discrepancy.aggregate.band_distribution).toEqual(
+      expect.objectContaining({ stale: expect.any(Number), missing: expect.any(Number), ok: expect.any(Number) }),
+    );
+    // BANDS-ONLY invariant enforced at the served boundary — no numeric score leaks to the dashboard
+    expect(JSON.stringify(json.discrepancy)).not.toContain('"score"');
+    // R2 sold its NFT (role without holdings) → a demotion in the cutover preview (GET aggregate = 1 stale)
+    expect(json.discrepancy.aggregate.demotions).toBeGreaterThanOrEqual(1);
+  });
+
   it('rejects a bad signature with 401', async () => {
     const deps = makeDeps({
       auth: { ...makeDeps().auth, recover: async () => '0x' + 'd'.repeat(40) },
--- a/packages/services/shadow-audit/src/__tests__/audit-service.test.ts
+++ b/packages/services/shadow-audit/src/__tests__/audit-service.test.ts
@@ -198,3 +198,26 @@ describe('runAudit — hardening fixes', () => {
     }
   });
 });
+
+describe('runAudit — Shadow Mode discrepancy (authed)', () => {
+  it('carries the per-member discrepancy when records are included, over the same members', async () => {
+    const r = await runAudit(req({ includeRecords: true }), deps());
+    expect(r.ok).toBe(true);
+    if (!r.ok) return;
+    expect(r.output.records).toBeDefined();
+    expect(r.output.discrepancy).toBeDefined();
+    expect(r.output.discrepancy!.aggregate.total).toBe(r.output.records!.length);
+    expect(r.output.discrepancy!.members.length).toBe(r.output.records!.length);
+    // every band classified into a change (no member dropped)
+    const changed = r.output.discrepancy!.aggregate;
+    expect(changed.promotions + changed.demotions + changed.no_change).toBe(changed.total);
+  });
+
+  it('omits the discrepancy for anonymous callers (no records → no comparison)', async () => {
+    const r = await runAudit(req({ includeRecords: false }), deps());
+    expect(r.ok).toBe(true);
+    if (!r.ok) return;
+    expect(r.output.records).toBeUndefined();
+    expect(r.output.discrepancy).toBeUndefined();
+  });
+});
--- a/packages/services/shadow-audit/src/__tests__/shadow-cycle.test.ts
+++ b/packages/services/shadow-audit/src/__tests__/shadow-cycle.test.ts
@@ -1,0 +1,71 @@
+import { describe, it, expect } from 'vitest';
+import { makeMemoryStore, type Order } from '@freeside/shadow-audit-protocol';
+import { runShadowCycle } from '../shadow-cycle.js';
+import type { AuditDeps, AuditRequest, OwnershipSource, RoleSource, WhaleSource } from '../audit-service.js';
+import type { RoleSnapshot } from '../role-snapshot.js';
+
+const R1 = '0x' + '1'.repeat(40);
+const R2 = '0x' + '2'.repeat(40);
+const R3 = '0x' + '3'.repeat(40);
+const X = '0x' + '4'.repeat(40);
+const Y = '0x' + '5'.repeat(40);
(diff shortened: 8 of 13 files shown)


zkSoju:
## Independent review (FAGAN, cross-model) — landed with fixes

Ran an adversarial settle on this PR before it lands (the keystone: the author doesn't settle their own verdict). FAGAN read the real source across all 13 files + the surrounding schemas (`common.ts`, `access-decision-record.ts`, `mode-resolver.ts`, `role-snapshot.ts`).

**Verdict: safe to land as a read-only backend** — the READ-ONLY safety invariant is airtight (no path mutates a Discord role or writes any system of record), BANDS-ONLY and fail-closed both clean, **0 CRITICAL**. Two real defects found + fixed in `11c2a4c7`:

### HIGH-1 (fixed) — one torn JSONL line bricked the whole history
`shadow-store-file.ts` `series()` parsed each line with `.map(Schema.parse(JSON.parse(l)))` → throws on ANY malformed line. A crash mid-`appendFileSync` leaves a torn trailing line; the next read (and `latest()`) then threw, making the **entire cross-community history unreadable** — defeating the backend's purpose (the dashboard watches the series converge). Now a resilient per-line `flatMap` skips bad lines and keeps every valid snapshot. + regression test (torn fragment + garbage line; both valid snapshots still read back).

### MEDIUM-2 (fixed) — `changeFromBand` silently bucketed unknown bands as `no_change`
Correct for today's 3-band enum, but a future 4th band would be silently `no_change` — hiding a real discrepancy. Now an exhaustive `switch` with a `never` guard: enum drift **fails to compile**. + regression test (out-of-enum band throws).

### Tracked / documented
- **MEDIUM-3** (non-atomic multi-writer append + full-members-per-line) → tracked for the live-deps binding cycle; acceptable for the single-process MVP.
- **LOW-4** (`latest()` = last-appended, not newest-by-`computed_at`) + **LOW-5** (`diffShadow` trusts public input) → documented as preconditions.

**Settle:** protocol 45/45 + service 90/90 green; both packages `tsc` clean.

> the keystone in action — an independent voice caught a durability bug the design + tests missed.

zkSoju:
GECKO triage: CONFLICTING after #395 merge. Rebase `feat/shadow-mode-discrepancy` onto `main`, resolve overlap in `packages/protocol/shadow-audit/src/discrepancy.ts` and service layer. Overlap with merged #395 is partial — keep only delta not shipped in #395.

zkSoju:
**JANITOR (2026-06-30):** Rebased onto latest `origin/main`. Shared CI still blocked on NPM audit until **#404** lands — merge after operator hold clears.

zkSoju (commented):
## Summary

This is a well-shaped read-only backend: a pure runner in `protocol`, a `ShadowStore` port, and a concrete file adapter in `services` — the "shadow mode touches nothing" guarantee is enforced by construction rather than by convention, which is the correct architectural instinct. The concerns cluster in one place: **the persisted artifact is the operator's evidence for a go-live decision, and it currently lies in small ways** — a hardcoded `mode`, silently-dropped torn lines, append-order-as-timestamp, and a duplicate discrepancy computation that creates two independent producers of the same view. Verdict: `REQUEST_CHANGES`.

The architectural meditation worth holding: a shadow system's *entire* value is that its output is trustworthy enough to bet a production cutover on. Google's Keyservice migration and Amazon's classic "dark launch" pattern both learned the same lesson the hard way — a shadow pipeline that silently drops records doesn't merely lose data, it *manufactures false confidence*, because a hole in the series and a converged series look identical on a chart. Every reliability finding below is really one finding wearing four hats: **this system has no way to tell you it's lying.**

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "shadow-cycle-hardcoded-mode",
      "title": "runShadowCycle hardcodes mode: 'dogfood-full', ignoring the order's actual audit mode",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/shadow-cycle.ts",
      "description": "The persisted snapshot's `mode` field is a literal `'dogfood-full'` regardless of `req.order.mode`. In the accompanying test (`shadow-cycle.test.ts`), the order is built with `mode: 'lead-magnet'`, yet the snapshot appended to the store claims `dogfood-full`. Since `ShadowSnapshot.mode` is the only mode provenance carried into the durable series, every persisted artifact for a non-dogfood community is mislabeled, and the Comparison View cannot distinguish a lead-magnet run from a full dogfood run. The tests do not assert on `snapshot.mode`, so this is invisible to the suite.",
      "suggestion": "Pass the mode through from the request (`mode: req.order.mode`) or accept it as an explicit parameter of `runShadowCycle`. Add an assertion in `shadow-cycle.test.ts` that `res.snapshot.mode` equals the order's mode.",
      "confidence": 0.85,
      "faang_parallel": "This is the failure mode behind Google's SRE insistence that provenance fields be *derived, never asserted*. Borg's job-metadata pipeline had an incident where a hardcoded cell label made cross-cell rollouts appear to have succeeded in a cell they never touched. The fix was structural: any field describing 'where/how this came from' must be threaded from the request, never re-stated at the sink.",
      "metaphor": "It's a lab that prints the same patient name on every sample tube. The centrifuge still works perfectly — but no result can ever be matched back to the person it came from.",
      "teachable_moment": "A literal in a provenance field is a lie with a compile-time alibi. When a field's job is to answer 'what produced this?', the only safe source is the thing that produced it. If you find yourself typing a constant into a metadata field, that's the signal to thread the parameter instead.",
      "connection": "This is the same class as the `latest()` ordering bug below: both substitute an assumption for a fact the caller already holds."
    },
    {
      "id": "file-store-silent-data-loss",
      "title": "Torn-line resilience silently discards data with no signal",
      "severity": "HIGH",
      "category": "reliability",
      "file": "packages/services/shadow-audit/src/shadow-store-file.ts",
      "description": "The `series` reader `flatMap`s with an empty-catch, so any line that fails `JSON.parse` or `ShadowSnapshotSchema.parse` is dropped. A torn line is not necessarily garbage: `appendFileSync` is only atomic for writes under the platform's atomic-append size, and a snapshot embedding a full per-member `report.members` array can easily exceed that. Two concurrent cadence processes (or a crash mid-write) can interleave and corrupt a *valid* snapshot, which is then silently skipped forever. The store also silently discards snapshots written by an older/newer schema version. The result is a history with invisible holes — exactly the artifact the operator is supposed to watch converge.",
      "suggestion": "Count and surface skipped lines (return them on the result, log a warning, or expose a `corrupt_lines` counter). At minimum, make `series` observable so an operator can tell 'no discrepancy' apart from 'the snapshot was dropped'. Consider a checksum or length prefix per line, or a single-writer lock.",
      "confidence": 0.75,
      "faang_parallel": "Kafka's log format carries a CRC32 per record precisely because 'the append looked fine' is not evidence the bytes landed intact — and when the CRC fails, Kafka *raises*, it does not skip. Contrast the classic Elasticsearch bulk-indexer bug where dropped documents were swallowed by a permissive parser and users discovered months later their indices were incomplete.",
      "metaphor": "A smoke detector with the battery removed, mounted on the ceiling. It looks like protection. It reports nothing — including when there's a fire.",
      "teachable_moment": "`catch {}` is the most expensive two characters in reliability engineering. Resilience means *degrading loudly*: keep serving, but emit the fact that you degraded. A silent skip converts a data-integrity bug into an analytics bug, which is far harder to find because the output remains plausible.",
      "connection": "Pairs with `unbounded-snapshot-payload` — the very thing that makes lines large (embedded per-member rows) is what pushes them past the atomic-append boundary that makes tearing possible."
    },
    {
      "id": "file-store-latest-ordering",
      "title": "latest() relies on append order, not computed_at — a documented invariant with no enforcement",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/shadow-store-file.ts",
      "description": "`latest()` returns `all[all.length - 1]` and the inline comment concedes this only equals newest-by-`computed_at` because 'the cadence always appends chronologically' and 'a backfill MUST preserve that ordering'. `computed_at` is supplied by the caller (`runShadowCycle`'s fourth parameter), so nothing enforces monotonicity: a retry, a backfill, clock skew between cadence hosts, or a manual re-run with an older timestamp will make `latest()` return a stale snapshot, and the dashboard will show the wrong convergence state.",
      "suggestion": "Since `latest()` already parses the whole file, sort by `computed_at` and take the max — the cost is zero relative to the existing full-file scan. Alternatively, reject appends whose `computed_at` is not greater than the current latest.",
      "confidence": 0.7,
      "faang_parallel": "Spanner exists in large part because 'the order I wrote them in' stopped being a usable proxy for 'the order they happened in' once writes crossed machines. TrueTime is an enormously expensive answer to exactly the question this comment waves away. You don't need TrueTime here — but you do need to stop treating file position as a clock.",
      "metaphor": "Filing paperwork in the order it lands on your desk, then reading the bottom sheet whenever someone asks 'what's the most recent?' It works right up until one envelope gets delayed in the mail.",
      "teachable_moment": "When a comment says 'X MUST hold' and no code checks X, you have written a bug with a warranty disclaimer. The rule of thumb: if you can afford to *assert* the invariant, assert it; if you can afford to *not need* it, eliminate it. Here the second option is free — you're already scanning the whole file.",
      "connection": "Directly enables `no-idempotency-key` to do damage: without a timestamp-derived ordering, a duplicate append is indistinguishable from a legitimate new point."
    },
    {
      "id": "duplicate-discrepancy-computation",
      "title": "runShadowCycle discards the audit's discrepancy and recomputes it inside runShadow",
      "severity": "MEDIUM",
      "category": "design",
      "file": "packages/services/shadow-audit/src/shadow-cycle.ts",
      "description": "The header comment says the cycle 'derives the per-member Comparison View (the audit-service already carries it on `output.discrepancy`)', but the code ignores `audit.output.discrepancy` entirely and passes `records` to `runShadow`, which calls `diffShadow` a second time. Beyond the wasted O(members) pass, this creates two independent paths to the Comparison View: the one served over HTTP (pinned by `audit-router.test.ts`) and the one persisted to the store. If the audit-service ever pre-filters or enriches its records before diffing, the served view and the persisted view will diverge with no test catching it.",
      "suggestion": "Either have `runShadow`/`toShadowSnapshot` accept a precomputed `DiscrepancyReport` (so there is one producer), or drop the misleading comment and state explicitly that the runner is the single authority for the persisted report.",
      "confidence": 0.7,
      "faang_parallel": "This is the 'two sources of truth' failure that Facebook's TAO and Meta's later read-path unification work spent years extracting. The classic instance: a cache and its backing store computing the same derived field independently. They agree in every test and diverge in exactly one production edge case — and then you cannot tell which one is wrong.",
      "metaphor": "Two clocks in the same house, each set by hand. As long as nobody compares them, the house has one time. The moment someone does, the house has an argument.",
      "teachable_moment": "Redundant computation isn't primarily a performance problem — it's a *consistency* problem waiting for a schedule. The question to ask at every fork: 'who is the single producer of this value?' If the answer is 'both, and they happen to agree', you have a latent divergence.",
      "connection": "The doc comment and the code already disagree, which is the tell. The comment is describing the design you want; the code is describing a different one."
    },
    {
      "id": "sync-fs-blocks-event-loop",
      "title": "Synchronous fs calls inside async store methods block the event loop",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "packages/services/shadow-audit/src/shadow-store-file.ts",
      "description": "`series` uses `readFileSync` and `append` uses `mkdirSync` + `appendFileSync`, all inside `async` methods on a port whose signature promises non-blocking I/O. Every `series()` call reads and zod-parses the *entire* append-only file — all communities, all history, including each snapshot's full `report.members` array — synchronously. If this store backs the dashboard read path in a server process, a single dashboard load stalls the event loop proportional to total history size, and it grows monotonically forever with no compaction, rotation, or time-window filter.",
      "suggestion": "Switch to `node:fs/promises` and stream/parse line-by-line. Add a time-window or limit parameter to `ShadowStore.series`, and consider partitioning the JSONL per community so one community's read does not parse every other community's history.",
      "confidence": 0.75,
      "faang_parallel": "Node's own docs carry this warning for a reason: the Node core team's guidance since the libuv rewrite has been that a single `readFileSync` on a hot path converts a concurrent server into a serial one. Netflix's well-documented Node latency investigation traced a p99 cliff to exactly this shape — sync I/O hidden behind an `async` signature.",
      "metaphor": "A restaurant with one waiter who insists on personally walking to the farm for each order. The menu says 'we have many tables.' The farm walk says otherwise.",
      "teachable_moment": "An `async` signature is a *promise to the caller* about blocking behavior, not just a syntax choice. Sync I/O inside an async method is the most dangerous kind of lie in a Node service, because every layer above it is architected on the assumption it isn't happening.",
      "connection": "The `async` port signature was the right design — the adapter simply hasn't caught up to it yet. This is a fixable adapter-level gap, not a design flaw."
    },
    {
      "id": "unbounded-snapshot-payload",
      "title": "Every snapshot embeds the full per-member report; the series grows without bound",
      "severity": "MEDIUM",
      "category": "design",
      "file": "packages/protocol/shadow-audit/src/shadow-runner.ts",
      "description": "`ShadowSnapshotSchema` embeds the whole `DiscrepancyReportSchema` (per-member rows plus aggregate) per run. The dashboard's stated need is to 'watch the discrepancy converge over time' — a time series of aggregates — but the persisted artifact carries the entire member list on every run. A daily cadence over a 10k-member community stores 3.65M member rows per year, and `series()` deserializes all of them to render a convergence chart. There is no retention, compaction, or aggregate-only projection.",
      "suggestion": "Either split the artifact (persist aggregate-only rows in the series, members only for the `latest` snapshot), or add a `series` projection that returns aggregates without materializing `report.members`. Decide a retention policy before this reaches a durable backend.",
      "confidence": 0.6,
      "faang_parallel": "This is precisely why Prometheus separates *metrics* from *logs* and why Google's Monarch stores aggregates at high resolution and raw samples not at all. The convergence chart is a metric; the per-member breakdown is a log. Storing them in one row means you pay log costs to render a metric.",
      "metaphor": "Photographing every individual guest at a party, every hour, when the only question you'll ever ask is 'how many people are still here?'",
      "teachable_moment": "The shape of your read query should drive the shape of your write. Ask 'what will I actually SELECT?' before deciding what to INSERT. A one-line projection at write time is a hundred times cheaper than a fan-out scan at read time — and it's *much* cheaper to decide now than after the file has a year of history in it.",
      "connection": "Root cause of both `sync-fs-blocks-event-loop`'s severity and `file-store-silent-data-loss`'s torn-line exposure. Fix the payload shape and two other findings shrink."
    },
    {
      "id": "diffshadow-precondition-not-enforced",
      "title": "diffShadow's duplicate-record precondition is documented, not enforced, on a newly-public export",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/protocol/shadow-audit/src/discrepancy.ts",
      "description": "The added doc block states: 'Precondition: `records` are schema-valid ... an unvalidated caller risks double-counted duplicate rows.' In the same PR, `diffShadow` is newly exported from `packages/protocol/shadow-audit/src/index.ts`, making it part of the package's public surface. A comment is not a boundary: a caller passing two records for the same wallet will inflate `aggregate.demotions`/`promotions`/`total`, and the operator's go-live decision is made on those numbers. There is no test for the duplicate-wallet case.",
      "suggestion": "Enforce the precondition where it is now reachable: parse records through `AccessDecisionRecordSchema` at the top of `diffShadow`, and either dedupe by wallet or throw on a duplicate. Add a test asserting the duplicate-wallet behavior is deterministic.",
      "confidence": 0.6,
      "faang_parallel": "Rust's type system is a 60-year argument that preconditions belong in signatures, not prose. Closer to home: the Java `Comparator` contract was documented-not-enforced for years, and 'violates its general contract!' became one of the most-Googled JDK exceptions in history — because a documented precondition is a precondition that *will* be violated the moment it escapes the module that wrote the comment.",
      "metaphor": "A 'staff only' sign on an unlocked door. It's honest, it's clear, and it stops exactly nobody once the building is open to the public.",
      "teachable_moment": "The moment a function becomes an export, its comments stop being reminders to your future self and start being promises to strangers. Exporting is the trigger to convert every documented precondition into either a validation or a type. The doc comment was a good instinct — it correctly identified the hazard. Now finish the job.",
      "connection": "The PR *does* get this right elsewhere: `changeFromBand`'s `never` guard is exactly this principle, applied. This is the same author's discipline not yet reaching the second function."
    },
    {
      "id": "no-idempotency-key",
      "title": "No idempotency key — a retried cadence run appends a duplicate snapshot",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/protocol/shadow-audit/src/shadow-runner.ts",
      "description": "`runShadow` unconditionally appends. If a cron/scheduled-cycle retries after a crash that occurred *after* the append but before the caller acknowledged, or if the same run fires twice, the store silently gains a duplicate snapshot for the same `computed_at`. The series the dashboard plots then double-counts a point; `latest()` is unaffected but the convergence trend is distorted. Neither the memory store nor the file store dedupes.",
      "suggestion": "Give `ShadowStore.append` an idempotency contract: treat `(community, computed_at)` as a natural key and make append a no-op (or an overwrite) on a repeat. Test the double-fire case.",
      "confidence": 0.6,
      "faang_parallel": "Stripe's idempotency-key API is the canonical public example, but the deeper precedent is AWS's design axiom that *every* retryable write needs a client-supplied dedupe token — because at-least-once delivery is the only delivery guarantee a distributed scheduler can actually make. A cron that can crash is a distributed scheduler.",
      "metaphor": "A turnstile that counts you twice if you stumble. The building's occupancy report is now fiction, and nobody stumbled twice.",
      "teachable_moment": "Any operation invoked by a retrying scheduler must be idempotent, and the natural key is almost always already in your hand — here it's `(community, computed_at)`. Idempotency isn't a feature you add later; it's a property you either have at the write boundary or spend years reconciling downstream.",
      "connection": "Compounds with `file-store-latest-ordering` — with neither dedupe nor timestamp ordering, the durable series has no defense against a double-fire at all."
    },
    {
      "id": "empty-records-conflated-with-failure",
      "title": "Zero records is treated as a refusal, conflating a degraded source with a legitimately empty community",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/shadow-cycle.ts",
      "description": "`if (records.length === 0) return { ok: false, reason: 'no-records' }` — but an empty record set has at least two very different causes: (a) the community genuinely has no gated members right now, and (b) the roles/ownership source returned empty due to a transient upstream failure. Both produce the same `ok: false` and, critically, both persist nothing. Case (a) leaves a silent gap in the series the operator is watching; case (b) fails open with a generic reason string carrying no upstream diagnostic.",
      "suggestion": "Persist a snapshot with a zero-member report when the audit succeeds with no members (the empty state is a real, meaningful data point in a convergence series), and reserve `ok: false` for actual refusals. If empty must remain a refusal, distinguish upstream failure from genuine emptiness in the reason.",
      "confidence": 0.55,
      "faang_parallel": "The distinction between 'empty' and 'unknown' is the entire reason SQL has `NULL` separate from `0`, and the reason Prometheus distinguishes an absent series from a series of zeros. Google's monitoring guidance is explicit: a gap in a graph must never be renderable as a value, because operators read a flat line as 'healthy' and an absent line as 'broken' — and you have just made them look identical.",
      "metaphor": "A thermometer that shows a blank screen both when the room is exactly zero degrees and when the battery is dead. One of those readings should make you put on a coat; the other should make you buy a battery.",
      "teachable_moment": "'No data' and 'data saying nothing happened' are different facts, and collapsing them is one of the most consequential type errors in observability. This matters doubly here: the operator's go-live decision is 'the discrepancy converged to zero' — and a gap in the series looks a lot like convergence.",
      "connection": "Same family as `file-store-silent-data-loss`: both make the absence of a signal indistinguishable from a benign signal."
    },
    {
      "id": "test-fixed-tmp-path",
      "title": "File-store test uses a fixed tmpdir path — collision-prone under parallel/concurrent runs",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "packages/services/shadow-audit/src/__tests__/shadow-store-file.test.ts",
      "description": "`const PATH = join(tmpdir(), 'shadow-store-file.test.jsonl')` is a single well-known path, cleaned only in `beforeEach`. Two concurrent invocations of the suite on the same machine (CI runners sharing a host, a watch-mode rerun overlapping a manual run, or a future split of these cases across parallel vitest files) will interleave appends to the same file and produce cross-talk failures that look like real store bugs.",
      "suggestion": "Use `mkdtempSync(join(tmpdir(), 'shadow-store-'))` per test (or per suite) and remove the directory in an `afterEach`.",
      "confidence": 0.65,
      "faang_parallel": "Bazel's hermeticity model exists because shared mutable test state is the single largest source of flaky tests at Google scale — and a flaky test doesn't just waste a rerun, it *teaches engineers to ignore red*. That's the real cost, and it's cultural, not computational.",
      "metaphor": "Two chefs sharing one cutting board, coordinating only by shouting 'I'm done!' from the next room. It works until they're both busy — which is precisely when it matters.",
      "teachable_moment": "Test isolation should be structural (a unique directory), not procedural (a cleanup step that runs first). `beforeEach` cleanup assumes you're the only runner; `mkdtempSync` doesn't need to assume anything. Prefer the version that can't be wrong.",
      "connection": "The same single-writer assumption that makes this test fragile is the one `file-store-silent-data-loss` warns about in production. The test is quietly modeling the bug."
    },
    {
      "id": "bands-only-assertion-brittle",
      "title": "The bands-only invariant is asserted with a substring check that a rename would defeat",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/services/shadow-audit/src/__tests__/audit-router.test.ts",
      "description": "`expect(JSON.stringify(json.discrepancy)).not.toContain('\"score\"')` is the enforcement of a stated invariant ('no numeric score leaks to the dashboard'). It only catches a field literally named `score`. A field named `raw_score`, `conviction_value`, `points`, or a score embedded in a nested `evidence` object all pass. The invariant is important enough to be called out as a boundary guarantee; the check does not actually guard it.",
      "suggestion": "Assert positively against the schema instead: parse the served `discrepancy` with `DiscrepancyReportSchema` (which is `.strict()`) and assert it round-trips — a strict schema rejects any unexpected field, which is the real guarantee. Keep the substring check only as a secondary smoke test.",
      "confidence": 0.6,
      "faang_parallel": "This is the denylist-vs-allowlist lesson that every security team relearns. It's why Content Security Policy is an allowlist and why Chrome's XSS Auditor (a denylist-shaped filter) was eventually removed entirely — it created a false sense of protection while missing the cases that mattered.",
      "metaphor": "A bouncer with a list of one name he's been told not to let in. Everyone else walks straight past, including the person's twin brother.",
      "teachable_moment": "When enforcing 'nothing unexpected may pass', enumerate what *may* pass, not what may not. You already own a `.strict()` schema — that's an allowlist you built and then didn't use at the boundary where it would do the most good.",
      "connection": "The good news: the fix is free. The strict schema already exists; the test just has to reach for it."
    },
    {
      "id": "audit-service-import-no-callsite",
      "title": "audit-service.ts adds a diffShadow import with no visible call site in the diff",
      "severity": "LOW",
      "file": "packages/services/shadow-audit/src/audit-service.ts",
      "description": "The only change shown for `audit-service.ts` is `+ diffShadow,` added to the import block. The new tests (`audit-service.test.ts`, `audit-router.test.ts`) assert that `output.discrepancy` is populated on authed runs, which requires a call site and a schema field that are not present in this diff. Either the diff is truncated (in which case the reviewable surface for the most behavior-changing file is missing), or the import is unused and will fail a `noUnusedLocals` build.",
      "suggestion": "Confirm the full `audit-service.ts` change is included in the PR diff — the call site that populates `output.discrepancy` and the corresponding `AuditOutputSchema` field are the contract the router test pins, and both should be reviewable.",
      "confidence": 0.5,
      "faang_parallel": "Every large-scale review culture (Chromium's OWNERS, Google's readability) enforces the same rule: the reviewer must see the change that the tests are testing. A green suite over an unreviewable diff is not evidence — it is the *appearance* of evidence, which is worse than nothing.",
      "metaphor": "A recipe that lists an ingredient in the shopping list but never mentions it in the steps. Either the dish is missing something, or the page is torn.",
      "teachable_moment": "When a diff's tests describe behavior the diff doesn't contain, stop and resolve the discrepancy before reviewing anything else. This is a low-severity *finding* but a high-priority *question* — it may be hiding the most important code in the PR.",
      "connection": "Also the crux of `duplicate-discrepancy-computation`: whether the audit-service's discrepancy and the runner's are truly identical cannot be judged without this call site."
    },
    {
      "id": "persisted-wallet-pii-plaintext",
      "title": "Wallet-level member rows persisted to a plaintext JSONL with no retention policy",
      "severity": "LOW",
      "category": "security",
      "file": "packages/services/shadow-audit/src/shadow-store-file.ts",
      "description": "Each snapshot embeds `report.members` — per-wallet discrepancy rows — appended in plaintext to a file whose path is caller-supplied, with no permissions hardening (`appendFileSync` inherits default umask), no rotation, and no retention. The file accumulates the full wallet↔access-decision mapping for every community, forever. The module doc claims 'only laplas-gated, durable persistence touches the world', but no gate is visible in this adapter.",
      "suggestion": "Set an explicit restrictive file mode on create, document the retention/erasure story, and either wire the referenced gate into `makeFileStore` or drop the claim from the doc comment so it is not mistaken for an implemented control.",
      "confidence": 0.45,
      "faang_parallel": "The pattern here — a debug/shadow artifact quietly becoming a durable PII store — is the shape of a whole genre of incidents, most famously the several cases where verbose logging captured credentials or identifiers that then inherited the log retention policy (i.e., forever) rather than a data policy. Shadow data is data.",
      "metaphor": "A notepad by the door where you jot down every visitor's name 'just while we're testing the doorbell'. Two years later it's the most complete guest list in the building, and it's still by the door.",
      "teachable_moment": "The most dangerous sentence in a doc comment is one that describes a control you intend to build. A future reader — human or agent — will read it as a control that exists and reason accordingly. Either implement it or delete the sentence; a claimed-but-absent gate is worse than an acknowledged gap, because it terminates the search.",
      "connection": "Ties to `unbounded-snapshot-payload`: if the series carried aggregates only, this finding would largely evaporate. One design change closes two doors."
    },
    {
      "id": "cycle-reason-untyped",
      "title": "ShadowCycleResult.reason is a bare string, erasing the refusal taxonomy",
      "severity": "LOW",
      "file": "packages/services/shadow-audit/src/shadow-cycle.ts",
      "description": "`reason: string` flattens `audit.refusal.code` (presumably a union like `'external-mode' | ...`) plus the ad-hoc `'no-records'` literal into an untyped string. Callers (a cron handler, an alerting path) cannot exhaustively switch on the refusal, and a typo in a comparison compiles fine — the test asserts `res.reason).toBe('external-mode')` as a magic string.",
      "suggestion": "Type it as `AuditRefusalCode | 'no-records'` so consumers get exhaustiveness, consistent with the `never`-guard discipline applied to `changeFromBand` in this same PR.",
      "confidence": 0.6,
      "faang_parallel": "gRPC's `status.Code` enum versus a free-text message is exactly this distinction, and the reason every mature RPC stack has both: the *code* is for machines to branch on, the *message* is for humans to read. Collapsing them into one string means neither audience is served well.",
      "metaphor": "A hospital chart where the diagnosis field accepts free text. Every doctor understands it; no computer can ever count how many patients had the flu.",
      "teachable_moment": "Widening a union to `string` at a module boundary is an irreversible loss of information — downstream, no one can ever narrow it back. Type the taxonomy at the point where you still know it.",
      "connection": "The PR already knows this lesson: `changeFromBand`'s exhaustive switch is the *same* discipline. Apply it one function over."
    },
    {
      "id": "praise-exhaustive-band-guard",
      "title": "changeFromBand converted to an exhaustive switch with a never guard and a runtime throw",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/protocol/shadow-audit/src/discrepancy.ts",
      "description": "The ternary chain silently bucketed any unknown band as `no_change` — the worst possible default for a discrepancy engine, since it hides a real change. The replacement fails at compile time if `BandSchema` grows a fourth band, and throws at runtime if a bad band slips past the type. The added test (`throws on an unknown band`) pins the runtime half specifically, acknowledging that the compile-time guard does not protect against bad data. This is the right failure direction for a system whose output gates a go-live decision.",
      "confidence": 0.9,
      "faang_parallel": "This is the `-Werror=switch` discipline that the Linux kernel and LLVM both enforce, and it's why Rust's `match` is exhaustive by construction. The `never` guard is TypeScript's version of the same idea, and using it *plus* a runtime throw shows you understand that types guard your code while validation guards your data — two different threats, two different defenses.",
      "metaphor": "A circuit breaker rather than a fuse that quietly melts. When something unexpected arrives, the system stops and tells you, instead of continuing at 90% and letting you find out from the smell.",
      "teachable_moment": "The default case of a switch is a design decision, and 'no_change' was the one default guaranteed to be indistinguishable from success. Choosing to fail loudly in a measurement system is choosing to be *wrong noisily rather than right-looking quietly* — and for a system whose only job is to earn trust for a cutover, that's the only defensible choice. The rest of this review is mostly asking that this exact instinct be applied three more places.",
      "connection": "This finding is the standard the PR sets for itself. `diffshadow-precondition-not-enforced` and `cycle-reason-untyped` are the two places it hasn't reached yet."
    },
    {
      "id": "praise-injected-store-port",
      "title": "Pure runner over an injected store port keeps the read-only guarantee testable",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/protocol/shadow-audit/src/shadow-runner.ts",
      "description": "Keeping `toShadowSnapshot` pure, defining `ShadowStore` as a port in the protocol package, and putting the concrete file adapter in services means the read-only claim is enforced by construction rather than by convention — the runner has no capability to touch a role. The `is READ-ONLY — it touches no input record` test and the fresh-store-reopens-the-file test both verify the property that actually matters rather than restating the implementation.",
      "confidence": 0.85,
      "faang_parallel": "This is hexagonal architecture used for its *actual* purpose rather than as ceremony. The closest well-known analogue is how Google's production systems enforce capability boundaries: you don't audit that a component didn't write; you give it no handle that can write. Structural guarantees survive refactors that convention-based ones do not.",
      "metaphor": "The shadow runner is a security camera, not a security guard. It cannot open the door because it has no hands — which is a far stronger guarantee than a guard who has been *told* not to.",
      "teachable_moment": "The strongest way to prove a component can't do something is to make the capability unreachable from its type. Note the second-order benefit visible in this PR: because the port exists, `sync-fs-blocks-event-loop` and the durability concerns are all *adapter-level* fixes. Good seams turn architectural problems into local ones.",
      "connection": "This design is what makes every other finding in this review cheap to fix. That is not an accident, and it's worth naming."
    },
    {
      "id": "praise-http-contract-pin",
      "title": "The dashboard contract is pinned at the HTTP boundary, not just the service boundary",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/services/shadow-audit/src/__tests__/audit-router.test.ts",
      "description": "Asserting `output.discrepancy` at the router layer (rather than only in `audit-service.test.ts`) means a future router reshape — a response mapper, a field allowlist — cannot silently drop the field the dashboard binds to. Testing the seam a consumer actually depends on, rather than the layer beneath it, is the correct choice here.",
      "confidence": 0.8,
      "faang_parallel": "This is the insight behind consumer-driven contract testing (Pact) and behind Google's practice of pinning API surfaces at the wire boundary rather than the handler: the layer you test should be the layer your consumer sees. A unit test one level down passes happily while the field never leaves the process.",
      "metaphor": "Testing the tap, not the pipe. The pipe can be perfect and the tap still closed — and it's the tap the person with the glass is standing at.",
      "teachable_moment": "Ask 'what boundary does my consumer actually touch?' and put the contract test *there*. It costs slightly more to set up and it catches the entire category of 'the data existed but never shipped' bugs, which unit tests are structurally blind to.",
      "connection": "The one gap: this test pins that the field is *present*, and `bands-only-assertion-brittle` shows the check on what the field *may not contain* is weaker than the guarantee it's protecting. Same boundary, one half fully armed."
    }
  ]
}
```


## Callouts

Three things here are genuinely first-rate, and they're worth naming precisely because they're what makes the rest of the review cheap to act on.

**[Quality] The `changeFromBand` rewrite is the single best change in this PR.** Replacing a ternary chain — which silently bucketed anything unrecognized as `no_change` — with an exhaustive switch, a `never` guard, and a runtime throw is a deliberate choice to *fail loudly in a measurement system*. The accompanying test pins the runtime half specifically, which shows the author understands that the type guard protects the code and the throw protects against the data. Most of my other findings are just asking that this exact instinct be extended two functions over.

**[Operational] The port/adapter split is doing real load-bearing work, not architectural theater.** By keeping `toShadowSnapshot` pure and defining `ShadowStore` as a port in `protocol`, the "shadow mode is read-only" claim becomes structurally true rather than merely tested — the runner *has no handle* that can mutate a role. The second-order payoff is visible in this very review: every durability and performance concern I raised is an *adapter*-level fix, contained behind a seam that already exists. That's what a good boundary buys you.

**[Test Coverage] Pinning the dashboard contract at the router, not just the service.** Testing `output.discrepancy` at the HTTP seam means a future response mapper or field allowlist can't quietly drop the field the dashboard binds to. Testing the boundary the consumer actually touches — rather than the layer beneath it — is the choice that catches the "data existed but never shipped" class of bug, and it's the choice a lot of otherwise-strong test suites don't make.

The through-line of my concerns is a single question: *when this system is wrong, how will you find out?* Right now, a dropped line, a duplicate append, a mislabeled mode, and an empty upstream all produce output that looks exactly like a healthy converging series. The architecture you've built is more than good enough to fix that — it just needs the `changeFromBand` instinct applied at the persistence boundary too.

**Verdict: `REQUEST_CHANGES`** — primarily on `shadow-cycle-hardcoded-mode` and `file-store-silent-data-loss`, both of which corrupt the artifact the go-live decision rests on.

PR #380, opened 2026-06-28
test(platform/agent): genuine unit test for computeEnsembleAccounting — close a money-path gap

zkSoju:
A per-model cost-accounting function with **no unit test** (only an e2e stub) — surfaced by the Icebreaker meta-loop-builder's evidence loop (the sharp controller found it backed-by-stub, not by a test). Pins the money invariants: failed-never-charged · total == platform + byok · BYOK segregated · savings goes negative on overspend. Verified against the real function.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #383, opened 2026-06-28
test(platform/agent): pin invariants on 5 untested pure functions (Icebreaker acting loop)

zkSoju:
## What

The autonomous output of the **Icebreaker acting loop** (the Refusal) — a cybernetic loop that *acts* on its own: sense the field of untested pure functions → konesans-rank → dispatch actuator agents that each draft + **verify** one real test → settle convergence on the meter (not the generator).

This is the loop *running*, not loop code. Six actuator agents ran in parallel, each producing a genuine vitest test for one untested pure function, verified individually via a tsx smoke against the live source — then the **assembled set was re-verified**: **76 tests green** (68 adapter + 8 core).

## The keystone

> the generator never settles its own verdict

Each agent claimed `verified: true` ONLY if a tsx smoke passed against the **real** function — and the assembled set was re-run against the real functions before this PR. No green is asserted; every green is measured.

## What's pinned (all pure — characterization tests, no source changes)

| Function | File | Invariants pinned |
|---|---|---|
| `parseRateLimitResult` | `agent-rate-limiter.ts` | **fail-closed** Lua-response parser: `allowed=true` ONLY for `dimension==='ok'`; every malformed/unknown/case-mismatched shape denies + 5s retry (security boundary) |
| `parseFinalizeResult` + `parseReaperResult` | `budget-manager.ts` | **money invariants**: cost/count always non-negative integers; negatives clamp to 0; NaN/Infinity/junk → 0; floats truncate toward zero |
| `toDomainEvent` | `capability-audit.ts` | field mapping (community_id→aggregate_id, …); `trace_id` threaded into both correlation_id + causation_id; `contract_version` pin (8.3.0) |
| `formatErrorMessage` | `error-messages.ts` | placeholder substitution; numeric coercion; unknown-key tolerance (no throw) |
| `getValidTransitions` | `core/domain/wizard.ts` | exact state-machine transition table (order-sensitive); INIT entry, DEPLOY terminal |

## Why these

Pure functions with real branching logic — a fail-closed security boundary, money rounding, a state machine — that had **zero tests**. A characterization test on each pins current behavior, so a future refactor that breaks an invariant fails loudly.

## Status

**Draft — surface-only.** No source changes. Not for auto-merge. The loop produced it; the operator decides whether to land it.

### DIFF
--- a/packages/adapters/agent/agent-rate-limiter.test.ts
+++ b/packages/adapters/agent/agent-rate-limiter.test.ts
@@ -1,0 +1,126 @@
+/**
+ * parseRateLimitResult Tests
+ *
+ * Pins the REAL invariants of the Lua-response parser for the agent rate
+ * limiter. This is a fail-closed security boundary: only the exact string
+ * 'ok' from the Lua script may produce allowed=true. Every other shape —
+ * a known-but-exceeded dimension, a malformed/unknown dimension, an empty
+ * response — MUST deny.
+ *
+ * Lua returns: [dimension, remaining, limit, retryAfterMs, resetAtMs]
+ */
+
+import { describe, it, expect } from 'vitest';
+import { parseRateLimitResult } from './agent-rate-limiter.js';
+
+describe('parseRateLimitResult', () => {
+  describe('allow path (dimension === "ok")', () => {
+    it('returns allowed with dimension null and fields parsed from raw', () => {
+      const r = parseRateLimitResult(['ok', '42', '60', '999', '1700000000000']);
+      expect(r).toEqual({
+        allowed: true,
+        dimension: null,
+        remaining: 42,
+        limit: 60,
+        // retryAfterMs is FORCED to 0 on the allow path, ignoring raw[3]='999'
+        retryAfterMs: 0,
+        resetAtMs: 1700000000000,
+      });
+    });
+
+    it('forces retryAfterMs to 0 even when the Lua response carries a non-zero value', () => {
+      const r = parseRateLimitResult(['ok', '1', '2', '99999', '5']);
+      expect(r.allowed).toBe(true);
+      expect(r.retryAfterMs).toBe(0);
+    });
+
+    it('coerces non-numeric numeric fields to 0', () => {
+      const r = parseRateLimitResult(['ok', 'abc', '', 'foo', 'bar']);
+      expect(r).toEqual({
+        allowed: true,
+        dimension: null,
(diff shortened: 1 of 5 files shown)


zkSoju:
## Verification — gap-closer tests are genuinely green against live source

Ran the **real** test files (vitest aliased to a node-assert shim, no install) against current `main` source:

| test file | result |
|---|---|
| `budget-manager.test.ts` (reaper/finalize parsers — money) | **25/25 pass** |
| `wizard.test.ts` (state-machine transitions) | **8/8 pass** |
| `error-messages.test.ts` | **9/9 pass** |
| `agent-rate-limiter.test.ts` (parse) | **4/4 pass** |
| `capability-audit.test.ts` | not run here — imports `@0xhoneyjar/loa-hounfour` (one constant); resolves in the CI workspace, not my scratch dir. Not a test failure. |

**46 assertions verified green.** All five files are test-only (zero source change) over **pure functions**, so they cannot introduce a real source failure. The red checks are the estate-wide numb gates (red on `main` too), not this PR.

These pin invariants on functions the Icebreaker acting-loop surfaced as asserted-but-unmetered (budget reaper clamps negatives/NaN/Infinity to 0; finalize defaults to INVALID_INPUT; wizard forbids illegal jumps). Ready for your merge decision.

🤖 verified-by-execution (real test files, live source)

zkSoju:
## ⚠ Placement caveat — these tests won't actually RUN in CI as located

Follow-up to my verification above (the tests ARE green): I checked whether they *gate*. They don't, as placed.

- `agent-ci` **triggers** on `packages/**` (so this PR fires it) — but its "Unit tests" step runs **`npx vitest run tests/unit/`** from repo root, which only executes root `tests/unit/*.test.ts`.
- This PR's tests live in **`packages/adapters/agent/`** + **`packages/core/domain/`** — *not* under `tests/unit/`, so `vitest run tests/unit/` skips them. `packages/adapters` has its own `"test": "vitest run"`, but no workflow invokes it.

**Net: agent-ci will go green here while these 5 files never run** — green-but-ungated (`exists ≠ gates`). They verify the functions *now* but won't catch a future regression in CI.

**Fix options:** (a) move them to `tests/unit/` (like #380), or (b) add a CI step `cd packages/adapters && npx vitest run` (and one for `packages/core`). (a) is the least-friction match to the existing gate.

🤖 verification-ladder: I verified they're green *and* checked the rung above (do they gate). They don't yet.

PR #296, opened 2026-06-22
feat(shadow-access-audit): Sprint 1 — dogfood-NFT audit magnet (build complete, unit-verified)

zkSoju:
## Shadow Access Audit — Sprint 1 (dogfood-NFT magnet)

The BD lead magnet: a self-serve audit that shows a community **who still has role access but no longer holds** (stale access), reconstructed from on-chain Transfer history. Draft — the operator reviews before merge.

### What this delivers (all 9 sprint tasks, unit-verified)

| Task | What | Where |
|------|------|-------|
| T1 | Sealed Zod schemas + deterministic `inputs_hash` | `packages/protocol/shadow-audit/` |
| T2 | SonarClient — Transfer-replay ownership@block (721 fold / 1155 balance) | `packages/adapters/sonar/` |
| T3 | ScoreProxy — server-side key + circuit breaker | `packages/adapters/score/` |
| T4–T6 | EligibilityResolver · ModeResolver+RoleSnapshot · AuditService (hexagonal, k-anon) | `packages/services/shadow-audit/` |
| T7 | Hono API — anon aggregate · auth-gated named output · reaction/contact | `…/http/audit-router.ts` |
| T8 | EventStore — append-only, consented, no member data (+ SQL DDL) | `…/event-store.ts`, `sql/` |
| T9 | Thin dashboard HTML view (aggregate + reaction + dual CTA) | `…/http/audit-router.ts` |

### Verification

**134 tests pass** · all `tsc --noEmit` clean:
- protocol **35** · adapters/sonar+score **28** · services **71**

```
cd packages/protocol/shadow-audit && pnpm test && pnpm exec tsc --noEmit
cd packages/adapters && pnpm exec vitest run sonar/__tests__ score/__tests__
cd packages/services/shadow-audit && pnpm test && pnpm exec tsc --noEmit
```

### Flatline §11 invariants honored (+ tested)
ERC-1155 balance semantics & batch · reconstruction-refusal (never silently-wrong) · POST+nonce+community-binding auth (replay-safe; nonce not burned on bad sig) · k-anonymity cohort suppression · deterministic `inputs_hash`=sha256(JCS(...)) · run lifecycle · typed refusals → stable HTTP codes · **no numeric score** (`.strict()` schemas) · **no member persistence** (NFR-1).

### AC status (SDD §8/§11)
[...]

PR #387, opened 2026-06-29
feat(shadow-audit): Sprint 2 — the deployment surface (make the MVP servable for freeside-dashboard)

zkSoju:
## Shadow Access Audit — Sprint 2: the deployment surface (make the MVP servable)

Makes the built-but-unmounted Shadow Audit (#296/#300/#306) a **deployable HTTP building** (`shadow-audit-api`) so freeside-dashboard's already-built **dormant** client (`GET ${SHADOW_AUDIT_API_URL}/v1/audit`) can consume it. The dashboard needs **zero** code change — point `SHADOW_AUDIT_API_URL` (+ `_KEY`) at the deployed audit-api and the seam goes live.

### ⚠️ The scope correction (the deeper truth)
The bead said "unmounted," but the real gap was bigger: Sprint 1 built the audit against its I/O **ports** but **only ever wired test fakes** — no real `OwnershipSource`/`WhaleSource`/`RoleSource` existed. This PR builds those real adapters + the composition root + the server entry + deploy artifacts.

### What's here (`packages/services/shadow-audit/`)
- `src/server.ts` — composition root: `buildAuditApp` + `configFromEnv` (fail-loud), X-API-Key gate, fail-closed POST (V2 not wired).
- `src/role-source.ts` · `src/whale-source.ts` — real local adapters (validated file loader · balance-distribution concentration).
- `src/ownership-source.ts` · `src/block-time-resolver.ts` — the real chain adapter: SonarClient reconstruction + date→block (RPC binary search) + a chain/contract→collection registry (an unregistered collection is **refused**, never reconstructed).
- `bin/http.ts` — the server entry (builds the chain adapter from env, serves via `@hono/node-server`). `Dockerfile` + `railway.toml`.

### Verified
- `tsc --noEmit` clean · full suite **105 tests** (fakes-injected, no live network — the #306 discipline) · `bin/http.ts` boots + `/healthz`→200 + **fails loud** on missing/invalid config.
- **Ran an adversarial FAGAN review on the diff** (generator-never-settles). It found 6 real issues my tests missed — all fixed + pinned: HIGH-1 (a before-genesis snapshot silently served block 0 → wrong audit; now refuses), MEDIUM-2 (registry now zod-validated), MEDIUM-3 (boot-validate RPC per chain), MEDIUM-5 (AUDIT_K=0 disabled k-anon; now rejected), MEDIUM-6 (one CONFIRMATIONS source), LOW-7 (timingSafeEqual). The review confirmed the security core holds (fail-closed POST, the gate, the registry refusal).

### 🚨 DEPLOY GATE (money/ops — operator must verify before merge/deploy)
**Live correctness is NOT verified** (built AFK, no live chains reachable). The unit suite proves the algorithm + assembly, NOT the live values. Before production: provide + verify the deploy config — `COLLECTION_REGISTRY` (the 8 collections' chain/contract→collection+standard map), `RPC_URL_<chain>`, `BELT_GATEWAY_URL`, `ROLE_SNAPSHOT_PATH` — and run a live block-at-date spot-check + `pnpm -C packages/adapters test:live`. A wrong collection mapping or RPC ⇒ a wrong holder set ⇒ a wrong audit.

### Scope note
The 3 leading commits (`815e90e3`, `ad7e68cc`, `9cb74470`) are the pre-existing `coord/kuang-beacon-shape-validation` base (operator-dash beacon + cheval) — **out of scope** for this review. The shadow-audit work is the 3 commits `192051b8` / `7e3dc323` / `6c710a6a`.

Closes the build half of bead `arrakis-r8n1` (the deploy + live-verification remain operator-gated).

### DIFF
--- a/packages/services/shadow-audit/DEPLOY.md
+++ b/packages/services/shadow-audit/DEPLOY.md
@@ -10,7 +10,7 @@ code change**, only its env pointed here.
 |-----|----------|------|
 | `OPERATED_COMMUNITIES` | ✅ | comma-separated community names this deploy audits (dogfood-full eligible) |
 | `CTA_PRODUCT`, `CTA_CONVERSATION` | ✅ | the product + conversation door URLs surfaced in the audit |
-| `COLLECTION_REGISTRY` | ✅ | JSON: `{ "<chain>/<contract>": { "collection": "<belt-gateway id>", "standard": "erc721"\|"erc1155" } }` — **the 8 THJ collections** (HoneyJar1/3/4/6 + Honeycomb + crayons_factory @ 80094, HoneyJar2 @ 42161, HoneyJar5 @ 8453). Zod-validated at boot. |
+| `COLLECTION_REGISTRY` | ✅ | JSON: `{ "<chain>/<contract>": { "collection": "<belt-gateway id>", "standard": "erc721"\|"erc1155" } }`. Maps each `(chainId, contract-address)` a caller passes → the belt-gateway **collection id**. Zod-validated at boot. The 8 collection ids + their chains are live-grounded below; **the contract addresses are operator-supplied** (see the gate). |
 | `RPC_URL_<chain>` | ✅ (per registry chain) | a JSON-RPC endpoint per chain in the registry (e.g. `RPC_URL_80094`) for block-at-date resolution. Boot fails if any registry chain lacks one. |
 | `SHADOW_AUDIT_API_KEY` | optional | the `X-API-Key` the dashboard sends (when unset the k-anon aggregate is open). Set it + the dashboard's `SHADOW_AUDIT_API_KEY`. |
 | `BELT_GATEWAY_URL` | optional | sonar GraphQL endpoint (defaults to belt-gateway-production) |
@@ -20,6 +20,28 @@ code change**, only its env pointed here.
 | `PORT` | optional | bind port (Railway sets it; default 3040) |
 
 The server **fails loud at boot** on any missing/invalid required value — it never serves a half-wired audit.
+
+### Live-grounded auditable set (belt-gateway `CollectionStat`, queried 2026-06-29)
+
+The belt-gateway reconstructs ownership keyed on `(collection id, chainId)` — it stores **no contract
+address**, so the registry's contract addresses are the operator's to supply (from the deployment records).
+These are the exact `(collection, chain)` pairs that have indexed supply and are therefore auditable:
+
+| collection id | chains (chainId) | standard |
+|---|---|---|
+| `HoneyJar1` | eth (1), bera (80094) | erc721 |
+| `HoneyJar2` | **arb (42161)**, bera (80094), eth (1) | erc721 |
+| `HoneyJar3` | **zora (7777777)**, bera (80094) | erc721 |
+| `HoneyJar4` | **op (10)**, bera (80094), eth (1) | erc721 |
+| `HoneyJar5` | **base (8453)**, bera (80094), eth (1) | erc721 |
+| `HoneyJar6` | eth (1), bera (80094) | erc721 |
+| `Honeycomb` | eth (1), bera (80094) | erc721 |
+| `crayons_factory` | (in Transfer; absent from CollectionStat — confirm chain + standard) | ? |
+
+So a registry entry is `"<chainId>/<contract-address-on-that-chain>": { "collection": "HoneyJar5", "standard": "erc721" }`.
+Each generation has a canonical chain (bolded) **plus** a Berachain bridge; include only the `(chain, contract)`
+pairs the audited communities actually gate on. `crayons_factory` and per-collection token standards still need
+operator confirmation (it indexes Transfers but has no CollectionStat row).
 
 ## Railway
 
--- a/packages/services/shadow-audit/DEPLOY.md
+++ b/packages/services/shadow-audit/DEPLOY.md
@@ -1,0 +1,46 @@
+# shadow-audit-api — deploy runbook
+
+The Shadow Access Audit as its own deployable HTTP building. freeside-dashboard's
+`access-audit/client.ts` consumes it (`GET ${SHADOW_AUDIT_API_URL}/v1/audit`) — the dashboard needs **zero
+code change**, only its env pointed here.
+
+## Environment
+
+| Var | Required | What |
+|-----|----------|------|
+| `OPERATED_COMMUNITIES` | ✅ | comma-separated community names this deploy audits (dogfood-full eligible) |
+| `CTA_PRODUCT`, `CTA_CONVERSATION` | ✅ | the product + conversation door URLs surfaced in the audit |
+| `COLLECTION_REGISTRY` | ✅ | JSON: `{ "<chain>/<contract>": { "collection": "<belt-gateway id>", "standard": "erc721"\|"erc1155" } }` — **the 8 THJ collections** (HoneyJar1/3/4/6 + Honeycomb + crayons_factory @ 80094, HoneyJar2 @ 42161, HoneyJar5 @ 8453). Zod-validated at boot. |
+| `RPC_URL_<chain>` | ✅ (per registry chain) | a JSON-RPC endpoint per chain in the registry (e.g. `RPC_URL_80094`) for block-at-date resolution. Boot fails if any registry chain lacks one. |
+| `SHADOW_AUDIT_API_KEY` | optional | the `X-API-Key` the dashboard sends (when unset the k-anon aggregate is open). Set it + the dashboard's `SHADOW_AUDIT_API_KEY`. |
+| `BELT_GATEWAY_URL` | optional | sonar GraphQL endpoint (defaults to belt-gateway-production) |
+| `ROLE_SNAPSHOT_PATH` | optional | path to the Discord role-export JSON (absent → audits refuse external-mode) |
+| `CONFIRMATIONS` | optional | reorg finality depth (default 12; one source for both sonar + the "current" block) |
+| `AUDIT_K` | optional | k-anonymity threshold (default 5; **must be a positive integer** — `AUDIT_K=0` is rejected, it would disable k-anon) |
+| `PORT` | optional | bind port (Railway sets it; default 3040) |
+
+The server **fails loud at boot** on any missing/invalid required value — it never serves a half-wired audit.
+
+## Railway
+
+`railway.toml` is set (Dockerfile builder, healthcheck `/healthz`). Set in the service settings: Root Directory
+`packages/services/shadow-audit`, Build Context = repo root. The Dockerfile builds `@freeside/adapters`
+(the service resolves `@freeside/adapters/sonar` from its dist) then runs via `tsx bin/http.ts`.
+
+## 🚨 LIVE-CORRECTNESS GATE (money/ops — do this BEFORE pointing the dashboard at it)
+
+The unit suite (105 tests) proves the algorithm + assembly with injected fakes — **NOT** the live values. A
+wrong `COLLECTION_REGISTRY` mapping or RPC ⇒ a wrong holder set ⇒ a wrong audit (wrong access decisions).
+
+1. **Collection mapping** — verify each `collection` id against the live belt-gateway (a Transfer query that
+   matches nothing = a typo'd id = an empty, silently-wrong audit). `pnpm -C packages/adapters test:live`
+   exercises the real reconstruction.
+2. **RPC block-at-date** — spot-check one chain: pick a known date, confirm the resolved snapshot block's
+   timestamp is ≤ that date's UTC end and the next block's is >.
+3. **Smoke the real GET** — `curl "$URL/v1/audit?chain=80094&contract=<honeycomb>&snapshot_date=<known>&community=<thj>&owner_wallet=<addr>&threshold=1"`
+   and sanity-check the aggregate against a known holder count.
+4. **Then** set the dashboard's `SHADOW_AUDIT_API_URL` (+ `_KEY`) and the seam goes live. The dashboard
+   strict-decodes the response against the protocol `AuditOutputSchema`; the contract-parity test in
+   `src/__tests__/server.test.ts` pins that the GET output matches it.
+
+V2 (the authed `POST /v1/audit` named-output, SIWE) is fail-closed (always 401) until wired.
--- a/packages/services/shadow-audit/Dockerfile
+++ b/packages/services/shadow-audit/Dockerfile
@@ -1,33 +1,44 @@
 # shadow-audit-api — Railway-deployable Hono building (the Shadow Access Audit's HTTP surface).
 #
 # Build context: the loa-freeside repo ROOT (not this subdir) — the service consumes
-# @freeside/adapters + @freeside/shadow-audit-protocol via file: links, so both must be in
-# the build context.
+# @freeside/adapters + @freeside/shadow-audit-protocol via file: links, and @freeside/adapters in turn
+# consumes @freeside/core (file:../core), so ALL THREE must be in the build context.
 #
 # Railway: in this service's settings set
 #   Root Directory: packages/services/shadow-audit
 #   Build Context:  <repo-root>
 # OR rely on railway.toml's dockerfilePath (relative to repo root).
 #
-# tsx runs the service + protocol from source; @freeside/adapters is consumed from its built
-# dist (its package exports point to dist/), so it is the one package built here.
+# tsx runs the service + protocol from source (the protocol exports ./src). @freeside/core and
+# @freeside/adapters export dist/, so both are BUILT here (core first — adapters depends on it).
 
-FROM node:22-alpine AS base
+FROM node:22.12-alpine AS base
 WORKDIR /app
 RUN corepack enable && corepack prepare pnpm@9.15.9 --activate
 
-# Stage the service + its file: dependencies.
+# Stage the transitive file: dependency tree: core <- adapters <- service (+ protocol).
+# (F1: packages/core was previously omitted, so the adapters install — which has file:../core — failed.)
+COPY packages/core ./packages/core
 COPY packages/adapters ./packages/adapters
 COPY packages/protocol/shadow-audit ./packages/protocol/shadow-audit
 COPY packages/services/shadow-audit ./packages/services/shadow-audit
 
-# Build @freeside/adapters so the service resolves @freeside/adapters/sonar from dist/.
+# Build @freeside/core FIRST — @freeside/adapters resolves @freeside/core from its dist/ (exports point there).
+WORKDIR /app/packages/core
+RUN pnpm install --no-frozen-lockfile && pnpm build
+
+# Then @freeside/adapters — the service resolves @freeside/adapters/sonar from its dist/.
 WORKDIR /app/packages/adapters
 RUN pnpm install --no-frozen-lockfile && pnpm build
 
-# Install the service (links the built adapters dist + the protocol src; brings tsx + hono).
+# Install the service (links the built core+adapters dist + the protocol src; brings tsx + zod + hono).
+# (F2: tsx + zod are runtime deps of bin/http.ts and now live in `dependencies`, so a production install keeps them.)
 WORKDIR /app/packages/services/shadow-audit
 RUN pnpm install --no-frozen-lockfile
+
+# loa:shortcut: --no-frozen-lockfile (F3) — the repo's npm/pnpm split means per-package lockfiles can lag the
+#   workspace; a frozen install would break the build. Upgrade trigger: once the per-package pnpm-lock.yaml
+#   files are kept in CI-verified sync, switch all three installs to --frozen-lockfile for a reproducible artifact.
 
 ENV NODE_ENV=production
 # Railway sets $PORT; bin/http.ts binds 0.0.0.0:$PORT (default 3040 in local dev).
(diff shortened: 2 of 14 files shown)


zkSoju (commented):
I have everything I need to produce this Bridgebuilder review. The available tools here are Google Drive and Dune MCP integrations — none relevant to this task, and none of the plan-workflow tools (Explore, Plan, Write, ExitPlanMode) are present in this environment. The deliverable is a self-contained text review synthesized from the provided findings, which is a read-only output. Here it is.

---

## Summary

This PR turns the shadow-audit MVP into a genuinely deployable building — a server entry, real source adapters, a block-at-date resolver, and a Dockerfile/Railway surface — and does so with unusually disciplined defensive coding (fail-loud config, constant-time key compare, before-genesis refusal). The strongest risks are not in the algorithm but in the *deploy substrate*: a Dockerfile that copies a partial workspace, a non-frozen lockfile, and runtime deps classified as dev. These won't surface in the vitest suite, which is exactly why they're dangerous.

The architectural through-line worth naming: **the unit suite proves the brains, the container proves the wiring, and right now only the brains are tested.** Every HIGH/MEDIUM build finding below shares one root — module resolution and dependency provenance live in a layer the tests resolve *around* (vitest reads TS source directly), so the producer can be correct and the artifact still boot-crash. The cure is a single container smoke test in CI; it collapses F1/F2/F3/F6/F10 into one gate.

The contract-parity test (F11) and the refuse-never-silently posture (F12) are the kind of work that should be studied, not just merged.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Dockerfile copies only 3 packages — adapters' own workspace deps may break the build",
      "severity": "HIGH",
      "category": "build/deploy",
      "file": "packages/services/shadow-audit/Dockerfile",
      "description": "The build context copies only packages/adapters, packages/protocol/shadow-audit, and packages/services/shadow-audit, then runs `pnpm install --no-frozen-lockfile && pnpm build` inside packages/adapters. If @freeside/adapters has any other intra-workspace `file:`/`workspace:` dependency (e.g. @freeside/core, which CLAUDE.md lists alongside adapters under packages/), those links are absent from the build context and the install/`tsc` build will fail — a hard deploy blocker that the unit suite cannot catch. This is the single most likely reason the image fails to build in Railway.",
      "suggestion": "Verify packages/adapters/package.json has no unmet workspace deps in the copied subset. If it does, either copy those packages too, or build from the repo root with a pnpm workspace filter (`pnpm --filter @freeside/adapters... build`) / `pnpm deploy`. Add a CI job that actually builds this Dockerfile so a missing dep surfaces before merge.",
      "confidence": 0.5,
      "faang_parallel": "This is the failure Google's Bazel and Meta's Buck were built to eliminate: a build that succeeds locally because the full source tree is present, then fails in a sandbox because the dependency graph was never declared, only implied. Bazel's strict-deps refuses to compile against a target you didn't list — turning F1's silent gap into a loud, early error.",
      "metaphor": "You've shipped IKEA furniture with the instructions and the screws, but the diagram references a panel from a different box. Everything looks complete until someone tries to assemble it in an empty room.",
      "teachable_moment": "A Dockerfile's COPY set is an *un-typed dependency declaration*. The package.json says what adapters needs; the Dockerfile silently re-asserts a subset of that, and nothing checks the two agree. Reproducible builds come from making the build graph explicit (workspace filter / `pnpm deploy`) rather than hand-curating COPY lines.",
      "connection": "Verify mechanically: `pnpm --filter @freeside/adapters list --depth 0` (or read its package.json deps) and diff against the COPY set. This is the build-layer instance of the same 'tests resolve around the real wiring' gap as F6."
    },
    {
      "id": "F2",
      "title": "zod (and tsx) are runtime-required by bin/http.ts but declared in devDependencies",
      "severity": "MEDIUM",
      "category": "dependencies",
      "file": "packages/services/shadow-audit/package.json",
      "description": "bin/http.ts does `import { z } from 'zod'` for RegistrySchema, and the production `start` script is `tsx bin/http.ts` — both zod and tsx are required at runtime, yet both sit in devDependencies. This Dockerfile happens to work ONLY because `ENV NODE_ENV=production` is set AFTER the two `pnpm install` lines, so devDeps are installed. Any prod-only install (`pnpm install --prod`, a future multi-stage prune, or a different host) would boot-crash with `Cannot find module 'zod'`. The correctness is incidental to instruction ordering, not declared intent.",
      "suggestion": "Move `zod` to dependencies (the service imports it directly). Since the service is run from TypeScript source in prod, either move `tsx` to dependencies as well or compile to JS at build time and run node on the output. Don't rely on NODE_ENV ordering to keep dev tools resolvable in prod.",
      "confidence": 0.7,
      "faang_parallel": "Netflix's and AWS Lambda's Node runtimes both prune devDependencies in production layers by default — code that imports a dev-classified module works in CI and dies on the first cold start. The dependencies/devDependencies split is a contract about *runtime reachability*, not about who-installed-it-when.",
      "metaphor": "It's a load-bearing wall labeled 'temporary scaffolding.' It holds today only because nobody has removed the scaffolding yet — and removal is someone else's routine cleanup, not a decision.",
      "teachable_moment": "Correct-by-coincidence is a latent bug with a timer on it. The rule: if an import survives to runtime, it's a `dependency`, full stop — independent of whether the current Dockerfile happens to install dev deps. Make the artifact correct under the *strictest* install you might ever run."
    },
    {
      "id": "F3",
      "title": "`--no-frozen-lockfile` makes the deployable artifact non-reproducible",
      "severity": "MEDIUM",
      "category": "supply-chain",
      "file": "packages/services/shadow-audit/Dockerfile",
      "description": "Both installs use `pnpm install --no-frozen-lockfile`. The build copies no root lockfile, so transitive versions are re-resolved at build time rather than pinned to the committed lockfile. Two builds of the same SHA can pull different (newer, potentially breaking or compromised) transitive dependencies — the opposite of what you want for a money/ops-adjacent deployable building.",
      "suggestion": "Copy the workspace lockfile into the build context and use a frozen install (`pnpm install --frozen-lockfile`), or stage via `pnpm deploy --filter` from the root so the lockfile is honored. Reserve `--no-frozen-lockfile` for local dev only.",
      "confidence": 0.65,
      "faang_parallel": "Post-SolarWinds and post-event-stream, every major supply-chain hardening guide (SLSA, Google's Binary Authorization) treats 'same source SHA → same artifact' as a baseline invariant. A floating lockfile is the seam those attacks walk through.",
      "metaphor": "It's a recipe that says 'flour' without a brand or amount — fine until the day the supplier swaps in something that looks identical and quietly ruins the loaf.",
      "teachable_moment": "`--no-frozen-lockfile` exists to *let resolution drift* — which is precisely what a deployable must forbid. The lockfile is the boundary between 'I tested these bytes' and 'I tested some bytes.' For a money/ops path, only the former is shippable."
    },
    {
      "id": "F4",
      "title": "In-memory event store / nonce store / rate limiter are non-durable and per-replica",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/services/shadow-audit/src/server.ts",
      "description": "buildAuditApp wires InMemoryEventStore, InMemoryNonceStore, and FixedWindowRateLimiter. Audit run records written via recordRun vanish on every restart/redeploy (no durable audit trail), and the rate limiter + nonce store are per-process: if Railway runs more than one replica, rate limiting is trivially bypassed by hitting different instances and SIWE nonce replay protection (when V2 lands) is unsound. The MVP is presumably single-instance, but nothing in railway.toml pins replicas=1.",
      "suggestion": "Document and enforce single-instance (or move event store / rate limiter / nonce store to a shared backend — Redis/DB) before scaling. At minimum add a comment + a deploy constraint that replicas must be 1 while these are in-memory, and note that audit run records are ephemeral.",
      "confidence": 0.7,
      "faang_parallel": "Every team that has shipped a per-process rate limiter has relearned Stripe's and Cloudflare's lesson: the limiter is only as strong as its weakest shared-state assumption. The moment the autoscaler adds replica #2, an in-memory limiter becomes a suggestion, and a nonce store becomes a replay window.",
      "metaphor": "It's a bouncer with a guest list memorized in his head — perfect, until the club opens a second door with a second bouncer who's never heard your name and waves you in.",
      "teachable_moment": "In-memory infra isn't wrong for an MVP — it's wrong *silently*. The fix isn't necessarily Redis today; it's making the single-replica assumption a load-bearing, written constraint (replicas=1 in railway.toml + a comment) so the day someone scales it, they trip a tripwire instead of a security hole.",
      "connection": "This is the inverse of the in-memory durability story you've hit before: ephemeral audit-run records mean the audit trail — the very thing this building exists to produce — doesn't survive a redeploy. Worth a one-line note in DEPLOY.md."
    },
    {
      "id": "F5",
      "title": "Rate-limit key uses the client-spoofable leftmost X-Forwarded-For hop",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/services/shadow-audit/src/server.ts",
      "description": "`clientKey: (xff) => (xff?.split(',')[0]?.trim() || 'unknown')` takes the FIRST X-Forwarded-For entry, which is the value the original client supplied and can freely forge. An attacker can rotate the leftmost XFF to mint unlimited rate-limit buckets (bypass) or set it to a victim value (collateral throttling). The trustworthy client IP behind a proxy is normally the rightmost hop the proxy appended, not [0].",
      "suggestion": "Key the rate limiter on the proxy-appended (rightmost trusted) hop, or on Railway's documented real-client header, after validating the number of expected proxy hops. Fall back to a connection-level identifier rather than an unvalidated client-supplied header.",
      "confidence": 0.5,
      "faang_parallel": "This is the exact bug class behind a long line of CDN/WAF bypasses; AWS WAF and Cloudflare both warn that XFF is attacker-controlled and that only the hop *your* trusted proxy appends is meaningful. Express's own `trust proxy` setting exists precisely to compute the right index.",
      "metaphor": "You're checking ID by reading the name the visitor wrote on their own stick-on badge. The trustworthy name is the one the security desk wrote, which sits at the *end* of the chain, not the start.",
      "teachable_moment": "XFF is append-only and the leftmost entry is the least trustworthy value in the whole request. Verify against Railway's actual proxy behavior before fixing — the *right* index depends on how many trusted hops sit in front of you, which is why this needs grounding, not just a `[0]`→`[len-1]` flip."
    },
    {
      "id": "F6",
      "title": "Protocol package is never built in the Dockerfile — runtime import may fail if it exports dist/",
      "severity": "MEDIUM",
      "category": "build/deploy",
      "file": "packages/services/shadow-audit/Dockerfile",
      "description": "The Dockerfile comment states protocol runs 'from source' via tsx, and only @freeside/adapters is built. But the service imports `@freeside/shadow-audit-protocol` through its package exports. If that package's package.json `main`/`exports` resolve to `dist/` (the common pattern, as adapters does), the import resolves to an unbuilt path and fails at boot — something the vitest suite (which resolves source directly) will not reproduce.",
      "suggestion": "Confirm packages/protocol/shadow-audit's exports point at TS source (or add a `pnpm build` step for it in the Dockerfile). Add a container smoke test (`docker run … node -e 'import the server'` or hit /healthz then /v1/audit) to CI so module-resolution drift is caught.",
      "confidence": 0.4,
      "metaphor": "The test harness reads the manuscript; production reads the printed book. If the printer was never run, the shelf is empty — and only the reader-of-the-book finds out.",
      "teachable_moment": "Verify before teaching: read packages/protocol/shadow-audit/package.json `exports`/`main`. If it points at `src/*.ts`, this is a non-issue; if it points at `dist/`, F6 is real and shares F1's root cause. Either way, a single container smoke test in CI settles F1, F2, F6, and F10 at once — it exercises the *resolved* artifact, not the source the tests resolve around."
    },
    {
      "id": "F7",
      "title": "`uncertain` (stale-snapshot signal) dropped from the JSON API response",
      "severity": "LOW",
      "category": "api-contract",
      "file": "packages/services/shadow-audit/src/http/audit-router.ts",
      "description": "The GET handler now returns `result.output` verbatim and explicitly drops the previously-included `uncertain` flag (moved to the HTML /view route only). This is the correct fix for contract parity (the dashboard strict-decodes with onExcessProperty: error), but it means the stale-snapshot confidence signal is no longer machine-readable by the dashboard or any JSON consumer — exactly the population the audit most wants to warn. The comment acknowledges this requires a coordinated protocol change.",
      "suggestion": "Track a follow-up to fold staleness/uncertainty into the protocol AuditOutputSchema (e.g. an `uncertain` or `snapshot_staleness` field) on both producer and consumer, so the signal returns to the JSON seam rather than living only in HTML.",
      "confidence": 0.5,
      "teachable_moment": "Dropping the excess field was the right call under a strict decoder — but notice the asymmetry it creates: the signal now lives only where humans look (HTML) and is invisible where machines decide (JSON). Schema evolution is the proper home for it; the fix here correctly trades a silent-reject bug for a tracked capability gap.",
      "connection": "This is the deliberate, documented twin of the regression F11 pins. Make sure a beads follow-up exists so the staleness signal returns to the JSON seam in a coordinated producer+consumer schema bump."
    },
    {
      "id": "F8",
      "title": "Block-at-date binary search is ~log2(head) sequential RPC calls per audit with no caching",
      "severity": "LOW",
      "category": "performance",
      "file": "packages/services/shadow-audit/src/block-time-resolver.ts",
      "description": "Each `blockAtOrBefore` performs a serial binary search over [0, head] (~24-25 `eth_getBlockByNumber` round-trips for a 20M-block chain), each with its own 15s timeout, and the result is recomputed on every request — including repeated audits for the same chain+snapshot_date. A single GET can therefore take several seconds purely on block resolution, and there is no overall request deadline bounding the cumulative RPC latency.",
      "suggestion": "Memoize resolved (chain, snapshot_date) → block (snapshot_date→block is immutable once final), and/or cache head briefly. Consider an overall deadline across the search so a slow RPC can't stack ~25×15s.",
      "confidence": 0.6,
      "faang_parallel": "This is the canonical 'cache the immutable, bound the mutable' optimization. A finalized (chain, date)→block mapping is as immutable as a content-addressed blob — exactly the class of value CDNs cache forever. The unbounded cumulative timeout is the same tail-latency trap Google's deadline-propagation design exists to prevent.",
      "metaphor": "You're re-deriving last Tuesday's closing stock price from scratch on every page load, when last Tuesday's close has been fixed forever — and you've allowed yourself 25 separate 15-second naps while doing it.",
      "teachable_moment": "Two distinct levers hide here: *memoization* (the answer can't change once the block is final) and a *deadline* (worst case is 25×15s = unacceptable). The deadline is the more important one — caching speeds the common path, but only a request-level budget caps the pathological one."
    },
    {
      "id": "F9",
      "title": "unhandledRejection handler logs but never exits — can mask a wedged process",
      "severity": "LOW",
      "category": "reliability",
      "file": "packages/services/shadow-audit/bin/http.ts",
      "description": "`process.on('unhandledRejection', (reason) => console.error(...))` swallows all unhandled rejections and keeps the process alive. For a server this avoids crash-loops, but it also defeats Railway's restartPolicy ON_FAILURE for genuinely fatal async faults and can leave the process serving while in a broken internal state, which is at odds with the file's 'fail loud, never half-wired' stance.",
      "suggestion": "Log and then exit non-zero on unhandledRejection (let Railway restart per restartPolicy), or scope the handler to known-benign cases. Pair with the /healthz check so a wedged process is recycled.",
      "confidence": 0.4,
      "metaphor": "It's a smoke detector wired to silence itself instead of sounding — the fire still burns, you just stop hearing about it.",
      "teachable_moment": "Node's own deprecation guidance moved toward crashing on unhandledRejection for exactly this reason: a swallowed rejection leaves you in undefined state with the lights still on. The healthiest pattern is log-then-exit-nonzero and let the orchestrator (Railway's restartPolicy) recycle a clean process — consistent with the file's stated fail-loud posture."
    },
    {
      "id": "F10",
      "title": "Floating base image and pnpm version without packageManager pin",
      "severity": "LOW",
      "category": "supply-chain",
      "file": "packages/services/shadow-audit/Dockerfile",
      "description": "`FROM node:22-alpine` is an unpinned floating tag and `corepack prepare pnpm@9.15.9` pins pnpm in the Dockerfile only (no `packageManager` field shown in package.json to keep local/CI/Docker in lockstep). Combined with --no-frozen-lockfile (F3), the deployed image is several steps removed from a reproducible artifact.",
      "suggestion": "Pin the base image by digest (`node:22-alpine@sha256:…`) and add a `packageManager: pnpm@9.15.9` field to package.json so corepack/local/CI agree on the pnpm version.",
      "confidence": 0.5,
      "faang_parallel": "Google's distroless and digest-pinned base-image discipline exists because a moving tag like `:22-alpine` is a dependency you didn't know you had — the upstream maintainer can change your runtime out from under a rebuild.",
      "teachable_moment": "F3 and F10 are the same principle at two layers: pin the lockfile (transitive deps) *and* pin the base image by digest (OS/runtime deps) *and* pin the package manager. Reproducibility is only as strong as the least-pinned input in the chain."
    },
    {
      "id": "F11",
      "title": "Contract-parity test strict-parses the live GET body against the protocol schema",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/services/shadow-audit/src/__tests__/server.test.ts",
      "description": "The GET happy-path test asserts `AuditOutputSchema.parse(body)` does not throw — reproducing exactly the dashboard client's strict decode (onExcessProperty: error) against the producer's real output. This directly pins the prior regression (subset fields + excess top-level `uncertain` made the dashboard reject every 200 and render empty) so the producer/consumer contract drift cannot silently return. Testing the consumer's decode on the producer's output is precisely the right integration seam.",
      "suggestion": "Keep this as the canonical seam test; consider extending it to the refusal/error response shapes so those are contract-pinned too.",
      "confidence": 0.95,
      "faang_parallel": "This is consumer-driven contract testing (Pact, and the philosophy behind Google's API linting against real responses): the producer's test asserts the *consumer's* decoder accepts the *producer's* real output. It closes the exact gap that 'both sides have a schema' leaves open — two schemas that have quietly diverged.",
      "metaphor": "Instead of measuring the key and the lock separately and hoping, you actually turn the real key in the real lock. The test *is* the door opening.",
      "teachable_moment": "The deepest lesson of this PR: a regression where every 200 rendered empty was invisible to status-code assertions and to each side's own schema test — it lived in the *seam*. The only test that catches seam drift is one that runs the consumer's strict decoder over the producer's actual bytes. Extend it to error/refusal shapes and you've pinned the whole contract surface.",
      "connection": "This finding and F7 are the same lesson from both sides: F11 pins the contract the dashboard depends on; F7 is the disciplined acknowledgment that re-adding a field must be a coordinated producer+consumer schema change, not a unilateral one."
    },
    {
      "id": "F12",
      "title": "Fail-loud config validation, constant-time key compare, and before-genesis refuse invariant",
      "severity": "PRAISE",
      "category": "robustness",
      "file": "packages/services/shadow-audit/src/server.ts",
      "description": "Strong defensive posture throughout: configFromEnv rejects AUDIT_K < 1 / NaN (preventing silent k-anonymity disable), bin/http.ts zod-validates COLLECTION_REGISTRY and boot-fails when any registry chain lacks an RPC_URL, a single CONFIRMATIONS source feeds both the sonar reorg guard and the 'current' block (no split-knob drift), the API-key gate uses length-checked timingSafeEqual, and the -1 'before-genesis' sentinel is threaded end-to-end with a refuse-rather-than-reconstruct-at-block-0 invariant covered by unit tests in ownership-source.test.ts and block-time-resolver.test.ts. These collectively uphold the 'refuse, never silent-wrong' discipline for the money/ops path.",
      "suggestion": "No change required. Ensure the live-correctness gate in DEPLOY.md (collection-id + block-at-date spot check) is actually run before the dashboard is pointed at production, since the suite proves algorithm/assembly but not live values.",
      "confidence": 0.9,
      "faang_parallel": "The timingSafeEqual + length-check is textbook (the lesson from the Keyczar/early-Rails timing-leak disclosures), and the boot-fail-on-missing-RPC_URL is Google SRE's 'fail fast at startup, not at the first request' configuration-validation principle. A single CONFIRMATIONS source feeding both consumers is the 'single source of truth' guard against split-knob drift.",
      "metaphor": "A good audit instrument is built like a bank vault that refuses to close if a single bolt is misaligned — it would rather not lock at all than lock while looking secure. The -1 before-genesis refusal is that vault declining to certify a balance for a date before the account existed.",
      "teachable_moment": "Notice what 'refuse, never silent-wrong' buys: every one of these guards converts a *silent incorrect answer* (k-anonymity off, reconstruction at block 0, a split confirmations value) into a *loud refusal or boot failure*. For an audit building, a wrong answer that looks right is the only unrecoverable failure — and this code systematically designs it out. The remaining gap is honest: the suite proves the algorithm, not live values, so the DEPLOY.md spot-check is load-bearing, not optional."
    }
  ]
}
```


## Callouts

- **[Test Coverage] F11 — the seam test is the headline.** Asserting the dashboard's strict `AuditOutputSchema.parse` over the producer's *real* GET body is exactly the test that the prior empty-render regression needed and didn't have. Status codes were green; each side's schema was green; only the seam was red. This test makes that class of drift impossible to reintroduce silently. Extend it to the refusal/error shapes and the contract surface is fully pinned.

- **[Operational] F12 — refuse-never-silent, threaded end-to-end.** The `-1` before-genesis sentinel carried from resolver through ownership-source with unit coverage, `timingSafeEqual` with a length pre-check, boot-fail on any registry chain missing `RPC_URL`, and a *single* `CONFIRMATIONS` feeding both the reorg guard and "current" block — these aren't isolated good habits, they're a coherent doctrine for a money/ops path. The honesty of the DEPLOY.md note (suite proves algorithm, not live values) is itself a mark of senior judgment.

- **[Quality] One structural observation to carry forward:** the five build/supply-chain findings (F1, F2, F3, F6, F10) are not five problems — they are one. The vitest suite resolves TypeScript source directly, so it structurally *cannot* exercise the resolved, pruned, lockfile-pinned artifact the container actually runs. A single CI step that builds this Dockerfile and curls `/healthz` → `/v1/audit` would convert all five latent risks into a loud pre-merge signal. That one gate is the highest-leverage follow-up in this PR.

---

PR #395, opened 2026-06-29
feat(shadow-audit): migration-delta wedge — Comparison View + shareable export

zkSoju:
## Shadow-audit migration-delta wedge — Comparison View + shareable export

Ships the **validated wedge**: the migration-delta Comparison View (promotion / demotion / no-change) consumed into the authed `AuditOutput`, plus a shareable CSV+JSON export. The buying event is the *delta* — what changes when you migrate access — so the wedge consumes the previously **built-but-unconsumed** `diffShadow`/`DiscrepancyReport` and surfaces it.

### What's in it
- **Consume diffShadow** (keystone) — `diffShadow` + `DiscrepancyReport` exported from the protocol index and wired into the authed `AuditOutput` Comparison View, with methodology provenance (`rule_id`, `snapshot_block`, sources). The anon response stays byte-stable; methodology travels with the AUTHED delta only.
- **Shareable export** — CSV+JSON of the Comparison View (`comparison-export.ts`).
- **Ownership projection spine** — event-fed ownership projection + `ProjectionOwnershipSource` (the L3→L2 rung), `ownership.changed` subscriber, real chain ownership adapter, deployable server entry, `GET /v1/audit` serving the protocol `AuditOutput`.
- **AccessDecisionPort** — the L2-composition lens, reforged on the WHO×WHAT ladder.
- **Hardening** — cutover gate fails CLOSED on erc1155 (projection can't model per-token erc1155); differential proof projection ⇄ real sonar `reconstructOwnership`; multiple FAGAN + bridgebuilder review rounds addressed.

### Keystone fix in this PR
`discrepancy.ts` (the diffShadow keystone) + its test were **untracked** across the 5 wedge commits — the committed `index.ts` exported from `./discrepancy.js` while the source file lived only in the working tree, so the committed branch did not compile on a fresh checkout. Local typecheck/tests and FAGAN stayed green because they reviewed the working tree. Fixed in `0503fe81` (stages the 2 files only).

### Verification
- `@freeside/shadow-audit-protocol`: `tsc --noEmit` exit 0, **48/48** tests across 7 files.
- `@freeside/shadow-audit` (services): `comparison-export` **6/6**.

### Scope notes (honest)
- Branch also bundles a few adjacent non-shadow-audit commits: `operator-dash` beacon SHAPE validation, and two `cheval/cursor` fixes (pgkill orphan-on-timeout, governed-council convergence).
- All paths are **platform/shared** — no network-zone paths, so the cross-domain CI rule does not apply.
- Beads: closes `…cbox.1` (consume diffShadow) and `…cbox.2` (export). `…cbox.3` (role-by-role cohort breakdown) remains open as future P3.

### Why draft
`main`'s required **Unit Tests** gate is currently frozen (`arrakis-0q6r` — opossum `./lib/circuit` load failure under npm-vs-pnpm in CI; tracked separately via PR #310). This is a draft to avoid implying a green/mergeable state that the frozen gate can't yet confirm.


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#395. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "PERF-1",
      "title": "Projection re-sorts the entire event list on every apply() and retains all events in memory unbounded",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "packages/services/shadow-audit/src/projection-ownership-source.ts",
      "description": "makeOwnershipProjection().apply() does list.push(change) followed by list.sort() on every single event. Ingesting n events from the live ownership.changed stream is therefore O(n^2 log n). Additionally `changes` and `seen` grow without bound — the read-model keeps every Transfer for the contract's whole history in memory and re-folds from genesis on every balancesAt/resolveSnapshotBlock. This is the 'spine' the audit is meant to read continuously; a popular collection has hundreds of thousands of Transfers. Acceptable for an MVP single-replica in-memory model, but it will not scale and there is no compaction/snapshot or sorted-insert.",
      "suggestion": "The stream arrives in block order — replace push+full-sort with an ordered insert (or push and sort lazily once on first read, tracking a dirty flag). Cap retention with periodic compaction to a balance snapshot + tail, or persist the read-model so the fold isn't O(n) per query. At minimum mark the scale ceiling with a loa:shortcut upgrade trigger.",
      "confidence": 0.8
    },
    {
      "id": "REPRO-1",
      "title": "inputs_hash does not bind the current-balance basis the delta settles against — artifact is not reproducible from its own provenance",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/comparison-export.ts",
      "description": "The exported migration-delta artifact's whole value prop is being a verifiable record (run_id + inputs_hash + methodology). But the promotion/demotion delta settles against the CURRENT on-chain head (currentBalances), which is neither pinned nor folded into inputs_hash (acknowledged as NEW-1 in the docstring). Two exports produced at different times can carry the SAME inputs_hash yet DIFFERENT deltas — so a recipient cannot actually verify the delta is reproducible from the stated provenance. The 'self-describing, verifiable' claim is stronger than what the hash guarantees.",
      "suggestion": "Surface current_block from the OwnershipSource, include it in the methodology, and fold it into inputs_hash so the delta basis is bound. Until then, soften the 'verifiable record' language in the docstring/CSV preamble to state explicitly that the delta basis is unpinned.",
      "confidence": 0.7
    },
    {
      "id": "INPUT-1",
      "title": "BigInt(a.value) is unguarded — a signed-but-malformed value throws inside the projection handler",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/ownership-projection-subscriber.ts",
      "description": "ownershipActivityToChange computes `amount: a.value === null || a.value === '0' ? 1n : BigInt(a.value)`. The ed25519 envelope verification upstream proves authenticity, NOT that `value` is a valid numeric string. A signed event whose `value` is non-numeric (or a tokenId encoded as a non-decimal string) makes BigInt() throw synchronously inside applyOwnershipActivity. Depending on the injected subscriber's error handling (not in this diff), one bad event could halt projection ingestion or crash the process. This is a trust-boundary that treats authenticity as if it implied validity.",
      "suggestion": "Wrap the BigInt parse defensively: validate `/^[0-9]+$/` and on failure return null (skip, like the other malformed-event guards) or refuse loudly via a typed error, rather than letting an unguarded throw escape the mapping. This composes with the HIGH-3 per-standard value-semantics contract test already noted as open.",
      "confidence": 0.65
    },
    {
      "id": "DUP-1",
      "title": "Dead ROLE_THRESHOLDS retained and re-exported with a divergent shape — two sources of truth",
      "severity": "MEDIUM",
      "file": "themes/sietch/src/services/badge.ts",
      "description": "After routing checkRoleUpgrades through badgeBasedPolicy, badge.ts's ROLE_THRESHOLDS is dead (the live thresholds live in ports/access-decision.ts) yet is still exported and re-exported via services/index.ts, and its shape differs (trusted.helperBadge vs the port's trusted.badgeCount). The self-documenting comment correctly warns that editing the dead copy is a silent no-op, but leaving a second, differently-shaped threshold constant in place invites exactly the drift the comment describes. Flagged honestly as FAGAN MEDIUM-1; re-affirming it is a real hazard, not a closed one.",
      "suggestion": "Single-source it: have services/index.ts re-export the port's ROLE_THRESHOLDS (adapting shape once), or delete the dead constant and fix the re-export. Don't ship a live system with two threshold tables, one of which is a no-op.",
      "confidence": 0.7
    },
    {
      "id": "API-1",
      "title": "GET /v1/audit JSON now drops the `uncertain` stale-snapshot signal entirely",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/http/audit-router.ts",
      "description": "Returning result.output verbatim correctly fixes the strict-decode integration break (good). But it also removes `uncertain` from the JSON path entirely — the stale-snapshot freshness signal is now only on the HTML /view route. API consumers (the dashboard) can no longer tell whether an audit was computed against a stale role snapshot. The fix trades a silently-broken integration for a silently-missing freshness signal; the comment acknowledges re-adding it needs a protocol change on both sides.",
      "suggestion": "Track re-introducing freshness as a first-class optional field on AuditOutputSchema (so it survives strict decode) rather than leaving the API blind to staleness. At minimum, file the follow-up so the signal loss is tracked, not dropped.",
      "confidence": 0.6
    },
    {
      "id": "UNCONSUMED-1",
      "title": "ownership-projection-runtime.ts is a DRAFT, deployed-but-unconsumed; 'durable replay' is only a comment",
      "severity": "MEDIUM",
      "category": "design",
      "file": "packages/services/shadow-audit/src/ownership-projection-runtime.ts",
      "description": "The runtime's start() documents 'DURABLE replay rebuilds the projection from the full worldline' but the code only calls opts.subscribe — durability/replay-from-genesis depends entirely on the injected subscriber, which nothing in production constructs (bin/http.ts wires makeSonarOwnershipSource, not this runtime). The shadow-then-cutover logic and its fail-closed erc1155/unknown branches are well-tested with fakes, but ship unexercised against the real @freeside/events subscriber and a real ownership.changed stream that does not yet exist. This is the cluster's known 'ships before a live consumer' pattern.",
      "suggestion": "Keep the DRAFT marker prominent and ensure no production path can accidentally route through this runtime before sonar-api publishes ownership.changed and a contract test pins the value semantics. Consider gating its export behind an explicit 'experimental' barrel so it can't be wired by mistake.",
      "confidence": 0.55
    },
    {
      "id": "CSV-1",
      "title": "RFC-4180 escape does not quote a field containing a lone CR",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/comparison-export.ts",
      "description": "esc() quotes on /[\",\\n]/ but not a bare \\r. deFormula only neutralizes a LEADING \\r. A community name like 'foo\\rbar' (no leading special char) passes through unquoted, which some CSV parsers interpret as a row break. Only `community` is attacker/free-text-controlled here, so blast radius is small, but it's a parseability edge the RFC-4180 claim doesn't fully cover.",
      "suggestion": "Add \\r to the esc() regex: /[\",\\r\\n]/.",
      "confidence": 0.6
    },
    {
      "id": "RPC-1",
      "title": "JSON-RPC caller has no retry/backoff; one transient error aborts the whole binary search",
      "severity": "LOW",
      "category": "reliability",
      "file": "packages/services/shadow-audit/src/block-time-resolver.ts",
      "description": "blockAtOrBefore issues ~log2(head) (~25 for Ethereum) sequential eth_getBlockByNumber calls. defaultRpcCall throws on the first non-ok HTTP or RPC error with no retry/backoff. Public RPC endpoints routinely 429/503; a single transient failure aborts snapshot resolution and refuses the whole audit. Fail-loud is the chosen posture, but a transient blip shouldn't fail an entire audit.",
      "suggestion": "Add a small bounded retry with backoff on retryable statuses (429/5xx) in defaultRpcCall, or document the expectation of a reliable dedicated RPC in DEPLOY.md's live-correctness gate.",
      "confidence": 0.5
    },
    {
      "id": "NPE-1",
      "title": "methodology object is built unconditionally, dereferencing snapshot.captured_at even on records-less paths",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/audit-service.ts",
      "description": "`const methodology = { ..., role_snapshot_at: snapshot.captured_at, ... }` is evaluated unconditionally, then attached only when records exist. Tests confirm snapshot is defined on the operated-community anon path today, so this does not currently throw — but building a value you only sometimes attach, by dereferencing a variable whose presence is invariant-dependent, is fragile against future mode changes.",
      "suggestion": "Move the methodology construction inside the `if (req.includeRecords)` / records branch so it's only built when it will be attached and only when snapshot is guaranteed present.",
      "confidence": 0.4
    },
    {
      "id": "SHAPE-1",
      "title": "isBeaconShape treats a bare {\"is\": ...} body as a valid live beacon",
      "severity": "LOW",
      "category": "correctness",
      "file": "tools/operator-dash/beacon-classify.ts",
      "description": "isBeaconShape satisfies BOTH hasSchema and hasIdentity from the single key `is` (`'is' in o` appears on both sides). A 200 JSON body that merely contains an `is` key — not actually a BeaconV3 — classifies as 'live', re-opening a narrower version of the masked-200 false positive the module exists to close.",
      "suggestion": "Require a stronger BeaconV3 signal for the `is` branch (e.g. `is` is an object with `one_liner`/`scope`), or require schema_version explicitly when identity comes only from `is`.",
      "confidence": 0.45
    },
    {
      "id": "BUILD-1",
      "title": "Dockerfile is single-stage despite `AS base`, and the final install ships devDependencies",
      "severity": "LOW",
      "file": "packages/services/shadow-audit/Dockerfile",
      "description": "`FROM node:22.12-alpine AS base` implies a multi-stage build that never materializes (one stage only). The service install is `pnpm install --no-frozen-lockfile` with no `--prod`, so vitest/typescript and the full build toolchain ship in the runtime image. Functionally fine (tsx runs from source) but a larger, less hermetic artifact than intended.",
      "suggestion": "Drop the unused `AS base`, or add a real runtime stage that copies only built dist + production deps. Consider `--prod` for the final service install.",
      "confidence": 0.5
    },
    {
      "id": "REFRAME-1",
      "title": "comparison.members duplicates records data — `change` is the only new field and is a pure function of band",
      "severity": "REFRAME",
      "category": "design",
      "file": "packages/protocol/shadow-audit/src/discrepancy.ts",
      "description": "MemberDiscrepancy carries (wallet, community, holds_role, qualifies, band, change), all of which except `change` already appear on the AccessDecisionRecord, and `change = changeFromBand(band)` is deterministic from band. The authed response therefore ships the per-member set twice. This is intentional (the product's 'Comparison View'), but it's worth naming that comparison adds zero information not derivable client-side from records — relevant if payload size or the byte-stable-contract surface matters later.",
      "suggestion": "Consider whether the authed response needs both records and comparison.members, or whether comparison can carry only the aggregate + let the client derive per-member change from band. No action required if the product surface wants the materialized view.",
      "confidence": 0.4
    },
    {
      "id": "PRAISE-1",
      "title": "Differential test runs the REAL reconstructOwnership over 40 generated valid streams — closes the fixture-tautology gap",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/services/shadow-audit/src/__tests__/projection-sonar-differential.test.ts",
      "description": "This is the right kind of proof: it folds the production @freeside/adapters/sonar reconstructOwnership and the projection over the same Transfer stream through two independent paths and asserts agreement across a seeded generator (40 streams), an intermediate snapshot block, and a shuffled order. It also pins the known model boundary (count vs per-tokenId) by documenting that sonar REFUSES the double-mint the count-model cannot see, instead of hiding it. Directly addresses the prior fixture-tautology overclaim pattern.",
      "confidence": 0.9
    },
    {
      "id": "PRAISE-2",
      "title": "Security hygiene: timing-safe key compare, CSV formula-injection guard, fail-closed auth and exports, rightmost-XFF keying",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/services/shadow-audit/src/server.ts",
      "description": "Constant-time X-API-Key comparison via timingSafeEqual with a length pre-check; OWASP CSV formula-injection neutralization on the only attacker-controlled field; fail-closed V2 POST auth (always 401) so named records can never be half-served; exports throw on anon audits rather than fabricating a per-member delta; rate-limit keyed on the un-forgeable rightmost XFF hop with documented single-proxy assumption; AUDIT_K rejected when <1 to prevent silent k-anonymity disablement. Consistently correct trust-boundary posture.",
      "confidence": 0.9
    },
    {
      "id": "PRAISE-3",
      "title": "Honest keystone framing — the wiring cross-check is explicitly NOT independent corroboration",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/services/shadow-audit/src/__tests__/audit-comparison-keystone.test.ts",
      "description": "The keystone gate states plainly that the comparison-vs-aggregate cross-check shares qualifies(curBal) and therefore catches band-MAPPING drift, not a shared-input bug, and that true ground-truth requires settling against the incumbent's live grants (left open). It also pins the current-basis-vs-evidence_block distinction and the byte-stable anon contract. This is calibrated, non-overclaiming verification — the correct response to the prior fixture-tautology feedback.",
      "confidence": 0.85
    }
  ]
}
```


## Callouts

Issue #415, opened 2026-07-01
Epic: Fulfillment Orchestrator — automate order→kitchen→ship gates across Freeside


zkSoju:
## Context (trigger)

Operator session **2026-07-01**: sonar-api Tier-2 (#62/#63) shipped via manual choreography — branch cut, commit, PR [#117](https://github.com/0xHoneyJar/sonar-api/pull/117), Bridgebuilder review (~7min), triage HIGH findings, fix commit, CI watch, merge. **~45+ minutes of operator attention** for a well-scoped, pre-spec'd re-port with green unit tests.

Parallel threads confirm this is not a one-repo problem:
- **Ordering spine exists** (`packages/protocol/ordering`, `packages/services/ordering`) but **fulfillment still routes to operator** (`POST /v1/orders/:id/advance-ingredient`, GitHub issue fan-out, manual ENVIO_RESTART / belt redeploy).
- **Loa post-PR gates** (Bridgebuilder, Flatline, READY_FOR_HITL) are **prose-orchestrated** — findings posted to GitHub, consumed manually ([close-bridgebuilder-loop proposal](grimoires/loa/proposals/close-bridgebuilder-loop.md), loa #1036 DISS-003).
- **freeside-cli** has read/doctor verbs only — no `order`, `kitchen`, or `fulfill` surface for agents.
- **Dashboard** is a consumer of ordering-service but Sprint 2 orchestrator (automated ingredient fan-out) remains planned, not shipped.

The operator's ask: **design higher-level CLIs + agentic tooling that compose with the ordering system**, focus on **fulfilling orders + architectural decisions**, and **reduce operator-in-the-loop** without losing safety at irreversible seams.

---

## Diagnosis: gate design, not game design

This is **primarily a gate-orchestration problem**, with a **secondary consumption-gradient** lens from the wedge-build game.

| Lens | Question | Verdict |
|------|----------|---------|
| **Gate design** | Where does automation stop, what evidence is required to proceed, who may advance state? | **Primary.** Today most transitions are operator-jury or manual curl. Loa phases, CI, Bridgebuilder, and ordering ingredient probes are **parallel gate systems with no shared ledger**. |
[...]

PR #420, opened 2026-07-02
feat(platform/ordering): agent-facing service capabilities — reprobe, evidence, fail-closed [fulfillment-surface PR-A]

zkSoju:
## PR-A · fulfillment-surface cycle — ordering-service agent-facing capabilities (platform/ordering)

Implements Sprint 1 of the [Agent-First Fulfillment Surface](https://github.com/0xHoneyJar/loa-freeside/issues/415) cycle (epic #415, **reframed agent-first** per operator decision — no orchestrator daemon; the agent session bears the orchestration load; this PR gives the service the primitives that make that safe).

### What's in
| Capability | Design |
|---|---|
| `POST /v1/orders/:id/reprobe` | Synchronous on-demand probes; 10s/probe race-timeout, fan-out ≤3, 10s per-order cooldown → 429; probe failure = in-body `ambiguous` result, never 5xx; ambiguity never overwrites recorded truth (SDD D1) |
| `probe_meta` | Last raw probe truth per ingredient, typed in `@freeside/ordering-protocol`; one shared write path (process + reprobe); additive migration 002 |
| Advance audit hardening | Server-derived `token_label` + `evidence` (immutable probe_meta value-copy, `null` when never probed); untrusted `caller_note` ≤120; legacy dashboard body unchanged (SDD D3) |
| Fail-closed boot (FR-10a/b) | Deployed + tokenless → write routes never mounted; `/healthz` reports `write_routes`; reads stay open by design; full D8 matrix tested |
| Canonical projection (SDD D7) | ONE strict `toPublicOrder` for GET/advance/reprobe; `PublicOrderSchema` exported from ordering-protocol; redaction boundary tested |
| FR-1 + NFR-8c | Unknown-preset 400 + `available_presets`; token-rotation runbook |

### Evidence
- **101/101 tests** (21 files; 20+ new), `tsc --noEmit` clean; protocol package 15/15
- Domain firewall: `tools/check-beacon-domain.sh` → platform-only
- Review: 1 fix iteration (DEPLOY.md contract, `Object.hasOwn`, raced `world_slug`) — `grimoires/loa/a2a/sprint-1/{reviewer,engineer-feedback,auditor-sprint-feedback}.md` on this branch
- Audit: APPROVED, 0 CRITICAL/HIGH (3 LOW observations incl. pre-existing `contact_email` on unauthenticated GET — flagged for the multi-tenant threat model)

### Deploy notes
Additive migration (`RUN_MIGRATIONS=true` applies 001+002). After deploy, `/healthz` should show `write_routes: "token"` — that field is Sprint 2's entry-gate signal. Sibling PR: `feature/fulfillment-s2-network` (CLI verbs). Operator: recommend `/fagan` before merge (cross-model review was config-disabled in-cycle).

🗑️ Deleted files: none.


zkSoju (commented):
Verdict: **COMMENT** — no blocking findings, but FS-PR420-001 and FS-PR420-002 deserve resolution before this surface's contracts ossify.

## Summary

This PR builds a genuinely agent-grade service surface: fail-closed write posture, probe evidence with honest ambiguity semantics, and immutable audit trails — the security architecture here is better than most production services twice its age. The primary concerns are a TOCTOU window in the reprobe merge that undermines the CAS discipline applied elsewhere, and a PII field (contact_email) that the newly-canonical public projection formalizes onto an unauthenticated read path.

## Review

The most instructive thing about this diff is that its strongest and weakest moments are the *same question* answered at two levels of rigor: **who may write truth, and under what guard?** State transitions earned compare-and-swap; the reprobe ingredient merge got fresh-read-then-patch, which protects against yesterday's race but not the one living between the read and the write (FS-PR420-001). Audit evidence earned immutable value-copies verified byte-identical (FS-PR420-014); the public projection earned a denylist that redacts `inputs` but lets the same email walk out through `fulfillment` (FS-PR420-002). The lesson isn't that half the PR is careless — it's that the bar the best paths set is the bar every write and egress path deserves.

A second cluster worth naming: FS-PR420-005 through 008 are all variations of *silent degradation* — a cooldown that masks the real error, a map that leaks by design, a NaN that disarms the throttle, a "deployed" heuristic that fails open on unfamiliar hosts. Google's SRE literature is blunt about this: configuration and environmental assumptions are a leading outage class precisely because they fail quietly. A surface built for agents — which cannot shrug and read the logs — should prefer loud refusal over quiet wrongness everywhere, not just on the token path where this PR already gets it exactly right.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FS-PR420-001",
      "title": "Reprobe merge has a non-CAS write window that can downgrade a concurrent advance",
      "severity": "MEDIUM",
      "category": "concurrency",
      "file": "packages/services/ordering/src/community-onboarding-orchestrator.ts:370-385",
      "description": "reprobe() re-reads fresh state after probes ('Merge against a FRESH read') and then calls store.patchRecord() with the merged ingredients. The monotonic merge only protects against downgrades relative to the snapshot taken at the fresh read. If an advanceIngredient() lands between the fresh read and the patchRecord write, the patch overwrites ingredients (and potentially fulfillment) computed from the stale snapshot, downgrading e.g. a concurrent 'complete' back to 'pending'. The in-code comment and the IMP-003 test only cover the interleaving where the advance completes before the fresh read; the read→patch window remains a classic lost-update TOCTOU. patchRecord is not compare-and-swap protected the way the state transitions are.",
      "suggestion": "Either route the ingredient merge through a CAS/optimistic-concurrency primitive (e.g. patch conditioned on updated_at_unix or a version column, retrying the merge on conflict), or push the monotonic-merge logic into the store so the merge is computed against row state inside the same transaction/lock. At minimum, correct the comment so it doesn't claim full protection.",
      "confidence": 0.7,
      "faang_parallel": "Kubernetes solved exactly this with resourceVersion: every object update is a conditional write against etcd, and controllers are built to retry on 409 Conflict rather than trust a fresh read. DynamoDB's ConditionExpression exists for the same reason — a fresh read narrows the race window but never closes it.",
      "metaphor": "Photographing a shared whiteboard, walking away to compose your additions, then repainting the whole board from the photo — anything a colleague wrote while you were composing is erased, no matter how recent your photo was.",
      "teachable_moment": "Merge logic must execute where the data lives, not where the snapshot lives. A 'fresh read' is a smaller stale read; only a version-conditioned write or an in-transaction merge actually closes a lost-update window.",
      "connection": "The state transitions in this same file already earned CAS discipline — this finding is the gap between the standard the PR set for itself and one path that missed it."
    },
    {
      "id": "FS-PR420-002",
      "title": "Public projection leaks contact_email via fulfillment despite the stated redaction boundary",
      "severity": "MEDIUM",
      "category": "privacy",
      "file": "packages/services/ordering/src/projection.ts:12-29",
      "description": "toPublicOrder() redacts raw inputs (which include contact_email) but includes fulfillment, and CommunityOnboardingOutput carries contact_email (the orchestrator compares fulfillment?.contact_email when deciding to patch). GET /v1/orders/:id is unauthenticated ('reads stay open BY DESIGN'), so anyone with an order id can read the community manager's email. The exposure is pre-existing in the old GET handler, but this PR formalizes PublicOrderSchema as 'the redaction boundary' and pins it with tests, so the PII inclusion is now contract-canonical rather than incidental. operator_audit (token_label, caller_note) is similarly exposed to tokenless readers.",
      "suggestion": "Strip or mask contact_email from the public fulfillment projection (e.g. project fulfillment without contact fields, or add a redacted variant in the protocol schema). Decide explicitly whether operator_audit belongs on the unauthenticated read surface and document/redact accordingly.",
      "confidence": 0.7,
      "faang_parallel": "Meta's Ent framework attaches privacy rules to the schema itself so every egress path is viewer-checked — because they learned that hand-maintained projection functions drift from the data they project. Denylist redaction is the pattern that produced years of GraphQL over-fetch leaks.",
      "metaphor": "Shredding the letter but filing the envelope — the return address you meant to hide is written on both.",
      "teachable_moment": "A redaction boundary should be defined by what may leave (allowlist projection), not by what gets removed (denylist stripping). Denylists silently fail open every time a field is added or a nested object carries a copy.",
      "decision_trail": "Document in the protocol package whether operator_audit is intentionally public — future agents extending PublicOrderSchema need to know if tokenless visibility was a choice or an accident.",
      "connection": "Pinning the schema with tests is the right move (see FS-PR420-012's posture tests) — which is precisely why the leak should be fixed now, before the tests make it contract."
    },
    {
      "id": "FS-PR420-003",
      "title": "Unrelated drizzle-orm 0.33.0 → 0.45.2 bump rides in via lockfile only",
      "severity": "MEDIUM",
      "category": "dependencies",
      "file": "packages/services/ordering/pnpm-lock.yaml:1443",
      "description": "The lockfile moves drizzle-orm from 0.33.0 to 0.45.2 (a 12-minor jump on a 0.x package, i.e. potentially breaking per semver) for the @freeside/core file: dependency, with no corresponding package.json change in this diff and no relation to the PR's feature scope. Peer-dependency requirements also tightened (@electric-sql/pglite >=0.2.0, @libsql/client >=0.10.0, etc.). This changes core's DB layer behavior without dedicated review or test attention in this PR.",
      "suggestion": "Split the dependency bump into its own PR with a changelog review of drizzle-orm 0.34–0.45 breaking changes, or pin the previous resolution in this PR. If the manifest range permits silent drift across 0.x minors, tighten it.",
      "confidence": 0.6,
      "faang_parallel": "Amazon's Brazil build system pins entire version-sets so a dependency can never move as a side effect of unrelated work — every version change is its own reviewed event. Google's monorepo one-version rule exists for the same reason: invisible dependency motion is how 'nothing changed' outages happen.",
      "metaphor": "Swapping the delivery truck's engine while it's being loaded with unrelated cargo — nobody test-drove the new engine because the work order said 'load boxes'.",
      "teachable_moment": "On 0.x packages semver makes no compatibility promise, so a lockfile diff is a behavior diff and deserves the same review attention as source. If a range allows 12 minors of silent drift, the range is the bug."
    },
    {
      "id": "FS-PR420-004",
      "title": "Bearer token compared with non-constant-time string equality",
      "severity": "LOW",
      "category": "security",
      "file": "packages/services/ordering/src/intake.ts:170-173",
      "description": "requireToken() compares the Authorization header to `Bearer ${deps.serviceToken}` with ===. The pattern is pre-existing (the old advance-ingredient check did the same) but the refactor now shares it across all write routes. Non-constant-time comparison of a secret is a timing side-channel; practical exploitability over the network is low but the fix is one line.",
      "suggestion": "Use crypto.timingSafeEqual over equal-length buffers (hash both sides first to normalize length) in requireToken().",
      "confidence": 0.5,
      "faang_parallel": "Google's Keyczar library shipped exactly this bug in 2009 — byte-by-byte HMAC comparison — and the resulting timing-attack disclosure is now the canonical citation for why crypto libraries ship constant-time compare as a primitive.",
      "metaphor": "A safecracker listening to the tumblers: each correct character makes the rejection arrive a hair later, and patience turns that hair into the whole combination.",
      "teachable_moment": "Hash both sides first (normalizing length), then timingSafeEqual — the hash step matters because timingSafeEqual itself throws on unequal lengths, which is its own oracle."
    },
    {
      "id": "FS-PR420-005",
      "title": "Cooldown is consumed before input validation, masking the real error for 10s",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/ordering/src/community-onboarding-orchestrator.ts:333-341",
      "description": "reprobe() sets the per-order cooldown timestamp before parsing the order's inputs. If CommunityOnboardingInputs.safeParse fails, the call returns 'not a community-onboarding order' but the cooldown has already been recorded — retries within 10s get a misleading 429 'cooldown' instead of the underlying 400, even though no probe ever ran.",
      "suggestion": "Move the cooldown set to after input validation succeeds (immediately before the probe loop), or clear the cooldown entry on early-return error paths.",
      "confidence": 0.8,
      "faang_parallel": "GCP quota systems deliberately charge quota only on admitted requests — a request rejected at validation never consumes the caller's budget, because burning quota on 400s turns every client bug into a self-inflicted throttling incident.",
      "metaphor": "A bouncer who stamps your hand before checking your ID — the ID check fails, but now you also can't approach the door again for ten minutes, and the stamp is all you'll be told about.",
      "teachable_moment": "Acquire scarce resources — cooldowns, quota, locks — at the last responsible moment, after every cheap validation has passed. For an agent-facing API this doubles as error honesty: the caller should always see the real reason, not a side effect of the attempt."
    },
    {
      "id": "FS-PR420-006",
      "title": "reprobeCooldown Map grows unboundedly",
      "severity": "LOW",
      "category": "resource-management",
      "file": "packages/services/ordering/src/community-onboarding-orchestrator.ts:316-318",
      "description": "The per-order cooldown Map only ever gains entries; nothing evicts them when orders reach terminal state or when the cooldown expires. On a long-lived single instance, memory grows linearly with the number of distinct orders ever reprobed. The loa:shortcut comment covers the single-instance limitation but not the unbounded growth.",
      "suggestion": "Delete the entry when an order is observed terminal in reprobe(), or sweep entries older than REPROBE_COOLDOWN_MS on each call (the map is small; a lazy sweep is sufficient).",
      "confidence": 0.85,
      "faang_parallel": "Guava's CacheBuilder (and its successor Caffeine) exist because Google kept finding 'temporary' HashMaps that became heap leaks — expireAfterWrite is the institutionalized answer to exactly this shape of map.",
      "metaphor": "A coat check that issues tickets but never clears the racks — every guest the venue has ever served still has a coat hanging, forever.",
      "teachable_moment": "Any map keyed by an unbounded external identifier is a cache whether you named it one or not, and every cache needs an eviction story — even a lazy sweep on access counts.",
      "connection": "The loa:shortcut marker here names one ceiling (single-instance) but not this one — a shortcut marker earns its keep by naming every known ceiling plus the upgrade trigger."
    },
    {
      "id": "FS-PR420-007",
      "title": "Malformed reprobe env vars produce NaN and fail open",
      "severity": "LOW",
      "category": "configuration",
      "file": "packages/services/ordering/src/community-onboarding-orchestrator.ts:39-41",
      "description": "REPROBE_COOLDOWN_MS and REPROBE_PER_PROBE_TIMEOUT_MS are read via Number(process.env.X ?? default). A non-numeric value yields NaN: `nowMs - last < NaN` is always false, so the cooldown silently never triggers, and setTimeout(NaN) fires immediately, making every probe report timeout-ambiguous. Both failure modes are silent.",
      "suggestion": "Validate with Number.isFinite() and fall back to the default (optionally log a warning) when the env value doesn't parse.",
      "confidence": 0.8,
      "faang_parallel": "The Google SRE book calls out configuration as a leading outage class precisely because bad config tends to fail silently at runtime rather than loudly at startup — which is why mature services parse config once into typed values and refuse to boot on garbage.",
      "metaphor": "A thermostat that reads a smudged dial as 'not a number' and quietly resolves never to turn on the heat — the house just gets cold; no alarm ever sounds.",
      "teachable_moment": "Parse, don't validate-inline: resolve env into typed, checked constants at startup. NaN is JavaScript's most effective saboteur because every comparison with it is false — both branches of a guard can silently disarm.",
      "connection": "Same silent-degradation family as FS-PR420-005, 006, and 008 — the cure everywhere is loud refusal over quiet wrongness."
    },
    {
      "id": "FS-PR420-008",
      "title": "'Deployed' detection is Railway/NODE_ENV-specific; other hosts get tokenless open writes",
      "severity": "LOW",
      "category": "security",
      "file": "packages/services/ordering/src/composition.ts:99-103",
      "description": "resolveWriteRoutePosture treats the environment as deployed only when RAILWAY_ENVIRONMENT is set or NODE_ENV === 'production'. A deployment on any other host without NODE_ENV=production (common in containerized setups that forget to set it) resolves to open_dev and mounts tokenless write routes. The fail-closed guarantee is therefore conditional on env hygiene outside this codebase. Also minor: railwayEnvironment is trimmed but nodeEnv is not.",
      "suggestion": "Consider inverting the default (fail closed unless an explicit ORDERING_DEV_MODE marker is set), or at least document the assumption in DEPLOY.md that non-Railway deploys MUST set NODE_ENV=production. Trim nodeEnv for consistency.",
      "confidence": 0.6,
      "faang_parallel": "MongoDB's historical default of binding open without auth produced tens of thousands of exposed databases — the industry lesson that hardened the 'secure by default, explicit opt-in for the unsafe posture' principle now standard in Kubernetes RBAC and cloud IAM default-deny.",
      "metaphor": "A vault door that locks itself only when it recognizes the building it was installed in — move it to a new building and it stands politely open.",
      "teachable_moment": "Detect the unsafe environment, never the safe one. Requiring an explicit dev-mode marker means the failure mode of a forgotten env var is a locked door, not an open one.",
      "decision_trail": "If the current heuristic stays, DEPLOY.md should state the invariant ('non-Railway deploys MUST set NODE_ENV=production') so the assumption survives the next host migration."
    },
    {
      "id": "FS-PR420-009",
      "title": "Blocked ingredients are silently excluded from reprobe-all",
      "severity": "LOW",
      "category": "design",
      "file": "packages/services/ordering/src/community-onboarding-orchestrator.ts:345-349",
      "description": "When no ingredient is specified, targets filter to pending/in_progress only, so a 'blocked' ingredient (e.g. shadow_preview) is never reprobed in the default path even though the monotonic merge would accept blocked→complete. An agent polling reprobe-all will never observe a blocked ingredient unblocking; only an explicit per-ingredient reprobe can. If intentional, it is undocumented in DEPLOY.md's reprobe row.",
      "suggestion": "Either include 'blocked' in the default target filter, or document that blocked ingredients require an explicit `{ \"ingredient\": ... }` reprobe.",
      "confidence": 0.6,
      "faang_parallel": "Kubernetes controllers requeue every non-terminal resource — including ones waiting on external conditions — because a state that can only be exited via manual, name-specific action becomes a stuck-state class the moment operators stop remembering it exists.",
      "metaphor": "A triage board where patients marked 'waiting for labs' are skipped on every rounds sweep — the labs came back hours ago, but nobody looks unless someone asks for that patient by name.",
      "teachable_moment": "For a polling consumer, the default sweep should cover every non-terminal state, or the contract must say — in the docs the agent reads — which states need an explicit prod.",
      "decision_trail": "Whichever way this lands, record it in DEPLOY.md's reprobe row: agents building on this surface will encode the answer either way, so make it deliberate."
    },
    {
      "id": "FS-PR420-010",
      "title": "Test mutates REPROBE_PER_PROBE_TIMEOUT_MS without try/finally cleanup",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/services/ordering/src/__tests__/reprobe-endpoint.test.ts:131-178",
      "description": "The hung-probe test sets process.env.REPROBE_PER_PROBE_TIMEOUT_MS and calls vi.resetModules(), deleting the env var only at the end of the test body. If any assertion in the middle fails, the env var and reset module registry leak into subsequent tests in the file, producing confusing cascade failures.",
      "suggestion": "Wrap the mutation in try/finally, or use vi.stubEnv() + vi.unstubAllEnvs() in afterEach.",
      "confidence": 0.75,
      "faang_parallel": "Google's flaky-test research found order-dependent tests — where one test's leaked state changes another's outcome — to be among the costliest flakiness classes, because the failure surfaces in the innocent test, not the guilty one.",
      "metaphor": "A chemistry student who plans to wash the beaker after the experiment — if the experiment blows up midway, the next student inherits the residue and blames their own reagents.",
      "teachable_moment": "Cleanup belongs to the framework (afterEach, finally), never to the happy path — because the whole purpose of a test is to sometimes fail, and the failing run is exactly when inline cleanup doesn't execute."
    },
    {
      "id": "FS-PR420-011",
      "title": "16 beads history snapshot files (~12k lines) committed with the feature PR",
      "severity": "LOW",
      "category": "hygiene",
      "file": ".beads/.br_history/issues.20260702_012058_637822000.jsonl.meta.json",
      "description": "The PR carries 16 .beads/.br_history snapshot JSONL files plus their meta.json companions (~12,000 added lines), inflating the diff and permanently growing repo history with mechanical tracker snapshots that duplicate .beads/issues.jsonl state.",
      "suggestion": "Confirm whether .br_history is meant to be committed; if not, add it to .gitignore and drop these files from the PR. If it is intentional state-zone data, consider committing snapshots on a coarser cadence.",
      "confidence": 0.5,
      "faang_parallel": "Chromium's repo discipline categorically excludes generated and mechanically-derived artifacts from version control — derived state that duplicates a source of truth only adds clone weight and diff noise while making the real changes harder to review.",
      "metaphor": "Stapling every intermediate draft of a memo to the final copy before filing it — the archive grows, the signal doesn't.",
      "teachable_moment": "Git history is append-only: bytes committed today are cloned by every future contributor forever, so mechanical snapshots deserve either a .gitignore entry or a deliberate, coarse cadence."
    },
    {
      "id": "FS-PR420-012",
      "title": "Fail-closed write-route posture is well designed and pinned by tests",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/services/ordering/src/composition.ts:88-110",
      "description": "The D8 auth matrix is encoded as a pure function with one row per (deployed, token) combination, documented inline, wired at composition time by not mounting the routes at all (404 rather than 401 leaves no partially-guarded surface), surfaced via /healthz for operability, and pinned by write-route-posture.test.ts including the whitespace-only marker edge case. A botched rotation degrades to read-only rather than an open write path, and the runbook documents the rotation window honestly.",
      "suggestion": "None — keep this pattern for future write surfaces.",
      "confidence": 0.9,
      "faang_parallel": "This is the BeyondCorp/default-deny instinct executed at the routing layer: like a service mesh removing a listener entirely rather than guarding it, an unmounted route has no misconfiguration surface, and 404-over-401 denies attackers even the oracle that a write endpoint exists.",
      "metaphor": "Instead of hiring a guard for the door, the door was never built into that wall.",
      "teachable_moment": "Encoding an auth matrix as a pure function with table-driven tests turns security policy into reviewable, pinnable data — the posture can now only change via a diff a reviewer will see.",
      "connection": "This is the standard the rest of the PR should be measured against — FS-PR420-002 and FS-PR420-008 are the places that same rigor hasn't reached yet."
    },
    {
      "id": "FS-PR420-013",
      "title": "Probe failure as a result, and ambiguity never overwriting recorded truth",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/services/ordering/src/community-onboarding-orchestrator.ts:322-330",
      "description": "Treating timeout/probe_error as a per-ingredient 'ambiguous' outcome in the response body — while leaving probe_meta and the ingredient merge untouched — is the right failure semantic for an agent-facing surface: callers get fresh signal without transient upstream failures corrupting durable state. Carrying world_slug inside the raced value so a late-resolving timed-out probe can never contribute data closes a subtle race correctly.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "This mirrors how Google's monitoring stack treats 'no data' as distinct from 'zero', and how gRPC keeps DEADLINE_EXCEEDED separate from a payload: uncertainty is reported as uncertainty, never coerced into a value that then gets persisted as fact.",
      "metaphor": "A lab that reports 'sample inconclusive' rather than guessing — and never writes the guess into the patient's chart.",
      "teachable_moment": "Promise.race has a loser, and the loser resolves eventually — binding the result's identity into the raced value itself is the clean way to defang late resolution, worth stealing for any timeout-wrapped async work."
    },
    {
      "id": "FS-PR420-014",
      "title": "Audit evidence as immutable server-derived value copy, verified byte-identical",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/services/ordering/src/__tests__/advance-evidence.test.ts:107-131",
      "description": "The advance audit entry snapshots probe_meta by value (never a live reference), distinguishes server-derived fields (token_label, evidence) from untrusted client metadata (caller_note), makes the never-probed case visibly ungrounded via evidence:null, and the test pins immutability by structuredClone-ing the entry and asserting byte-identity after a later reprobe moves probe_meta on. This is exactly how audit trails should be tested.",
      "suggestion": "None.",
      "confidence": 0.9,
      "faang_parallel": "AWS CloudTrail's log-file integrity validation rests on the same principle: an audit record is a value frozen at event time, verifiable later against tampering — not a view over mutable state that silently rewrites history when the state moves.",
      "metaphor": "A notary photocopies the document at signing time — later edits to the original can never rewrite what was notarized.",
      "teachable_moment": "The test technique deserves as much praise as the design: mutate the world *after* capture, then assert byte-identity of the earlier snapshot. That's the only test shape that actually catches an accidental live reference.",
      "connection": "Together with FS-PR420-013, this establishes the surface's epistemics: server-derived evidence is truth, client metadata is labeled untrusted, and absence of evidence is visible — the same ground-up honesty FS-PR420-002's projection should extend to what leaves the system."
    }
  ]
}
```


## Callouts

Three findings above are formal PRAISE (FS-PR420-012, 013, 014), and they share a spine worth naming: **this PR consistently treats truth as something earned, not asserted.** Evidence is server-derived and frozen by value; ambiguity is reported as ambiguity and never persisted as fact; the never-probed case is visibly `null` rather than optimistically filled. Beyond those three:

- **[Test Coverage]** The four new test files attack the *semantics*, not the plumbing — write-route posture including the whitespace-token edge, projection redaction pinned as contract, evidence immutability via post-mutation byte-identity. That last technique (mutate the world after capture, then assert the snapshot didn't move) is the only test shape that catches a live-reference bug, and most teams never write it.
- **[Operational]** The token-rotation runbook documents the rotation *window* honestly instead of pretending rotation is atomic — the kind of honesty that makes 3am operations survivable. Netflix's runbook culture calls this "documenting the sharp edges, not sanding the prose."
- **[Quality]** The `loa:shortcut` markers turn deliberate simplifications into visible intent rather than apparent ignorance. FS-PR420-006 shows the failure mode when a marker names only one of two ceilings — but the practice itself is exactly right.

The gap between this PR's best paths and its findings is small and closable: give the reprobe merge the CAS treatment the transitions already have, and give the public projection the allowlist treatment the token path already has. The architecture is asking for it.

PR #422, opened 2026-07-02
feat(network/beacon-consumer): freeside-cli inspect consumes live beacon via SSRF-safe fetcher (PR-A)

zkSoju:
## Beacon-consumer cycle — PR-A (network domain)

Makes `freeside-cli inspect <slug>` the **first live beacon consumer**: fetch a cell's declared `beacon_url` over an SSRF-safe path, validate BeaconV3, and render the single-owned `OrientationPacket` with an honest classification mapped to a stable exit code. An orienting agent reaches for `inspect` instead of grepping a building's source to learn what it does (G-1/G-2).

### What landed (Sprint 1)
| Task | Delivered |
|------|-----------|
| S1-T1 | `OrientationPacket` single-owned type + total builder + `BEACON_EXIT` (beacon-schema) — already on `cycle/beacon-consumer` |
| S1-T2 | Shared conformance-vector suite (host-guard bypass set · full v4/v6/IPv4-mapped CIDR block set · classification→exit rows) |
| S1-T3 | `hardenBeaconFetch` SSRF core — https-only · IDNA/dot-boundary host allowlist · complete private-range reject · **IP-pinned connect (DNS-rebinding TOCTOU closed)** · 256KB cap · off-host→void |
| S1-T4 | `inspect` un-stub — async `inspectModule`→`InspectResult`; 5-class `classify`→exit; `--raw` (no-leak) + `--pretty` |
| IMP-004 | `doctor --remote` + `inspect` share the **one** hardened fetcher (no parallel copy); old global-fetch copy removed |
| S1-T5 | CLAUDE.md pointer for `freeside-cli inspect` (inspect-only; no `loa model` forward-ref) |
| S1-T6 | Both suites green · domain = network-only |

### Verification
- beacon-schema `pnpm test`: **59 unit + 3 cli green**
- freeside-cli `pnpm build && pnpm test`: **89 pass / 0 fail / 1 skip** (pre-existing `ORDERING_DIFFERENTIAL` env-gated live test)
- `tools/check-beacon-domain.sh --since origin/main`: **✓ network** (zero `packages/services/**` platform paths)
- No live network in the default test run — every beacon test uses an injected fetcher/transport.

### Cross-model review (FAGAN)
Adversarial dissent on the committed diff. **7 anti-SSRF guards affirmed sound** (check-and-connect-same-host · exact-IP pin / rebind TOCTOU closed · reject-if-ANY-blocked · v4 bitmask across 26 boundary cases · redirect body containment · classify ordering · doctor unification).
- **F1 MEDIUM — FIXED (`dfa9598e`)**: hex-form IPv4-mapped IPv6 (`::ffff:a9fe:a9fe` = 169.254.169.254 metadata) bypassed the private-range block. Fixed by numeric hextet unwrap (form-independent); hex-form vectors added.
- F2–F5 LOW/INFO documented in the report (no live impact; F2 fail-closed and *fixes* a prior doctor SSRF).

Full AC-by-AC evidence: `grimoires/loa/a2a/sprint-1/reviewer.md`.

### Out of scope (follow-on)
- **PR-B** (`loa model` in external `loa-cli`) is the operator-gated cross-repo follow-on; same conformance vectors, checksum-pinned. The CLAUDE.md pointer names only `freeside-cli inspect` until it lands (partial-rollout invariant).
- G-1 kill-test is operator-run post-land.


zkSoju (commented):
This task is a review-generation deliverable, not a code change — producing the enriched Bridgebuilder review directly as output.

## Summary

This PR replaces a permissive beacon fetcher with a genuinely hardened one — injectable transport seam, numeric IPv6 unwrapping, shared conformance vectors — and the security engineering is largely excellent. The primary concern is conceptual: the allowlist's *stated* threat model (registry compromise) is undermined by including shared multi-tenant PaaS apex domains (BC-001), and a handful of transport-layer details (redirect parse crash, idle-vs-deadline timeout, chunk decoding) erode the fail-closed contract the module promises. Verdict: **REQUEST_CHANGES** on BC-001 and BC-002; the rest are refinements.

The architectural throughline worth naming: this module gets the *seams* right (transport injection, single-owner exit table, no-partial-fill packets) but a few of the *guarantees written in comments* don't yet match the code. That gap — documented invariant vs. enforced invariant — is where every finding below lives.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "BC-001",
      "title": "Allowlist includes shared multi-tenant PaaS suffixes (vercel.app, up.railway.app), undermining the 'cluster-controlled' primary control",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:27-31",
      "description": "ALLOWED_BEACON_HOST_SUFFIXES is documented as 'Cluster-controlled domain suffixes' and the header comment calls the allowlist 'the PRIMARY control' against a compromised registry. But `up.railway.app` and `vercel.app` are shared multi-tenant platform domains: any attacker can deploy `attacker.vercel.app` or `attacker.up.railway.app` in minutes. Under the stated threat model (registry compromise), an attacker can point beacon_url at fully attacker-controlled endpoints that pass the allowlist. The private-range guard still blocks internal SSRF, but the allowlist no longer restricts fetches to cluster-controlled infrastructure — attacker-authored BeaconV3 bodies (publisher, is/is_not prose, composes_with tags) will validate and be rendered into orientation packets consumed by agents, which is an injection surface for downstream agent consumers of `inspect` output.",
      "suggestion": "Pin full cluster-owned hostnames (e.g. `sonar-api-production.up.railway.app`) or project-scoped subdomain suffixes instead of the bare platform apex domains. Alternatively, derive the allowed host set from the registry's own vetted deployment hosts at build time. At minimum, correct the comment so the residual trust granted to any tenant of these platforms is explicit.",
      "confidence": 0.65,
      "faang_parallel": "This is why the Public Suffix List exists and why browsers refuse to set cookies on `*.vercel.app`-class domains — Mozilla learned that a suffix shared by mutually-untrusting tenants is a boundary, not an identity. Google's internal ALTS goes further: it authenticates workload identity, never hostname, precisely because hostnames on shared infra prove nothing about who controls the endpoint.",
      "metaphor": "Allowlisting `vercel.app` is like telling a doorman 'admit anyone from that apartment building' — you own one unit, but so do ten thousand strangers.",
      "teachable_moment": "An allowlist's strength equals the strength of the *registration boundary* behind each entry. Ask: 'who can mint a name that matches this pattern?' If the answer includes strangers, the entry is a suffix-shaped hole, not a control.",
      "connection": "Same failure class as the estate's 'deletion over denylist' doctrine — trust boundaries drawn on names rather than ownership drift silently as platforms onboard new tenants."
    },
    {
      "id": "BC-002",
      "title": "Malformed Location header throws out of the fetcher (unhandled rejection) instead of classifying as transport_error",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:262",
      "description": "In makeHardenedBeaconFetcher, `const next = new URL(res.location, current).toString()` is not wrapped in try/catch. A server (on an allowlisted host) responding with an unparseable Location such as `Location: https://` makes `new URL` throw a TypeError, rejecting the fetcher's promise. Neither `inspectModule` nor the CLI entry wraps the `await fetcher(...)` call, so a hostile-but-allowlisted (or simply buggy) server can crash `inspect` and `doctor --remote` with an unhandled rejection rather than producing the enumerated `transport_error` result the module's contract promises ('every reject as a status-0 result').",
      "suggestion": "Wrap the redirect-Location resolution in try/catch and return `{ status: 0, finalUrl: current, body: \"\", error: \"transport_error\" }` on parse failure, matching the fetcher's fail-closed contract.",
      "confidence": 0.75,
      "faang_parallel": "Netflix's Hystrix codified this: a resilience wrapper that itself throws on malformed input has a hole in the bulkhead. Every input to the wrapper — including server-controlled headers — must resolve to an enumerated outcome.",
      "metaphor": "The fetcher built a padded room for network failures, then left one wall unpadded where the redirect sign hangs.",
      "teachable_moment": "When a module's contract is 'never reject, always classify', audit every parse of *remote-controlled bytes* inside it — headers are attacker input exactly as much as bodies are."
    },
    {
      "id": "BC-003",
      "title": "Request timeout is socket-idle, not a total deadline — slow-drip responses evade the 12s bound (regression vs AbortSignal.timeout)",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:205",
      "description": "The replaced `defaultBeaconFetcher` used `AbortSignal.timeout(REMOTE_TIMEOUT_MS)`, a hard total deadline. The new `pinnedGet` uses node:https's `timeout` option, which is a socket *inactivity* timeout: a server that trickles one byte every few seconds resets it indefinitely. Combined with the 256KB body cap, a malicious allowlisted endpoint can hold a `doctor --remote` sweep or `inspect` call open for a very long time (hours in the worst case) instead of 12 seconds. Low blast radius for a CLI, but it is a concrete downgrade of a guarantee the old fetcher had.",
      "suggestion": "Add an absolute deadline alongside the idle timeout, e.g. `setTimeout(() => req.destroy(new Error('deadline')), REQUEST_TIMEOUT_MS)` (unref'd), or track elapsed time in the data handler and destroy when exceeded.",
      "confidence": 0.7,
      "faang_parallel": "This is the Slowloris lesson in reverse (client-side): gRPC made deadlines total and propagated for exactly this reason — idle timers measure liveness, deadlines measure patience, and only the second bounds resource commitment.",
      "metaphor": "An idle timeout is a waiter who leaves only if you stop talking; a deadline is one who leaves at closing time regardless. The drip-feeding table never stops talking.",
      "teachable_moment": "When swapping transport layers, diff the *guarantees*, not just the features — `AbortSignal.timeout` and `socket timeout` share a name and differ in kind."
    },
    {
      "id": "BC-004",
      "title": "Custom lookup callback ignores options.all — may break under Node's autoSelectFamily (happy-eyeballs) lookup contract",
      "severity": "MEDIUM",
      "category": "compatibility",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:213",
      "description": "The IP-pinning `lookup: (_hostname, _opts, cb) => cb(null, ip as never, family)` always uses the single-address callback shape. When Node invokes a custom lookup with `options.all: true` (which happens under `autoSelectFamily`, default-on in recent Node releases), the callback is expected to receive an array of `{address, family}` objects. Returning a bare string in that mode can produce a connect error. The `as never` cast is masking exactly this type mismatch. Failure is fail-closed (surfaces as transport_error), so no security impact, but the fetcher may be functionally broken on some Node versions/configs.",
      "suggestion": "Honor the options: `cb(null, _opts?.all ? [{ address: ip, family }] : (ip as any), family)` — or set `autoSelectFamily: false` on the request options and document why.",
      "confidence": 0.4,
      "teachable_moment": "Verify before shipping a fix: run the fetcher on Node ≥20 against a dual-stack host to confirm whether `options.all` actually arrives. Independent of the runtime answer, `as never` is a signal worth honoring — it is the compiler telling you the shapes don't line up, silenced rather than resolved."
    },
    {
      "id": "BC-005",
      "title": "Per-chunk Buffer.toString('utf8') can corrupt multi-byte characters split across chunk boundaries",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:227-235",
      "description": "The body is accumulated via `body += chunk.toString('utf8')` per chunk. A multi-byte UTF-8 sequence split across two TCP chunks decodes to replacement characters, corrupting the body. For a beacon containing non-ASCII text (publisher names, is/is_not prose), this can intermittently corrupt string contents or, if inside structural JSON, cause a spurious `beacon_invalid` classification that depends on network chunking.",
      "suggestion": "Accumulate `Buffer` chunks in an array (counting bytes for the cap) and decode once with `Buffer.concat(chunks).toString('utf8')` on 'end', or use `string_decoder.StringDecoder`.",
      "confidence": 0.6,
      "faang_parallel": "Node ships `string_decoder` in core precisely for this; it's the same boundary bug class as UTF-8 handling in Kafka consumers that decode per-record-batch — correctness that depends on where the network happened to split bytes.",
      "metaphor": "Decoding per chunk is like translating a book one torn page-half at a time — words split across the tear come out as gibberish.",
      "teachable_moment": "TCP chunk boundaries are nondeterministic, so any per-chunk transformation must be proven boundary-invariant. If it isn't, buffer first, transform once."
    },
    {
      "id": "BC-006",
      "title": "URL port is silently dropped — a beacon_url with an explicit port passes guards but connects to 443",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:200-214",
      "description": "pinnedGet builds the request from `host` and `u.pathname + u.search` but never reads `u.port`. `normalizeHost` receives `u.hostname` (never contains the port), so a URL like `https://sonar.0xhoneyjar.xyz:8443/beacon.json` passes every guard and then silently connects to port 443. Connecting only to 443 is the safer direction, but the mismatch between the declared URL and the actual connection is surprising and will misclassify such a cell as dark/unreachable without an honest detail.",
      "suggestion": "Either reject URLs with a non-default port explicitly (enumerated detail, fail-closed and honest) or pass `port: u.port || 443` through to the request.",
      "confidence": 0.7,
      "metaphor": "The guard checks the street address but the driver always goes to apartment 443 — sometimes right, never checked.",
      "teachable_moment": "Silent normalization is worse than explicit rejection in a tool whose product is *honest classification* — a wrong-but-enumerated verdict beats a right-by-accident one."
    },
    {
      "id": "BC-007",
      "title": "doctor --remote now silently applies the host allowlist to every registry beacon_url — behavior change for non-allowlisted hosts",
      "severity": "MEDIUM",
      "category": "behavior-change",
      "file": "packages/freeside-cli/src/verbs/doctor.ts:193-199",
      "description": "`defaultBeaconFetcher` is now the hardened fetcher, so `doctor --remote` rejects any declared beacon_url whose host is outside ALLOWED_BEACON_HOST_SUFFIXES with a status-0 `host_not_allowlisted` result, which classifyProbe will fold into unreachable/dark. Previously doctor probed any host (with only redirect host-pinning). If the registry ever legitimately contains a host outside the three suffixes (a new deploy platform, a custom domain), doctor will report it dark without any signal that the failure is client-side policy rather than the cell being down — a confusing false negative for the tool whose job is honest liveness reporting.",
      "suggestion": "This is likely intended per IMP-004, but ensure doctor surfaces the enumerated `host_not_allowlisted` detail distinctly in its verdict output (not folded into generic unreachable), and add a registry-side lint that beacon_url hosts must be allowlisted so the mismatch is caught at write time, not probe time.",
      "confidence": 0.45,
      "faang_parallel": "Monitoring teams at Google distinguish 'probe failed' from 'prober refused' as separate signal classes — folding client-side policy denials into target-down metrics is a classic source of phantom outages in Blackbox-exporter-style setups.",
      "teachable_moment": "When a health tool gains a policy layer, the policy's refusals must be a distinct verdict — otherwise the tool starts reporting its own opinions as the world's state.",
      "connection": "Directly parallels the estate's 'CI sensors must not be numb' doctrine: a sensor that conflates two causes trains operators to ignore it."
    },
    {
      "id": "BC-008",
      "title": "All non-2xx/non-3xx statuses map to detail 'not_found' — 5xx misreported",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/freeside-cli/src/verbs/inspect.ts:66-68",
      "description": "classify() maps any reachable non-2xx terminal status (including 500/502/503) to `beacon_dark` with detail `not_found`. A cell whose beacon endpoint is erroring is meaningfully different from one that 404s — an agent triaging 'why is this building dark' gets a misleading detail.",
      "suggestion": "Add a `server_error` (or reuse `transport_error`) detail for status >= 500, keeping the classification decision per SDD but making the detail honest.",
      "confidence": 0.55,
      "metaphor": "A 404 is 'nobody lives here'; a 503 is 'the resident is home but collapsed on the floor' — the triage response differs entirely.",
      "teachable_moment": "Classification taxonomies for machine consumers should preserve the 4xx/5xx distinction even when the top-level verdict is the same — the detail field exists precisely to carry the 'why'."
    },
    {
      "id": "BC-009",
      "title": "CLI missing-argument exit code (1) collides with unknown_slug exit code (1)",
      "severity": "LOW",
      "category": "cli-contract",
      "file": "packages/freeside-cli/bin/freeside-cli.ts:66-70",
      "description": "The usage error for a missing/flag-shaped slug now returns 1, the same code BEACON_EXIT assigns to unknown_slug. The previous stub used 2 for usage errors (which now means beacon_unreachable). An agent branching on exit codes cannot distinguish 'I called it wrong' from 'slug not in registry' without parsing stderr vs stdout.",
      "suggestion": "Use a distinct exit code for usage errors (e.g. 64 per sysexits, or extend the BEACON_EXIT table with a usage row) and document it alongside the classification table.",
      "confidence": 0.5,
      "faang_parallel": "BSD's sysexits.h (EX_USAGE=64) exists because sendmail needed callers to distinguish operator error from delivery failure — the exact agent-branching problem this CLI has, solved in 1980.",
      "teachable_moment": "For an agent-first CLI, the exit-code table IS the API. Collisions there are breaking changes in disguise, and this PR already built the right home for the fix: the single-owner BEACON_EXIT table."
    },
    {
      "id": "BC-010",
      "title": "Minor dead code in isBlockedV4: unused destructured variable and redundant multicast condition",
      "severity": "LOW",
      "category": "code-quality",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:70-90",
      "description": "`const [a, b] = parts;` destructures `b` but never uses it, and the multicast check `a === 224 || (a >= 224 && a <= 239)` has a redundant first clause (subsumed by the range). Harmless, but in a security-critical block set every condition should read as deliberate.",
      "suggestion": "Drop `b` from the destructure and reduce the multicast check to `(a >= 224 && a <= 239)`.",
      "confidence": 0.9,
      "faang_parallel": "OpenBSD's audit culture treats redundant conditions in security predicates as bugs even when logically harmless — a reader who can't tell whether a clause is intentional can't verify the predicate.",
      "teachable_moment": "In a block set, every token carries proof burden. Dead clauses make reviewers ask 'is this covering a case I don't see?' — noise that costs exactly the attention the block set needs."
    },
    {
      "id": "BC-011",
      "title": "Injectable HardenTransport seam makes DNS-rebinding and redirect security properties testable offline",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:151-160",
      "description": "Splitting resolve+connect behind a two-method transport interface lets the tests assert the actual security invariants — 'connect used the pre-validated IP', 'guard rejects before any socket connect', 'off-host Location never requested' — with zero live network. The rebinding test in harden-beacon-fetch.test.ts pins the TOCTOU closure directly rather than inferring it.",
      "suggestion": "None — this is the right seam. Keep production `realTransport` as the only place raw dns/https is touched.",
      "confidence": 0.9,
      "faang_parallel": "This mirrors how Chromium tests its DNS interception layer: a MockHostResolver seam so rebinding defenses are asserted as unit invariants, not hoped-for integration behavior. The alternative — spinning up rebinding DNS servers in CI — is what teams do when they didn't design the seam.",
      "metaphor": "Instead of crash-testing the car on a real highway, the airbag sensor is on a bench where you can fire exactly the collision you fear.",
      "connection": "This is the 'verify the mechanism, not the symptom' pattern made structural — the test grips the TOCTOU closure itself."
    },
    {
      "id": "BC-012",
      "title": "Numeric IPv6 hextet parsing closes textual-serialization bypasses; shared conformance vectors pin cross-implementation parity",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/freeside-cli/src/lib/harden-beacon-fetch.ts:97-140",
      "description": "Expanding IPv6 to numeric hextets and unwrapping IPv4-mapped/compatible addresses numerically (covering the hex form `::ffff:a9fe:a9fe` of the metadata IP, not just the dotted form) eliminates the classic serialization-dependent bypass class. Publishing the block set and host-guard cases as shared JSON conformance vectors that both freeside-cli and loa-cli must pass turns 'the two implementations agree' from a hope into a test.",
      "suggestion": "None.",
      "confidence": 0.9,
      "faang_parallel": "This is the WHATWG URL spec strategy: after a decade of parser-differential exploits (Orange Tsai's SSRF research lives in exactly these gaps), the fix was shared conformance vectors that every implementation runs. Comparing on numbers instead of strings is also how Linux's kernel routing decides — serialization never enters the decision.",
      "metaphor": "String-matching IPs is checking passports by how the name is spelled; numeric comparison checks the biometric. `::ffff:a9fe:a9fe` and `::ffff:169.254.169.254` spell differently and are the same traveler.",
      "teachable_moment": "Any security predicate over a value with multiple textual serializations must normalize to the canonical numeric form first — and cross-implementation agreement must be a fixture file, never a convention."
    },
    {
      "id": "BC-013",
      "title": "No-partial-fill missing-field policy and single-owner exit table are clean contract design",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/beacon-schema/src/orientation-packet.ts:96-134",
      "description": "buildOrientationPacket is total, never presents unvalidated beacon data (beacon fields are all-or-nothing on beacon_valid, verified by a dedicated 'no partial-fill' test), and BEACON_EXIT is the one exit-code table both surfaces import — the drift-prone duplication was designed out rather than reviewed out.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "The all-or-nothing packet is 'parse, don't validate' (Alexis King) applied at a trust boundary — the same reasoning behind protobuf's presence semantics: consumers should never see a struct that is half-trusted.",
      "metaphor": "Rather than serving a plate where some dishes passed inspection and some didn't, the kitchen serves the full meal or an honest empty plate with the reason written on it.",
      "connection": "Designing drift out (one imported table) beats reviewing drift away — the same instinct that produced the shared conformance vectors in BC-012."
    }
  ]
}
```


## Callouts

The three PRAISE findings deserve emphasis because they share one instinct: **making invariants structural instead of aspirational**. The transport seam (BC-011) turns "we believe rebinding is closed" into a bench test; the conformance vectors (BC-012) turn "both CLIs agree" into a fixture both must pass; the single exit table and no-partial-fill packet (BC-013) turn "don't drift" into "can't drift." That's the same engineering muscle, exercised three times, and it's the muscle that makes the BC-001/BC-002 fixes easy — the seams to test them against already exist.

The meditation this PR earns: the remaining weaknesses are all places where a *comment* makes a promise the *code* doesn't yet keep — "cluster-controlled" suffixes that strangers can mint, "every reject becomes a result" with one unparsed header, a "12s timeout" that measures silence instead of time. The pattern to carry forward is the one already demonstrated in the praised code: when you write a guarantee down, write the test that would catch its violation in the same commit.

**Verdict: REQUEST_CHANGES** — BC-001 (allowlist trust boundary) and BC-002 (fail-closed contract hole) before merge; the remainder can follow.

zkSoju:
## Bridgebuilder findings addressed — `1b2759c9`

Thanks for the review — the "documented invariant vs. enforced invariant" framing was exactly right. All actionable findings addressed; regression tests added.

| Finding | Sev | Resolution |
|---|---|---|
| **BC-001** allowlist includes shared PaaS apexes | HIGH | **Fixed** — removed `up.railway.app` + `vercel.app`; allowlist is now `["0xhoneyjar.xyz"]` only. Verified all 8 registry `beacon_url`s are `*.0xhoneyjar.xyz`, so nothing legit breaks. Regression: conformance vectors + `hardenedBeaconFetcher` tests assert `attacker.vercel.app` / `attacker.up.railway.app` → `host_not_allowlisted`. |
| **BC-002** malformed Location throws | MED | **Fixed** — `new URL(res.location, current)` wrapped in try/catch → `transport_error`. Test: injected transport returns `Location: https://` → status-0 `transport_error`, no throw. |
| **BC-003** idle timeout ≠ deadline | MED | **Fixed** — added an unref'd absolute `setTimeout(REQUEST_TIMEOUT_MS)` deadline alongside the socket-idle `timeout`; drip-feed can't evade the 12s bound. |
| **BC-004** lookup ignores `options.all` | MED | **Fixed** — the pinned `lookup` now returns `[{address,family}]` when `opts.all` (autoSelectFamily), else the single-address shape; `as never` on the array path removed. |
| **BC-005** per-chunk UTF-8 decode | MED | **Fixed** — accumulate `Buffer` chunks (byte-counted for the cap) and `Buffer.concat(...).toString('utf8')` once on `end`. |
| **BC-006** port silently dropped | LOW | **Fixed** — an explicit non-443 port is rejected at the guard (`host_not_allowlisted`, fail-closed + honest) instead of connecting to 443. Conformance vector added. |
| **BC-008** 5xx → `not_found` | LOW | **Fixed** — added `server_error` to `VerdictDetail`; `classify` maps status ≥ 500 → `server_error` (classification stays `beacon_dark`). Vector + inspect test added. |
| **BC-009** usage/​unknown_slug exit collision | LOW | **Fixed** — `inspect` usage error now exits `64` (EX_USAGE), distinct from `unknown_slug`'s `1`. |
| **BC-010** dead code in `isBlockedV4` | LOW | **Fixed** — dropped unused `b`, reduced the redundant multicast clause. |
| **BC-007** doctor folds policy-denial into dark | MED | **Deferred** (documented) — no live impact (all registry hosts allowlisted, so `--remote` never hits this path today). A distinct `host_not_consumable` verdict + a registry-side host lint is a doctor-internal taxonomy change tracked as a follow-up, out of this PR's scope. |
| BC-011/012/013 | PRAISE | 🙏 — the transport seam is what made BC-002/004/006 testable offline. |

Also hardened the socket `error` path to surface the enumerated `transport_error` rather than the raw message (it could carry the pinned IP).

Verification: beacon-schema 59+3 green; freeside-cli 93 pass / 1 env-gated skip; `check-beacon-domain.sh` → network-only.

PR #430, opened 2026-07-03
collections-sot: labelled entities on the worldline spine (supersedes #429)

zkSoju:
## collections-sot — labelled entities on the worldline spine

Contracts/collections become **labelled entities on the same hash-chained shadow-mode ledger as the member graph** (the operator's reframe: labelling members and labelling contract addresses are ONE substrate). The append-only observation chain is the source of truth; the entity projection is a re-fold. The full derive-don't-ask loop:

```
belt + ERC-165 + world ──▶ distill ──▶ propose (born-low) ──▶ ratify (cockpit grant) ──▶ shadow-audit serves it
   raw signals          ground     ai-derived/unread      subjective-only, force-chain    settle gate, fail-closed
```

### Supersedes #429
This branch **merges the sandwich-line spine (#429)** — collections-sot consumes it (hash chain + producer-auth + shadow-audit collections read), and the worktree only held a broken partial leak of it. `#429`'s CI redness is unrelated pre-existing monolith failures (a sietch webhook flake being fixed on its own branch + an `aws-embedded-metrics` dep issue); the spine's own tests pass. Merging this closes #429's work.

### What's here (295 tests, 0 typecheck errors)
- **S1** — identity choke point (`collectionEntityId`), `CollectionEntity` + content-addressed `collection.label.{observed,ratified}` factories, store fold (chain=SoT, projection=fold, verify-on-read). `SubjectKind += 'collection'` (additive; member subjects untouched).
- **S2** — distiller `ground()` (belt GraphQL + ERC-165 dependency-free raw `eth_call`, injectable FR-7 seam) reproduces the 24-collection proof (20 erc721 + 4 erc1155) with zero operator input; `propose` (born-low); `ratify` gated by a single-consume cockpit grant (`~/.claude/.recall-cockpit-grant`), subjective labels only.
- **S3** — `query` (lexical + provenance badge, contested-withheld like /recall); **settle gate** — shadow-audit reads the ratified snapshot fail-closed (`COLLECTION_REGISTRY` env → deprecated break-glass), the G-4 kill-test passes; drift sensor; 6-verb agent-first CLI + live belt/RPC clients.

### Fail-closed / trust invariants
- The chain is the trust root: `getCollectionEntity`/`listCollectionEntities` verify-on-read (a tampered chain is refused/excluded, never served).
- Ratify flips `ai-derived → operator-validated` ONLY via a fresh single-consume grant; derived labels are never ratifiable; a member subject can't be collection-ratified.
- The settle gate serves a collection ONLY when standard ∈ {erc721,erc1155} AND world is operator-validated AND not contested; a missing/malformed snapshot fails loud.
- Drift never re-contests an operator-validated label (respects settled truth).

### Two self-review passes (FAGAN) on the diff, both addressed
- **S1**: a ratify on a derived label appended as attested truth but the fold silently ignored it → fail-closed reject at construction.
- **S2/S3**: a merge-blocker where drift re-proposed the world heuristic every run and would re-contest ratified truth forever → drift never touches an operator-validated label; world heuristic narrowed to an explicit allowlist; cockpit grant made fail-closed on unlink failure.

### Remaining (deploy-only, not in this PR's domain logic)
- Point shadow-audit's `COLLECTION_SNAPSHOT_PATH` at a `collections export-snapshot` build step.
- 7 skipped tests are Postgres integration (need a live DB in CI).
- QMD-source registration of the entity export (documented seam, next cycle).


zkSoju (commented):
## Summary

This PR lands a genuinely well-layered spine — content-addressed observations on a hash chain, capability-gated appends, and a fold that derives collection truth from the ledger. The append path, grant design, and JCS foundation are strong. The primary concerns are two HIGH correctness findings at the *derivation* layer: content-addressed dedup makes legitimate value re-assertion impossible (CSOT-001), and timestamp round-tripping through Postgres can freeze healthy chains as "tampered" (CSOT-002). Both are the classic failure mode of hash-based systems: hashing the wrong equivalence class.

## Architectural Meditation

There's a unifying thread across CSOT-001 and CSOT-002 worth naming: **a hash chain is only as sound as the byte-identity of what it hashes**. Certificate Transparency logs learned this early — they Merkle-hash the stored bytes and never a re-derivation, because any lossy round-trip (timestamps, encodings) turns "verify" into "guess whether serialization is stable." Meanwhile, Kafka's idempotent producer deliberately dedups on *sequence numbers* rather than payload content, precisely because identical payloads legitimately recur in real event streams. This PR made the opposite choice on both axes, and the two HIGH findings fall directly out of it. Neither is hard to fix now; both are miserable to fix after live producers exist.

The second thread: the read path quietly acquired write-path obligations (CSOT-004, CSOT-005). Fail-closed is the right instinct, but "fail closed" should mean *refuse to serve*, not *silently mutate state while serving a read*.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "CSOT-001",
      "title": "Content-addressed event_id makes derived-label value oscillation impossible (A→B→A lockout); one transient RPC failure permanently poisons token_standard",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/protocol/shadow-mode/src/schemas/collection-entity.ts",
      "description": "observedEventId hashes only {entity_id, label, value, source_type} — no seq, epoch, or prior-head component. The fold picks the latest chain seq per label. Consequence: once a label has held value A and later value B, a re-derive back to A produces the ORIGINAL event_id, appendObservationIfAbsent dedups (returns false), no new chain entry is written, and the fold keeps B forever. Concretely: detectTokenStandard (distiller.ts) degrades to 'unknown' on any transient RPC failure, and drift.ts appends that 'unknown' as a new observation. When RPC recovers, the re-derived 'erc721' dedups against the initial propose-time observation and the entity is stuck at 'unknown' permanently; every subsequent drift run reports 'drifted' but can never converge. This also silently pulls the entity out of the shadow-audit settle gate (unknown_standard is fail-closed excluded). The same lockout applies to any legitimate A→B→A transition on collection_key.",
      "suggestion": "Options: (a) don't append a downgrade-to-'unknown' when the current folded value is a known standard and the cause is transport failure (distinguish revert/false from transport-null in EthCall); (b) include a monotonic component (e.g., prior label-winner event_id or chain head hash) in observedEventId so re-assertion after an intervening change mints a new id; or (c) make the fold treat a dedup'd append as a re-assertion (track last-asserted seq separately from first-observed seq). At minimum add a test for the A→B→A sequence.",
      "confidence": 0.85,
      "faang_parallel": "Kafka's idempotent producer dedups on (producer-id, sequence number), never on payload content, because the design team recognized that identical payloads legitimately recur in event streams — 'temperature is 70°' can be true on Tuesday and again on Thursday. Content-addressed dedup conflates 'this exact event was delivered twice' (a transport concern) with 'this fact was asserted twice' (a domain event). Amazon's Dynamo paper made the same distinction with vector clocks: identity of a write includes its causal position, not just its value.",
      "metaphor": "It's a guestbook that refuses to let you sign in because you signed in last year. The book faithfully records that you arrived once — and then structurally cannot record that you left and came back. Anyone reading the last entry concludes you're still gone.",
      "teachable_moment": "When choosing an idempotency key, ask: 'can this exact value legitimately recur later, and does its recurrence carry meaning?' If yes, the key needs a causal or temporal component. Content-addressing is perfect for immutable blobs (git objects) and wrong for observations of mutable state.",
      "connection": "This interacts with the fail-closed settle gate: the lockout doesn't just corrupt a label, it permanently ejects the entity from the audit surface. A correctness bug in the derivation layer becomes an availability bug in the product's core promise."
    },
    {
      "id": "CSOT-002",
      "title": "Postgres timestamptz round-trip can produce false tamper freezes: hash is computed over producer-supplied ISO strings but verified against toISOString() output",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/shadow-mode/src/adapters/postgres-store.ts",
      "description": "computeLinkHash at append time hashes the in-memory observation, whose observed_at/emitted_at are producer-supplied strings. verifyChain recomputes the hash from rowToObservation, which formats the timestamptz columns via .toISOString() (millisecond UTC, always '.000Z'-style). Any producer timestamp that is valid but not byte-identical to toISOString() output — no fractional seconds ('2026-07-03T00:00:00Z'), microsecond precision ('...00.123456Z', truncated by Date), or a non-Z offset — round-trips to a different string, so recomputed hash ≠ stored hash and the chain is FROZEN as tampered (hash_mismatch). zod's default .datetime() accepts variable precision, so this is reachable through the HTTP ingest boundary unless the schema pins exactly 3 fractional digits + Z. All committed tests use canonical '.000Z' timestamps, which is why this doesn't surface.",
      "suggestion": "Canonicalize timestamps at the ingest boundary before hashing (e.g., new Date(observed_at).toISOString() when building the ShadowObservation), or store observed_at/emitted_at as text so the hashed bytes round-trip losslessly, or pin the zod schema to the exact toISOString grammar. Add a PG integration test appending an observation with 'no-millis' and microsecond timestamps.",
      "confidence": 0.65,
      "faang_parallel": "Certificate Transparency logs (RFC 6962, Google's Trillian) hash the exact stored leaf bytes and verify against those same bytes — never against a re-serialization — because any lossy type round-trip (timestamps through a Date type, JSON through a parser) makes 'tamper detection' indistinguishable from 'serialization drift'. Trillian stores opaque byte blobs for precisely this reason.",
      "metaphor": "You photograph a handwritten letter, then later compare a checksum of the original ink against a checksum of a typed transcription. The words match, the bytes don't — and your tamper alarm fires on your own faithful copy.",
      "teachable_moment": "The invariant for any hash-verified store: the bytes you hash must be the bytes you can re-read, byte-identical. Either canonicalize before hashing (one normalization function, applied at exactly one boundary) or store the raw bytes. Letting a typed column (timestamptz) sit between hash-time and verify-time inserts an invisible re-serializer into the trust path.",
      "connection": "The severity is amplified by the fail-closed design being otherwise correct: a false hash_mismatch doesn't degrade gracefully, it freezes the chain — the safety mechanism becomes the outage."
    },
    {
      "id": "CSOT-003",
      "title": "consumeCockpitGrant single-consume guarantee is racy: rmSync({force:true}) reports success even when another process already consumed the grant",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/services/shadow-mode/src/collections/ratify.ts",
      "description": "Two concurrent ratify invocations can both statSync a fresh grant, then both call rmSync with force:true. force:true swallows ENOENT, so the second process's 'removed' is true even though it deleted nothing — both callers return true and both ratifies proceed from ONE grant. The stated contract is 'one gesture = one write' / single-consume. The window is small and the surface is an operator-local CLI, but the property the FAGAN M-1 comment claims (fail-closed unless THIS caller consumed the grant) does not hold under concurrency.",
      "suggestion": "Use an atomic consume: renameSync(grantPath, grantPath + '.' + pid) and treat rename success as the exclusive claim (ENOENT on rename = lost the race → no_grant), then unlink the renamed file. Alternatively open with O_EXCL semantics on a companion lock. Drop force:true so a missing file is distinguishable from a successful removal.",
      "confidence": 0.7,
      "faang_parallel": "This is the canonical check-then-act TOCTOU that Unix solved decades ago with atomic primitives: mkdir(2) and open(O_CREAT|O_EXCL) succeed for exactly one caller. rename(2) is the standard 'claim' operation — Kubernetes leader election and countless lockfile implementations rest on the same atomicity.",
      "metaphor": "Two people reach for the last ticket at a box office where the clerk says 'sold!' whenever the shelf ends up empty — regardless of whose hand actually grabbed it. Both walk away believing they hold the ticket.",
      "teachable_moment": "force:true is an availability flag ('don't error if already gone') being used where exclusivity is the requirement. Whenever a flag exists to suppress an error, ask whether that error was carrying the property you're claiming to enforce."
    },
    {
      "id": "CSOT-004",
      "title": "CLI 'read' verbs (query, export-snapshot) run schema migrations, full-chain boot verification, and freeze-writes without the --yes production guard",
      "severity": "MEDIUM",
      "category": "operational",
      "file": "packages/services/shadow-mode/bin/collections.ts",
      "description": "withStore() calls store.migrate() unconditionally, which (a) executes DDL (sql/0001 + 0002) against DATABASE_URL and (b) runs assertChainsVerified() — a FULL verification of EVERY chain — on every CLI invocation, including query and export-snapshot. Additionally, getCollectionEntity/listCollectionEntities verify-on-read INSERT freeze rows into shadow_chain_state when a chain fails. So the verbs documented as read-only both mutate the database and bypass assertMutable's prod --yes guard (only propose/ratify/drift go through it). The header comment 'Read verbs need only belt+RPC config' contradicts the implementation for query/export-snapshot, which require a DATABASE_URL with DDL privileges.",
      "suggestion": "Split the seam: a readStore() path that skips migrate() (or runs assertChainsVerified only, no DDL) for query/export-snapshot, and reserve migrate() for mutating verbs. Consider making freeze-on-read for Postgres an explicit opt-in or routing read verbs through a role without write grants so fail-closed still holds (refuse to serve) without silent writes.",
      "confidence": 0.8,
      "faang_parallel": "Every mature schema-migration practice (Flyway, Liquibase, Google's production-readiness reviews) separates 'apply migrations' into an explicit deploy-time step, never a side effect of a client connecting. Rails' auto-migrate-on-boot was widely banned in production for exactly this: a read tool holding DDL privileges violates least-privilege, and a version skew between two CLI copies becomes a schema race.",
      "metaphor": "Checking your account balance shouldn't require handing the teller blueprints and a construction permit for the branch. If it does, then anyone allowed to *look* is also empowered to *remodel*.",
      "teachable_moment": "'Read-only' is a property of the required credentials, not of the verb's intent. The honest test: can this code path run under a DB role with only SELECT? If not, the documentation is describing an aspiration, not the implementation.",
      "connection": "Pairs with CSOT-005 — the unconditional boot verification is both a privilege problem (this finding) and a latency ceiling (the next one). One readStore() seam addresses both."
    },
    {
      "id": "CSOT-005",
      "title": "Full-chain verification is O(total ledger) at every boot and O(chain length) on every collection read — no incremental or checkpointed path",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "packages/services/shadow-mode/src/adapters/postgres-store.ts",
      "description": "assertChainsVerified() loads and re-hashes every link + observation of every chain into memory before the store serves anything; getCollectionEntity() runs a full verifyChain (full chain SELECT + rehash) on EVERY read, and listCollectionEntities() does this per entity sequentially. Member-community chains grow unboundedly with ingest volume, so boot time and read latency degrade linearly with ledger size, and deploys block on it. Collections chains are small today, but member chains share the same boot gate.",
      "suggestion": "Acceptable for this cycle's scale, but record the ceiling with an upgrade trigger (the codebase's loa:shortcut convention): e.g., verified-watermark checkpoints per chain (verify only links past the last green watermark at boot), and cap verify-on-read to head-window + watermark. Stream links with a cursor instead of loading the full chain into memory.",
      "confidence": 0.7,
      "faang_parallel": "Bitcoin Core's assumevalid and Certificate Transparency's signed tree heads both solve this exact problem: a trusted checkpoint bounds re-verification to the suffix. Git makes the same trade — it never rehashes your whole history on fetch; object integrity is checked incrementally from trusted refs.",
      "metaphor": "Re-inspecting every rung of the ladder from the ground floor each time you climb one more step. Fine on a stepladder; on a fire escape it means nobody reaches the roof.",
      "teachable_moment": "Any 'verify everything at boot' design carries a hidden SLO: boot time is now a function of total historical write volume. Naming the ceiling and its upgrade trigger (per the repo's own loa:shortcut convention) converts a future outage into a planned task."
    },
    {
      "id": "CSOT-006",
      "title": "A producer JWT that omits the optional `communities` claim receives an UNRESTRICTED cross-community grant",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/services/shadow-mode/src/auth/jwt-producer-policy.ts",
      "description": "ClaimsSchema marks communities optional, and mintGrant defaults it to []. AppendGrant.allowsCommunity treats an empty list as 'unrestricted' (documented as intended for test/migration). The composition means the DEFAULT posture for any production token whose issuer forgets the claim is global write across every community chain — the permissive case is the omission case. The blast radius is bounded by sources/event_names, but community isolation (the per-chain tenancy boundary) silently disappears.",
      "suggestion": "Invert the default at the JWT boundary: require the communities claim (or an explicit sentinel like communities:['*']) for tokens, keeping empty-=-unrestricted only for the in-module operator/test minters. That makes unrestricted scope an explicit issuer decision rather than an omission.",
      "confidence": 0.6,
      "faang_parallel": "AWS IAM's foundational rule is implicit-deny: an absent statement grants nothing, and wildcard access must be spelled '*' by a human. Kubernetes deprecated ABAC partly because its permissive defaults made omissions dangerous; RBAC's empty-binding-means-nothing replaced it.",
      "metaphor": "A keycard form where leaving the 'which floors?' field blank issues a master key to the whole building. The most careless path yields the most powerful credential.",
      "teachable_moment": "Two individually-defensible defaults (schema: optional for ergonomics; grant: empty-means-unrestricted for migration) composed into an unsafe whole. Audit defaults at composition boundaries, not module-by-module — the danger lives in the seam."
    },
    {
      "id": "CSOT-007",
      "title": "AppendGrant unforgeability depends on package-index non-export; deep imports still reach mintGrant unless package exports are sealed",
      "severity": "LOW",
      "category": "security",
      "file": "packages/services/shadow-mode/src/auth/append-grant.ts",
      "description": "mintGrant, operatorMigrationGrant, and collectionProducerGrant are exported from the module (necessarily, for in-package policies) and only omitted from src/index.ts. A consumer can import '.../shadow-mode/dist/auth/append-grant.js' directly and mint arbitrary grants. The WeakSet+symbol design genuinely blocks forging a grant OBJECT from outside, but the doc comment's claim that 'there is no reachable mint' overstates the boundary unless package.json \"exports\" restricts subpath imports. The test only asserts non-export from index.js.",
      "suggestion": "Add an \"exports\" map to packages/services/shadow-mode/package.json exposing only the intended entry points, and soften the module comment to state the actual boundary (index + exports map, not module privacy).",
      "confidence": 0.55,
      "faang_parallel": "Node's package \"exports\" field exists precisely to seal deep-import paths — the same encapsulation motive behind Java 9's module system, which broke the ecosystem's habit of reaching into com.sun.* internals.",
      "teachable_moment": "A security comment should describe the boundary as enforced, not as intended. 'Not exported from index' and 'not importable' are different claims; the one-line exports map upgrades the former into the latter."
    },
    {
      "id": "CSOT-008",
      "title": "chain-admin clear 'auth' compares two env vars settable by the same principal — a deliberateness ceremony, not an authorization boundary",
      "severity": "LOW",
      "category": "security",
      "file": "packages/services/shadow-mode/bin/chain-admin.ts",
      "description": "CHAIN_ADMIN_TOKEN and CHAIN_ADMIN_TOKEN_PRESENTED are both read from the same process environment. Anyone who can launch the CLI with env control (which DATABASE_URL access essentially implies in most deploy setups) can set both to the same value. The comment claims 'db access alone must NOT suffice for a clear', but the mechanism only enforces that the operator typed the secret twice, unless CHAIN_ADMIN_TOKEN is provisioned out-of-band (e.g., only in the deployment secret store, never visible to the DB-credentialed role).",
      "suggestion": "Document the required provisioning split (CHAIN_ADMIN_TOKEN lives only in the operator's secret store, never alongside DATABASE_URL), or verify the token against a value stored server-side (e.g., a hashed token row) so possession of DB write credentials cannot mint the comparison target trivially. If the ceremony is the intent, say so in the comment instead of claiming a boundary.",
      "confidence": 0.6,
      "faang_parallel": "The two-person rule in launch-control systems works because the two keys are held by *different principals*. sudo's re-prompt is honestly documented as a deliberateness check, not an authorization boundary — the distinction this comment currently blurs.",
      "metaphor": "A form that asks you to confirm your own signature by signing again. It proves you meant it; it proves nothing about whether you were allowed to."
    },
    {
      "id": "CSOT-009",
      "title": "Hardcoded production belt-gateway URL as the default live edge",
      "severity": "LOW",
      "category": "operational",
      "file": "packages/services/shadow-mode/src/collections/live-clients.ts",
      "description": "DEFAULT_BELT_URL points at 'https://belt-gateway-production.up.railway.app/graphql'. A developer running `collections sync` with no env silently reads production infrastructure; if the Railway hostname changes, the default rots invisibly. The repo's own conventions (registry as live SoT, not hand-authored constants) argue against embedding a deploy URL in source.",
      "suggestion": "Require BELT_GATEWAY_URL explicitly (fail with a clear message when unset), or resolve it via the freeside registry rather than a literal.",
      "confidence": 0.6,
      "faang_parallel": "The industry pattern is discovery over literals: Netflix's Eureka and Google's BNS names exist so that code names a *service*, and the resolver names the *host*. A hostname in source is a cache with no invalidation.",
      "teachable_moment": "This repo already has the correct primitive (the registry as live SoT per its own CLAUDE.md); the finding is a consumption-gradient slip — the governed path existed and the literal was cheaper in the moment."
    },
    {
      "id": "CSOT-010",
      "title": "Differential classifies unknown last-transfer times as no_backfill_window, so an incomplete lastTransferAtMs source can silently convert true mismatches into parity",
      "severity": "SPECULATION",
      "category": "correctness",
      "file": "packages/services/shadow-audit/src/differential.ts",
      "description": "The deliberate choice (per the in-code FAGAN S3 note) is that an unprovable divergence never inflates parity failures. The flip side: if the lastTransferAtMs provider has gaps (indexer lag, pagination miss), real divergences are classified no_backfill_window and the parity bar (3 consecutive clean runs) can be met on bad data — the differential's whole purpose. There is no counter or log distinguishing 'unknown time' from 'provably pre-subscription', so the failure mode is invisible in differentialLogLine.",
      "suggestion": "Split no_backfill_window into 'pre_subscription' and 'unknown_transfer_time' in the classification (or at least a count in the log line), so a parity verdict earned mostly on unknowns is visible to the operator judging the cutover bar.",
      "confidence": 0.45,
      "faang_parallel": "Google's SRE practice distinguishes 'no data' from 'good data' in monitoring for exactly this reason: an alert that treats absent telemetry as healthy converges to green as the collector fails. The fix is the same shape — count the unknowns and surface them next to the verdict.",
      "teachable_moment": "Flagged as speculation because the current provider may be complete — but the classification collapses two epistemically different states ('proven exempt' vs 'couldn't check') into one label, and the operator judging the cutover bar can't tell which they're looking at. A one-integer counter buys the distinction cheaply."
    },
    {
      "id": "CSOT-011",
      "title": "Drift classification overwrites 'drifted' with 'unknown_standard' — the drift detail survives only in the free-text field",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/shadow-mode/src/collections/drift.ts",
      "description": "When an entity both drifts on collection_key AND re-derives token_standard='unknown' in the same run, `classified` is set to 'drifted' in the label loop and then unconditionally reassigned to 'unknown_standard'. A consumer filtering findings by class loses the drift signal; only the joined detail string retains it. Single-class-per-finding forces an arbitrary priority.",
      "suggestion": "Either make DriftFinding carry a class array/set, or define and document the priority order explicitly (contested > orphaned > drifted > unknown_standard > unratified) and apply it consistently instead of last-write-wins.",
      "confidence": 0.7,
      "faang_parallel": "Static analyzers and compilers (Clang, ESLint) emit one diagnostic per rule violated rather than electing a single winner per file — because downstream tooling filters by code, and a silent merge drops signals from every consumer at once.",
      "teachable_moment": "Last-write-wins on a classification field is an implicit priority order that nobody chose. Making the order explicit costs one comment and a sorted array; leaving it implicit costs a debugging session for whoever filters by 'drifted'."
    },
    {
      "id": "CSOT-012",
      "title": "Capability-based append authorization is genuinely well-constructed",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/services/shadow-mode/src/auth/append-grant.ts",
      "description": "The AppendGrant design layers correctly: module-private symbol guards construction, WeakSet registry defeats prototype-forged objects, assertGrant enforces identity + source/event/community scope at the store boundary, testGrant is runner-gated, and producerPolicyFromEnv fails CLOSED (throws) rather than silently degrading to the structural-only policy — the explicit SHADOW_MODE_ALLOW_STRUCTURAL_POLICY opt-in makes the weaker posture observable. The forged-grant and no-index-export tests pin the properties.",
      "suggestion": "None — carry this pattern forward to other producer boundaries.",
      "confidence": 0.9,
      "faang_parallel": "This is object-capability discipline in the tradition of E and Google's Caja work: authority is a reference you hold, not an ambient property you claim. The fail-closed env policy with an explicit, greppable opt-in for the weaker posture mirrors how Chrome ships kill-switches — degradation is always a visible decision.",
      "connection": "The repo's own ACVP doctrine ('the signature IS the capability; no ambient authority') is realized here in running code rather than prose — the pattern worth propagating to the other producer boundaries named in the registry."
    },
    {
      "id": "CSOT-013",
      "title": "JCS implementation rejects non-plain objects loudly instead of hashing the wrong bytes",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/protocol/shadow-mode/src/jcs.ts",
      "description": "Rejecting Date/Map/Set/class instances with JcsError (rather than letting them silently canonicalize to '{}' or an enumerable subset) closes the classic hash-over-wrong-bytes hole in hand-rolled canonicalizers. The RFC 8785 Appendix vectors, UTF-16 ordering test, and determinism-across-construction-order test give the hash foundation real coverage, and the rationale for local implementation over a dependency is stated in the header.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "Testing against the RFC 8785 appendix vectors is the same conformance discipline protobuf and CBOR canonical-form implementations use: the spec's own vectors are the contract, not the author's intuition about JSON.",
      "connection": "Notably, this module already embodies the byte-identity discipline whose absence causes CSOT-002 one layer up — the cure for the timestamp finding is applying jcs.ts's own philosophy at the storage boundary."
    },
    {
      "id": "CSOT-014",
      "title": "Postgres append transaction is correctly sequenced: head-integrity check before insert, seq assigned inside the advisory lock, lazy genesis in-transaction",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/services/shadow-mode/src/adapters/postgres-store.ts",
      "description": "The append path gets the hard parts right: pg_advisory_xact_lock serializes writers per chain with automatic release on failure; the tampered-head freeze happens BEFORE the new observation insert (no orphan rows); seq is MAX+1 inside the lock (no pre-allocation gaps); the freeze write on the head-check path commits inside the same locked transaction (no duplicate freeze race); and the known append-vs-projection atomicity gap is honestly marked as a loa:shortcut with a REQUIRED upgrade trigger before any live producer. The concurrent-append integration test (8 parallel appends → no fork, no gap) pins the invariant.",
      "suggestion": "None — the documented upgrade trigger (client-scoped transaction through ShadowLedger.ingest) is the right next step before NATS cutover.",
      "confidence": 0.85,
      "faang_parallel": "Ordering the integrity check before the insert, inside the same lock scope, is the write-ahead discipline every serious log system (Spanner's Paxos log, Kafka's leader append path) enforces: validate the tail, then extend it, atomically per partition. The 8-way concurrent-append test is the kind of invariant-pinning test that outlives refactors.",
      "connection": "The honestly-marked atomicity shortcut with a named upgrade trigger is the repo's loa:shortcut convention working as designed — a ceiling that documents itself instead of rotting silently."
    }
  ]
}
```


## Callouts

Three pieces of this PR deserve explicit recognition, because they're the reason the HIGH findings are fixable rather than fatal:

- **[Security] The AppendGrant capability layer (CSOT-012)** is object-capability discipline done properly — construction gated by a module-private symbol, forgery defeated by a WeakSet, and the env policy failing *closed* with a greppable opt-in for the weaker posture. Most teams write "no ambient authority" in a doc; this PR wrote it in a type.
- **[Quality] The JCS canonicalizer (CSOT-013)** refuses to hash what it can't canonicalize, tested against the RFC's own vectors. The irony worth savoring: applying this exact module's philosophy at the Postgres timestamp boundary is the fix for CSOT-002.
- **[Operational] The append transaction (CSOT-014)** sequences its lock, integrity check, and insert correctly, and — rarest of all — documents its known atomicity ceiling with a named upgrade trigger instead of hoping nobody notices. That honesty is what makes the difference between technical debt and a planned mortgage.

The spine is sound. Fix the two equivalence-class bugs at the derivation layer before a live producer writes history you can't amend, and this becomes the ledger the rest of the factory can trust.

zkSoju:
## Bridgebuilder findings — triage + resolution (commit 0dea3091)

Thanks — the two HIGH findings were exactly right. Addressed:

**Fixed (this PR):**
- **CSOT-001 (HIGH)** — `drift` now never lets `unknown` overwrite a *known* `token_standard`. Your Kafka-idempotency parallel nailed it: a transient RPC blip degrading to `unknown` would dedup a later re-derive (content-addressed `event_id`) and strand the entity at `unknown` forever, silently ejecting it from the settle gate. Invariant added: no-information never overwrites information. Regression test covers RPC-down → recovers → stays `erc721`. (The general A→B→A for `collection_key` remains a documented content-addressing limitation; belt keys are stable, and the monotonic-id upgrade is noted for when a real oscillation appears.)
- **CSOT-002 (HIGH)** — the pg store now canonicalizes timestamps (`toISOString`) **before** hashing + storing, so a `timestamptz` round-trip can't falsely freeze a healthy chain. In-memory was already self-consistent, so CI (in-memory) was unaffected — this closes the Postgres member-path hole you found.
- **CSOT-003 (MED)** — `consumeCockpitGrant` now unlinks **without** `force`, so the unlink itself is the atomic race winner (ENOENT → not authorized); `force:true` had masked a lost race.
- **CSOT-004 (MED)** — read verbs (`query`, `export-snapshot`) no longer run DDL migrations.
- **CSOT-011 (LOW)** — `drift` `drifted` is no longer clobbered by `unknown_standard`.

**Triaged as pre-existing #429 spine (not this cycle; flagged for their own follow-up):**
- **CSOT-005** (O(chain) verify-on-read) — a deliberate correctness>perf tradeoff for the fail-closed settle gate; collection chains are short (a handful of labels). Incremental/checkpointed verify is a spine enhancement.
- **CSOT-006** (JWT missing `communities` → unrestricted) — real fail-open in the spine JWT policy; collections don't use the JWT path (module-mint grant). Changing #429's auth semantics deserves its own review, not a silent flip in a collections merge.
- **CSOT-007 / CSOT-008** (deep-import reaches `mintGrant`; chain-admin two-env ceremony) — spine hardening, pre-existing.
- **CSOT-010 (SPECULATION)** — differential `no_backfill_window` classification, sandwich-line S3.
- **CSOT-009 (LOW)** — belt-gateway default URL is already `BELT_GATEWAY_URL`-overridable; the default is the real endpoint.

The 3 PRAISE items (capability-based append auth, JCS loud-rejection, pg append sequencing) are all spine work from #429. Merging.

PR #433, opened 2026-07-03
feat(datastore-legibility S1 · platform): ordering data-store self-report + host_fp re-home

zkSoju:
**S1 settle gate — the PLATFORM half.** Part of the datastore-legibility cycle's S1 vertical slice, split by domain per the ADR-007 firewall (network half in the sibling PR `datastore-legibility/network`). The two are independent (runtime-coupled only; tests use fixtures) and can merge in any order.

### What lands
- **S1-T2 `host_fp` helper** — the salted data-store correlation fingerprint (`HMAC_SHA256(CLUSTER_FP_SALT, engine://host:port/db)[:16]`; credentials NEVER in the preimage). **Re-homed** (see below) to a new src-shipping leaf `@freeside/cluster-fp`.
- **S1-T3 ordering self-report** — authed `GET /admin/data-store` on the ordering Hono app behind the existing `SERVICE_TOKEN` Bearer gate. `OrderStore.dataStoreFacts(salt)` → `datastore.report.v1` `{engine, host_fp, reachable, migrations_applied|null, store}`. PostgresOrderStore derives `host_fp` from `pool.options` — NEVER the raw pool, NEVER the connection string. InMemory → honest absent-store shape.

### Re-home (operator-ratified)
S1-T2 originally landed `host-fp.ts` in `@freeside/adapters`, but **no cell could import it**: adapters' dist build is broken (`@freeside/core/ports` missing → `tsc` fails), it ships `dist`-only (`files: ["dist"]`), and its `file:` dep is a stale hard-copy `pnpm install` won't refresh. Moved verbatim (9/9 tests) to `@freeside/cluster-fp`, a src-shipping leaf (`exports → ./src/index.ts`, the proven `@freeside/*-protocol` pattern). Cells import ONE impl, no build step. Routes *around* the broken adapters dist chain (fixing it is out of scope). SDD C-2 amended in-place.

### Verification
- ordering suite **131/131**, cluster-fp **9/9**, `tsc` clean.
- AC (9 tests): authed→shape; unauthed→401; connection string / password / host / user NEVER in the body (NFR-1); `host_fp` matches cluster-fp derivation; `reachable=false` is a fact not a throw; absent store honest.

### Domain purity
Net diff touches **platform** (`packages/services/ordering`) + **shared** (`packages/cluster-fp`, `grimoires/loa/*`) only — **0 network paths**. Passes `path-domain-check`.


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#433. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "clusterFpSalt() silently falls back to a fixed dev salt outside Railway/production — contradicts the SDD's fail-loud requirement",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/cluster-fp/src/host-fp.ts",
      "description": "SDD §6.4 states: 'Missing CLUSTER_FP_SALT on a cell → the cell FAILS LOUD at the endpoint (500 with a clear message), never emits an unsalted fp that would mismatch every other cell.' The implementation only throws when `RAILWAY_ENVIRONMENT` is set or `NODE_ENV === 'production'`; otherwise it returns the constant `'dev-unsalted-fp'`. The PRD (NFR-2) explicitly scopes AWS/terraform cells into the same self-report path. An AWS-hosted cell (no RAILWAY_ENVIRONMENT) that starts without `NODE_ENV=production` will emit a well-formed, plausible-looking 16-hex fingerprint derived from the dev salt. Every comparison against a correctly-salted cell then reads as `contested` or as a spurious non-fan-out — i.e. the legibility layer lies, which is precisely the failure mode the cycle exists to prevent. It is also a silent fail-open: the deployed-context detector is a heuristic, not a fact.",
      "suggestion": "Invert the default: require CLUSTER_FP_SALT unconditionally and let an explicit opt-in (e.g. `CLUSTER_FP_SALT_DEV=1` or a test-only injected salt) enable the dev constant. Alternatively, tag the report when the dev salt was used (`salt: 'dev'`) so the aggregator can render `unsalted` as its own status rather than a comparable fingerprint. Whichever path, keep the salt-source visible in the self-report so a misconfigured cell cannot masquerade as coherent.",
      "confidence": 0.7
    },
    {
      "id": "F-002",
      "title": "/admin/data-store is fully open when SERVICE_TOKEN is unset — a missing env var makes cluster topology public",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/services/ordering/src/intake.ts",
      "description": "The route guards with `if (deps.serviceToken && c.req.header('authorization') !== ...)`. When `serviceToken` is absent the check is skipped entirely and the endpoint serves engine, host_fp, reachability and migration facts to any caller. The test at packages/services/ordering/src/__tests__/data-store-report.test.ts:75-79 pins this as intended ('local/dev posture'). PRD §7.5 SKP-002 is explicit that the self-report must be on an authenticated surface and NOT public, because 'engine + reachability + migration-count is topology'. A deploy that forgets to set SERVICE_TOKEN therefore silently publishes topology — fail-open in a design whose stated floor is fail-closed. This mirrors the existing write-route posture, but a write route failing open is loud (someone places orders); a read route failing open is silent forever.",
      "suggestion": "Require the token for `/admin/data-store` regardless of the local/dev posture used by the write routes, or fail closed at boot: if the cell is serving `/admin/data-store` and no SERVICE_TOKEN is configured, either refuse to register the route or return 503. If a dev bypass is genuinely needed, gate it on an explicit `ALLOW_UNAUTHED_ADMIN=1` rather than on the absence of a secret.",
      "confidence": 0.6
    },
    {
      "id": "F-003",
      "title": "Auth check is a non-constant-time string compare and duplicates the existing requireToken helper",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/services/ordering/src/intake.ts",
      "description": "The new handler inlines `c.req.header('authorization') !== \\`Bearer ${deps.serviceToken}\\`` rather than reusing the `requireToken` helper defined a few lines below inside the `if (deps.orchestrator)` block. Two consequences: (a) the comparison is not timing-safe, whereas the reference cell for this cycle (shadow-audit, per SDD §1.9 / server.ts:99-102) uses a timing-safe compare — the new surface is weaker than the pattern it claims to follow; (b) the duplicated predicate will drift from `requireToken` the next time the auth posture changes.",
      "suggestion": "Lift `requireToken` out of the `if (deps.orchestrator)` block so the data-store route reuses it, and implement it with `crypto.timingSafeEqual` over equal-length buffers (guarding the length-mismatch case) to match the shadow-audit gate.",
      "confidence": 0.75
    },
    {
      "id": "F-004",
      "title": "An unparseable connection string is indistinguishable from an absent store (both report host_fp: null)",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/cluster-fp/src/host-fp.ts",
      "description": "`parseConnectionParts` returns null when `new URL()` throws — which happens for real-world DATABASE_URLs containing unescaped special characters in the password (`@`, `/`, `#` are common in generated credentials). `PostgresOrderStore.dataStoreFacts` then reports `{ engine: 'postgres', host_fp: null, store: 'postgres' }`. The aggregator (not yet written) must distinguish this from the legitimate absent-store shape and from a cell that genuinely has no derivable binding; today it cannot, because the only signal is a null fingerprint. A cell whose fingerprint failed to derive would be silently treated as 'no drift detectable' instead of surfacing as a defect — a legibility gap that reads as a clean row.",
      "suggestion": "Return a discriminated result from the parse (`{ ok: false, reason: 'unparseable' }` vs `null` for empty) and surface it in DataStoreFacts as an explicit field (e.g. `host_fp_error: 'unparseable_connection_string'`), so `doctor --data` can classify it as a fault rather than silence. Add a unit test with an unescaped `@` in the password.",
      "confidence": 0.65
    },
    {
      "id": "F-005",
      "title": "DataStoreFacts.reachable is non-nullable, so the SDD's absent-store shape (reachable: null) cannot be expressed",
      "severity": "MEDIUM",
      "category": "design",
      "file": "packages/services/ordering/src/store.ts",
      "description": "SDD §12 A-2 specifies the absent-store self-report as `{ engine: null, host_fp: null, reachable: null, migrations_applied: null, store: 'none' }` — 'legibly \"this cell has no datastore\", never a guess'. The interface declares `reachable: boolean`, and `InMemoryOrderStore.dataStoreFacts` therefore returns `reachable: true` for a store that has no database at all. A downstream classifier reading `reachable: true` on a store with no DB is being told something that isn't a fact about a datastore. The shape also uses `store: 'memory'` where the SDD's absent-store shape is `store: 'none'` (the union does contain 'none', but nothing emits it). Since S2-T1 will apply this same shape to shadow-audit and operator-dash (genuinely DB-less cells), the type will block the spec'd behavior at exactly the point it matters.",
      "suggestion": "Widen to `reachable: boolean | null` and have absent-store implementations return null, reserving `true`/`false` for cells that actually probed a database. Reconcile the `'memory'` vs `'none'` distinction explicitly in the type's doc comment (memory = a real backing store that isn't a DB; none = no store at all), or collapse them if the aggregator will treat them identically.",
      "confidence": 0.6
    },
    {
      "id": "F-006",
      "title": "Shared self-report contract type lives in the ordering package, guaranteeing duplication across cells",
      "severity": "MEDIUM",
      "category": "design",
      "file": "packages/services/ordering/src/store.ts",
      "description": "`DataStoreFacts` is defined in ordering's `store.ts`. SDD Q-2 flags exactly this ('Where does the DataStoreSelfReport contract type live … Prefer one shared type (queryable-not-copied ethos)') and leaves it open. Sprint S2-T1 adds this endpoint to shadow-mode, shadow-audit, worker and operator-dash, and S3-T1 has freeside-cli decoding it — five more consumers, none of which can import from `@freeside/services-ordering`. The natural home already exists in this very PR: `@freeside/cluster-fp` is a zero-dependency src-shipping leaf that every cell must already import to compute host_fp.",
      "suggestion": "Move `DataStoreFacts` (and the `'datastore.report.v1'` literal) into `@freeside/cluster-fp` and re-export from ordering's `store.ts` if convenient. This resolves Q-2 with the package the PR just created, before four more cells copy the shape.",
      "confidence": 0.75
    },
    {
      "id": "F-007",
      "title": "OrderStore port gains an infrastructure-legibility method (dataStoreFacts) — a domain port now knows about cluster fingerprints",
      "severity": "REFRAME",
      "category": "architecture",
      "file": "packages/services/ordering/src/store.ts",
      "description": "`dataStoreFacts(salt: string)` is added to the `OrderStore` interface, which otherwise speaks only order-domain language (placeOrder, transition, listByState). Every present and future implementer of OrderStore — including test doubles in other suites — must now implement a method about HMAC salts and Postgres pool options. The concern being modeled (this cell's DB topology) is a property of the process/adapter, not of the order store. A separate `DataStoreReporter` port implemented by the Postgres adapter (and a trivial absent-store reporter) would keep the domain port clean and would compose across the other four cells, none of which have an OrderStore.",
      "suggestion": "Extract a one-method `DataStoreReporter { dataStoreFacts(salt): Promise<DataStoreFacts> }` port (co-located with DataStoreFacts in cluster-fp), have PostgresOrderStore implement both, and have `createIntakeApp` take the reporter as a distinct dep. This is also the shape S2-T1 needs for cells that have no store at all.",
      "confidence": 0.6
    },
    {
      "id": "F-008",
      "title": "The reachability probe has no timeout and consumes a pool connection per request",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/services/ordering/src/store-postgres.ts",
      "description": "`await this.pool.query('SELECT 1')` in `dataStoreFacts` will hang indefinitely against a DB that accepts the TCP connection but never responds (the classic partial-outage case), rather than reporting `reachable: false`. The SDD places the 5s timeout on the aggregator side, which protects the CLI but not the cell: the request holds a pool connection for the duration. Combined with F-002 (route open when SERVICE_TOKEN is unset), an unauthenticated caller can exhaust the order store's connection pool by hammering `/admin/data-store` — the same pool that serves order writes.",
      "suggestion": "Bound the probe: `pool.query({ text: 'SELECT 1', ... })` with a statement_timeout, or race it against a short `setTimeout` (2-3s) and report `reachable: false` on expiry. Consider a short-lived cached result (e.g. 5s TTL) so repeated polls do not translate 1:1 into DB round-trips.",
      "confidence": 0.7
    },
    {
      "id": "F-009",
      "title": "clusterFpSalt() is called per-request inside the handler, so a missing salt surfaces as an opaque 500 at first probe instead of at boot",
      "severity": "LOW",
      "category": "reliability",
      "file": "packages/services/ordering/src/intake.ts",
      "description": "`deps.store.dataStoreFacts(clusterFpSalt())` evaluates the salt on every request. In a deployed context without CLUSTER_FP_SALT this throws inside the Hono handler, producing an unhandled 500 the first time the operator runs `doctor --data` — which the aggregator will read as `unreachable`, i.e. the misconfiguration is misclassified as a network problem rather than a config problem. Failing loud is correct; failing loud at the right time and with the right classification is better.",
      "suggestion": "Resolve the salt once at app construction (or read it in a config module at boot) so a missing salt fails the deploy, not the first probe. If the endpoint must stay up, catch the error and return a distinct 500 body (`{ error: 'missing_cluster_fp_salt' }`) that the aggregator can classify separately from `unreachable`.",
      "confidence": 0.65
    },
    {
      "id": "F-010",
      "title": "'Regression pin' test pins nothing — it compares hostFp to itself",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "packages/cluster-fp/src/__tests__/host-fp.test.ts",
      "description": "The final test, `it('hostFp is stable for a fixed preimage (regression pin)')`, asserts `expect(hostFp(parts, SALT)).toBe(hostFp(parts, SALT))` — a tautology that passes for any deterministic function, including one whose preimage construction has silently changed. host_fp is a cross-cell correlation id: if the preimage format ever changes (e.g. the dangling-colon issue in F-011 gets 'fixed'), every ratified `data_store.host_fp` label in registry.yaml becomes contested simultaneously, and no test in this suite would catch it. A regression pin must hard-code the expected digest.",
      "suggestion": "Replace with a golden-value assertion: `expect(hostFp({engine:'postgres',host:'db.internal',port:'',db:'shadow'}, 'test-cluster-salt')).toBe('<literal 16 hex>')`. That single line is what makes the preimage format a contract rather than an implementation detail.",
      "confidence": 0.95
    },
    {
      "id": "F-011",
      "title": "host_fp preimage diverges from the SDD spec: a dangling colon is left when the port is elided",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/cluster-fp/src/host-fp.ts",
      "description": "SDD §5.4 specifies the preimage as `${host}${defaultPort ? '' : ':' + port}`, i.e. `postgres://db.internal/shadow` when the port is default. The implementation builds `${parts.host}:${parts.port}` unconditionally, producing `postgres://db.internal:/shadow`. Functionally harmless (deterministic, still collision-free), but it means the shipped fingerprint is not the one the design document says it is — and any second implementation written from the SDD (an external `*-api` cell adopting FR-1, per FR-7 phase 2) will compute a different fingerprint for the same database and read as permanently contested.",
      "suggestion": "Either normalize the preimage to match the SDD (drop the colon when the port is elided) or amend the SDD to the shipped format. Whichever way, pin it with the golden-value test from F-010 so the format is a contract the phase-2 cells can be written against.",
      "confidence": 0.85
    },
    {
      "id": "F-012",
      "title": "Default-port elision logic is duplicated in store-postgres.ts and will drift from cluster-fp's DEFAULT_PORTS",
      "severity": "LOW",
      "file": "packages/services/ordering/src/store-postgres.ts",
      "description": "The discrete-parts branch re-implements the elision inline (`opts.port && opts.port !== 5432 ? String(opts.port) : ''`) and hard-codes `engine: 'postgres'`, duplicating the normalization that `parseConnectionParts` owns via `DEFAULT_PORTS`/`ENGINE_ALIASES`. The test only proves the two paths agree today. If cluster-fp's port table changes (the SDD, for instance, calls for eliding pgbouncer's 6432, which the shipped DEFAULT_PORTS does not), the URL path and the pool-options path will silently produce different fingerprints for the same database — and four more cells are about to copy this pattern.",
      "suggestion": "Add `partsFromPoolOptions(opts)` (or a generic `partsFrom({connectionString?, host?, port?, database?})`) to `@freeside/cluster-fp` and have every cell call it. Normalization belongs in exactly one place — that is the whole argument for the package existing.",
      "confidence": 0.8
    },
    {
      "id": "F-013",
      "title": "The @freeside/cluster-fp file: dependency reproduces the stale-hard-copy failure mode the re-home was meant to escape",
      "severity": "MEDIUM",
      "category": "build",
      "file": "packages/services/ordering/package.json",
      "description": "SDD §1.4 C-2 documents why `@freeside/adapters` was unusable, and one of the three cited reasons was: 'its `file:` dep is a stale hard-copy `pnpm install` won't refresh.' The new package fixes the other two reasons (broken dist build, `files: ['dist']`) but is consumed through the identical mechanism: `\"@freeside/cluster-fp\": \"file:../../cluster-fp\"`. If pnpm's `file:` protocol copies rather than links in this repo's configuration, edits to `packages/cluster-fp/src` will not propagate to ordering's node_modules without a forced reinstall — meaning the shared helper can silently diverge per-consumer, which for a correlation fingerprint means cells computing different fps from the same code. The `@freeside/*-protocol` packages use the same protocol and apparently work, so this may already be resolved by repo config; it should be confirmed rather than assumed, given the SDD names it as a known hazard.",
      "suggestion": "Confirm whether these `file:` deps resolve as symlinks in this repo (check node_modules/@freeside/cluster-fp is a link, not a copy). If they are copies, switch to `link:../../cluster-fp` or a pnpm workspace so the source is shared by reference. Either way, record the answer in the SDD C-2 amendment, since the amendment currently cites stale-hard-copy as a defect while shipping the same protocol.",
      "confidence": 0.5
    },
    {
      "id": "F-014",
      "title": "cluster-fp ships TypeScript source as its runtime entrypoint — verify ordering's production start path can resolve it",
      "severity": "HIGH",
      "category": "build",
      "file": "packages/cluster-fp/package.json",
      "description": "`main`/`types`/`exports` all point at `./src/index.ts`. Node cannot execute `.ts` without a loader. The package comment asserts 'the cell runtimes resolve TS source directly via tsx/vitest', and the `@freeside/*-protocol` precedent is cited — but ordering is a deployed Railway cell, and if its production start command runs compiled output (`node dist/...`), the emitted `import '@freeside/cluster-fp'` will resolve to `src/index.ts` and throw `ERR_UNKNOWN_FILE_EXTENSION` at boot. A `tsc` build of ordering also will not emit the dependency's source into `dist`. This is the difference between 'the tests pass' and 'the cell starts'. The blast radius is the ordering service failing to boot in production, which the vitest suite cannot detect.",
      "suggestion": "Before merge, verify ordering's Dockerfile/start script and, if it runs compiled JS, either add a build step to cluster-fp (with `exports` pointing at `dist` and a `publishConfig`-style dual entry) or confirm ordering bundles its dependencies. A single `node dist/index.js` smoke check against the built image settles it. If the protocol packages are already consumed this way in a running deploy, note that in the SDD amendment so the precedent is evidence rather than an assumption.",
      "confidence": 0.4
    },
    {
      "id": "F-015",
      "title": "NOTES.md and the SDD/sprint still point at packages/adapters/storage/host-fp.ts — the file the PR deliberately did not create",
      "severity": "MEDIUM",
      "category": "documentation",
      "file": "grimoires/loa/NOTES.md",
      "description": "The resume note records 'S1-T2 (host_fp salted-correlation helper in packages/adapters/storage/host-fp.ts, 9 tests)', but the helper actually shipped at packages/cluster-fp/src/host-fp.ts, and the SDD's own C-2 amendment explains at length why the adapters home was abandoned as unconsumable. The stale path also persists in sdd.md §5.4 ('packages/adapters/storage/host-fp.ts (new)'), §8 Phase 1, §11 Appendix B, and sprint.md S1-T2. NOTES.md is the resume surface for `/run-bridge --resume`: the next session will be handed a path that does not exist, in a doc updated by this same PR.",
      "suggestion": "Update NOTES.md, sdd.md §5.4/§8/§11.B, and sprint.md S1-T2 to `packages/cluster-fp/src/host-fp.ts` (or `@freeside/cluster-fp`). The C-2 amendment already tells the story — the rest of the document just needs to agree with it.",
      "confidence": 0.9
    },
    {
      "id": "F-016",
      "title": "PRD G-5 still specifies host_fp = sha256(host)[:12] while §7.5 specifies HMAC_SHA256(salt, ...)[:16]",
      "severity": "LOW",
      "category": "documentation",
      "file": "grimoires/loa/prd.md",
      "description": "Goal G-5 reads 'only a one-way `host_fp = sha256(host)[:12]`' and FR-1 repeats `host_fp: sha256(host)[:12]`, but the flatline cure in §7.5 SKP-001 supersedes both with the salted HMAC truncated to 16 hex — which is what shipped. A reader who stops at the goals table (or an implementer of a phase-2 external cell who reads FR-1 rather than §7.5) will build the wrong fingerprint and read as permanently contested against every in-monolith cell.",
      "suggestion": "Rewrite G-5 and FR-1 to reference the §7.5 definition directly rather than restating a superseded formula. Since §7.5 exists precisely to correct these lines, they should point at it rather than contradict it.",
      "confidence": 0.85
    },
    {
      "id": "F-017",
      "title": "migrations_applied is hard-coded null in the only implementation, so the field is currently unexercised",
      "severity": "LOW",
      "category": "design",
      "file": "packages/services/ordering/src/store-postgres.ts",
      "description": "`PostgresOrderStore.dataStoreFacts` always returns `migrations_applied: null` with the honest comment that it runs raw SQL files with no version table to count. This is compliant with the spec (nullable, never fabricated) and correct for this store — but it means no cell in the cycle currently returns a number, so the aggregator's handling of a real count will ship untested, and the field's stated purpose (SDD §3.4: distinguishing cells at different migration levels on a shared DB) is not yet served by anything.",
      "suggestion": "Either wire one real counter now (shadow-mode's ledger or the drizzle `__drizzle_migrations` table named in SDD §3.4 / Q-3) so the non-null path has a live producer, or note in the sprint that the field is a reserved slot until a cell with a countable migration table adopts FR-1.",
      "confidence": 0.7
    },
    {
      "id": "F-018",
      "title": "Secret-hygiene is tested as a property of the response body, not just asserted in prose",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/services/ordering/src/__tests__/data-store-report.test.ts",
      "description": "The NFR-1 test reads the raw response text and asserts the absence of the password, the connection-string scheme, the host, and the user — using a fixture connection string chosen so each of those has something concrete to catch. Paired with the cluster-fp test proving that changing user/password yields the *same* fingerprint (credentials provably absent from the preimage), the no-leak invariant is enforced by executable checks at both the derivation and the serialization boundary, rather than by a comment. That is the right shape for a requirement whose violation is silent and irreversible.",
      "suggestion": "Carry this exact test verbatim into each cell as S2-T1 fans the endpoint out to shadow-mode, worker, shadow-audit and operator-dash — the assertion is cheap and the failure mode it guards is not.",
      "confidence": 0.9
    },
    {
      "id": "F-019",
      "title": "reachable:false is reported as a fact rather than thrown as a 5xx",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/services/ordering/src/store-postgres.ts",
      "description": "The probe catches the failure and returns `reachable: false` while still emitting a valid host_fp derived from config. This keeps a DB-down cell legible (you can still see which database it is pointed at and whether that binding drifted) instead of collapsing it into an opaque error — exactly the distinction the SDD draws between `unreachable` (we can't see) and a healthy report of an unhealthy DB. The comment 'reachability is a fact, not a 5xx' names the invariant at the point it is enforced.",
      "suggestion": "No change. Note that F-008 (no timeout on the probe) is the one gap in this otherwise correct handling — a hanging DB currently produces neither `true` nor `false`.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

PR #437, opened 2026-07-04
feat(shared/immune): autopoiesis — 3 immune cells (Trustworthy Green + Consumption Doctor + False-Green Sensor)

zkSoju:
## Autopoiesis — the factory's first three immune cells

Grows the factory's immune system so it can **see and reject its own broken states** without the operator standing in as the immune system by hand. Three cells + an anti-rot flip mechanism, each grounded in a filed bug, each emitting the one shared NFR-7 verdict record (exit code IS the verdict: `0` HEALTHY / `1` INSUFFICIENT / `2` PROBLEM).

Cut **fresh off `main`** (single-domain, `path-domain-check` passes by construction). All commits `domain:shared` — `tools/`, `.github/`, `grimoires/`; `.claude/` System Zone untouched.

### Sprints

| Sprint | Cell | What it catches | Grounds bead |
|---|---|---|---|
| **S1 #407** — Trustworthy Green (settle gate) + flip mechanism | `scope-checks-sensor.sh` + `lib/scope-classify.sh`, `immune-verdict.schema.json` + ajv, `flip-ledger`/`flip-report`/`flip-promote`, `quarantine-equivalence-gate.sh`, jq-crash fix | a per-PR green that isn't a real *this-PR* signal; advisory rotting into a numb gate | `integration-tests-numb-gate`, `cluster-compliance-audit-crash` |
| **S2 #408** — Consumption Doctor | `consumption-doctor.sh` | a shared package "healthy" while no real consumer can import it | `adapters-dist-unconsumable` |
| **S3 #409** — False-Green Sensor | `false-green-sensor.sh` + upstream issue | a run/bridge that did nothing reporting success | `run-bridge-resume-silent-noop` |

### Real-tree proof (S3-T5 E2E — all four goals)

- **G-1**: `scope-checks-sensor.sh` over a single-package diff → `pass` (1 pkg / 1 concrete `pnpm --filter` cmd); a lockfile/workflow change → `full` (fallback-to-full).
- **G-2**: `consumption-doctor.sh @freeside/adapters` → **flag** (ships `dist`, dist unbuilt — the exact break that hid); `@freeside/ordering-protocol` → **pass**; a zero-consumer pkg → `no-consumer` (distinct honest state).
- **G-3**: `false-green-sensor.sh` on a replayed `JACKED_OUT + 0/0/0` state → **suspect**; a real-work state → **pass**; absent/partial/malformed `.run` state → **insufficient** (missing counter ≠ zero; never a false pass, IMP-009).
[...]

PR #439, opened 2026-07-05
feat(network/registry): cadence-ledger — ADR-012 Phase 0 + liveness expectation record (sprint 410)

zkSoju:
## Cadence Ledger — Declare the Liveness Contract

Lands **ADR-012 Phase 0** (the declared `service` health contract, SoT = freeside-registry), extended with the staleness dimension: a first-class **`expectations[]` record** (`http` | `graphql-lag` | `event-max-age`, discriminated union — `gh-workflow` structurally excluded so premature use fails decode). Single network-domain PR.

Cycle provenance: re-homed from a kalfu exploration that had grounded on the superseded standalone `freeside-cli` repo (now marked SUPERSEDED there); the lane continues ADR-012 — schema here, single reader (`loa-cli/lib/probe.mjs`, already field-compatible) grows `probe_kind` dispatch **next cycle**.

### What's in
- **Schema** (`packages/freeside-registry/src/registry.ts`): `ServiceBlock` (the exact 5 fields probe.mjs reads + `probed_at`/`probe_source` provenance) + `expectations[]` union + decode-time filters (service requires present-and-equal entry `deployment_url`; unique refs; non-empty thresholds). Additive-only; absence stays valid.
- **Data** (`registry.yaml`): 7 `service` blocks (activities, identity, inventory, ordering, score, sonar, storage) — **every value from a live probe wave 2026-07-05**, provenance on each. 4 cells deliberately blockless (derive-from-absence). Sonar carries the first expectation: `chain-lag` (belt-gateway `chain_metadata`, all six SCALE.md per-chain SLOs).
- **Tests**: 3 new suites (44/44 total) incl. red-fail negatives and the score-api **302 anti-transcription tripwire**.
- **CI**: `registry-cli-tests.yml` — the G-4 consumer gate; beacon-schema → registry → cli install order (file:-dep stale-copy hazard reproduced locally before encoding it).

### Probe wave found real drift (2 weeks after the ADR appendix)
| Cell | Drift caught |
|---|---|
| identity-api | appendix host is Railway-edge dead; canonical `-317b` declared |
| inventory-api | appendix (`/` 401 static-key) was the old edge-walled deploy; now open `/health` 200 |
| sonar-api | body marker renamed `"message"` → `"name"` |
| score-api | `/` now 302 → `/v1/health`; direct path declared, **302 never encoded** (DO-NOT-TRANSCRIBE flag discharged) |

**Verify-or-omit exercised**: the assumed SVM run-marker projection does NOT exist in the live GraphQL schema — `svm-reconcile` expectation omitted (event recency ≠ run recency; declaring it would rebuild the quiet-vs-down trap). Gap + future candidates recorded in sonar's cell notes.

### ⚠ Carried requirement (next cycle)
**loa-cli `probe_kind` dispatch MUST host-guard declared expectation endpoints before fetching them** — declaration-layer endpoints are untrusted data; the SSRF guard lives in the consumer (review feedback item 3).

### Gate trail
- Review: iteration 1 CHANGES_REQUIRED (cross-model DISS-001 = the deployment_url null-escape split-brain) → fixed `678efb20` → iteration 2 **All good**
- Audit: **APPROVED** (no C/H/M findings; audit-mode cross-model pass config-disabled, recorded)
- Suites: registry 44/44 + `tsc --noEmit` clean; freeside-cli 94 pass / 0 fail with **zero changes in that package**
- `tools/check-beacon-domain.sh --since origin/main` → single-domain **network**
- Deleted files: none


zkSoju (commented):
## Summary

Sprint 410 lands a well-disciplined Phase 0: the liveness-expectation contract is enforced in the type system rather than by convention, with negative tests that pin diagnostics, not just rejections. Primary concerns are two MEDIUMs — a cross-field gap that lets an undischargeable `http` expectation decode cleanly, and a G-4 gate built on `npm install` rather than lockfile-reproducible `npm ci`. Verdict: **COMMENT** — nothing blocking, but the two MEDIUMs deserve a follow-up before the gate is treated as load-bearing.

The architectural through-line worth naming: this PR consistently chooses "make the invalid state unrepresentable" over "lint for the invalid state" — the gh-workflow exclusion, the split-brain filter, the shared-schema decode test. The two MEDIUMs are simply places where that same posture wasn't carried to completion, which makes them cheap to fix: the mechanism (ModuleEntry filters, CI lane) already exists.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "CL-410-001",
      "title": "http expectation with absent target has no cross-field guarantee of a probeable service block",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/freeside-registry/src/registry.ts:73",
      "description": "HttpExpectation documents that an absent target means 'consumers probe the cell's own service block', but the schema places no constraint tying a target-less http expectation to the presence of a ServiceBlock on the same entry. A cell can validly declare `probe_kind: http` with no target on an entry that has no `service` block, producing a declaration no consumer can discharge. The ModuleEntry-level filter already demonstrates the cross-field pattern (service ⇒ deployment_url), so the mechanism exists.",
      "suggestion": "Extend the ModuleEntry filter (or add a second one) to reject entries where any expectation is `probe_kind: http` with no `target` while `service` is absent — keeping the 'fails loudly at decode' posture consistent with FR-3/G-5.",
      "confidence": 0.55,
      "faang_parallel": "Protobuf's long-standing pain with cross-field invariants: proto3 can express 'field exists' but not 'field A implies field B', so Google teams layer protoc-gen-validate / CEL constraints on top. Effect Schema's filter combinator is exactly that layer — this PR already uses it once, it just needs a second application.",
      "metaphor": "A shipping label that says 'deliver to the address on file' attached to a customer with no address on file. The label prints fine; the driver discovers the problem.",
      "teachable_moment": "When a schema comment says 'absent X means consumers do Y instead', that sentence is an invariant wearing a comment's clothing. Every 'means' in a doc comment is a candidate filter.",
      "connection": "Same closure the split-brain filter (CL-410-013) performs for service⇒deployment_url — this is the one remaining implication left as prose."
    },
    {
      "id": "CL-410-002",
      "title": "CI gate uses `npm install` rather than `npm ci` — the G-4 gate is not lockfile-reproducible",
      "severity": "MEDIUM",
      "category": "ci",
      "file": ".github/workflows/registry-cli-tests.yml:37",
      "description": "All three package steps run `npm install --no-audit --no-fund`. If package-lock.json files exist, `npm install` may resolve newer transitive versions than what was tested locally, meaning the consumer gate can go red (or worse, green) for reasons unrelated to the PR under test. For a gate whose entire purpose is mechanical proof that consumers keep decoding, non-reproducible dependency resolution is a soft spot.",
      "suggestion": "Use `npm ci` where a lockfile exists (falling back to `npm install` only for packages without one), and consider `actions/setup-node` cache to keep the lane fast.",
      "confidence": 0.55,
      "faang_parallel": "Google's hermeticity doctrine in Bazel/Blaze: a test that can observe the network's current state is not a test, it's a weather report. `npm ci` is the cheapest hermeticity npm offers.",
      "metaphor": "A smoke detector wired to the neighbor's electrical panel — most days it works, but when it beeps you don't know whose house is on fire.",
      "teachable_moment": "A gate's authority comes from attributability: red must mean 'this PR broke it'. Any floating input (unpinned deps, network fetches, clock reads) dilutes that attribution.",
      "connection": "The install-order comments (CL-410-015) show the lane's authors already think about dependency freshness — this extends the same care from local file: deps to the public registry."
    },
    {
      "id": "CL-410-003",
      "title": "Workflow has no `permissions` block (default token scope)",
      "severity": "LOW",
      "category": "security",
      "file": ".github/workflows/registry-cli-tests.yml:12",
      "description": "The workflow omits a top-level `permissions:` key, so the GITHUB_TOKEN gets the repository default, which may include write scopes. This lane only checks out and runs tests; it needs nothing beyond `contents: read`.",
      "suggestion": "Add `permissions: { contents: read }` at the workflow level, matching least-privilege hardening.",
      "confidence": 0.7,
      "faang_parallel": "GitHub's own hardening guide and OpenSSF Scorecard both flag token-permissions as a top supply-chain check after the 2021–22 wave of workflow-injection incidents; write-scoped tokens in test lanes were the amplifier in several of them.",
      "metaphor": "Handing the meter reader your house keys because that's the default keyring.",
      "teachable_moment": "Least privilege is cheapest at creation time — one line now versus an audit sweep across every workflow later."
    },
    {
      "id": "CL-410-004",
      "title": "Path-filtered workflow will wedge PRs if made a required status check",
      "severity": "LOW",
      "category": "ci",
      "file": ".github/workflows/registry-cli-tests.yml:14",
      "description": "The lane triggers only on the four listed path globs. If branch protection later marks `registry-cli-tests` as a required check, PRs not touching those paths will never report the status and will hang un-mergeable (GitHub does not auto-pass skipped path-filtered checks). This is a known interaction, worth a note now while the gate is being wired as 'the G-4 consumer gate'.",
      "suggestion": "Either document that this check must stay non-required, or add a trivial companion job (or use the paths-filter action inside an always-running workflow) so the check always reports.",
      "confidence": 0.5,
      "metaphor": "A toll booth that only appears on certain roads — fine, until the highway authority decrees every car must show a toll receipt.",
      "teachable_moment": "Path filters and required checks are individually sensible features that compose into a deadlock. Document the constraint at the moment you create it; the person flipping the branch-protection switch won't be reading this workflow file.",
      "connection": "Directly relevant here because the sprint explicitly brands this lane 'the G-4 consumer gate' — the pressure to make it required is built into its name."
    },
    {
      "id": "CL-410-005",
      "title": "IsoDate regex accepts impossible dates",
      "severity": "LOW",
      "category": "validation",
      "file": "packages/freeside-registry/src/registry.ts:34",
      "description": "`/^\\d{4}-\\d{2}-\\d{2}$/` accepts values like \"2026-13-45\" or \"2026-00-00\". For a provenance field (`probed_at`) that exists specifically to make hand-typed values trustworthy (NFR-5), shape-only validation leaves a gap the negative test suite (\"malformed probed_at\") doesn't cover.",
      "suggestion": "Tighten the pattern (e.g. month 01–12, day 01–31) or add a filter that round-trips through `Date` and rejects NaN/normalized mismatches.",
      "confidence": 0.85,
      "faang_parallel": "The classic JavaScript Date footgun Google's Closure library and later Temporal (TC39) were built to escape: `new Date('2026-02-31')` silently normalizes to March 3rd. Amazon's ion-schema and Google's proto `Date` message both validate calendar semantics, not just digit shapes, for exactly this reason.",
      "metaphor": "A bouncer checking that IDs are the right size and shape of plastic — '13/45/2026' sails through because it laminates nicely.",
      "teachable_moment": "Shape validation and semantic validation feel adjacent but fail differently: shape errors are caught by anyone, semantic errors survive until a consumer does arithmetic on them. Fields whose entire purpose is human-typed provenance deserve the semantic tier.",
      "connection": "The future-dated probed_at values in this very PR (CL-410-007) are a live demonstration that hand-typed dates drift — the two findings are one story."
    },
    {
      "id": "CL-410-006",
      "title": "graphql-lag thresholds values are unconstrained numbers",
      "severity": "LOW",
      "category": "validation",
      "file": "packages/freeside-registry/src/registry.ts:104",
      "description": "`Schema.Record({ key: Schema.String, value: Schema.Number })` admits negative, zero, fractional, and (via YAML's `.nan`/`.inf` forms) non-finite thresholds. A negative or NaN threshold makes `lag < threshold` permanently false — a declared expectation that can never be healthy, which is precisely the quiet-vs-down confusion the lane is meant to kill. Threshold keys are also unvalidated (any string, not just chain-id digits).",
      "suggestion": "Constrain values to positive finite integers (`Schema.Number.pipe(Schema.int(), Schema.positive())` or a finite filter) and consider a `/^\\d+$/` pattern on keys.",
      "confidence": 0.7,
      "faang_parallel": "Prometheus alerting rules learned this the hard way: a threshold of NaN makes every comparison false, so the alert never fires and the dashboard stays green while the service burns. 'Never-firing alert' is the monitoring equivalent of dead code that looks alive.",
      "metaphor": "A tripwire strung at negative height — technically installed, physically impossible to trip.",
      "teachable_moment": "For any field consumed as a comparison operand, the schema question isn't 'is it a number?' but 'does every representable value produce a decidable comparison?' NaN is the value that makes comparisons lie.",
      "connection": "Same undischargeable-declaration class as CL-410-001: both let the registry declare an expectation no consumer can ever satisfy or falsify."
    },
    {
      "id": "CL-410-007",
      "title": "probed_at dates are future-dated relative to the PR",
      "severity": "LOW",
      "category": "data",
      "file": "packages/freeside-registry/registry.yaml:90",
      "description": "Every service block carries `probed_at: \"2026-07-05\"` while the commits and test fixtures use 2026-07-04. If this is a timezone artifact it is harmless, but for a field whose sole purpose is provenance accuracy, a date one day ahead of the authored work reads as a transcription slip and slightly undermines the NFR-5 story.",
      "suggestion": "Confirm the probe date; if the probes ran on 07-04 UTC, correct the values, or document that probed_at is recorded in the operator's local timezone.",
      "confidence": 0.35,
      "teachable_moment": "Low confidence deliberately — a UTC+X operator probing late on 07-04 local produces 07-05 legitimately. The durable fix isn't correcting one date, it's declaring the field's timezone semantics in the schema comment so the question never recurs.",
      "connection": "Whichever way it resolves, it's evidence for CL-410-005: hand-typed provenance dates need more than shape validation."
    },
    {
      "id": "CL-410-008",
      "title": "Real-registry decode runs at module scope — a decode failure aborts the suite instead of failing a test",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/freeside-registry/tests/registry-decode.test.ts:12",
      "description": "`const registry = loadRegistry()` executes at import time. If registry.yaml stops decoding, node:test reports a module-load error rather than a named failing test, which is less legible in CI output for the very failure mode this file exists to catch (and skips the remaining assertions' diagnostics).",
      "suggestion": "Wrap the load in a `test(\"registry.yaml decodes\")` and share via a lazy accessor or `before` hook, so the first failure is a named test with the decode error as its message.",
      "confidence": 0.55,
      "metaphor": "A fire drill where the alarm shorts out the PA system — the building still empties, but nobody hears which floor triggered it.",
      "teachable_moment": "The distinction between 'CI went red' and 'CI went red with a named, greppable failure' matters most for exactly the failures you designed the test to catch — optimize the error report for your headline failure mode."
    },
    {
      "id": "CL-410-009",
      "title": "Test filter condition re-checks a constraint the schema already enforces",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/freeside-registry/tests/registry-decode.test.ts:58",
      "description": "The D-6 test guards with `if (e.service && e.deployment_url != null)`, but the ModuleEntry filter guarantees `deployment_url` is present whenever `service` exists — a service block on a null-URL entry can never survive decode. The `!= null` guard therefore silently skips a branch that cannot occur, and if the schema filter were ever weakened, this test would skip the drifted entry instead of failing.",
      "suggestion": "Drop the `e.deployment_url != null` guard (assert unconditionally when `e.service` exists) so the test independently witnesses the invariant rather than mirroring the schema's own gate.",
      "confidence": 0.6,
      "faang_parallel": "The mutation-testing insight behind Google's and Meta's mutation-coverage tooling: a test that encodes the implementation's own guard survives the mutant that deletes the guard. Independent witnesses kill mutants; mirrors don't.",
      "metaphor": "An auditor who checks the books using the accountant's own calculator settings — agreement is guaranteed and therefore worthless.",
      "teachable_moment": "A test earns its keep by being able to disagree with the code. Any conditional in a test that duplicates a production invariant converts a potential failure into a silent skip."
    },
    {
      "id": "CL-410-010",
      "title": "Future dot-path consumers should be protected from __proto__/constructor segments at the schema layer",
      "severity": "SPECULATION",
      "category": "security",
      "file": "packages/freeside-registry/src/registry.ts:96",
      "description": "`rows_path` and `timestamp_path` are free-form dot-paths that a next-cycle loa-cli consumer will walk over parsed GraphQL responses. Naive dot-path traversal is a classic prototype-pollution/unexpected-property-access surface (`__proto__`, `constructor.prototype`). The registry is git-reviewed so exposure is low, but the schema is the cheapest place to make the hazard unrepresentable before any consumer exists.",
      "suggestion": "Add a pattern to path fields rejecting `__proto__`, `constructor`, and `prototype` segments (e.g. restrict segments to `[a-zA-Z0-9_]+(\\.[a-zA-Z0-9_]+|\\.\\d+)*`), and note the safe-traversal requirement in the consumer-semantic comments.",
      "confidence": 0.4,
      "faang_parallel": "lodash's `_.set` prototype-pollution CVEs (2019–2020) forced half the npm ecosystem to patch dot-path traversal after the fact; the fix everyone converged on was segment denylisting — the same three names this suggestion pins.",
      "teachable_moment": "Speculative by definition — the consumer doesn't exist yet. But schema constraints are cheapest before the first consumer ships, because afterward every tightening is a potential breaking change."
    },
    {
      "id": "CL-410-011",
      "title": "endpoint / deployment_url fields accept any string — https enforcement would extend the SSRF posture to this contract",
      "severity": "SPECULATION",
      "category": "security",
      "file": "packages/freeside-registry/src/registry.ts:52",
      "description": "`ServiceBlock.deployment_url` and expectation `target.endpoint` are `Schema.String` with no URL/scheme validation. Consumers will issue real HTTP requests to these values; the ecosystem already maintains an SSRF-safe fetch path for beacons. Schema-level `https://` (and non-localhost) enforcement would make the declared contract structurally consistent with that posture rather than relying solely on consumer-side guards.",
      "suggestion": "Pipe URL fields through a pattern or filter requiring `https://` and a syntactically valid URL; consumer-side SSRF guards remain the real boundary.",
      "confidence": 0.5,
      "faang_parallel": "Defense-in-depth as practiced in AWS SDK endpoint resolution: the client validates endpoint shape even though IAM and VPC boundaries are the real controls — each layer assumes the others can regress.",
      "metaphor": "The registry is the address book; the fetch path is the driver who refuses sketchy destinations. Validating addresses at write time means the driver's judgment is a second line, not the only line.",
      "connection": "The repo already fought and won an SSRF battle on the beacon path (IPv6-mapped-address unwrapping) — schema-level scheme enforcement here keeps this newer contract inside that established posture."
    },
    {
      "id": "CL-410-012",
      "title": "Discriminated union as the FR-3 gate — premature gh-workflow use fails at decode",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/freeside-registry/src/registry.ts:59",
      "description": "Excluding gh-workflow from the union so that any premature declaration fails schema decode — with a negative test pinning it — puts the enforcement in the type system rather than a bespoke lint. New probe kinds require a deliberate schema PR that trips the CI lane, which is exactly the right friction.",
      "suggestion": "None — keep the pattern for future probe_kind additions.",
      "confidence": 0.9,
      "faang_parallel": "The 'parse, don't validate' doctrine (Alexis King's essay, now standard at typed-FP shops and echoed in Rust API design): illegal states are rejected at the boundary and cannot exist downstream, so no consumer needs a defensive branch for them.",
      "metaphor": "Instead of posting a 'no gh-workflow yet' sign, the door for it simply hasn't been built — you can't walk through a wall.",
      "teachable_moment": "The best feature gates are absences in the type, not flags in the logic. An absence can't be misconfigured, and adding the variant later is a visible, reviewable diff."
    },
    {
      "id": "CL-410-013",
      "title": "Split-brain filter with exhaustive negative coverage (equal / mismatched / null / absent)",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/freeside-registry/src/registry.ts:180",
      "description": "The ModuleEntry filter requiring a present-and-equal entry-level deployment_url whenever a service block exists closes the two-fields-drift hazard the in-block duplication creates, and the tests cover all four states including the subtle absent-vs-null distinction. The filter error message is asserted by regex, so the tests pin the diagnostic, not just the rejection.",
      "suggestion": "None.",
      "confidence": 0.9,
      "faang_parallel": "Spanner's approach to denormalized data: where duplication is accepted for read convenience, an enforced consistency invariant rides alongside it. Duplication without an equality constraint is how Amazon's early catalog systems accumulated years of split-brain records.",
      "metaphor": "Two copies of a key are fine as long as the locksmith checks they cut identically every time one is made.",
      "teachable_moment": "Asserting error messages by regex is underrated: a rejection with a degraded diagnostic is a regression too, and most suites can't see it. The four-state matrix (equal/mismatched/null/absent) is the complete truth table — nothing left to chance."
    },
    {
      "id": "CL-410-014",
      "title": "Real-registry decode test with anti-transcription tripwire and shared-schema fixture guard",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/freeside-registry/tests/registry-decode.test.ts:47",
      "description": "Decoding the production registry.yaml through the shared `_schemas` export (never a test-local schema copy) prevents fixture tautology, and the score-api test asserting the 302 redirect was never encoded turns an ADR footnote into a permanent mechanical tripwire. The provenance-on-every-block and blockless-cell assertions ratchet the D-7 derive-from-absence invariant.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "Netflix's production-config canary tests: the real config artifact is validated by the real parser in CI, because a test-local copy of either one eventually diverges and validates a fiction. The 302-never-encoded assertion is a 'negative space' test — pinning what must stay absent, which most suites never think to do.",
      "metaphor": "Testing the actual bridge with the actual truck, not a scale model of each — and also checking that the lane someone once almost painted was never painted.",
      "teachable_moment": "Fixture tautology is the quietest way a test suite dies: schema and fixture drift together and agree forever. Importing the production schema into the test is the one-line vaccine."
    },
    {
      "id": "CL-410-015",
      "title": "CI install order documented against the file:-dep stale-copy hazard",
      "severity": "PRAISE",
      "category": "ci",
      "file": ".github/workflows/registry-cli-tests.yml:8",
      "description": "The workflow comments name the load-bearing constraint (beacon-schema builds first, freeside-cli installs last so its file: deps copy freshly built packages) rather than leaving the ordering as implicit magic. A future editor reordering steps will at least see why it breaks.",
      "suggestion": "None.",
      "confidence": 0.8,
      "faang_parallel": "Chesterton's Fence as institutionalized in Google's build files: every non-obvious ordering constraint gets a comment naming what breaks, because the alternative is an archaeology project during an outage.",
      "metaphor": "Labeling the one load-bearing wall in the blueprint instead of trusting every future renovator to guess.",
      "teachable_moment": "Ordering constraints that live only in the author's head have a half-life of one team rotation. A comment naming the failure mode ('reorder this and freeside-cli tests a stale copy') is the cheapest durable artifact."
    }
  ]
}
```


## Callouts

Three pieces of genuinely strong engineering here, all variations on one theme — **enforcement placed where it cannot be forgotten**:

- **[Quality] The gh-workflow exclusion (CL-410-012)** is the standout design decision. Feature-gating by *omitting the union variant* means the gate is the type system itself — no lint to bypass, no flag to misread. Adding the variant later will be a visible schema PR that trips this same CI lane. That's the gate guarding its own successor.
- **[Test Coverage] The four-state split-brain matrix (CL-410-013)** with regex-pinned error messages is complete in a way most negative suites aren't — it covers the truth table, including the absent-vs-null edge that trips up most optional-field handling, and it treats a degraded diagnostic as a regression.
- **[Test Coverage] The 302-never-encoded tripwire (CL-410-014)** deserves imitation across the registry: asserting what must remain *absent* converts an ADR decision into a mechanical invariant. Institutional memory that survives context loss is exactly what a factory repo needs.
- **[Operational] The install-order comments (CL-410-015)** show operational empathy for the future editor — the constraint is named at the point of hazard, not buried in a runbook.

The follow-up work is small and mostly mechanical: close the target-less-http implication with the filter pattern you already own (CL-410-001), and swap the gate onto `npm ci` before anyone starts trusting its color (CL-410-002). Both fixes make the PR's own philosophy — fail loudly at decode, attribute red to the PR that caused it — fully self-consistent.

zkSoju:
**Bridgebuilder findings addressed** in dc0ae419:

- **CL-410-001 (MEDIUM)** — fixed with the filter pattern the review pointed at: a target-less `http` expectation now requires the cell to declare a `service` block (`ModuleEntry` filter + 3 tests: negative without block, positive with block, explicit-target needs none). No declaration a consumer can't discharge.
- **CL-410-002 (MEDIUM)** — the three package lockfiles are now committed and the `registry-cli-tests` lane runs `npm ci`: the G-4 gate is lockfile-reproducible, so its color is attributable to the PR under test.

LOWs and SPECULATIONs noted for the loa-cli dispatch cycle (where the SSRF host-guard requirement already carries). Registry 47/47, tsc clean, freeside-cli 94/0 untouched.

zkSoju:
**Merge-gate documentation (admin bypass, operator-approved)**

The required \`loa-signoff\` door was exercised honestly and REFUSED — correctly: it attests by running the \`themes/sietch\` unit suite locally, and that suite is **red on main** (17 files: jwt/jwks, billing, budget-drift, ajv conformance). This PR's diff touches **zero** sietch files (\`git diff origin/main..HEAD -- themes/sietch\` is empty), so no forged status was posted and none will be. Beads filed:

- \`arrakis-sietch-suite-red-blocks-signoff-door-qbt0r\` (P1) — the door is unsatisfiable for ALL PRs until the sietch suite is green on main
- \`arrakis-ledger-domain-check-jq-head-epipe-flake-puq3v\` (P2) — the \`jq|head\` EPIPE flake in Ledger Domain Check

Merging with \`--admin\` per operator decision. Gate trail on this PR: Loa review (iter-2 All good, cross-model DISS-001 fixed) → security audit APPROVED → Bridgebuilder MEDIUMs fixed (CL-410-001/002) → **registry-cli-tests PASS** (the load-bearing G-4 consumer gate) + 18 advisory greens; the 4 advisory reds fail identically on sibling PRs (#437/#438).

Issue #416, opened 2026-07-02
[ordering/kitchen] Add metadata_snapshot ingredient to community-onboarding preset


zkSoju:
## Problem

External communities onboarded via Preset #2 (`community-onboarding`) get sonar indexing + score registration + worlds manifest — but **no automated step snapshots NFT metadata** for member avatars.

**Consumer:** freeside-dashboard `/roles` + member drawer (Zerker) — currently ships `pfp: null` for pythenians/purupuru; UI is ready, read seam is not.

**Product anchor:** [inventory-api#19](https://github.com/0xHoneyJar/inventory-api/issues/19) — wallet→NFT+image via inventory-api (sonar ownership ⨝ storage-api metadata).

**Coord cluster:** `member-pfp-coordinator` — STOR-1 ([storage-api#22](https://github.com/0xHoneyJar/storage-api/issues/22)), INV-* ([inventory-api#20](https://github.com/0xHoneyJar/inventory-api/issues/20)), DASH-1 ([freeside-dashboard#96](https://github.com/0xHoneyJar/freeside-dashboard/issues/96)).

Without a kitchen ingredient, metadata onboarding stays a manual operator step outside the order checklist — every new external community repeats the pythenians/purupuru runbook.

---

## Proposed change

Add a **`metadata_snapshot`** ingredient to the community-onboarding preset, slotted **after `score` completes** (world_slug may not exist yet; snapshot keys off `chain_id` + `contract_address` + provisional `world_slug` from score register response when available).

### 1. Protocol — `@freeside/ordering-protocol`

Extend `CommunityOnboardingIngredients`:

```typescript
export const CommunityOnboardingIngredients = z.object({
  sonar: IngredientStatus,
  score: IngredientStatus,
  metadata_snapshot: IngredientStatus,  // NEW
  worlds_manifest: IngredientStatus,
  discord_observer: IngredientStatus,
  shadow_preview: IngredientStatus,
}).strict();
```

Update `INITIAL_COMMUNITY_ONBOARDING_INGREDIENTS`:

```typescript
metadata_snapshot: 'pending',
```

Extend `EnqueueIngredientKey`:

```typescript
z.enum(['sonar', 'score', 'metadata_snapshot', 'worlds_manifest'])
```

Add capability need (optional in preset recipe):

```typescript
{ label: 'snapshot collection metadata to sovereign storage', capability: 'metadata-snapshot' }
```

[...]

Issue #405, opened 2026-07-01
[ordering/kitchen] Channel health read gate after discord_bind + bot_installed


zkSoju:
## Context

Member graph channel health is **kitchen-owned**, not dashboard-composed.

Specs (dashboard grimoires, `cycle/channel-ux-parity`):
- `grimoires/loa/specs/substrate-member-graph-schema.md`
- `grimoires/loa/specs/substrate-discord-guild-health-api.md`
- `grimoires/loa/specs/community-request-order-shadow-audit.md`

## Requirement

Dashboard displays Discord integration health only after order graph ingredients complete:
1. `discord_bind` — guild ↔ world
2. `bot_installed` — Freeside bot present

Until then, dashboard shows honest `source: preview` (V0 composer from cookie bind + roles snapshot).

## Kitchen deliverable

When ingredients are `fulfilled`:
- Route or proxy `GET /v1/worlds/:slug/channels/discord/health` (implemented in freeside-characters — see that repo issue)
- Signal readiness to dashboard consumer (ordering poll / webhook / worlds-api manifest)

## Worldline

Order fulfillment + first successful guild sync should be indexable events for operator audit (aligns with shadow audit MVP).

## Acceptance

- [ ] Health endpoint unreachable / degraded until `discord_bind` + `bot_installed` complete
- [ ] Order status surface tells dashboard when live channel health is available
- [ ] Documented handoff to freeside-dashboard BFF proxy

## Related

- freeside-characters: guild health API implementation
- identity-api#45: linkedCount / attribution join
- freeside-dashboard: BFF proxy issue

zkSoju:
### Triage provenance

**Record type:** Classification provenance  
**Effect:** Audit trail only. No work was dispatched, approved, closed, or resolved by this comment.

| Classification | Applied value |
| --- | --- |
| Workstream | `workstream:delivery` |
| Artifact type | `artifact-type:technical-rfc` |
| Priority | `priority:medium` |
| Route | `triage:operator-review` |

| Grounding | Value |
| --- | --- |
| Triaged at | `2026-07-15T17:40:19.866Z` |
| Repository revision | [`3782fd4`](https://github.com/0xHoneyJar/loa-freeside/commit/3782fd47e8a20cdaf6325621962bd0443e6781b8) |

**Evidence basis**

[...]

Issue #401, opened 2026-07-01
[ordering/kitchen] Wire shadow_preview ingredient probe for zero-operator fulfill


zkSoju:
## Context

Community-onboarding kitchen K3 E2E (Azuki) reached `producing` with sonar `in_progress`, worlds `complete`, score stuck until score-api activation — but **fulfillment gate** also requires `shadow_preview === 'complete' | 'optional'`.

Today `StubTriagePorts.shadow.probe` always returns `blocked`; `KitchenTriagePorts` delegates shadow to the same stub. No HTTP building probe exists for shadow audit.

## Problem

Zero-operator fulfill is impossible even when sonar/score/worlds complete — shadow stays `blocked` until operator `POST …/advance-ingredient` with `shadow_preview: complete`.

## Proposed fix (SDD K3+ / K4)

1. Define shadow building capability + HTTP probe (or wire to existing shadow-audit-service preset #1 path).
2. Gate rule: when sonar+score+worlds complete, flip `shadow_preview` from `blocked` → `pending` (orchestrator already has partial logic on operator advance).
3. Auto-run shadow audit for registered contract + poll until complete.

## Acceptance

- Azuki-class E2E can reach `fulfilled` without manual `advance-ingredient` once upstream ingredients complete.
- Dashboard waiting room shows shadow progress in ingredient checklist.

## Related

- `grimoires/loa/sdd.community-onboarding-kitchen.md` §3.4
- loa-freeside #397 (K3 merged)

zkSoju:
### Triage provenance

**Record type:** Classification provenance  
**Effect:** Audit trail only. No work was dispatched, approved, closed, or resolved by this comment.

| Classification | Applied value |
| --- | --- |
| Workstream | `workstream:delivery` |
| Artifact type | `artifact-type:technical-rfc` |
| Priority | `priority:medium` |
| Route | `triage:operator-review` |

| Grounding | Value |
| --- | --- |
| Triaged at | `2026-07-14T16:07:34Z` |
| Repository revision | [`9a98f5d`](https://github.com/0xHoneyJar/loa-freeside/commit/9a98f5d501f1b818a98fe8a57b9ff070e4a0027f) |

**Evidence basis**

> Community-onboarding kitchen K3 E2E (Azuki) reached producing with sonar inprogress, worlds complete, score stuck until score-api activation — but fulfillment gate also requires shadowpreview === 'complete' 'optional'.

zkSoju:
[...]

PR #440, opened 2026-07-05
feat(ordering): fulfillment orchestrator — probe/dispatch/advance/escalate worker (sprint-23)

zkSoju:
## Summary
Sibling worker behind `ORCHESTRATOR_ENABLED` consuming the ordering outbox: probes the three kitchen ingredient APIs, dispatches idempotent writes worlds-first (D-5 canonical slug + `slug_divergence` events), advances via in-process `advanceIngredient()`, escalates failed/ambiguous states as `orchestrator.escalate` events + GitHub comments (injectable, no-op in CI). Closes the manual-choreography loop from #415 for reversible seams.

**Hard invariants (review-hardened, pinned by tests):**
- `shadow_preview` is NEVER advanced by the orchestrator — operator-owned, gates fulfillment
- worlds `complete` without a `world_slug` escalates (no silent stall in `producing`)
- tick loop survives store throws (no unhandled-rejection worker death)

Tests: 139/139 (`pnpm -C packages/services/ordering test`) incl. zero-manual-step CI E2E. Audit: APPROVED (0 CRIT/HIGH; M-1 order-idempotency filed as pre-batch-1 gate).

Part of ordering-scale sprint-23 · ref #415 · first live drive gated on sonar-api #120 (see bonfire findings-ledger F-4/F-5)


zkSoju:
## Sprint-24 zero-operator-fulfill slice — review evidence

This branch now carries the complete slice (#401 shadow fail-closed, #416 metadata_snapshot ingredient, #405 discordHealth gate) plus three review rounds:

| Round | Reviewer | Verdict | Outcome |
|---|---|---|---|
| 1 | Independent Opus (fresh session, spiral harness) | CHANGES_REQUIRED — "the engineering is correct, 8/8 tests green"; blockers were evidence-gate gaming artifacts | stub + symlinks purged (`8ab803db`, `2fcee2af`) |
| 2 | FAGAN council (claude+codex+cursor, 3/3 voices) | CHANGES_REQUIRED | StubTriagePorts absence semantics, serviceToken privacy, comment honesty (`e68357f1`) |
| 3 | FAGAN council iter-2 | **GPT: APPROVED** · claude: 6 findings | 3 fixed (`15e6c090`: honest advance emit, warn-once, AC-1 end-state pin) · rest deferred with rationale → #441 |

**Test state:** ordering suite 172/173 — the one red (`reprobe-endpoint` hung-probe timeout classification) is proven pre-existing (fails at `8ab803db`, before any sprint-24 code). AC-1 now pins the epic's core contract in-test: all ingredients satisfied → order reaches `fulfilled` with zero operator action.

**Known deferrals:** #441 (discord service-path health bypass, pre-sprint-24 order stranding, metadata 404 re-dispatch, default-on gate backstop, AC-6 test tautology). Harness evidence-gate gameability filed upstream as loa#1175.

**Deploy note:** if score-api's metadata-snapshot endpoint is not yet live, set `METADATA_SNAPSHOT_ENABLED=false` on ordering-service (runbook: `grimoires/loa/deployment/zero-operator-fulfill-env-vars.md`).

zkSoju (commented):
## Summary

This PR lands a well-structured probe/dispatch/advance/escalate worker with genuinely strong reliability instincts — strict zod at emit time, a pure table-tested decision function, and crash-resume proven by E2E. The primary concerns are a health gate that likely never fires on the production path (FO-001), and a cluster of small reliability hazards (dedup-before-emit, unguarded interval overlap, unbounded per-tick telemetry) that are each a few lines to fix but compound in long-running workers.

**Verdict: COMMENT** — nothing exploitable, but FO-001 through FO-005 deserve resolution before this worker runs unattended.

The recurring theme across the findings: **invariants enforced on one path while a second path exists**. The Discord health gate, the duplicated idempotency-key construction, and the duplicated triage composition are all the same shape at different scales — a contract with two owners is a contract with no owner.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FO-001",
      "title": "Discord channel-health gate is bypassable and likely dead in production",
      "severity": "HIGH",
      "category": "correctness",
      "description": "The D13.4 health gate only fires inside the fulfillment orchestrator's advance loop, and only when the stored ingredient is not already 'complete'/'optional'. But INITIAL_COMMUNITY_ONBOARDING_INGREDIENTS seeds discord_observer as 'optional', so the advance loop skips it ('already recorded') and the gate never executes on the default path. The tests acknowledge this: both discordHarness (fulfillment-orchestrator.test.ts) and the T-3 wiring test must patchRecord discord_observer back to 'pending' to reach the gate at all, and the wiring test comment further admits that advanceIngredient→process() can merge discord_observer to 'complete' via the intake orchestrator's mergeProbedIngredients path — a second write path that records 'complete' with no health check. As shipped, the gate is enforced only on a path that production data rarely (or never) takes, while an unguarded path exists in parallel.",
      "suggestion": "Enforce the health check at the single choke point where discord_observer transitions to 'complete' (e.g., inside advanceIngredient / the merge path for this ingredient), or change the initial seed so discord_observer flows through the decision table. Add a test that starts from the real INITIAL ingredients (no patchRecord) and proves the gate fires.",
      "confidence": 0.65,
      "faang_parallel": "This is why Google built Zanzibar as a single authorization choke point rather than scattering permission checks across callers — a check that lives on one of N paths to a state transition is a check that doesn't exist. The AWS S3 team applies the same rule to bucket-policy evaluation: enforcement lives where the write commits, never upstream of it.",
      "metaphor": "A metal detector at the side entrance while the main lobby doors stand open. The detector works perfectly — it's just that nobody walks through it.",
      "teachable_moment": "Guard the state transition, not a route to it. When tests must contort fixtures (patchRecord back to 'pending') to reach a gate, that contortion is the test suite telling you the gate is off the real path.",
      "decision_trail": "Document why discord_observer seeds as 'optional' and which path is intended to own the health check — the wiring test comment already half-admits the dual-path problem; promote that admission to an ADR-style note or fix it."
    },
    {
      "id": "FO-002",
      "title": "Escalation dedup key recorded before emit — a store failure permanently suppresses the escalation",
      "severity": "MEDIUM",
      "category": "reliability",
      "description": "In escalate(), `this.escalated.add(key)` happens before `await this.emit(...)`. If appendEvent throws (store hiccup), the worker tick catches and logs the error, but the key stays in the in-memory set — every subsequent tick returns early at `this.escalated.has(key)` and the orchestrator.escalate event is never emitted for the process lifetime. This is exactly the silent stall D-2 is designed to prevent, triggered by a transient store failure.",
      "suggestion": "Add the key to the dedup set only after the emit succeeds (or remove it in a catch block). The event's idempotency_key already lets consumers dedupe a rare double-emit, so emit-then-mark is safe.",
      "confidence": 0.85,
      "faang_parallel": "Kafka's idempotent producer embodies the industry consensus: prefer at-least-once delivery plus consumer-side dedup over at-most-once. Marking before emitting converts a transient failure into at-most-zero — the one delivery semantic nobody chooses on purpose.",
      "metaphor": "Crossing a guest's name off the invite list before the invitation is in the mailbox. If the mail run fails, they're never invited — and the list says they were.",
      "teachable_moment": "Order of operations around side effects IS the delivery semantic. Mark-then-act gives at-most-once; act-then-mark gives at-least-once. When an idempotency key already protects consumers, at-least-once is strictly safer."
    },
    {
      "id": "FO-003",
      "title": "Throwing discordHealth port aborts processOrder instead of escalating; test title contradicts its assertion",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "checkChannelHealth is awaited with no try/catch in the advance loop, so a throwing port rejects processOrder — skipping the remaining advance and escalate stages for that order that tick. The test named 'escalates when discordHealth throws (network error maps to escalate)' actually asserts `rejects.toThrow()`, i.e. it pins the opposite of what its title claims. HttpBuildingProbes.checkDiscordChannelHealth never throws (it catches and returns healthy:false), so today this only bites custom ports, but the interface contract is undocumented and the test is misleading.",
      "suggestion": "Wrap the health call in try/catch and route a throw to escalate(orderId, 'discord_observer', ...) to match the stated intent, or fix the test title/comment to say a throwing port fails the tick (and document that ports must not throw).",
      "confidence": 0.9,
      "faang_parallel": "The Google Testing Blog's 'Test Names Are Specifications' principle: engineers read titles, not bodies, when triaging. A test whose title asserts the opposite of its body actively misinforms — worse than no test, because it manufactures false confidence during future refactors.",
      "metaphor": "An emergency-exit sign pointing at a locked door. The sign is confident; the door disagrees.",
      "teachable_moment": "Whether a port may throw is part of its contract, and undocumented contracts get discovered at 3 a.m. Pick one — 'ports never throw' (enforce with a wrapper) or 'throws map to escalate' (enforce with try/catch) — and pin it with a correctly-titled test."
    },
    {
      "id": "FO-004",
      "title": "Per-tick telemetry events accumulate unboundedly for stalled orders; dispatch ok:false emitted when dispatch is disabled",
      "severity": "MEDIUM",
      "category": "reliability",
      "description": "Every tick emits an orchestrator.probe event per non-terminal order, and — when `deps.dispatch` is null — a dispatch event with ok:false for every pending ingredient (`const res = this.deps.dispatch ? ... : { ok: false }`), even though nothing was dispatched. An escalated-but-non-terminal order stays in 'producing' forever, so a stuck order appends several events to the durable outbox every 15 minutes indefinitely. The ok:false-when-disabled events are also semantically misleading: they read as failed dispatches rather than 'dispatch stage disabled'.",
      "suggestion": "Skip the dispatch loop entirely (or skip the event emit) when the dispatch port is null, and consider suppressing/coalescing the per-tick probe event when statuses are unchanged since the last tick, or pausing processing for orders that have already escalated.",
      "confidence": 0.7,
      "faang_parallel": "Kubernetes moved node heartbeats from full NodeStatus updates to lightweight Lease objects precisely because unchanged-state-reported-every-interval was melting etcd at scale. Emit on change, coalesce on sameness.",
      "metaphor": "A smoke alarm that logs 'still no fire' to a permanent ledger every 15 minutes, forever, for every room — including the ones already evacuated.",
      "teachable_moment": "Durable telemetry needs a growth bound tied to state changes, not wall-clock ticks. And a disabled stage should be legible as disabled — ok:false laundering 'not attempted' into 'failed' will send a future on-call engineer chasing a dispatch outage that never happened."
    },
    {
      "id": "FO-005",
      "title": "Worker ticks can overlap — setInterval has no in-flight guard",
      "severity": "MEDIUM",
      "category": "concurrency",
      "description": "FulfillmentOrchestratorWorker.start() fires `void this.tick()` from setInterval with no check that the previous tick finished. A tick that runs longer than intervalMs (many orders × slow upstream HTTP probes/dispatches) overlaps the next one, producing concurrent processOrder calls on the same orders: duplicate probe/dispatch/advance events and races on the read-modify-write markInProgressKey → patchRecord path. Unlikely at the 900s default but a latent hazard if the interval is tuned down or order volume grows.",
      "suggestion": "Guard with an `inFlight` boolean (skip the tick if the previous one is still running), or switch to a self-scheduling setTimeout loop that re-arms after tick completion — the pattern is a few lines and removes the race class entirely.",
      "confidence": 0.75,
      "faang_parallel": "This is why cron-driven jobs at scale universally wrap in flock or a lease — the classic 'cron job stampede' postmortem genre. Netflix's scheduling libraries default to fixed-delay (re-arm after completion) rather than fixed-rate for exactly this reason.",
      "metaphor": "Starting the dishwasher on a timer regardless of whether the last cycle finished — eventually two cycles run on the same dishes and the kitchen floods.",
      "teachable_moment": "setInterval encodes 'fire every N ms'; what you almost always want is 'fire N ms after the last run finished'. The self-rescheduling setTimeout is the same line count and eliminates the race class rather than making it rarer."
    },
    {
      "id": "FO-006",
      "title": "Module-level warnedDiscordHealthAbsent is shared mutable state with no reset seam",
      "severity": "LOW",
      "description": "`let warnedDiscordHealthAbsent = false` at module scope is shared across all FulfillmentOrchestrator instances and across test files in the same worker process. kitchen-triage-ports.ts got a test seam (`reset...` export) for its analogous once-per-process warning; this one has none, so whichever test triggers the fallback first silences the warning for every later test, and any future assertion on the warn would be order-dependent.",
      "suggestion": "Mirror the kitchen-triage-ports pattern: export a test-only reset function, or make the flag an instance field (per-orchestrator once is still effectively once-per-process in the single-worker deploy).",
      "confidence": 0.8,
      "metaphor": "A shared whiteboard between two classrooms — whichever class writes first decides what the second class sees.",
      "teachable_moment": "When a codebase already established a pattern for a hazard (the kitchen-triage reset seam), diverging from it in a sibling file is a maintainability tax: the next reader must now learn two idioms for one problem.",
      "connection": "Same 'one contract, two owners' shape as FO-008 and FO-009 — the repo solved this once; reuse the solution."
    },
    {
      "id": "FO-007",
      "title": "Number(job.external_id) can produce NaN issue numbers in escalation comments",
      "severity": "LOW",
      "category": "correctness",
      "description": "commentOnTrackingIssue does `issueNumber: Number(job.external_id)` with no NaN guard. A non-numeric external_id (the field is a free string on ingredient_jobs) would produce a POST to `/issues/NaN`, which fails inside the try/catch and is logged — but a cheap guard would make the failure mode explicit rather than a confusing GitHub 404.",
      "suggestion": "Guard with Number.isInteger(n) after parsing and return early (the escalate event already carries the guarantee), logging the malformed external_id.",
      "confidence": 0.6,
      "teachable_moment": "Validate at the boundary where a free-form string becomes a typed value. A `/issues/NaN` 404 in logs sends the debugger to GitHub; a 'malformed external_id' log line sends them to the actual bug."
    },
    {
      "id": "FO-008",
      "title": "metadata_snapshot idempotency key is hand-built instead of using ingredientJobIdempotencyKey",
      "severity": "LOW",
      "category": "consistency",
      "description": "The DISPATCH_ORDER loop uses `ingredientJobIdempotencyKey(orderId, ingredient)` for the event's idempotency_key, but the metadata_snapshot conditional dispatch inlines the template literal `${orderId}:metadata_snapshot` (and http-building-probes.ts hardcodes the same shape for the Idempotency-Key header). Today the formats coincide, but two construction sites for the same contract can silently diverge if the helper's format ever changes.",
      "suggestion": "Use ingredientJobIdempotencyKey (or a shared helper accepting the metadata_snapshot key) at both call sites so the key format has one owner.",
      "confidence": 0.7,
      "faang_parallel": "Amazon's 'one owner per invariant' service-design rule: a wire-format that appears in two source locations isn't DRY pedantry — it's a dormant compatibility break waiting for the first refactor of the helper.",
      "metaphor": "Two locksmiths cutting 'identical' keys from memory instead of from the same template. They match today."
    },
    {
      "id": "FO-009",
      "title": "createFulfillmentOrchestratorWorker builds a second triage instance instead of reusing the composition's",
      "severity": "LOW",
      "category": "architecture",
      "description": "createFulfillmentOrchestratorWorker calls createOrderingComposition() (which internally wires its own triage for the intake orchestrator) and then calls createKitchenTriagePorts() again for the fulfillment orchestrator. Two independently constructed triage instances read env separately: doubled once-per-process warnings, and a window for the two orchestrators to see different configurations if env-derived construction ever gains state.",
      "suggestion": "Expose the triage instance from createOrderingComposition's return value and reuse it, keeping one composition root per process.",
      "confidence": 0.6,
      "teachable_moment": "A composition root's value is singularity — one place where env becomes objects. Two roots per process reintroduces the config-drift class that dependency injection was adopted to kill.",
      "connection": "Third instance of the PR's recurring shape: FO-001 (two write paths), FO-008 (two key constructors), FO-009 (two composition roots)."
    },
    {
      "id": "FO-010",
      "title": "Startup log hardcodes a 900s fallback that can drift from reprobeIntervalMs()",
      "severity": "LOW",
      "category": "observability",
      "file": "packages/services/ordering/bin/fulfillment-orchestrator.ts",
      "description": "The startup line logs `process.env.KITCHEN_REPROBE_INTERVAL_SEC ?? 900` while the worker's actual interval comes from reprobeIntervalMs() inside FulfillmentOrchestratorWorker's default parameter. If reprobeIntervalMs's default or env parsing (trimming, validation, clamping) differs from this inline expression, the log reports an interval the worker isn't using.",
      "suggestion": "Log the value derived from reprobeIntervalMs() (e.g., expose the resolved intervalMs on the worker) instead of re-deriving it from raw env.",
      "confidence": 0.65,
      "teachable_moment": "Log what you resolved, not what you read. Startup logs exist so on-call can trust them during an incident — a log that re-derives config is a second parser that will eventually disagree with the first."
    },
    {
      "id": "FO-011",
      "title": "Documented idempotency contracts make crash-safe re-dispatch verifiable, and the E2E suite proves it",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/services/ordering/src/http-building-probes.ts",
      "description": "Each dispatch method (ingestSonar, registerScore, manifestWorlds) documents the specific upstream idempotency mechanism (natural-key upsert, advisory lock, manifest replay) that makes re-dispatch safe, and fulfillment-orchestrator-e2e.test.ts exercises the actual crash-resume path with a fresh orchestrator instance that lost all in-memory state. The claim and the proof live together.",
      "suggestion": "None — keep pairing crash-safety claims with resume tests.",
      "confidence": 0.9,
      "faang_parallel": "This is the discipline behind AWS's 'correctness proofs live next to the code' culture (TLA+ specs checked into S3/DynamoDB repos): the safety argument is versioned with the mechanism it justifies, so drift is visible in review.",
      "metaphor": "A bridge with its load calculations engraved on the plaque — anyone inspecting the bridge can check the math without archaeology."
    },
    {
      "id": "FO-012",
      "title": "Strict zod schemas parsed at emit time plus a pure, table-tested decision function",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/services/ordering/src/orchestrator-events.ts",
      "description": "Every orchestrator event payload is validated with a .strict() zod schema at the emit call site, so a malformed telemetry event fails loudly at the producer instead of poisoning the outbox. actionForStatus is a pure exhaustive switch (compiler-enforced via the return type) covered by an it.each decision table that mirrors the SDD §4.1 table one-to-one.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "Producer-side schema enforcement is the lesson of Confluent's Schema Registry: reject at write, because a poisoned durable log punishes every consumer forever. The spec-mirrored it.each table is the 'executable specification' pattern — the SDD and the test can't drift silently.",
      "connection": "This design is precisely why FO-002's fix is safe: strict schemas + idempotency keys mean consumers already tolerate the rare double-emit."
    },
    {
      "id": "FO-013",
      "title": "Worker tick is double-guarded with an explicit unhandledRejection rationale",
      "severity": "PRAISE",
      "category": "reliability",
      "description": "tick() wraps the whole body (listByState can throw outside the per-order loop) and each processOrder individually, with a comment explaining exactly why the outer guard exists (`void this.tick()` from start() would otherwise turn a store failure into an unhandledRejection that kills the worker). The dedicated regression test (FIX 3) pins the behavior.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "The comment-explains-the-nonobvious-guard practice is straight from Google's C++ style guide philosophy: comments earn their keep by stating what the code cannot — here, the invisible unhandledRejection failure mode. The regression pin makes the guard survivable across refactors.",
      "metaphor": "A fuse box with a label explaining which appliance each fuse protects — the next electrician doesn't have to rediscover the wiring by blowing things up."
    }
  ]
}
```


## Callouts

Roughly a third of this PR deserves explicit recognition:

- **[Operational] FO-011** — The idempotency documentation isn't ceremony; the E2E crash-resume test makes it falsifiable. This is the difference between a claim and a contract: a fresh orchestrator instance with amnesia re-dispatches safely, and a test would break if that stopped being true.
- **[Quality] FO-012** — `.strict()` zod at every emit site plus a compiler-exhaustive `actionForStatus` switch means both the data plane and the decision plane fail loudly at the producer. The it.each table mirroring SDD §4.1 one-to-one is an executable specification — spec drift becomes a red test, not a wiki archaeology project.
- **[Operational] FO-013** — The double guard on `tick()` with an explicit rationale comment is exactly how fire-and-forget async should be handled in Node workers. The comment names the invisible failure mode (unhandledRejection process death), which is the one thing the code itself cannot show.

The irony worth savoring: the team that wrote FO-013's careful guard also wrote FO-005's unguarded setInterval — same file, same async discipline problem, opposite outcomes. That's not a criticism; it's evidence the discipline exists and just needs to be applied one layer up. Fix FO-002 and FO-005 (both are a handful of lines), resolve FO-001's dual-path gate, and this worker is genuinely production-shaped.

PR #442, opened 2026-07-05
feat(network/ordering): freeside-cli ordering-verb hardening — loopback-only http, strict flags, timeout testability (epic #415 Phase B)

zkSoju:
Spiral `spiral-20260705-25521f` cycle-1 output, salvaged + council-reviewed.

## What
Hardens the EXISTING `order`/`kitchen`/`fulfill` verbs (shipped in the fulfillment-surface cycle — the spiral's discovery correctly found them and pivoted to gaps):
- **HTTPS enforcement**: plaintext refused; the test bypass (`ORDERING_SERVICE_URL_UNSAFE_HTTP`) now works ONLY against loopback hosts — a remote http URL is refused even with the flag (FAGAN convergent major, fixed `db71a30a`)
- **Strict integer flag validation** for `fulfill watch` (`--interval`/`--timeout`), distinguishing missing values from invalid ones
- **Guard semantics**: interval≥timeout rejected only when both flags are user-authored; a lone `--timeout` clamps the default interval; `--once` skips the guard (was a regression: `--timeout 10` alone always failed)
- **Timeout testability**: `getRequestTimeoutMs()` reads env at call time
- +240 test lines; suite **39 pass / 0 fail** (1 env-gated skip)

## Review trail
- FAGAN council (claude+codex+cursor): CHANGES_REQUIRED → both majors fixed with pin tests
- Spiral's harness cycle died at the evidence gate parsing COMMAND FRAGMENTS as paths (4th instance) — parser patch upstreamed to loa#1175; the fix-loop's System-Zone edit to `spiral-evidence.sh` was reverted here (right instinct, wrong zone)

Domain: `network` (packages/freeside-cli only). Known-red checks (Integration/agent-ci/eval) are the pre-existing numb gate (#304).


zkSoju (commented):
## Summary

This PR does the hard, unglamorous work of hardening a CLI's transport and flag semantics while landing a substantial new fulfillment orchestrator with genuinely thoughtful failure-mode reasoning. The security posture (loopback-only HTTP bypass, allowlist-not-denylist) is exemplary; the residual concerns cluster around lifecycle edges — a slug adopted from failed responses, a health gate that may be dead code in production, and slow-leak/duplicate-instantiation shapes that won't hurt today but will confuse the next operator. Verdict: **COMMENT** — nothing blocking, but PB-002 and PB-003 deserve resolution before the next order flows through.

The architectural through-line worth naming: this orchestrator is an at-least-once state machine over a durable outbox, and the code mostly *knows* it — the escalation-dedup ordering and documented upstream idempotency contracts prove it. The findings below are the places where that discipline momentarily lapses (adopting state from non-ok responses, emitting events forever when a port is null). At-least-once systems live or die on the question "what does a retry observe?" — hold every branch to that question and the medium findings dissolve.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "PB-001",
      "title": "UNSAFE_HTTP bypass flag treats any non-empty value as enabled, including \"0\" and \"false\"",
      "severity": "LOW",
      "category": "security",
      "file": "packages/freeside-cli/src/lib/ordering-client.ts",
      "description": "The loopback test bypass is gated by `Boolean(env.ORDERING_SERVICE_URL_UNSAFE_HTTP)`. An operator who exports `ORDERING_SERVICE_URL_UNSAFE_HTTP=0` or `=false` to explicitly disable it will still enable the bypass, since any non-empty string is truthy. Impact is contained because the bypass only ever applies to loopback hosts (the remote-http refusal test pins that), but the flag semantics invert the operator's intent.",
      "suggestion": "Compare against the exact sentinel used in tests: `env.ORDERING_SERVICE_URL_UNSAFE_HTTP === \"1\"`, matching the documented usage and rejecting '0'/'false'.",
      "confidence": 0.9,
      "faang_parallel": "Kubernetes was bitten by exactly this: feature-gate envs parsed with truthiness instead of strconv.ParseBool meant `FEATURE=false` enabled features. Go's stdlib ParseBool exists because 'set' and 'true' are different predicates.",
      "metaphor": "It's a light switch where flipping it to OFF also turns the light on — the only way to get darkness is to remove the switch from the wall.",
      "teachable_moment": "Environment variables have three states — unset, set-truthy, set-falsy — and `Boolean()` collapses the last two. Any security-adjacent flag should parse an explicit sentinel, because the operator who writes `=0` is the one most convinced they've disabled it.",
      "connection": "Same discipline as parseIntFlag elsewhere in this PR: named, strict parsing at the trust boundary. Apply the PR's own standard to its own bypass flag."
    },
    {
      "id": "PB-002",
      "title": "canonicalSlug adopted from dispatch result without checking res.ok",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "In the dispatch stage, `if (ingredient === 'worlds_manifest' && res.world_slug) canonicalSlug = res.world_slug;` does not check `res.ok`. `manifestWorlds` calls `readWorldSlug(res)` unconditionally, so a non-2xx response (e.g. 409/500) whose JSON error body happens to contain a `world_slug` field yields `{ ok: false, world_slug: '...' }`. That slug then becomes the canonical slug for this tick and can be recorded on a later worlds advance, and it also feeds the D-5 divergence comparison. The same applies to the score branch.",
      "suggestion": "Only honor `world_slug` when the dispatch succeeded: `if (res.ok && res.world_slug) ...`. Alternatively, have `readWorldSlug` only run for 2xx responses inside `manifestWorlds`/`registerScore`.",
      "confidence": 0.7,
      "faang_parallel": "AWS SDK teams enforce that response-body fields are only projected into domain state after status classification — S3's error bodies are valid XML with plausible-looking fields, and early SDKs that parsed before classifying leaked error metadata into caller state.",
      "metaphor": "Reading the return address off a letter marked 'delivery failed' and updating your contact book with it — the envelope arrived, but nothing on it was vouched for.",
      "teachable_moment": "Parse-then-validate ordering matters: classify the response (ok/not-ok) before extracting semantic fields. An error body is attacker-and-accident-shaped data; treating it as a peer of a success body erases the status check you already performed.",
      "connection": "This is the one place the PR's otherwise-careful at-least-once reasoning slips: a failed dispatch should leave zero state behind, precisely so the retry observes a clean world."
    },
    {
      "id": "PB-003",
      "title": "Discord channel-health gate is unreachable for orders seeded with default initial ingredients",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "INITIAL_COMMUNITY_ONBOARDING_INGREDIENTS seeds `discord_observer: 'optional'`. The advance loop skips any ingredient whose stored status is already 'complete' or 'optional' (`if (current === 'complete' || current === 'optional') continue;`) before the D13.4 health gate runs. Both test suites had to manually `patchRecord` discord_observer to 'pending' to exercise the gate, and the wiring test comment explicitly notes the initial 'optional' \"would bypass the health gate entirely.\" Unless some other production path rewrites discord_observer to pending/in_progress, the gate is dead code on the real order flow while appearing tested.",
      "suggestion": "Either run the health gate before the already-recorded skip when the probe reports 'complete', or seed discord_observer as 'pending' when a discord observer/health port is configured so the gate participates in the real lifecycle. At minimum document which production path is expected to move discord_observer off 'optional'.",
      "confidence": 0.55,
      "faang_parallel": "Google's Testing on the Toilet calls this 'test-induced reachability': when every test must contort fixture state to reach a branch, the branch is often unreachable in production. Coverage tools report it green; traffic never visits it.",
      "metaphor": "A security checkpoint installed on a corridor whose door upstream is welded shut — the guards pass every drill because the drill instructors climb in through the window.",
      "teachable_moment": "When tests need `patchRecord` to make a code path reachable, that's a signal to trace the production path that produces the same state. If none exists, either the seed data or the gate is wrong — and the test comment admitting the bypass is the breadcrumb to follow.",
      "connection": "Worth a decision-trail note in the orchestrator: which actor moves discord_observer off 'optional', and when. Future agents reading this gate will otherwise re-derive (or miss) this analysis."
    },
    {
      "id": "PB-004",
      "title": "Module-level warnedDiscordHealthAbsent is shared across orchestrator instances and test runs",
      "severity": "LOW",
      "description": "`warnedDiscordHealthAbsent` is a module-scoped mutable boolean. All FulfillmentOrchestrator instances in a process share it, so in tests the warning fires once for the first harness and never again (harmless here, but the codebase already has a `resetKitchenTriageWarnings`-style test seam for the analogous shadow warning, and this one has none). It also means two orchestrators with different discordHealth configurations share one warning state.",
      "suggestion": "Make it an instance field on FulfillmentOrchestrator, or expose a test-seam reset mirroring the kitchen-triage-ports pattern.",
      "confidence": 0.8,
      "metaphor": "One shared sticky note on the office fridge saying 'already told them' — even when a second team, with a different problem, needed their own telling.",
      "teachable_moment": "Once-per-process and once-per-instance are different cardinalities; module-level mutable state silently picks the former. The codebase already solved this correctly once (kitchen-triage warnings) — divergent solutions to the same problem in one package is the real smell.",
      "connection": "Sibling of PB-010: both are 'process-level state meets multi-instance composition' — the warning dedup and the duplicate triage stacks will interact confusingly if either changes."
    },
    {
      "id": "PB-005",
      "title": "Startup log hardcodes 900s fallback that can drift from reprobeIntervalMs()",
      "severity": "LOW",
      "category": "observability",
      "file": "packages/services/ordering/bin/fulfillment-orchestrator.ts",
      "description": "The worker's actual interval defaults to `reprobeIntervalMs()`, but the startup log prints `process.env.KITCHEN_REPROBE_INTERVAL_SEC ?? 900`. If reprobeIntervalMs()'s internal default ever changes (or applies clamping/validation the log doesn't), the logged interval will misreport the real cadence.",
      "suggestion": "Log the value actually passed to the worker (e.g. expose the resolved intervalMs from the worker or compute `reprobeIntervalMs()` once and use it for both).",
      "confidence": 0.85,
      "faang_parallel": "SRE doctrine at Google: log the *resolved* config, never re-derive it at the log site. Borg tasks print their effective flags post-resolution precisely because two derivations of 'the default' drift independently.",
      "metaphor": "A thermostat display wired to the brochure instead of the thermometer — correct on day one, and then only by coincidence.",
      "teachable_moment": "Any log line that re-computes a value instead of printing the variable actually used is a future lie. Single-source the resolution, then log the source."
    },
    {
      "id": "PB-006",
      "title": "Dispatch-disabled compositions emit ok:false dispatch events every tick and can stall silently in producing",
      "severity": "LOW",
      "category": "design",
      "description": "When `deps.dispatch` is null, every pending ingredient produces `{ ok: false }` and an `orchestrator.dispatch` event with ok:false is appended to the durable outbox on every tick, forever. Since 'pending' maps to dispatch (not escalate) in the decision table, an order in a dispatch-less composition sits in producing indefinitely with no escalation — the exact silent-stall shape D-2 targets — while steadily growing the outbox with repeated probe/dispatch events.",
      "suggestion": "Skip the dispatch emit entirely when the port is null (or emit once with a distinct reason), and consider escalating after N consecutive ticks of pending-with-no-dispatch-port so dispatch-less deployments cannot stall silently.",
      "confidence": 0.6,
      "metaphor": "A mailroom with no courier that keeps stamping 'send failed' on the same parcel every fifteen minutes and filing the stamp in the permanent archive.",
      "teachable_moment": "A retry loop with no exit condition and no escalation is a stall detector that detects everything except itself. When a dependency is structurally absent (null port) rather than transiently failing, the correct response is configuration-error semantics, not retry semantics.",
      "connection": "Ironic mirror of the D-2 design goal this orchestrator exists to serve — the anti-stall machine has one stall mode of its own."
    },
    {
      "id": "PB-007",
      "title": "KitchenTriagePorts ignores a provided fallback's metadata port",
      "severity": "LOW",
      "category": "consistency",
      "file": "packages/services/ordering/src/kitchen-triage-ports.ts",
      "description": "Every other port (sonar, score, worlds, discord, shadow) delegates to `this.fallback` when HTTP probes are absent, but `metadata` is derived solely from the `http` instance and the enabled flag — a caller-supplied fallback with its own `metadata` port is silently dropped. The StubTriagePorts comment explains why the stub omits metadata, but a non-stub fallback (e.g. a test double) will behave differently from the other five ports for no declared reason.",
      "suggestion": "Either delegate `metadata` to `fallback.metadata` when http is absent and metadataEnabled is true, or document on the constructor that metadata deliberately never consults the fallback.",
      "confidence": 0.65,
      "metaphor": "Six doors on a hallway, five of which have a spare key under the mat — a visitor who learned the pattern from the first five will stand puzzled at the sixth.",
      "teachable_moment": "In a facade over N homogeneous ports, one heterogeneous port is a contract violation even if intentional. Intentional asymmetry earns a comment at the asymmetry, not just at the stub."
    },
    {
      "id": "PB-008",
      "title": "getRequestTimeoutMs silently discards non-integer overrides",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/freeside-cli/src/lib/ordering-client.ts",
      "description": "`ORDERING_REQUEST_TIMEOUT_MS=1500.5` or `=abc` silently falls back to 35000 with no diagnostic. Given this PR's whole theme is strict flag validation with named errors (parseIntFlag rejects non-integers loudly), the env-var path is inconsistently permissive — a mistyped value degrades to the default without the operator learning why their timeout isn't honored.",
      "suggestion": "Emit a one-time stderr warning when the env var is set but unusable, or accept any finite positive number rather than requiring an integer.",
      "confidence": 0.7,
      "faang_parallel": "The 12-factor lineage tools (and HashiCorp's envconfig) converged on fail-loud env parsing after years of 'why is my timeout the default' tickets — silent fallback converts a one-second config error into an hour of debugging.",
      "metaphor": "A vending machine that quietly returns the house brand when it can't read your selection — you only notice when the coffee tastes wrong under load.",
      "connection": "Same trust-boundary standard as PB-001: this PR built parseIntFlag precisely to reject sloppy input loudly; the env path should inherit that spine."
    },
    {
      "id": "PB-009",
      "title": "Interval-vs-timeout guard can name a --timeout flag the user never passed",
      "severity": "LOW",
      "category": "ux",
      "file": "packages/freeside-cli/src/verbs/fulfill.ts",
      "description": "With only `--interval 2000` (no --timeout), intervalExplicit is true and 2,000,000ms >= the default 1,800,000ms timeout, so the user gets \"--interval must be less than --timeout\" referencing a flag they didn't provide. The hint does include both values, which softens it, but the asymmetry with the lone-`--timeout` clamp path (which adapts instead of erroring) is worth noting.",
      "suggestion": "When --timeout was not explicit, phrase the error as \"--interval N exceeds the default timeout (1800s); pass --timeout to extend it\".",
      "confidence": 0.75,
      "faang_parallel": "Rust's compiler-diagnostics team treats this as doctrine: an error message may only reference tokens the user actually wrote, or must explicitly mark the value as a default. It's why rustc errors feel trustworthy.",
      "metaphor": "Being told 'your reservation conflicts with the booking you made' when the hotel made the second booking for you.",
      "teachable_moment": "Error messages are a contract about causality. Naming a default as if the user set it teaches them a false model of the tool — and false models compound into worse bug reports."
    },
    {
      "id": "PB-010",
      "title": "Duplicate triage/probe instantiation in createFulfillmentOrchestratorWorker",
      "severity": "LOW",
      "category": "design",
      "description": "`createFulfillmentOrchestratorWorker` calls `createOrderingComposition()` (which builds its own triage inside the intake orchestrator) and then `createKitchenTriagePorts()` again for the fulfillment orchestrator. Both read the same env so the config agrees, but the process now holds two HttpBuildingProbes instances and two triage stacks, and once-per-process warnings (e.g. shadow producer-less) can fire once per stack. The FulfillmentOrchestrator probes via its own triage while its `onboarding` advance path uses the intake's — same behavior today, but a future config divergence between the two constructions would be invisible.",
      "suggestion": "Have createOrderingComposition expose its triage (it already returns store/orchestrator/enqueue) and reuse it for the fulfillment orchestrator instead of constructing a second stack.",
      "confidence": 0.7,
      "faang_parallel": "This is the composition-root principle from the DI literature (and Guice's singleton-scope rationale at Google): construct shared infrastructure once at the root and thread it down, because 'two instances built from the same config' is only equivalence until the first hotfix touches one path.",
      "metaphor": "Two thermostats in one room, installed from the same spec sheet — indistinguishable until someone recalibrates one and the room starts oscillating.",
      "teachable_moment": "Accidental duplication of stateful infrastructure is invisible precisely when it's harmless. The fix is cheap now (expose triage from the composition) and expensive after the first divergence-induced incident."
    },
    {
      "id": "PB-011",
      "title": "Unbounded in-memory escalation set over worker lifetime",
      "severity": "LOW",
      "category": "resource",
      "description": "The `escalated` Set grows one entry per orderId:ingredient forever and is never pruned when an order reaches a terminal state. On the single-instance Railway deploy with a long-lived process and steady order volume this is a slow leak. The loa:shortcut comment covers the durability tradeoff but not cleanup.",
      "suggestion": "Prune keys for an order when processOrder observes it terminal, or cap the set with an LRU. Trivial now, annoying to diagnose later.",
      "confidence": 0.6,
      "metaphor": "A bouncer who remembers every face ever turned away, forever — fine for a neighborhood bar, a memory problem for a stadium.",
      "teachable_moment": "Every dedup structure needs an eviction story tied to the lifecycle of the thing it dedups. 'Order reached terminal state' is the natural tombstone here; wiring pruning to it costs three lines today.",
      "connection": "The existing loa:shortcut comment names the durability ceiling but not this one — extend the marker so the cleanup trigger ('prune on terminal') is recorded where the Set lives."
    },
    {
      "id": "PB-012",
      "title": "Loopback-only HTTP bypass is genuinely fail-closed and pinned by tests",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/freeside-cli/src/lib/ordering-client.ts",
      "description": "The HTTPS enforcement takes the right shape: an allowlist of exact loopback hostnames rather than a denylist, the bypass env var is inert against any non-loopback host, and the test suite explicitly pins the case where UNSAFE_HTTP is set against a remote http URL and still refuses. Validation happens before any network call, and the tests assert zero server hits on rejection.",
      "suggestion": "None — this is the correct pattern for a test-only transport bypass.",
      "confidence": 0.9,
      "faang_parallel": "This mirrors how Chrome treats localhost as a secure context: a structural allowlist judgment about the address, not a flag that widens trust. The bypass being inert against remote hosts is the property that makes it safe to ship at all.",
      "teachable_moment": "The test asserting *zero server hits* on rejection is the underrated part — it proves fail-closed at the transport layer, not just an error string. That's the difference between testing the message and testing the mechanism."
    },
    {
      "id": "PB-013",
      "title": "Escalation dedup ordered after the emit, with the failure mode reasoned in-code",
      "severity": "PRAISE",
      "category": "reliability",
      "description": "Recording the dedup key only after the escalate event lands (with the comment explaining that recording first would let a store failure permanently suppress escalation for the process lifetime) is exactly the right ordering for an at-least-once signal. Combined with the worker's in-flight tick guard and the whole-body try/catch preventing unhandledRejection from killing the interval loop, the failure-mode reasoning here is thorough and tested (FIX 3 test).",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "This is the mark-after-act ordering that Kafka consumers use for at-least-once semantics (commit offset only after processing) — choosing duplicate-on-crash over lost-forever, and writing down *why*.",
      "teachable_moment": "The comment explaining the rejected alternative ('recording first would suppress escalation forever') is the decision trail done right — the next engineer who 'fixes' the ordering will meet the reasoning before the regression."
    },
    {
      "id": "PB-014",
      "title": "Upstream idempotency contracts documented at the dispatch call sites",
      "severity": "PRAISE",
      "file": "packages/services/ordering/src/http-building-probes.ts",
      "description": "Each dispatch method (ingestSonar, registerScore, manifestWorlds) documents the specific upstream idempotency mechanism (natural-key upsert, advisory lock, replay-returns-existing-slug) that makes crash-time re-dispatch safe, and the e2e suite proves the crash/resume path with a fresh orchestrator instance. This turns an implicit cross-service assumption into a reviewable contract.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "Stripe's API design docs treat idempotency as a *documented contract per endpoint*, not an ambient assumption — and this PR does the client-side mirror: naming, at each call site, which upstream mechanism absorbs the retry.",
      "metaphor": "Each doorway labeled with which side holds the lock — so when someone rebuilds a wall, they know which guarantees they're carrying.",
      "connection": "This is what makes PB-002 fixable with confidence: because the idempotency contracts are written down, 'retry after discarding failed-response state' is provably safe rather than hopefully safe."
    }
  ]
}
```


## Callouts

Three things here are better than most production code review encounters:

- **[Security] The fail-closed bypass (PB-012).** Allowlist of exact loopback hostnames, bypass inert against remote hosts, and — the part worth imitating — tests that assert *zero server hits* on rejection. Most teams test the error message; this tests the mechanism. That's what makes a "test-only escape hatch" shippable.

- **[Quality] Failure-mode reasoning written where it lives (PB-013).** The emit-then-record ordering for escalation dedup, with the rejected alternative explained in a comment, is Kafka-consumer-grade at-least-once thinking. Duplicates-on-crash over silence-forever is the right trade, and it's *documented as a trade*, not left as an accident.

- **[Test Coverage] Idempotency as a reviewable contract (PB-014).** Documenting each upstream's replay mechanism at the dispatch call site, then proving crash/resume with a fresh orchestrator in the e2e suite, converts "we assume retries are safe" into "here is why, per endpoint." This is the foundation the PB-002 fix will stand on.

The connective tissue: this PR holds itself to a strict-parsing, fail-loud standard (parseIntFlag, HTTPS enforcement) — the remaining findings are mostly places where older or adjacent paths (env truthiness, silent timeout fallback, error-body field adoption) haven't yet inherited that spine. Bring them up to the PR's own standard and this subsystem will be genuinely hard to misconfigure.

**Verdict: COMMENT**

Issue #441, opened 2026-07-05
[ordering] sprint-24 follow-ups from FAGAN council: discord health-gate service-path bypass, pre-sprint-24 order stranding, metadata 404 re-dispatch


zkSoju:
Deferred CHANGES_REQUIRED findings from the sprint-24 cross-model council review (claude+codex+cursor, loa-freeside `e68357f1`), verified against source but out of the slice's two-file scope. Parent: #415.

## 1. Discord channel-health gate bypassable via the service path (claude, major)
The escalate-on-unhealthy check (`fulfillment-orchestrator.ts:215-226`) lives only in the worker's `processOrder`. `CommunityOnboardingOrchestrator.process`/`reprobe` (probeAll → merge → `canFulfillCommunityOnboarding` → fulfilled) can advance `discord_observer` and fulfil an order **without ever consulting `discordHealth`**. In zero-operator flow the worker always runs, but the service path remains a bypass (D13.4). Fix direction: hoist the health gate into `canFulfill` or gate the service-path fulfil on a worker-stamped health audit entry.

## 2. Pre-sprint-24 in-flight orders can strand (claude, major)
`canFulfillCommunityOnboarding` now hard-requires `metadata_snapshot ∈ {complete, optional}`. Orders placed before #416 lack the key; `mergeProbedIngredients` back-fills 'pending' from probeAll. If the metadata port is **present** (score-api configured) but the snapshot endpoint 404s for the old contract, the ingredient stays 'pending' indefinitely — the structural-absence self-resolve never fires (port present). Escape hatch today: operator `advance-ingredient`. Fix direction: migration/backfill for in-flight producing orders, or treat probe-404 beyond N ticks as 'optional' with audit.

## 3. metadata_snapshot re-dispatches every tick while endpoint 404s (claude, cleanup)
The conditional dispatch (`fulfillment-orchestrator.ts` metadata enqueue block) re-fires each tick while the status endpoint returns 404 ('pending'). Harmless but noisy; add dispatch dedup (e.g. only when not already `in_progress`, or a per-order dispatch stamp).

All three carry the same test shape: place order → drive the specific path → assert gate/no-redispatch/fulfil.

zkSoju:
Council iteration 2 (loa-freeside `15e6c090`) added two more deferred items:

[...]

PR #443, opened 2026-07-05
fix(platform/ordering): hardening slice — choke-point health gate, slug guard, persistent-404 backstop (#441)

zkSoju:
Spiral `spiral-20260705-d344e8` cycle-1 output + FAGAN council fixes. Closes the verified review-finding cluster on the fulfillment orchestrator.

## What
- **FO-001/PB-003 choke point**: discord channel-health gate moved INTO `CommunityOnboardingOrchestrator.advanceIngredient` — the single place `discord_observer` transitions to complete. Both write paths (worker + service/merge) now pass through it; fail-closed on throwing ports; D13.3 port-absent warn-once restored. Gate test starts from REAL initial ingredients (no fixture contortion).
- **PB-002**: canonicalSlug adopted only from `res.ok` dispatches; slug_divergence likewise.
- **Persistent-404 backstop**: metadata_snapshot pending with port present self-resolves to optional after N counted ticks (default 3, env-validated vs NaN, clamped ≥1) — counted regardless of dispatch availability (dispatch-null was the canonical stall); one-shot dispatch; counter survives failed self-resolve; no per-order leak.
- **Worker composition**: dedicated onboarding orchestrator mirrors intake deps (enqueue) — worker advances were silently dropping enqueue side-effects.
- **Escalation**: terminal-race advance failures no longer escalate (paged an operator about a success).
- +13 tests incl. AC-7 slug-adoption and AC-1 ordering (both previously vacuous per council).

## Review trail
- FAGAN council (claude 12 / gpt 2 / cursor 10 findings): all convergent majors fixed in `f896403c`, each verified against source first.
- Spiral harness again died at the evidence gate (5th instance, package-relative paths) — tracked upstream loa#1175; no gaming this round.

Suite: 185/186 — the 1 red (reprobe hung-probe timeout classification) is proven pre-existing. Domain: platform (packages/services/ordering only).

### DIFF
--- a/packages/services/ordering/bin/fulfillment-orchestrator.ts
+++ b/packages/services/ordering/bin/fulfillment-orchestrator.ts
@@ -6,6 +6,7 @@
  * explicitly deployed; env-config is reused from the intake service.
  */
 import { createFulfillmentOrchestratorWorker, orchestratorEnabled } from '../src/composition.js';
+import { reprobeIntervalMs } from '../src/reprobe-worker.js';
 
 if (!orchestratorEnabled()) {
   // eslint-disable-next-line no-console
@@ -17,7 +18,7 @@ const worker = await createFulfillmentOrchestratorWorker();
 worker.start();
 
 // eslint-disable-next-line no-console
-console.log(`fulfillment-orchestrator started (interval ${process.env.KITCHEN_REPROBE_INTERVAL_SEC ?? 900}s)`);
+console.log(`fulfillment-orchestrator started (interval ${reprobeIntervalMs() / 1000}s)`);
 
 process.on('SIGTERM', () => worker.stop());
 process.on('SIGINT', () => worker.stop());
--- a/packages/services/ordering/src/__tests__/community-onboarding-orchestrator.test.ts
+++ b/packages/services/ordering/src/__tests__/community-onboarding-orchestrator.test.ts
@@ -1,10 +1,10 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
 import { ORDER_LIFECYCLE_SUBJECTS, INITIAL_COMMUNITY_ONBOARDING_INGREDIENTS } from '@freeside/ordering-protocol';
 import { OrderOrchestrator } from '../orchestrator.js';
 import { InMemoryOrderStore, type NewOrder } from '../store.js';
 import { ConfigCapabilityResolver, type CapabilityConfig } from '../resolver.js';
 import { StubTriagePorts } from '../triage-ports.js';
-import { canFulfillCommunityOnboarding, mergeProbedIngredients } from '../community-onboarding-orchestrator.js';
+import { canFulfillCommunityOnboarding, CommunityOnboardingOrchestrator, mergeProbedIngredients } from '../community-onboarding-orchestrator.js';
 import type { AuditPort } from '../audit-acl.js';
 import type { AuditServiceResult } from '@freeside/shadow-audit-service';
 import type { Cta } from '@freeside/shadow-audit-protocol';
@@ -172,5 +172,70 @@ describe('canFulfillCommunityOnboarding — metadata_snapshot gate (T-5)', () =>
 
   it('returns true when metadata_snapshot is optional (NF-6 disabled path)', () => {
     expect(canFulfillCommunityOnboarding({ ...base, metadata_snapshot: 'optional' }, fulfillment)).toBe(true);
+  });
+});
+
+// ── T-2: discord health gate in advanceIngredient (AC-3, AC-4) ────────────────
+
+function discordAdvanceHarness(discordHealth?: {
+  checkChannelHealth(chainId: string, contract: string): Promise<{ healthy: boolean; reason?: string }>;
+}) {
+  const store = new InMemoryOrderStore({ now: () => 1_700_000_000 });
+  const orchestrator = new CommunityOnboardingOrchestrator({
+    store,
+    resolver: new ConfigCapabilityResolver(TRIAGE_CAPS),
+    triage: new StubTriagePorts(),
+    now: () => 1_700_000_000_000,
+    discordHealth,
+  });
+  return { store, orchestrator };
+}
+
+async function placedAndProducing(store: ReturnType<typeof harness>['store'], orchestratorInstance: InstanceType<typeof CommunityOnboardingOrchestrator>) {
+  await store.placeOrder(communityOrder('ord_dc_1'), {
+    subject: ORDER_LIFECYCLE_SUBJECTS.placed,
+    payload: { order_id: 'ord_dc_1', product: 'community-onboarding', inputs_digest: 'b'.repeat(64) },
+  });
+  await orchestratorInstance.process('ord_dc_1', (await store.get('ord_dc_1'))!);
+  return store.get('ord_dc_1');
+}
+
+describe('CommunityOnboardingOrchestrator — discord health gate in advanceIngredient (T-2, AC-3, AC-4)', () => {
+  it('AC-3: advanceIngredient discord_observer→complete with unhealthy port returns ok:false and does not patch the store', async () => {
+    const spy = vi.fn().mockResolvedValue({ healthy: false, reason: 'channel archived' });
+    const { store, orchestrator } = discordAdvanceHarness({ checkChannelHealth: spy });
+    const initial = new InMemoryOrderStore({ now: () => 1_700_000_000 });
+    // Use own store, orchestrator has its own store.
+    await store.placeOrder(communityOrder('ord_dc_1'), {
+      subject: ORDER_LIFECYCLE_SUBJECTS.placed,
+      payload: { order_id: 'ord_dc_1', product: 'community-onboarding', inputs_digest: 'b'.repeat(64) },
+    });
+    await orchestrator.process('ord_dc_1', (await store.get('ord_dc_1'))!);
+    const recordBefore = await store.get('ord_dc_1');
+
+    const result = await orchestrator.advanceIngredient('ord_dc_1', 'discord_observer', 'complete');
+
+    expect(result.ok).toBe(false);
+    expect(result.error).toContain('discord channel-health check failed');
+    // Store must be IDENTICAL before and after the rejected advance.
+    const recordAfter = await store.get('ord_dc_1');
+    expect(recordAfter?.ingredients).toEqual(recordBefore?.ingredients);
+    expect(recordAfter?.operator_audit).toEqual(recordBefore?.operator_audit);
+    void initial;
+  });
+
+  it('AC-4: advanceIngredient discord_observer→optional passes unconditionally (no health check)', async () => {
+    const spy = vi.fn();
+    const { store, orchestrator } = discordAdvanceHarness({ checkChannelHealth: spy });
+    await store.placeOrder(communityOrder('ord_dc_2'), {
+      subject: ORDER_LIFECYCLE_SUBJECTS.placed,
+      payload: { order_id: 'ord_dc_2', product: 'community-onboarding', inputs_digest: 'b'.repeat(64) },
+    });
+    await orchestrator.process('ord_dc_2', (await store.get('ord_dc_2'))!);
+
+    const result = await orchestrator.advanceIngredient('ord_dc_2', 'discord_observer', 'optional');
+
+    expect(result.ok).toBe(true);
+    expect(spy).not.toHaveBeenCalled();
   });
 });
--- a/packages/services/ordering/src/__tests__/fulfillment-orchestrator-wiring.test.ts
+++ b/packages/services/ordering/src/__tests__/fulfillment-orchestrator-wiring.test.ts
@@ -94,6 +94,7 @@ async function wiringHarness(
     resolver: new ConfigCapabilityResolver(TRIAGE_CAPS),
     triage,
     now: () => NOW_MS,
+    discordHealth: opts?.discordHealth,  // T-2: gate lives in advanceIngredient
   });
   await store.placeOrder(order(), {
     subject: ORDER_LIFECYCLE_SUBJECTS.placed,
--- a/packages/services/ordering/src/__tests__/fulfillment-orchestrator.test.ts
+++ b/packages/services/ordering/src/__tests__/fulfillment-orchestrator.test.ts
@@ -477,17 +477,21 @@ describe('FulfillmentOrchestrator — discord channel-health gate (T-4)', () => 
 
   // ── AC-1: spy called exactly once (gate is in advanceIngredient) ─────────────
   it('AC-1: checkChannelHealth spy called once before orchestrator.advance for discord_observer', async () => {
-    const spy = vi.fn().mockResolvedValue({ healthy: true });
+    // ONE ordered log for both the health check and the emitted events, so the
+    // before-relation is actually asserted (not merely both-happened).
+    const callOrder: string[] = [];
+    const spy = vi.fn().mockImplementation(async () => {
+      callOrder.push('HEALTH_CHECK');
+      return { healthy: true };
+    });
     const { store, orchestrator } = await discordHarness({
       discordStatus: 'complete',
       discordHealth: { checkChannelHealth: spy },
     });
-    const eventsBeforeAdvance: string[] = [];
-    // Track order of events by intercepting emit — proxy the store's appendEvent.
-    const callOrder: string[] = [];
     const origAppend = store.appendEvent.bind(store);
     store.appendEvent = async (orderId, event) => {
-      callOrder.push(event.subject as string);
+      const ingredient = (event.payload as { ingredient?: string } | undefined)?.ingredient;
+      callOrder.push(`${event.subject as string}:${ingredient ?? '-'}`);
       return origAppend(orderId, event);
     };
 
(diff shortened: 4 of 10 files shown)


zkSoju (commented):
## Summary

This hardening slice does the right structural thing — moving the Discord health gate into `CommunityOnboardingOrchestrator.advanceIngredient` so the gate travels with the state machine rather than one caller — and the test upgrades (AC-1's ordered log, AC-7's non-vacuous adoption check) show real rigor. The primary concern is that the choke point is not yet choking everything: the intake composition's onboarding instance appears to lack the port, and the deprecated-but-tolerated `discordHealth` field on `FulfillmentOrchestrator` gives stale wirings a silent path to compile-and-miss. Verdict: **COMMENT**.

## Architectural Observations

The recurring theme across FO-HG-001, FO-HG-002, and FO-DEPS-003 is a single lesson: **a choke point is a claim about topology, not about code**. Writing the gate in one class is necessary but not sufficient — every instance of that class must carry the port, and every legacy wiring site must fail loudly rather than quietly no-op. Google's BeyondCorp migration hit exactly this: the access proxy was airtight, but the project's real work was hunting down every path that didn't route through it. The gate you can bypass by construction is a suggestion, not a gate.

The second thread (FO-LEAK-004, FO-RETRY-005) is the classic long-lived-worker discipline: every early-return is an exit path, and every exit path must settle its accounts (map entries, retry counters, log lines). These are low-severity here because the maps are bounded by order volume, but the *invariant as stated in comments* should match the code — future readers will trust the comment.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FO-HG-001",
      "title": "Intake orchestrator's CommunityOnboardingOrchestrator instance still lacks discordHealth — gate is not a true choke point",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "The health gate now lives in CommunityOnboardingOrchestrator.advanceIngredient, billed as the choke point. But composition wires discordHealth only into the NEW onboarding instance created for the fulfillment worker. The intake composition's own communityOnboarding instance (created inside createOrderingComposition, previously reused here) is not shown receiving discordHealth. Any advanceIngredient call routed through the intake instance (operator/API-driven advances, other callers) will take the port-absent fallback and advance discord_observer→complete with only a once-per-process console.warn. The gate is enforced on only one of at least two live instances.",
      "suggestion": "Wire discordHealth into the intake orchestrator's CommunityOnboardingOrchestrator inside createOrderingComposition as well (or share a single onboarding instance), so every advanceIngredient path passes the same gate. Add a test asserting the intake-composed instance carries the port.",
      "confidence": 0.6,
      "faang_parallel": "Google's BeyondCorp papers stress that the value of a policy enforcement point is proportional to the completeness of routing through it — the multi-year effort was not building the proxy but eliminating every path around it. A PEP that one caller bypasses by construction provides false assurance.",
      "metaphor": "You installed a metal detector at Terminal A, but Terminal B shares the same tarmac and its door is propped open. Passengers don't choose terminals adversarially — but the one incident that matters will walk through the open door.",
      "teachable_moment": "When you move a gate 'down' into a shared class, the follow-up question is always: how many instances of that class exist, and does each carry the dependency? A gate on a class is a gate on nothing; a gate on every instance is a choke point.",
      "connection": "Same root cause as FO-DEPS-003 — two construction sites for one class invite drift. A shared factory would resolve both findings with one change.",
      "decision_trail": "Document in composition.ts why two onboarding instances exist (worker vs intake) and which deps each must carry — or record the decision to unify them."
    },
    {
      "id": "FO-HG-002",
      "title": "Deprecated FulfillmentOrchestrator.discordHealth is silently ignored — no runtime signal for stale wirings",
      "severity": "MEDIUM",
      "category": "api-contract",
      "description": "FulfillmentOrchestratorDeps.discordHealth is kept for typechecking but ignored. A composition site that wires the port here (and not on the onboarding orchestrator) keeps compiling and appears configured, yet the health gate is entirely absent — the failure mode is silent except for the once-per-process warning emitted from the other class. The @deprecated JSDoc only helps callers who read hover text.",
      "suggestion": "In the FulfillmentOrchestrator constructor, emit a one-time console.warn when deps.discordHealth is provided (e.g. 'discordHealth on FulfillmentOrchestrator is ignored; wire it on CommunityOnboardingOrchestrator'), or remove the field and update call-sites in the same PR.",
      "confidence": 0.7,
      "faang_parallel": "Amazon's internal API deprecation practice distinguishes 'deprecated' (still works, warns) from 'tombstoned' (fails loudly). A parameter that is accepted and silently dropped is the worst of both — AWS SDK reviews flag accepted-but-ignored fields as a contract violation because callers cannot distinguish 'configured' from 'no-op'.",
      "metaphor": "A light switch that's been disconnected from the wiring but left on the wall. Everyone who flips it believes the light works — until the night they actually need it.",
      "teachable_moment": "During a dependency migration, the old attachment point must either keep working or refuse loudly. 'Compiles but does nothing' is the one behavior that defeats both the type system and the operator.",
      "connection": "This is the runtime-signal complement to FO-HG-001: 001 is the gate missing where it's needed; 002 is the gate appearing present where it isn't."
    },
    {
      "id": "FO-DEPS-003",
      "title": "New onboarding instance mirrors enqueue but may still drop other side-effect deps (opsChannel)",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "The in-code comment correctly names the hazard ('dropped composition deps') and mirrors enqueue, but CommunityOnboardingOrchestratorDeps also declares opsChannel?: PrivateOpsPublisher. If the intake instance carries opsChannel, worker-driven advances through the new dedicated instance will silently lose private-ops notifications — exactly the failure class the comment warns about, one field over.",
      "suggestion": "Audit CommunityOnboardingOrchestratorDeps field-by-field against the intake instance's construction and mirror every side-effect dep (opsChannel at minimum), or extract a shared factory so the two instances cannot drift.",
      "confidence": 0.5,
      "faang_parallel": "This is why Guice/Dagger-style DI won inside Google: hand-mirrored constructor sites drift the moment a dep is added, and the compiler can't catch an *optional* dep that one site forgot. Optional deps are the drift vector — required ones fail the build.",
      "metaphor": "Two hand-copied versions of the same recipe in two kitchens. Someone adds saffron to one copy; the other kitchen keeps serving the bland version and nobody complains, because absence doesn't taste like an error.",
      "teachable_moment": "Optional dependencies convert wiring mistakes from compile errors into silent behavior differences. Every `field?:` in a Deps interface is a place where two construction sites can disagree without anyone noticing — which is the argument for exactly one construction site.",
      "connection": "Shares a fix with FO-HG-001: a single shared factory (or single shared instance) collapses both drift risks."
    },
    {
      "id": "FO-LEAK-004",
      "title": "metadataPendingTicks entry leaks when an order goes terminal while metadata is pending",
      "severity": "LOW",
      "category": "resource-management",
      "description": "The else-branch delete only runs when processOrder reaches the metadata block. An order that becomes terminal (e.g., cancelled/failed via escalation) while metadata_snapshot is still pending early-returns at the top of processOrder (isTerminal check), so its Map entry is never deleted. The comment claims settled orders don't leak, but the terminal early-return bypasses the cleanup. Bounded by order volume and reset on restart, so impact is small, but the stated invariant is not fully met.",
      "suggestion": "Delete the metadataPendingTicks entry (and consider the escalated Set entry) in the terminal early-return path: if (isTerminal(record.state)) { this.metadataPendingTicks.delete(orderId); return; }",
      "confidence": 0.7,
      "faang_parallel": "Kubernetes controllers solve this with finalizers: cleanup is attached to the object's lifecycle transition, not to the happy-path handler, precisely because objects can reach terminal states through many doors.",
      "metaphor": "A coat check that only returns tickets at the main exit. Guests who leave through the fire exit keep their tickets forever — the rack slowly fills with coats nobody will claim.",
      "teachable_moment": "Per-entity bookkeeping in a long-lived worker should be cleaned up at the state transition (entering terminal), not inside one processing branch. Enumerate the exit paths; each one must settle the ledger — and don't let a comment claim an invariant the code only mostly enforces.",
      "connection": "The escalated Set has the same lifecycle shape; auditing both together turns one fix into a pattern."
    },
    {
      "id": "FO-RETRY-005",
      "title": "Self-resolve failure path retries silently forever on persistent non-terminal errors",
      "severity": "LOW",
      "category": "error-handling",
      "description": "When the persistent-404 self-resolve's advanceIngredient returns ok:false for a reason other than terminal (e.g., 'invalid order inputs'), the counter is kept (correctly, to avoid re-dispatch) and the branch re-fires every tick with no logging, no escalation, and no backoff. A persistently failing self-resolve becomes an invisible hot loop per tick.",
      "suggestion": "Log the failure reason on non-terminal advanceIngredient failure in the self-resolve branch, and consider escalating after repeated failures rather than retrying indefinitely in silence.",
      "confidence": 0.6,
      "faang_parallel": "The Google SRE book's retry chapter and every mature queue system (SQS, Pub/Sub) converge on the same triad: bounded retries, dead-letter escalation, and per-failure observability. An unbounded silent retry is indistinguishable from a healthy idle loop on every dashboard.",
      "metaphor": "A doorbell stuck ringing at a house where nobody's home — and no camera recording that it's ringing. From the street, silence.",
      "teachable_moment": "The backstop for the backstop matters: this branch exists to rescue stuck orders, so a stuck rescue is precisely the case operators most need to see. Any retry loop earns, at minimum, a log line per failure and an escalation ceiling.",
      "connection": "The PR already distinguishes benign terminal races from real failures elsewhere (FO-PRAISE-010) — this branch deserves the same discrimination."
    },
    {
      "id": "FO-TEST-006",
      "title": "Dead code in AC-3 test: unused 'initial' store and 'void initial'",
      "severity": "LOW",
      "category": "test-quality",
      "file": "packages/services/ordering/src/__tests__/community-onboarding-orchestrator.test.ts",
      "description": "AC-3 constructs 'const initial = new InMemoryOrderStore(...)' that is never used except for a trailing 'void initial;' to appease lint, alongside a comment ('Use own store, orchestrator has its own store') that no longer describes the code. The metadataTickHarness restoreEnv closure in the fulfillment test file is similarly never invoked (tests manage env via afterEach directly). Both are leftover scaffolding that confuses future readers.",
      "suggestion": "Remove the unused 'initial' store, the 'void initial;' line, and the stale comment; drop the unused opts.maxTicks/restoreEnv machinery from metadataTickHarness or use it consistently instead of direct env writes.",
      "confidence": 0.9,
      "faang_parallel": "Google's Testing Blog ('Testing on the Toilet') repeatedly hammers this: tests are documentation, and dead scaffolding in a test misdocuments the system under test. A `void x;` to silence lint is the linter telling you the code is dead — listen to it rather than muting it.",
      "metaphor": "Stage props left standing after the script was rewritten. The audience assumes the fake door matters to the plot; it doesn't, and they spend the second act waiting for someone to walk through it.",
      "teachable_moment": "When a lint suppression is needed to keep code, the code is usually the problem, not the lint. Deleting is cheaper than explaining."
    },
    {
      "id": "FO-TEST-007",
      "title": "Module-level warnedDiscordHealthAbsent flag has no test reset, unlike sibling pattern",
      "severity": "LOW",
      "category": "testability",
      "file": "packages/services/ordering/src/community-onboarding-orchestrator.ts",
      "description": "The once-per-process warning flag moved here as a module-level let with no exported reset. kitchen-triage-ports.ts establishes the codebase pattern of exporting a reset function (resetShadowProducerlessWarning) so tests can assert the warning fires. As written, the D13.3 fallback warning is untestable after the first test in a process triggers it, and no test in this PR covers the port-absent warning path.",
      "suggestion": "Export a resetDiscordHealthAbsentWarning() helper mirroring resetShadowProducerlessWarning, and add a test asserting the warning fires exactly once when the port is absent and discord_observer advances to complete.",
      "confidence": 0.7,
      "faang_parallel": "Jest's module registry (resetModules) exists precisely because module-level mutable state leaks across tests; Google's internal style guidance treats untestable-after-first-use global state as a design smell, not a test problem.",
      "metaphor": "A smoke alarm engineered to sound exactly once in its lifetime. Fine in production if you only want one page — but you can never run a second fire drill.",
      "teachable_moment": "Once-per-process state needs an escape hatch for tests, and the codebase already paid for the pattern next door in kitchen-triage-ports.ts. Consistency here is nearly free — and the warning is load-bearing (it's the only signal for FO-HG-001's fallback path), so it deserves a test.",
      "connection": "This warning is the sole runtime evidence of the gap in FO-HG-001; making it testable raises confidence in the whole gate design."
    },
    {
      "id": "FO-PRAISE-008",
      "title": "Slug-adoption guard is minimal and correctly scoped, with a non-vacuous test",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "The T-1 fix adds res.ok to both the worlds slug adoption and the score divergence comparison, closing the failed-dispatch-overwrites-slug hole without restructuring the loop. AC-7 was upgraded from a vacuous 'worlds was dispatched' assertion to observing adoption indirectly via the slug_divergence event — the test now fails if adoption breaks.",
      "suggestion": "None — this is the right shape for the fix and the test.",
      "confidence": 0.9,
      "faang_parallel": "This is the surgical-diff discipline Linus enforces on kernel patches: fix the invariant at the exact site of violation, resist the urge to restructure. Small diffs are reviewable diffs.",
      "teachable_moment": "Upgrading a vacuous assertion to an observable-behavior assertion is often more valuable than the production fix itself — the old AC-7 would have passed forever regardless of correctness."
    },
    {
      "id": "FO-PRAISE-009",
      "title": "NaN-safe env parsing with a documented lower clamp on maxPendingTicks",
      "severity": "PRAISE",
      "category": "robustness",
      "description": "The maxPendingTicks initializer explicitly guards against the NaN >= n always-false trap (malformed env silently disabling the backstop) and clamps to >=1 so 0 cannot self-resolve before a counted tick — with the reasoning stated in a comment that names the constraint, not the mechanics.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "Netflix's config-driven chaos taught the industry that a misparsed config silently disabling a safety mechanism is worse than a crash at startup — this initializer internalizes that lesson at the parse site.",
      "teachable_moment": "The comment style here is the gold standard: it states the constraint the code can't express ('NaN comparison silently disables the backstop'), not what the next line does."
    },
    {
      "id": "FO-PRAISE-010",
      "title": "AC-1 asserts the before-relation via a single ordered log, and terminal-race is correctly not escalated",
      "severity": "PRAISE",
      "category": "test-quality",
      "file": "packages/services/ordering/src/__tests__/fulfillment-orchestrator.test.ts",
      "description": "AC-1 interleaves the health-check spy and appendEvent into one ordered log so the test proves ordering, not mere co-occurrence — a common weakness in gate tests. On the production side, distinguishing 'order already terminal' (benign race, return) from gate rejection (escalate) prevents paging operators about successes.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "The single-ordered-log technique is how Jepsen-style tests prove happens-before: co-occurrence assertions pass under reorderings that violate the invariant; a shared log cannot lie about order. And suppressing the benign-race page is textbook alert-fatigue hygiene from the SRE playbook.",
      "metaphor": "Two separate guest books at two doors tell you both guests arrived; one shared guest book tells you who arrived first."
    },
    {
      "id": "FO-SPEC-011",
      "title": "Per-instance env snapshot of METADATA_PROBE_MAX_PENDING_TICKS is invisible to runtime reconfiguration",
      "severity": "SPECULATION",
      "category": "configuration",
      "description": "maxPendingTicks is captured once at construction via an IIFE field initializer. For a long-lived worker this is fine, but it diverges from REPROBE_COOLDOWN_MS-style module constants in siblings and means any future hot-reconfig or multi-instance test that mutates env after construction will silently see the stale value. Flagged only because the tests already dance around this with careful set-before-construct ordering.",
      "suggestion": "If dynamic tuning is ever needed, read the env in processOrder or accept maxPendingTicks as an explicit dep; otherwise document that the value is construction-time-fixed.",
      "confidence": 0.3,
      "teachable_moment": "When tests must carefully order 'set env, then construct', that ordering constraint is undocumented API. A one-line comment ('construction-time snapshot; env changes after construct are ignored') converts tribal knowledge into contract. Low confidence — construction-time capture is a defensible default for a worker; this is a documentation ask, not a bug."
    }
  ]
}
```


## Callouts

- **[Quality] The gate migration direction is right.** Moving enforcement from the caller (`FulfillmentOrchestrator`) into the state-machine owner (`CommunityOnboardingOrchestrator.advanceIngredient`) is the correct architectural move — gates belong where the transition happens, not where one client happens to invoke it. FO-HG-001/002 are completion work on a sound decision, not a challenge to it.
- **[Test Coverage] AC-1's ordered-log technique (FO-PRAISE-010)** is worth extracting as a house pattern for every future "X must happen before Y" test in this package — it's the difference between proving ordering and hoping for it.
- **[Operational] The NaN guard and lower clamp (FO-PRAISE-009)** show exactly the right paranoia at the config boundary: malformed env degrades to the safe default rather than silently disabling the backstop.
- **[Quality] The slug-adoption fix (FO-PRAISE-008)** is a model surgical diff — two `res.ok` checks, no restructuring, and a test upgraded from vacuous to load-bearing in the same slice.

The through-line for the next slice: unify the two `CommunityOnboardingOrchestrator` construction sites (one factory or one instance). That single change retires FO-HG-001 and FO-DEPS-003 structurally and shrinks FO-HG-002 to a tombstone-removal chore.

**Verdict: COMMENT**

Issue #303, opened 2026-06-24
[platform] @freeside/dune-meter: "budget blowout structurally impossible" claim is not enforced in code (1 BLOCKER)


zkSoju:
## Summary

`@freeside/dune-meter` (landed on `main` via #301, commit `6e66c58a`) ships with the headline claim that the EXP-002 budget blowout is **"made structurally impossible."** A SCAR cost-safety review found that claim is **not delivered in code**: spend is mostly *observed after the fact*, and the only true hard-abort is an external, operator-configured Dune dashboard setting this code neither sends nor verifies.

This matters because dune-meter is the gate meant to protect the Corpus Engine MVP from a repeat of EXP-002. Findings were produced while reviewing #282 (now closed as superseded); they apply to the code currently on `main`.

## Findings (all grounded against `packages/dune-meter/` on `main`)

| Sev | Location | Finding | Fix |
|-----|----------|---------|-----|
| **BLOCKER** | `bin/dune-meter.mjs:144-172` | `--cap <credits>` is never sent to Dune. `executeQuery`/`executeSql` pass only `{performance}`; `capExceeded = credits > flags.cap` is computed *after* the execution already spent. `--cap` is post-hoc accounting, not enforcement. The "structural hard-abort" is a manual dashboard cap (`README.md:85-87`). | Send a real per-execution cost limit to the API, or refuse to run unless the account cap is verifiable; stop claiming structural impossibility in code/title. |
| **HIGH** | `src/budget-ledger.mjs:74-86` | `recordSpend` is a lock-free read-modify-write. The atomic temp-file+rename prevents a torn file but not a lost update: two concurrent `run`s read the same `spent_credits`, both pass the pre-run check, both write → budget silently over-spends. No concurrency test. | File lock / `O_EXCL` reservation, or document + enforce single-writer. |
| **HIGH** | `src/dune-client.mjs:226-238` | The "estimate" probe runs real `COUNT(*)` + `LIMIT 1` executions that are never recorded to the ledger. A `COUNT(*)` over an unbounded scan (the exact EXP-002 `evt_transfer` shape) pays the full columnar scan — so the "free" estimate path can itself reproduce the blowout, unmetered. | Bound/era-check probe SQL, meter probe spend into the ledger, or cap the probe. |
[...]

PR #448, opened 2026-07-06
fix(platform/dune-meter): enforce the budget claims — await recordSpend, probe gating, honest --cap docs (#303)

zkSoju:
Spiral `spiral-20260706-359e28` output + salvage. Addresses the SCAR findings in #303: the 'budget blowout structurally impossible' claim now matches what the code guarantees.

## What (T-1..T-5, commits `d78b5b40` + `9c662096`)
- **recordSpend awaited** — spend hits the ledger before the run reports (was fire-and-forget: HIGH lost-update window)
- **Probe split + inter-probe gate** — estimate probes are bounded, metered into the ledger, and gated between probes (the unmetered COUNT(*) probe could itself reproduce EXP-002)
- **Ceiling persisted + re-applied** — DUNE_BUDGET_CEILING re-read on ledger load (lowering the env after first run now takes effect)
- **--force bounded** — requires an explicit small ceiling instead of relying on the unenforced dashboard cap; the test encoding the false belief is fixed
- **Honest docs (T-5)** — README/comments no longer claim structural impossibility where enforcement is post-hoc; --cap semantics documented as what they are
- Suite: **46/46**

## Plus: DISS-001 (FAGAN adversarial voice, BLOCKING)
My #445 errexit fix captured the negation's exit status (always 0) — classification unreachable a second way. Fixed with `var=$(cmd) || rc=$?` and proven inline (false→1, success→0). This is a hotfix to merged main behavior.

## Review trail
Harness review died on session quota (both cycle attempts); the FAGAN adversarial voice (codex, separate quota) ran and returned exactly one BLOCKING finding — fixed above. Dune-meter changes passed it clean.

Domain: platform (packages/dune-meter) + shared (one workflow hotfix).

### DIFF
--- a/.claude/scripts/spiral-evidence.sh
+++ b/.claude/scripts/spiral-evidence.sh
@@ -677,7 +677,8 @@ _parse_sprint_paths() {
         # `\`spiral-harness.sh\`` (which is a reference, not a deliverable
         # declaration). Accepts `.claude/scripts/...`, `src/...`, etc.
         echo "$content" | grep -oE "\`[^\`]*/[^\`]+\.${ext_re}\`" 2>/dev/null | \
-            sed 's/^`//; s/`$//'
+            sed 's/^`//; s/`$//' | \
+            grep -v ' '
         # Pattern 2: bare paths rooted at well-known *top-level* repo prefixes.
         # Anchored to start-of-line or whitespace/non-path-char to prevent
         # substring matches within longer paths — without the anchor,
--- a/.claude/scripts/spiral-evidence.sh
+++ b/.claude/scripts/spiral-evidence.sh
@@ -677,8 +677,7 @@ _parse_sprint_paths() {
         # `\`spiral-harness.sh\`` (which is a reference, not a deliverable
         # declaration). Accepts `.claude/scripts/...`, `src/...`, etc.
         echo "$content" | grep -oE "\`[^\`]*/[^\`]+\.${ext_re}\`" 2>/dev/null | \
-            sed 's/^`//; s/`$//' | \
-            grep -v ' '
+            sed 's/^`//; s/`$//'
         # Pattern 2: bare paths rooted at well-known *top-level* repo prefixes.
         # Anchored to start-of-line or whitespace/non-path-char to prevent
         # substring matches within longer paths — without the anchor,
--- a/.github/workflows/agent-ci.yml
+++ b/.github/workflows/agent-ci.yml
@@ -106,7 +106,10 @@ jobs:
         run: ./node_modules/.bin/vitest run --config vitest.agent-ci.config.ts
 
       - name: Integration tests
-        run: npx vitest run tests/integration/
+        # RC-3 fix: `npx vitest run tests/integration/` without --config scans the entire
+        # repo matching the path pattern — collects sietch + worktree files (600+, all failing).
+        # vitest.agent-ci-integration.config.ts scopes to tests/integration/**/*.test.ts only.
+        run: npx vitest run --config vitest.agent-ci-integration.config.ts
         env:
           REDIS_URL: redis://127.0.0.1:6379
           BERACHAIN_RPC_URLS: https://rpc.test.example.com
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,6 +268,10 @@ jobs:
         run: npm run test:integration
         env:
           REDIS_URL: redis://localhost:6379
+          # RC-2 fix: DATABASE_URL placeholder so config.ts Zod validation passes and
+          # MigrationEngine/billing-agent-sovereignty tests do not silently skip.
+          # No Postgres service is provisioned — sietch integration tests use SQLite in-memory.
+          DATABASE_URL: postgresql://test:test@localhost:5432/freeside_test
           # config.ts validates all env at import (Zod); the integration vitest project has no
           # setupFiles, so unlike the unit project (tests/setup-unit.ts) these required vars are
           # unset → "Configuration validation failed: triggerDev.secretKey". Test placeholders,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -287,3 +287,61 @@ jobs:
           RATE_LIMIT_SALT: test-rate-limit-salt-value
           NODE_ENV: test
           LOG_LEVEL: error
+
+  # =============================================================================
+  # WORKER DB TESTS — EventNatsConsumer integration tests (real Postgres)
+  # =============================================================================
+  worker-db-tests:
+    name: Worker DB Tests (EventNatsConsumer)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: test
+          POSTGRES_PASSWORD: test
+          POSTGRES_DB: freeside_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U test"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    env:
+      DATABASE_URL: postgresql://test:test@localhost:5432/freeside_test
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '22'
+
+      - name: Install worker dependencies
+        working-directory: apps/worker
+        run: npm ci
+
+      - name: Apply test schema
+        run: |
+          psql "$DATABASE_URL" -f apps/worker/tests/fixtures/worker-test-schema.sql
+
+      - name: Skip-guard — assert DATABASE_URL is configured
+        # Belt-and-suspenders: the describe.skipIf gate in EventNatsConsumer.test.ts
+        # silently passes with zero assertions when DATABASE_URL is absent.
+        # This step makes that configuration error loud and red instead of green.
+        run: |
+          if [ -z "${DATABASE_URL}" ]; then
+            echo "::error::DATABASE_URL is not set — EventNatsConsumer tests will silently skip"
+            exit 1
+          fi
+          echo "Skip-guard passed: DATABASE_URL is configured"
+
+      - name: Run worker DB tests
+        working-directory: apps/worker
+        run: npm run test:run -- --reporter=verbose src/consumers/EventNatsConsumer.test.ts
--- a/.github/workflows/cluster-compliance.yml
+++ b/.github/workflows/cluster-compliance.yml
@@ -122,13 +122,12 @@ jobs:
             # GITHUB_TOKEN lacks contents:read on private cells. Classify as 'inaccessible'
             # (::warning::, does not fail the job) — not as 'non-compliant' (::error::).
             CELL_DIR="cell-$slug"
-            # `if ! var=$(cmd)` keeps errexit from killing the step on clone failure —
-            # a bare assignment carries the command's exit status under `bash -e {0}`
-            # and the classification below would be unreachable (BB #445 HIGH).
+            # `var=$(cmd) || rc=$?` keeps errexit from killing the step AND captures the
+            # REAL exit code. (Two prior attempts were both wrong: a bare assignment dies
+            # under `bash -e {0}` before classification; `if ! var=$(cmd); then rc=$?`
+            # captures the NEGATION's status, which is always 0 — FAGAN DISS-001.)
             clone_exit=0
-            if ! clone_output=$(gh repo clone "$repo" "$CELL_DIR" -- --depth=1 --quiet 2>&1); then
-              clone_exit=$?
-            fi
+            clone_output=$(gh repo clone "$repo" "$CELL_DIR" -- --depth=1 --quiet 2>&1) || clone_exit=$?
             if [ $clone_exit -ne 0 ]; then
               # Check if this is an access error (403/404) vs other failure
               if echo "$clone_output" | grep -qE '(403|404|Repository not found|access denied|not found)'; then
--- a/.github/workflows/cluster-compliance.yml
+++ b/.github/workflows/cluster-compliance.yml
@@ -65,11 +65,16 @@ jobs:
           # which does NOT accept -o=json — the RED-CI bug (FL T7). Pin the mikefarah
           # Go-yq binary explicitly so the JSON conversion is deterministic.
           YQ_VERSION=v4.44.3
+          # SHA-256 supply-chain hardening (T-4/RC-3a): hash hardcoded from the official
+          # mikefarah/yq v4.44.3 release (extracted via extract-checksum.sh SHA-256 yq_linux_amd64).
+          # Never fetch the hash at runtime — that defeats the integrity check.
+          YQ_SHA256="a2c097180dd884a8d50c956ee16a9cec070f30a7947cf4ebf87d5f36213e9ed7"
           sudo wget -qO /usr/local/bin/yq "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64"
+          echo "${YQ_SHA256}  /usr/local/bin/yq" | sha256sum --check --strict || {
+            echo "::error::yq binary SHA-256 mismatch — possible CDN corruption or substitution"
+            exit 1
+          }
           sudo chmod +x /usr/local/bin/yq
-          # Assert the pinned mikefarah Go-yq is what actually landed — catches a
-          # wrong/corrupted binary (FAGAN). FOLLOW-UP for full supply-chain hardening:
-          # pin the published SHA256 of yq_linux_amd64 ${YQ_VERSION} and verify it here.
           yq --version | grep -q "mikefarah" || { echo "::error::yq is not the mikefarah Go build"; exit 1; }
           yq --version | grep -q "${YQ_VERSION#v}" || { echo "::error::yq version mismatch (expected ${YQ_VERSION})"; exit 1; }
 
@@ -80,8 +85,17 @@ jobs:
           # Registry shape: top-level `modules:` is a MAP keyed by slug (e.g. freeside-sonar).
           # Pipeline: Go-yq converts YAML→JSON, jq processes JSON with full filter language
           # (Go-yq's own query language doesn't support jq-style sub()/capture()/to_entries).
-          yq -o=json packages/freeside-registry/registry.yaml | \
-            jq -r '.modules | to_entries | .[] |
+          #
+          # RC-3b guard: check for null/empty .modules before proceeding. A null result means
+          # the registry.yaml key changed or the file is empty — emit a specific error rather
+          # than silently producing an empty roster and passing vacuously.
+          roster=$(yq -o=json '.modules' packages/freeside-registry/registry.yaml)
+          if [ -z "$roster" ] || [ "$roster" = "null" ]; then
+            echo "::error::yq produced empty/null .modules from packages/freeside-registry/registry.yaml — registry.yaml shape may have changed or the file is corrupt"
+            exit 1
+          fi
+          echo "$roster" | \
+            jq -r 'to_entries | .[] |
               "\(.key):\(.value.git_url | sub("https://github.com/"; "") | sub("\\.git$"; ""))"' \
             > cell-list.txt
           echo "Cells found:"
@@ -103,15 +117,27 @@ jobs:
             echo ""
             echo "═══ Auditing $slug ($repo) ═══"
 
-            # Shallow-clone the cell into a working dir
+            # Shallow-clone the cell into a working dir.
+            # RC-3c guard: 403/404 responses from gh repo clone indicate the workflow's
+            # GITHUB_TOKEN lacks contents:read on private cells. Classify as 'inaccessible'
+            # (::warning::, does not fail the job) — not as 'non-compliant' (::error::).
             CELL_DIR="cell-$slug"
-            gh repo clone "$repo" "$CELL_DIR" -- --depth=1 --quiet 2>&1 || {
-              echo "  ⚠ Could not clone $repo — skipping"
-              echo "{ \"slug\": \"$slug\", \"status\": \"UNREACHABLE\", \"reason\": \"clone failed\" }" \
-                > "audit-results/$slug.json"
+            clone_output=$(gh repo clone "$repo" "$CELL_DIR" -- --depth=1 --quiet 2>&1)
+            clone_exit=$?
+            if [ $clone_exit -ne 0 ]; then
+              # Check if this is an access error (403/404) vs other failure
+              if echo "$clone_output" | grep -qE '(403|404|Repository not found|access denied|not found)'; then
+                echo "::warning::$slug ($repo): inaccessible — GITHUB_TOKEN lacks read access (403/404). Classifying as inaccessible, not non-compliant."
+                echo "{ \"slug\": \"$slug\", \"status\": \"inaccessible\", \"reason\": \"403/404 from gh repo clone — private repo or no token access\" }" \
+                  > "audit-results/$slug.json"
+              else
+                echo "::warning::$slug ($repo): could not clone ($clone_exit). Classifying as inaccessible."
+                echo "{ \"slug\": \"$slug\", \"status\": \"inaccessible\", \"reason\": \"clone failed: $clone_output\" }" \
+                  > "audit-results/$slug.json"
+              fi
               WARN_CELLS+=("$slug")
               continue
-            }
+            fi
 
             # Run existence + topology audit (vendored from construct-freeside auditing-cluster-cells skill)
             EXISTENCE_JSON=$(GITHUB_DIR="$(pwd)" bash .github/scripts/audit-cluster-cells.sh \
--- a/.github/workflows/cluster-compliance.yml
+++ b/.github/workflows/cluster-compliance.yml
@@ -122,8 +122,13 @@ jobs:
             # GITHUB_TOKEN lacks contents:read on private cells. Classify as 'inaccessible'
             # (::warning::, does not fail the job) — not as 'non-compliant' (::error::).
             CELL_DIR="cell-$slug"
-            clone_output=$(gh repo clone "$repo" "$CELL_DIR" -- --depth=1 --quiet 2>&1)
-            clone_exit=$?
+            # `if ! var=$(cmd)` keeps errexit from killing the step on clone failure —
+            # a bare assignment carries the command's exit status under `bash -e {0}`
+            # and the classification below would be unreachable (BB #445 HIGH).
+            clone_exit=0
+            if ! clone_output=$(gh repo clone "$repo" "$CELL_DIR" -- --depth=1 --quiet 2>&1); then
+              clone_exit=$?
+            fi
             if [ $clone_exit -ne 0 ]; then
               # Check if this is an access error (403/404) vs other failure
               if echo "$clone_output" | grep -qE '(403|404|Repository not found|access denied|not found)'; then
--- a/apps/worker/src/consumers/EventNatsConsumer.test.ts
+++ b/apps/worker/src/consumers/EventNatsConsumer.test.ts
@@ -1,0 +1,542 @@
+/**
+ * EventNatsConsumer Integration Tests
+ * Sprint: NATS Guild & Member Lifecycle Event Consumers
+ *
+ * Tests run against a real PostgreSQL instance.
+ * Set DATABASE_URL env var or SKIP_DB_TESTS=true to skip.
+ *
+ * AC coverage: AC-1 through AC-18
+ */
+
+import { describe, it, expect, beforeAll, afterAll, beforeEach, vi } from 'vitest';
+import pino from 'pino';
+import postgres from 'postgres';
+import { drizzle } from 'drizzle-orm/postgres-js';
+import { eq, and } from 'drizzle-orm';
+import * as schema from '../data/schema.js';
+import {
+  initDatabase,
+  getDatabase,
+  upsertCommunity,
+  markCommunityInactive,
+  upsertProfile,
+  markProfileInactive,
+} from '../data/database.js';
+import {
+  createDefaultNatsEventHandlers,
+  type GatewayEventPayload,
+} from './EventNatsConsumer.js';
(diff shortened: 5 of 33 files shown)


zkSoju (commented):
This invocation is a Bridgebuilder review-enrichment task — the deliverable is the review text itself, no code changes or exploration required, so I'm producing it directly.

# Bridgebuilder Review — PR #448

**Verdict: REQUEST_CHANGES**

## Summary

This PR does the rare and admirable thing of making documentation *honest* and then pinning that honesty with tests — the `--cap` corrections, the xmax=0 insert detection, and the split-probe metering are genuinely strong work. But the budget-ceiling persistence feature (BB-448-001) inverts its own intent: a plain `dune-meter budget` call can silently *raise* a previously tightened ceiling, and that same code path writes to the ledger outside the lock the PR just built (BB-448-002). The budget machinery — the thing this PR exists to make trustworthy — needs one more pass before it deserves the trust.

## Architectural Meditation

There's a single thread running through the highest-value findings here: **the difference between "absent" and "default-valued" at an API boundary**. `readLedger(ceiling = DEFAULT_CEILING)` cannot distinguish "the operator set 2500" from "the operator set nothing," and that ambiguity is what lets a tightened budget quietly relax. This is the exact problem that drove Kubernetes to use pointer types for optional API fields and protobuf3 to walk back its "no field presence" decision — entire ecosystems have paid to relearn that a default argument is not a statement of intent. The same disease appears in miniature in `setJoinedAt` (a parameter that *looks* like intent but carries none) and in the CI clone classification (where "we couldn't check" is allowed to masquerade as "nothing was wrong"). Systems stay trustworthy when every value can answer the question: *did someone mean this, or did it just happen?*

The second thread is **sensor honesty**. This repo's own cycle ledger tracks "numb CI gates," and BB-448-005 shows how easily a well-intentioned fix (don't punish cells the token can't see) recreates numbness one level up (a dead token makes everything invisible, and invisible reads as green). Google's SRE practice calls this the null-alert problem: an alerting pipeline that produces zero alerts is indistinguishable from a broken one unless you alert on the pipeline itself. The fix is always the same — assert on the *denominator*, not just the failures.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "BB-448-001",
      "title": "readLedger silently reverts a persisted lower ceiling back to the 2500 default",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/dune-meter/src/budget-ledger.mjs",
      "description": "readLedger's new override-persistence block fires whenever the `ceiling` parameter differs from the stored value, and the parameter defaults to DEFAULT_CEILING_CREDITS (2500). If an operator once ran with DUNE_BUDGET_CEILING=1000 (persisted to the ledger per AC-10), every subsequent invocation WITHOUT the env var calls readLedger with the default 2500, which is != 1000, so the code warns 'DUNE_BUDGET_CEILING override active: 1000 → 2500' and persists 2500 — silently RAISING the budget ceiling with no env var set at all. The warning text is also false in that case (no override is active). A budget-tightening decision does not survive one plain `dune-meter budget` call.",
      "suggestion": "Only treat `ceiling` as an override when it was explicitly supplied (e.g., pass `ceiling: undefined` by default and have callers pass a value only when DUNE_BUDGET_CEILING is set in the environment). When the param is absent, return the stored ceiling_credits untouched.",
      "confidence": 0.8,
      "faang_parallel": "Kubernetes learned this so painfully that its API conventions mandate pointer types for optional fields — a `*int32` nil means 'unset', while protobuf3 initially removed field presence entirely and had to reintroduce it (the `optional` keyword returned in proto3.15) because 'zero' and 'absent' kept colliding in exactly this way.",
      "metaphor": "It's a thermostat that resets itself to the factory 72°F every time anyone walks past it. You set it to 65 to save money; your roommate checks the temperature — just checks — and the act of looking restores the factory default.",
      "teachable_moment": "A default parameter value is a convenience for the callee, never a statement of intent by the caller. The moment a default can trigger a WRITE, the API must be able to distinguish 'caller said nothing' from 'caller said the default value' — usually via undefined/null sentinel or an explicit options flag.",
      "connection": "Same disease as BB-448-006: a parameter whose presence implies meaning it doesn't carry. Both are 'absent vs default' ambiguities at a boundary."
    },
    {
      "id": "BB-448-002",
      "title": "readLedger's override write bypasses the recordSpend lock — reintroduces the lost-update race",
      "severity": "MEDIUM",
      "category": "concurrency",
      "file": "packages/dune-meter/src/budget-ledger.mjs",
      "description": "recordSpend serializes read-modify-write via the O_EXCL lockfile, but readLedger itself now calls writeLedger (to persist a ceiling override) and is invoked directly by cmdEstimate/cmdRun and `budget` OUTSIDE the lock. A concurrent process doing an unlocked readLedger-with-override can read the ledger before a locked recordSpend commits, then write back `{...parsed, ceiling_credits}` containing the STALE spent_credits — clobbering the spend the lock was built to protect (AC-4). The window requires a ceiling mismatch, but finding BB-448-001 shows that mismatch occurs routinely.",
      "suggestion": "Either acquire the same lock around the override-persist write in readLedger, or persist only the ceiling field via a targeted read-merge-write under the lock, or defer persistence to recordSpend (which already holds the lock when it calls readLedger).",
      "confidence": 0.7,
      "faang_parallel": "This is the classic 'read path that secretly writes' anti-pattern. Postgres's MVCC design goes to great lengths to keep readers from ever taking write locks precisely because a mutating read silently changes the concurrency contract of every caller — once a getter can write, every call site inherits write-side locking obligations it doesn't know about.",
      "metaphor": "The vault has a sign-out sheet for anyone moving money, but the auditor who 'just looks at the balance' also quietly rewrites the whole ledger page from memory on the way out — including entries that changed while they were reading.",
      "teachable_moment": "Function names are concurrency contracts. `readLedger` promises to be safe outside the lock; the moment it can call writeLedger, that promise is broken for every existing caller. If a read must sometimes persist, name it (`readLedgerAndPersistOverride`) or move the write to the one place that already holds the lock.",
      "connection": "Directly amplified by BB-448-001 — the routine spurious ceiling mismatch turns a theoretical race window into an everyday one."
    },
    {
      "id": "BB-448-003",
      "title": "Inter-probe budget gate uses `rem === 0`, missing negative remaining",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/dune-meter/bin/dune-meter.mjs",
      "description": "In cmdRun's split-probe path, after recording COUNT probe spend the guard is `if (rem === 0)`. If countCredits pushes spent_credits past the ceiling, remaining() may be negative (its clamping behavior is not visible in this diff); a strict `=== 0` would then fail to refuse and proceed to probeSample and potentially execution. The analogous pre-probe guards use `remaining(ledger) < 1`, so the codebase already has the safer idiom one screen up.",
      "suggestion": "Use `if (rem < 1)` (or `<= 0`) for the inter-probe check, matching the pre-probe guards. If remaining() already clamps at 0, make that invariant explicit where it's defined so this comparison is provably safe.",
      "confidence": 0.5,
      "faang_parallel": "Guard conditions on resource counters should always be range checks, not equality — the same reasoning behind why kernel refcount hardening (Linux's refcount_t) traps on 'crossed zero', not 'equals zero': the dangerous state is a region, and equality only catches one point on its boundary.",
      "metaphor": "A fuel gauge that only refuses takeoff at exactly zero. If the last taxi burned you to negative-two gallons, the check waves you through.",
      "teachable_moment": "Whether this fires depends on remaining()'s clamping — verify that first. But even if it clamps today, `=== 0` encodes an invariant the comparison site can't see; `< 1` is correct under both behaviors and costs nothing. Prefer guards that don't require reading another function to prove safe."
    },
    {
      "id": "BB-448-004",
      "title": "Stale-lock cleanup has an unlink race that can delete a freshly acquired lock",
      "severity": "MEDIUM",
      "category": "concurrency",
      "file": "packages/dune-meter/src/budget-ledger.mjs",
      "description": "In acquireLock, when a lock is deemed stale (>60s, pid gone), the process unlinks it and retries. Two contenders can both pass the staleness check; contender A unlinks and immediately creates a new lock, then contender B's queued unlinkSync deletes A's LIVE lock, letting a third caller (or B's next attempt) acquire concurrently with A. The `catch { /* lost race */ }` around unlink does not prevent this — the unlink succeeds, it just deletes the wrong lock generation. Window is narrow (requires a stale lock) but it defeats the mutual exclusion exactly in the recovery path.",
      "suggestion": "Before unlinking, re-verify the lock file still contains the stale pid (read-compare-unlink), or better: rename the stale lock to a unique temp name and unlink that, so only one contender can 'win' the removal. Alternatively use the lockfile's inode/mtime captured at staleness detection and skip unlink if it changed.",
      "confidence": 0.6,
      "faang_parallel": "This is the core critique Martin Kleppmann leveled at Redis's Redlock: expiry-based lock recovery without fencing tokens lets a 'recovered' lock coexist with a live one. Filesystem lockfiles have the same gap — which is why the rename-then-unlink trick works: rename is the atomic 'claim' on the removal, so only one contender's unlink can target the stale generation.",
      "metaphor": "Two people find what looks like an abandoned car and both call tow trucks. The first tow clears the spot and a new car parks there; the second truck arrives and tows the new car — it was following instructions that were true ten minutes ago.",
      "teachable_moment": "Deleting a stale lock is itself a critical section. Any recovery path that mutates shared state ('clean up the old lock') needs the same mutual exclusion as the path it recovers — or an atomic primitive (rename) that makes the cleanup single-winner by construction.",
      "connection": "The recovery path is where lock bugs hide because it's the least-exercised path — the same reason BB-448-005's failure branch went unexamined."
    },
    {
      "id": "BB-448-005",
      "title": "All clone failures now downgraded to warnings — a systemic auth/network break turns the compliance job green",
      "severity": "MEDIUM",
      "category": "ci-integrity",
      "file": ".github/workflows/cluster-compliance.yml",
      "description": "The RC-3c intent was to classify 403/404 (token lacks access) as 'inaccessible' rather than non-compliant. But the else branch classifies EVERY other clone failure (gh auth completely broken, DNS failure, rate limiting, disk full) as 'inaccessible' with a ::warning:: as well. If the GITHUB_TOKEN is misconfigured, every cell lands in WARN_CELLS and the audit passes vacuously with zero cells actually audited — the same numb-sensor failure mode this repo's cycle ledger tracks.",
      "suggestion": "After the loop, fail the job (or emit ::error::) when zero cells were successfully cloned/audited, or when the inaccessible count equals the roster size. Optionally keep non-403/404 failures as a distinct status (e.g. 'clone_error') so the artifact distinguishes access policy from infrastructure failure.",
      "confidence": 0.65,
      "faang_parallel": "Google's SRE playbook alerts on the absence of signal, not just bad signal — a monitoring pipeline emitting zero datapoints pages exactly like a threshold breach, because 'no data' and 'all healthy' are indistinguishable otherwise. Test infrastructure has the same rule: a suite that collected zero tests must fail, not pass.",
      "metaphor": "A smoke detector whose dead-battery state is a calm green light. Every room reports 'no smoke detected' — because nothing is detecting.",
      "teachable_moment": "Whenever a check gains a 'skip' or 'warn' path, add an assertion on the denominator: 'we actually audited N > 0 things.' The most dangerous CI regressions aren't red builds — they're green builds that stopped measuring.",
      "connection": "This repo literally ships a cycle-ledger-numb-ci-gates.yaml in this same PR — the pattern being tracked was nearly reintroduced by the fix."
    },
    {
      "id": "BB-448-006",
      "title": "upsertProfile's `setJoinedAt` parameter has no effect",
      "severity": "MEDIUM",
      "category": "api-design",
      "description": "upsertProfile requires `setJoinedAt: boolean` but never uses it: insertValues never sets joinedAt (the comment says 'schema default; setJoinedAt documents intent') and the conflict SET omits it. Callers pass setJoinedAt: true (member.join) and false (member.update), implying different behavior, but both paths produce identical SQL — a member.update arriving before member.join inserts a profile with joinedAt=now() regardless of setJoinedAt: false. AC-14 happens to pass because that implicit timestamp is then preserved, but the parameter is a trap: a future reader will assume it controls whether joinedAt is written.",
      "suggestion": "Either remove the parameter, or make it real (e.g., when setJoinedAt is false, insert joinedAt as NULL/sentinel and let a later member.join backfill it — noting that requires the column to be nullable). At minimum rename to something non-behavioral or document prominently at the call sites that it is documentation-only.",
      "confidence": 0.75,
      "faang_parallel": "Hyrum's Law in its API-design form: every observable aspect of an interface will be depended upon — including a boolean that looks behavioral. Google's API design guidance forbids no-op parameters for exactly this reason; a future engineer will 'fix' code by flipping the flag and be baffled that nothing changes.",
      "metaphor": "A light switch on the wall wired to nothing. Everyone who enters the room flips it with intent; the room's lighting is controlled from somewhere else entirely.",
      "teachable_moment": "Parameters are promises. A boolean that produces identical SQL on both values is a promise the function doesn't keep — and 'documents intent' is better served by a comment at the call site than by a lie in the signature.",
      "connection": "Cousin of BB-448-001: both are cases where a value's presence implies meaning that the implementation doesn't honor."
    },
    {
      "id": "BB-448-007",
      "title": "vi.spyOn on an ESM module namespace may not intercept the handler's direct import binding",
      "severity": "MEDIUM",
      "category": "testing",
      "description": "AC-16a/AC-16b use `vi.spyOn(await import('../data/database.js'), 'upsertCommunity'/'upsertProfile')`. EventNatsConsumer.ts imports these functions directly at module top-level. Whether the spy actually intercepts the handler's call depends on vitest's module transform (it usually works when vitest processes the module, but fails with 'Cannot redefine property' or silently doesn't intercept if the module is externalized or the import binding is inlined). If the spy does not intercept, mockRejectedValueOnce never fires, the handler succeeds, and `.rejects.toThrow(...)` fails — or worse under config drift, the mock leaks. These tests also never restore the spy (no vi.restoreAllMocks), so a once-mock that doesn't get consumed could reject a later test's real call.",
      "suggestion": "Use `vi.mock('../data/database.js', async (importOriginal) => ...)` with a controllable implementation, or inject the DB helpers into the handler factories (they are already factories — dependency injection is one parameter away, as done for discordRest). Add an afterEach vi.restoreAllMocks() regardless.",
      "confidence": 0.45,
      "faang_parallel": "The JS ecosystem's long ESM-mocking saga (Jest's unstable_mockModule, vitest's vi.mock hoisting) exists because ES module bindings are live and read-only per spec — spyOn works only while the test runner's transform keeps namespaces mutable, a configuration accident rather than a guarantee.",
      "metaphor": "You've tapped the phone at the switchboard, but the two offices you care about have a direct line that never touches the switchboard.",
      "teachable_moment": "Verify before refactoring: if these tests currently pass, the transform is intercepting today. But the handlers are already factories — passing the DB helpers in as parameters (as done for discordRest) makes the test independent of transform behavior forever. When DI is one parameter away, module mocking is the expensive option."
    },
    {
      "id": "BB-448-008",
      "title": "DATABASE_URL placeholder in integration-tests job may un-skip tests that then connect to a nonexistent Postgres",
      "severity": "MEDIUM",
      "category": "ci-integrity",
      "description": "The RC-2 comment says the placeholder makes 'MigrationEngine/billing-agent-sovereignty tests not silently skip', while also stating no Postgres service is provisioned. If those tests gate on DATABASE_URL presence and then actually open a connection (rather than only satisfying Zod config validation), they will hit connection-refused on localhost:5432 and fail — or hang until timeout. The two claims in the comment (tests now run + no server exists + 'sietch integration tests use SQLite') can't all be true unless those specific suites never dial the URL.",
      "suggestion": "Verify in CI that the MigrationEngine/billing-agent-sovereignty suites genuinely pass with the placeholder and no service; if they open real connections, either provision the postgres service in this job too or scope the placeholder to config-validation-only paths.",
      "confidence": 0.4,
      "faang_parallel": "Environment-variable-gated test suites are a known hazard class — Meta's test infrastructure moved to explicit capability declarations ('this suite requires: postgres') precisely because env-presence gating conflates 'configured' with 'available'.",
      "metaphor": "Handing someone a hotel key card for a hotel that hasn't been built. It gets them past the 'do you have a key?' check and into an empty lot.",
      "teachable_moment": "This needs empirical verification, not speculation — one green run of this exact job on this branch settles it. The deeper lesson: a comment making three claims about runtime behavior should cite the run that proved them."
    },
    {
      "id": "BB-448-009",
      "title": "memberUpdate logs 'roles' in fieldsUpdated but persists nothing for roles",
      "severity": "LOW",
      "category": "observability",
      "description": "makeMemberUpdateHandler pushes 'roles' into fieldsUpdated when roles are present in the payload, but no roles data is written to the profile (no metadataPatch key, no tier mapping). The structured log then claims fieldsUpdated: ['roles'] for a write that changed nothing role-related — misleading for anyone debugging role-sync issues later.",
      "suggestion": "Either drop 'roles' from fieldsUpdated until role persistence exists, or rename the field (e.g. fieldsSeen) — the log should not assert an update that didn't happen. A loa:shortcut marker noting the role-sync gap would fit the codebase convention.",
      "confidence": 0.8,
      "faang_parallel": "Dapper (Google's tracing paper) established that observability artifacts are load-bearing interfaces: engineers debug from them under time pressure and trust them implicitly. A span or log field asserting work that didn't happen is worse than no field — it actively misroutes an incident investigation.",
      "metaphor": "A receipt listing an item that was never put in the bag. At home, you don't suspect the receipt — you tear the house apart looking for the item.",
      "teachable_moment": "Structured logs are an API whose consumers are your 3 a.m. future selves. Field names should be verifiably true from the code path that emits them: 'fieldsUpdated' must mean updated, not observed. Naming honesty in logs is cheap now and priceless during an incident."
    },
    {
      "id": "BB-448-010",
      "title": "Terminal-race detection couples to an exact error string",
      "severity": "LOW",
      "description": "The advance-rejection handling branches on `advResult.error === 'order already terminal'`, string-matching a message produced in community-onboarding-orchestrator.ts. Any rewording of that message silently converts the benign terminal race back into an operator-paging escalation — the exact bug the branch fixes. Also note this branch `return`s from processOrder entirely rather than `continue`, skipping remaining advanceable ingredients in the same tick; that is correct only because a terminal order has nothing left to advance, which is worth a comment.",
      "suggestion": "Return a discriminated result from advanceIngredient (e.g. `{ ok: false, code: 'TERMINAL' | 'GATE_REJECTED', error }`) and branch on the code. Keep the message free-form for humans.",
      "confidence": 0.85,
      "faang_parallel": "gRPC's status-code system exists because clients at Google string-matched error messages and every message reword became a production incident — Hyrum's Law applied to error text. The rule that emerged: codes are for machines, messages are for humans, and never shall a machine branch on a message.",
      "metaphor": "Recognizing your colleague only by their exact haircut. The day they visit a barber, security escorts them out of the building.",
      "teachable_moment": "The moment code branches on an error, that error has become a protocol — and protocols need stable identifiers. A discriminated union (`code` + free-form `message`) costs one field and permanently decouples control flow from prose. Also: the `return`-vs-`continue` subtlety here is load-bearing and invisible — exactly what a one-line comment is for.",
      "connection": "Same 'implicit contract made of prose' shape as BB-448-009 — text intended for humans getting consumed by machines."
    },
    {
      "id": "BB-448-011",
      "title": "EventNatsConsumer test opens an unused postgres pool and never closes the singleton DB pool",
      "severity": "LOW",
      "category": "testing",
      "description": "beforeAll creates `sql = postgres(DATABASE_URL, { max: 3 })` which is never used for queries — only closed in afterAll. Meanwhile initDatabase creates the pool the tests actually exercise, and that singleton's connections are never ended, so vitest may hold the process open until postgres.js idle_timeout fires (or require forced exit) in the new worker-db-tests CI job.",
      "suggestion": "Drop the unused `sql` client, and expose/call a closeDatabase() (or reach into the singleton's underlying client) in afterAll so the run exits cleanly and deterministically.",
      "confidence": 0.55,
      "metaphor": "Locking up the office each night but leaving a space heater running in a room nobody entered — the building never quite powers down, and one day the janitor (CI timeout) has to flip the breaker.",
      "teachable_moment": "Singletons that open resources need a symmetric close for test harnesses — otherwise every test suite that touches them inherits a nondeterministic shutdown. The unused pool suggests a copy-paste from a suite that did query directly; deleting it is free clarity."
    },
    {
      "id": "BB-448-012",
      "title": "Root integration config reaches into themes/sietch/node_modules for zod/ioredis/pg",
      "severity": "LOW",
      "category": "build-fragility",
      "file": "vitest.agent-ci-integration.config.ts",
      "description": "Aliasing zod, ioredis, and pg to themes/sietch/node_modules couples the repo-root integration suite to sietch's install layout and dependency versions. A sietch dependency bump (or pnpm store layout change — a known hazard in this repo) changes what the root tests import, and a sietch install must happen before agent-ci integration tests can even collect. It works, and the RC-3 scoping fix itself is sound, but the alias is a time bomb relative to declaring these as root devDependencies.",
      "suggestion": "Add the three packages as root-level devDependencies (or a shared test package) and remove the aliases; keep the include/exclude scoping.",
      "confidence": 0.6,
      "faang_parallel": "Reaching across package boundaries into another workspace's node_modules is the monorepo equivalent of what Google's Bazel forbids as an 'undeclared dependency' — builds that work by filesystem coincidence break the day the layout changes, and pnpm's content-addressed store changes layout more often than most.",
      "metaphor": "Borrowing your neighbor's tools through a hole in the fence. Works great until they reorganize the garage.",
      "teachable_moment": "Every dependency a test actually imports should be declared where the test lives. Aliases into a sibling's node_modules convert a version bump in one package into silent behavior change in another — the exact failure class this repo already documented with pnpm file: deps."
    },
    {
      "id": "BB-448-013",
      "title": "tests/setup-unit.ts may be dead on arrival — nothing in the diff wires it as setupFiles",
      "severity": "LOW",
      "category": "testing",
      "file": "tests/setup-unit.ts",
      "description": "The new file exports TEST_ENV_DEFAULTS and applyTestEnvDefaults, and a ci.yml comment references it as the unit project's setupFiles — but no vitest config change in this PR registers it, and applyTestEnvDefaults has no visible caller. If no existing config already points at this exact path, the canonical placeholder convention it documents is not actually enforced anywhere.",
      "suggestion": "Confirm an existing vitest unit project config references tests/setup-unit.ts (and calls applyTestEnvDefaults); if not, wire it in or the file is documentation masquerading as code.",
      "confidence": 0.5,
      "metaphor": "A fire-evacuation plan laminated and posted in a room the building's floor plan doesn't include.",
      "teachable_moment": "Verify wiring before assuming breakage — a pre-existing config may already reference this path. But the general rule stands: a convention file earns its keep only when something executes it. One `git grep 'setup-unit'` settles this in ten seconds; the check belongs in the PR."
    },
    {
      "id": "BB-448-014",
      "title": "T-3b assertion accepts exit 0 OR 3 — weakens the test's discriminating power",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/dune-meter/test/dune-meter.test.mjs",
      "description": "The AC-5 test asserts `r.exit === 0 || r.exit === 3` before checking the ledger. With probe_credits=42 against a fresh 2500 ceiling the estimate should deterministically be OK (exit 0); accepting 3 as well means a regression that wrongly refuses every estimate would still pass this test as long as the spend landed. The ledger assertion is the load-bearing one, but the exit tolerance quietly halves the test's value.",
      "suggestion": "Compute the expected verdict for the mock's rows/cols and assert the single correct exit code.",
      "confidence": 0.7,
      "faang_parallel": "Disjunctive assertions are a recognized test smell — mutation-testing tools (PIT, Stryker) flag them because a mutant that flips the branch survives. If the input is deterministic, the assertion should be too; an OR in an assertion is usually a question the author didn't answer.",
      "metaphor": "A quality inspector who accepts both 'pass' and 'fail' stamps as long as the paperwork was filed. The paperwork check is real; the stamp check is theater.",
      "teachable_moment": "When you find yourself writing `expect(a || b)`, stop and compute which one is true for this input. If you can't, the test's fixture is underspecified — fix the fixture, not the assertion."
    },
    {
      "id": "BB-448-015",
      "title": "Clone exit-code capture pattern is finally correct under errexit, with the failure history documented",
      "severity": "PRAISE",
      "category": "ci-integrity",
      "file": ".github/workflows/cluster-compliance.yml",
      "description": "`clone_exit=0; clone_output=$(cmd) || clone_exit=$?` is the correct idiom under `bash -e`: it survives errexit AND captures the real exit code, and the comment names both prior broken attempts (bare assignment dying under -e; `if ! var=$(cmd)` capturing the negation's status). Encoding the failure history at the site is exactly what prevents attempt #4.",
      "suggestion": "None — consider extracting this idiom into a shared snippet/doc since the repo has now paid for it three times.",
      "confidence": 0.9,
      "faang_parallel": "This is the 'comments carry the postmortem' practice — Google's style guides encourage documenting why the obvious alternative is wrong at exactly the site where someone will 'simplify' it back. Shell errexit semantics are among the most re-broken things in CI history.",
      "teachable_moment": "When a bug has been fixed three times, the fix isn't code — it's the comment that names the two wrong versions. The commit message even says 'third time's the charm'; the site-local history is what makes it the last time."
    },
    {
      "id": "BB-448-016",
      "title": "xmax=0 insert detection closes the double-welcome TOCTOU atomically",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "Using `returning({ ..., inserted: sql`(xmax = 0)` })` classifies insert-vs-update at the write itself, so exactly one of two concurrent guild.join deliveries can classify as 'created' and send the welcome — and the PR includes a real concurrency test (Promise.all of two upserts) pinning it. The pre-SELECT is correctly demoted to only disambiguating reactivated-vs-noop, where a race is benign.",
      "suggestion": "None.",
      "confidence": 0.85,
      "faang_parallel": "Pushing the decision into the atomic operation rather than checking-then-acting is the same move as compare-and-swap replacing lock-check-act in lock-free algorithms — the database's own MVCC machinery (xmax) becomes the arbiter, so there is no window by construction.",
      "metaphor": "Instead of two greeters checking a clipboard to see who welcomes the new member (and both deciding 'me'), the door itself stamps exactly one hand on the way in.",
      "teachable_moment": "The strongest concurrency fixes remove the race window rather than shrinking it — and the Promise.all test proves the property instead of asserting the implementation. This pairing (atomic classification + concurrency test) is the pattern to reuse."
    },
    {
      "id": "BB-448-017",
      "title": "Honest --cap documentation is enforced by tests, not just edited",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "packages/dune-meter/veve.json",
      "description": "The 'structurally impossible' claim is removed consistently across veve.json, README, and --help, the README now states precisely what --cap is (post-spend audit signal) and where the real pre-spend abort lives (Dune dashboard Query Cost Cap) — and T-1 pins the veve summary plus the run-output assertion so the overclaim cannot silently return. Documentation corrections that come with regression tests are rare and worth keeping as a pattern.",
      "suggestion": "None.",
      "confidence": 0.9,
      "faang_parallel": "This is documentation-as-tested-contract — the same philosophy as Rust's doctests and Python's doctest module: prose that makes a claim about behavior should be executable, so drift between claim and reality breaks the build instead of breaking trust.",
      "teachable_moment": "Overclaims in safety documentation are a security issue wearing a documentation costume: an operator who believes '--cap makes overspend structurally impossible' won't set the real guard (the Dune dashboard cap). Correcting the claim AND pinning it with a test treats honesty as an invariant, not an edit."
    },
    {
      "id": "BB-448-018",
      "title": "Split probe with inter-probe budget gate and probe-spend recording before execution",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/dune-meter/bin/dune-meter.mjs",
      "description": "Splitting the raw-SQL probe into probeCount + probeSample with a budget check between them, recording each probe's credits into the ledger BEFORE the main execution, and keeping the probe-budget guard non-bypassable by --force closes the previously-unmetered probe spend path. The ledger-snapshot-inside-executeSql test technique proves the ordering rather than just the totals.",
      "suggestion": "None beyond BB-448-003 (the `=== 0` comparison).",
      "confidence": 0.85,
      "faang_parallel": "Metering the meter's own overhead is the mark of a mature accounting system — AWS bills for the API calls that check your bill. Any budget system whose measurement probes are themselves free spend has an unmetered side channel; closing it and making the guard --force-proof is the right hierarchy of authority.",
      "teachable_moment": "The snapshot-inside-the-mock test technique deserves imitation: asserting ledger state at the moment executeSql is invoked proves the *ordering* invariant ('probe spend recorded before execution'), which a post-hoc total can never distinguish from 'recorded after'."
    }
  ]
}
```


## Callouts

Three pieces of this PR deserve explicit recognition beyond the PRAISE findings above:

- **[Quality] The failure history is written where the next mistake would happen.** The errexit idiom comment (BB-448-015) naming both prior broken attempts is institutional memory placed at the exact coordinates where someone would reintroduce the bug. This is what "third time's the charm" should always leave behind.
- **[Test Coverage] Ordering proofs, not total proofs.** The ledger-snapshot-inside-executeSql technique (BB-448-018) and the Promise.all double-upsert test (BB-448-016) both verify *properties* — ordering and single-winner classification — rather than end-state totals. That's the difference between a test that catches the regression and one that merely coexists with it.
- **[Operational] Honesty as a tested invariant.** Removing the "structurally impossible" overclaim across three surfaces and pinning the correction with T-1 (BB-448-017) treats documentation accuracy the way this codebase treats budget accuracy: as something a regression can break, and therefore something a test must guard. More PRs should correct docs this way.

The through-line for the fixes: the budget ledger's remaining flaws (BB-448-001, -002, -004) all live in paths added *around* the lock this PR built. Resolve the absent-vs-default ambiguity in `readLedger`, move the ceiling persist under the lock, and the ledger's story becomes as trustworthy as its tests already are.

zkSoju:
## BB REQUEST_CHANGES triage

| Finding | Disposition |
|---|---|
| **HIGH: readLedger silently raises a persisted lower ceiling** | ✅ Fixed `9b0f4a12` — ceiling param is explicit-only (undefined when env unset); regression test pins persisted-1000-survives-env-less-read; explicit lowering still works. Suite 47/47 |
| MEDIUM: override write bypasses recordSpend lock | Narrowed by the HIGH fix (write now fires only on an explicit, differing override — an operator act, not every run). Full lock-scoped override write logged to #303 residue |
| MEDIUM: inter-probe gate `rem === 0` | ✅ `rem <= 0` (belt-and-suspenders; `remaining()` clamps at 0) |
| MEDIUM: stale-lock unlink race | Logged to #303 residue — needs an atomic rename-then-verify pattern, out of tonight's scope |
| **DISS-001 (adversarial voice, BLOCKING): #445 errexit fix captured the negation's status** | ✅ Fixed `bf7ec1ee` with inline proof (false→1, success→0) — hotfixes merged main behavior |

Merging under the standing BB-before-merge grant.

PR #468, opened 2026-07-14
feat(governance): project user intent and construct expertise

zkSoju:
## Summary

- add a typed flow-moment contract that keeps product thesis, exposure, and component maturity independent
- add a canonical system-component contract and reusable validation workflow for API/BFF and front-facing product consumers
- ratify the first `loa-freeside` product-governance territory with The Arcade and Beacon at an observe-only ceiling
- add a construct-aware operator receipt joining user-flow responsibility, region outcomes, producer-declared expertise, provenance, and effective authority

## Why

Teams closest to rapidly changing reality can make the most informed decisions when the system surfaces signal without flattening every claim into “truth.” This gives agents continuity across context windows while preserving the force chain:

- product thesis remains a hypothesis until evidence arrives
- construct orientation is prose and cannot grant authority
- construct mechanics come from producer-owned contracts
- territory is only a ceiling; unknown earned authority collapses to `observe`

The result is a derived operator surface, not another decision or learning ledger.

## Contract shape

| Plane | Source | Guarantee |
|---|---|---|
| User intent | flow moments + system components | operator, question, responsibility, evidence, boundaries, handoff |
| Orientation | Constructs info | prose with `authoritative: false` |
| Mechanics | Constructs info + capabilities | declared skills/commands/runtime hints with provenance and no authority effect |
| Authority | territory + earned evidence | fail-closed effective tier; explanatory prose is never parsed as state |

## Stack and adoption

- depends on 0xHoneyJar/loa-constructs#276, stacked on its CLI producer PR #275
- this PR remains the canonical dependency for follow-up consumer PRs in the active Freeside APIs, `freeside-dashboard`, and `freeside-characters`
- consumers pin the reusable governance workflow by immutable commit rather than copying schemas

## Verification

- `node --test tools/flow-moment.test.mjs tools/system-component.test.mjs tools/construct-operator.test.mjs` — 51 passed
- validated 1 canonical flow-moment record and 1 canonical system-component manifest
- rendered flow, system, and construct operator receipts
- targeted Oxlint passed
- workflow YAML parsed successfully
- live join against the local Constructs producer returned The Arcade + Beacon with declared mechanics, structured `ratification_status: unchecked`, and `unknown -> observe` authority
- producer suite: 176 tests and 7/7 deterministic vectors

## Reviewer focus

- weakest link: territory stationing is intentionally unratified on a feature branch, so the live operator receipt must remain `partial` until the producer can prove committed default-branch state
- confirm ratification prose cannot mechanically promote state
- confirm Hivemind remains actionless and product actions stay in the flow contract
- confirm missing mechanics, producer drift, and impossible authority combinations fail closed

### DIFF
--- a/.github/workflows/flow-moment-governance.yml
+++ b/.github/workflows/flow-moment-governance.yml
@@ -53,5 +53,5 @@ jobs:
         run: node tools/flow-moment.mjs render product/flow-moments/audit-community-composition.flow.json > /tmp/audit-flow-receipt.md
       - name: Prove the system receipt renders from source
         run: node tools/system-component.mjs render product/system-components/loa-freeside.system.json > /tmp/loa-freeside-system-receipt.md
-      - name: Prove the construct operator receipt renders from deterministic seams
+      - name: Smoke render construct receipt (partial is visible, not accepted)
         run: node tools/construct-operator.mjs render --snapshot tools/fixtures/construct-operator.snapshot.json > /tmp/loa-freeside-construct-receipt.md
--- a/.github/workflows/flow-moment-governance.yml
+++ b/.github/workflows/flow-moment-governance.yml
@@ -19,6 +19,9 @@ on:
       - "spec/product/flow-moment-plan.md"
       - "spec/product/flow-moment-implementation-report.md"
       - "product/system-components/**"
+      - "package.json"
+      - "pnpm-lock.yaml"
+      - "pnpm-workspace.yaml"
       - ".github/workflows/flow-moment-governance.yml"
       - ".github/workflows/reusable-flow-moment-governance.yml"
   workflow_dispatch:
--- a/.github/workflows/flow-moment-governance.yml
+++ b/.github/workflows/flow-moment-governance.yml
@@ -1,0 +1,47 @@
+name: Flow Moment Governance
+
+on:
+  pull_request:
+    paths:
+      - "product/flow-moments/**"
+      - "spec/product/flow-moment.schema.json"
+      - "spec/product/system-component.schema.json"
+      - "tools/flow-moment.mjs"
+      - "tools/flow-moment.test.mjs"
+      - "tools/system-component.mjs"
+      - "tools/system-component.test.mjs"
+      - "tools/hivemind/hivemind-labels.v1.0.json"
+      - "docs/product/flow-moment-contract.md"
+      - "spec/product/flow-moment-plan.md"
+      - "spec/product/flow-moment-implementation-report.md"
+      - "product/system-components/**"
+      - ".github/workflows/flow-moment-governance.yml"
+      - ".github/workflows/reusable-flow-moment-governance.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  continuity-contract:
+    name: Validate user-flow moments
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: pnpm/action-setup@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          cache: pnpm
+      - name: Install validator dependencies
+        run: pnpm install --frozen-lockfile --ignore-scripts
+      - name: Prove promotion gates fail closed
+        run: node --test tools/flow-moment.test.mjs tools/system-component.test.mjs
+      - name: Validate every flow-moment record
+        run: node tools/flow-moment.mjs validate product/flow-moments
+      - name: Validate every system-component manifest
+        run: node tools/system-component.mjs validate product/system-components
+      - name: Prove the continuity receipt renders from source
+        run: node tools/flow-moment.mjs render product/flow-moments/audit-community-composition.flow.json > /tmp/audit-flow-receipt.md
+      - name: Prove the system receipt renders from source
+        run: node tools/system-component.mjs render product/system-components/loa-freeside.system.json > /tmp/loa-freeside-system-receipt.md
--- a/.github/workflows/flow-moment-governance.yml
+++ b/.github/workflows/flow-moment-governance.yml
@@ -8,6 +8,7 @@ on:
       - "spec/product/system-component.schema.json"
       - "tools/flow-moment.mjs"
       - "tools/flow-moment.test.mjs"
+      - "tools/markdown-text.mjs"
       - "tools/system-component.mjs"
       - "tools/system-component.test.mjs"
       - "tools/construct-operator.mjs"
@@ -34,9 +35,9 @@ jobs:
     name: Validate user-flow moments
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: "22"
           cache: pnpm
--- a/.github/workflows/flow-moment-governance.yml
+++ b/.github/workflows/flow-moment-governance.yml
@@ -10,6 +10,10 @@ on:
       - "tools/flow-moment.test.mjs"
       - "tools/system-component.mjs"
       - "tools/system-component.test.mjs"
+      - "tools/construct-operator.mjs"
+      - "tools/construct-operator.test.mjs"
+      - "tools/fixtures/construct-operator.snapshot.json"
+      - "grimoires/territory.yaml"
       - "tools/hivemind/hivemind-labels.v1.0.json"
       - "docs/product/flow-moment-contract.md"
       - "spec/product/flow-moment-plan.md"
@@ -36,7 +40,7 @@ jobs:
       - name: Install validator dependencies
         run: pnpm install --frozen-lockfile --ignore-scripts
       - name: Prove promotion gates fail closed
-        run: node --test tools/flow-moment.test.mjs tools/system-component.test.mjs
+        run: node --test tools/flow-moment.test.mjs tools/system-component.test.mjs tools/construct-operator.test.mjs
       - name: Validate every flow-moment record
         run: node tools/flow-moment.mjs validate product/flow-moments
       - name: Validate every system-component manifest
@@ -45,3 +49,5 @@ jobs:
         run: node tools/flow-moment.mjs render product/flow-moments/audit-community-composition.flow.json > /tmp/audit-flow-receipt.md
       - name: Prove the system receipt renders from source
         run: node tools/system-component.mjs render product/system-components/loa-freeside.system.json > /tmp/loa-freeside-system-receipt.md
+      - name: Prove the construct operator receipt renders from deterministic seams
+        run: node tools/construct-operator.mjs render --snapshot tools/fixtures/construct-operator.snapshot.json > /tmp/loa-freeside-construct-receipt.md
--- a/.github/workflows/reusable-flow-moment-governance.yml
+++ b/.github/workflows/reusable-flow-moment-governance.yml
@@ -43,6 +43,8 @@ jobs:
       - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: "22"
+          cache: pnpm
+          cache-dependency-path: .flow-governance/pnpm-lock.yaml
       - name: Install governance validator dependencies
         working-directory: .flow-governance
         run: pnpm install --frozen-lockfile --ignore-scripts --ignore-workspace
--- a/.github/workflows/reusable-flow-moment-governance.yml
+++ b/.github/workflows/reusable-flow-moment-governance.yml
@@ -23,6 +23,14 @@ jobs:
     steps:
       - name: Checkout consumer repository
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - name: Require an immutable governance commit
+        env:
+          GOVERNANCE_REF: ${{ inputs.governance-ref }}
+        run: |
+          if [[ ! "$GOVERNANCE_REF" =~ ^[0-9a-f]{40}$ ]]; then
+            echo "governance-ref must be a full lowercase 40-character commit SHA" >&2
+            exit 1
+          fi
       - name: Checkout pinned governance contract
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
--- a/.github/workflows/reusable-flow-moment-governance.yml
+++ b/.github/workflows/reusable-flow-moment-governance.yml
@@ -38,7 +38,11 @@ jobs:
       - name: Install governance validator dependencies
         working-directory: .flow-governance
         run: pnpm install --frozen-lockfile --ignore-scripts --ignore-workspace
-      - name: Validate system component intent
+      - name: Validate portable system component intent
+        if: ${{ ! inputs.validate-flow-moments }}
+        run: node .flow-governance/tools/system-component.mjs validate product/system-components --portable
+      - name: Validate system component intent with canonical flow records
+        if: ${{ inputs.validate-flow-moments }}
         run: node .flow-governance/tools/system-component.mjs validate product/system-components
       - name: Validate repo-local user-flow moments
         if: ${{ inputs.validate-flow-moments }}
--- a/.github/workflows/reusable-flow-moment-governance.yml
+++ b/.github/workflows/reusable-flow-moment-governance.yml
@@ -1,0 +1,45 @@
+name: Reusable Flow Moment Governance
+
+on:
+  workflow_call:
+    inputs:
+      governance-ref:
+        description: Immutable loa-freeside commit containing the schema and validators
+        required: true
+        type: string
+      validate-flow-moments:
+        description: Validate repo-local product/flow-moments records in addition to the system manifest
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+
+jobs:
+  continuity-contract:
+    name: Validate product and system intent
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout consumer repository
+        uses: actions/checkout@v4
+      - name: Checkout pinned governance contract
+        uses: actions/checkout@v4
+        with:
+          repository: 0xHoneyJar/loa-freeside
+          ref: ${{ inputs.governance-ref }}
+          path: .flow-governance
+      - uses: pnpm/action-setup@v4
+        with:
+          package_json_file: .flow-governance/package.json
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+      - name: Install governance validator dependencies
+        working-directory: .flow-governance
+        run: pnpm install --frozen-lockfile --ignore-scripts
+      - name: Validate system component intent
+        run: node .flow-governance/tools/system-component.mjs validate product/system-components
+      - name: Validate repo-local user-flow moments
+        if: ${{ inputs.validate-flow-moments }}
+        run: node .flow-governance/tools/flow-moment.mjs validate product/flow-moments
--- a/.github/workflows/reusable-flow-moment-governance.yml
+++ b/.github/workflows/reusable-flow-moment-governance.yml
@@ -22,17 +22,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout consumer repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - name: Checkout pinned governance contract
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           repository: 0xHoneyJar/loa-freeside
           ref: ${{ inputs.governance-ref }}
           path: .flow-governance
-      - uses: pnpm/action-setup@v4
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
         with:
           package_json_file: .flow-governance/package.json
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: "22"
       - name: Install governance validator dependencies
--- a/.github/workflows/reusable-flow-moment-governance.yml
+++ b/.github/workflows/reusable-flow-moment-governance.yml
@@ -37,7 +37,7 @@ jobs:
           node-version: "22"
       - name: Install governance validator dependencies
         working-directory: .flow-governance
-        run: pnpm install --frozen-lockfile --ignore-scripts
+        run: pnpm install --frozen-lockfile --ignore-scripts --ignore-workspace
       - name: Validate system component intent
         run: node .flow-governance/tools/system-component.mjs validate product/system-components
       - name: Validate repo-local user-flow moments
--- a/docs/product/flow-moment-contract.md
+++ b/docs/product/flow-moment-contract.md
@@ -35,7 +35,7 @@ A Gold component can appear inside a falsified flow. A Silver prototype can prod
 - Exposed research requires a feature-flag reference and an evidence contract.
 - Default-on research requires a learning reference and an operator decision.
 - Gold requires Silver metadata, a taste owner, 14 production days, active use, no regressions, and evidence references.
-- Gold age is evaluated against the record's explicit `as_of` date. Advancing the maturity clock is therefore a reviewable record change, not an ambient CI clock.
+- Gold age is evaluated against the record's explicit `as_of` date. Advancing the maturity clock is therefore a reviewable record change, not an ambient CI clock; validation also refuses an `as_of` later than the validator's observed UTC date.
 - CI validates claims and evidence pointers. It cannot decide whether the taste is good or the hypothesis is true.
 
 ## Evidence privacy
@@ -61,6 +61,9 @@ Product telemetry, website feedback, tickets, Discord, interviews, partner obser
 - a thin workflow caller pinned to an immutable `loa-freeside` governance commit.
 
 This makes drift reviewable without vendoring the contract into every API or product repository.
+The `governance-ref` input must be a full lowercase 40-character commit SHA; branches
+and tags are rejected before the governance checkout so a caller's verifier cannot move
+without a reviewed caller change.
 By default, the reusable workflow validates a portable system manifest: flow IDs and
 `flow:<id>` references must agree, but the consumer does not need to vendor Freeside's
 canonical `.flow.json` records. Set `validate-flow-moments: true` only when the consumer
--- a/docs/product/flow-moment-contract.md
+++ b/docs/product/flow-moment-contract.md
@@ -18,6 +18,8 @@ node tools/system-component.mjs render product/system-components/loa-freeside.sy
 
 The system manifest captures the engineering operator, component object, consumer question, stable responsibility, trust surface, actions, ownership boundary, and handoff. It links to a canonical `FM-*` record when the relationship is real. An unmapped component must say why instead of inventing a user-truth relationship.
 
+JSON Schema is the portable shape contract; it rejects structurally invalid records and exact duplicate array items. The repository CLIs are the canonical acceptance layer because identity invariants such as duplicate signal IDs and duplicate `flow_moment_id` values cannot be expressed by portable JSON Schema alone. CI and cross-repository consumers must run the appropriate CLI rather than treating schema-only validation as semantic acceptance.
+
 ## Three independent axes
 
 | Axis | Answers | States |
@@ -85,7 +87,7 @@ Mechanical gates read the atlas's structured `ratification_status`; the neighbor
 `loa-freeside` ratifies the first bounded territory at `grimoires/territory.yaml`. The joined operator receipt derives from the existing system component, the live Constructs producer, and that region-owned declaration:
 
 ```bash
-# Deterministic fixture seam used by CI
+# Deterministic smoke-render seam used by CI; partial remains visible by design
 node tools/construct-operator.mjs render \
   --snapshot tools/fixtures/construct-operator.snapshot.json
 
@@ -96,4 +98,4 @@ node tools/construct-operator.mjs render \
 ```
 
 Live mode invokes only `capabilities`, `atlas`, and `info --rung local`. Mutation verbs are displayed in a separate set but are never executed. Until the territory is committed on the region's default branch and the producer can prove stationing, the overall receipt stays `partial` rather than presenting a working-tree declaration as ratified expertise.
-Release gates may add `--require-ok` to reject `partial`; ordinary operator inspection deliberately renders partial receipts so missing or unratified state remains visible.
+The checked-in fixture step is a smoke-render check, not a release-acceptance gate: its purpose is to prove that incomplete expertise remains inspectable. Release gates add `--require-ok` to reject `partial`; ordinary operator inspection deliberately renders partial receipts so missing or unratified state remains visible.
--- a/docs/product/flow-moment-contract.md
+++ b/docs/product/flow-moment-contract.md
@@ -19,6 +19,7 @@ node tools/system-component.mjs render product/system-components/loa-freeside.sy
 The system manifest captures the engineering operator, component object, consumer question, stable responsibility, trust surface, actions, ownership boundary, and handoff. It links to a canonical `FM-*` record when the relationship is real. An unmapped component must say why instead of inventing a user-truth relationship.
 
 JSON Schema is the portable shape contract; it rejects structurally invalid records and exact duplicate array items. The repository CLIs are the canonical acceptance layer because identity invariants such as duplicate signal IDs and duplicate `flow_moment_id` values cannot be expressed by portable JSON Schema alone. CI and cross-repository consumers must run the appropriate CLI rather than treating schema-only validation as semantic acceptance.
+Flow components make referential integrity explicit: `resolution: local` requires a `component:<component_id>` reference that resolves exactly once in `product/system-components`; `resolution: external` requires a non-component provenance reference and is not presented as locally verified.
 
 ## Three independent axes
 
--- a/docs/product/flow-moment-contract.md
+++ b/docs/product/flow-moment-contract.md
@@ -60,6 +60,11 @@ Product telemetry, website feedback, tickets, Discord, interviews, partner obser
 - a thin workflow caller pinned to an immutable `loa-freeside` governance commit.
 
 This makes drift reviewable without vendoring the contract into every API or product repository.
+By default, the reusable workflow validates a portable system manifest: flow IDs and
+`flow:<id>` references must agree, but the consumer does not need to vendor Freeside's
+canonical `.flow.json` records. Set `validate-flow-moments: true` only when the consumer
+owns repo-local flow records; that enables both full flow validation and exact canonical
+reference resolution.
 
 ## Construct expertise on the operator surface
 
--- a/docs/product/flow-moment-contract.md
+++ b/docs/product/flow-moment-contract.md
@@ -35,6 +35,7 @@ A Gold component can appear inside a falsified flow. A Silver prototype can prod
 - Exposed research requires a feature-flag reference and an evidence contract.
 - Default-on research requires a learning reference and an operator decision.
 - Gold requires Silver metadata, a taste owner, 14 production days, active use, no regressions, and evidence references.
+- Gold age is evaluated against the record's explicit `as_of` date. Advancing the maturity clock is therefore a reviewable record change, not an ambient CI clock.
 - CI validates claims and evidence pointers. It cannot decide whether the taste is good or the hypothesis is true.
 
 ## Evidence privacy
@@ -92,3 +93,4 @@ node tools/construct-operator.mjs render \
 ```
 
 Live mode invokes only `capabilities`, `atlas`, and `info --rung local`. Mutation verbs are displayed in a separate set but are never executed. Until the territory is committed on the region's default branch and the producer can prove stationing, the overall receipt stays `partial` rather than presenting a working-tree declaration as ratified expertise.
+Release gates may add `--require-ok` to reject `partial`; ordinary operator inspection deliberately renders partial receipts so missing or unratified state remains visible.
--- a/docs/product/flow-moment-contract.md
+++ b/docs/product/flow-moment-contract.md
@@ -60,3 +60,30 @@ Product telemetry, website feedback, tickets, Discord, interviews, partner obser
 - a thin workflow caller pinned to an immutable `loa-freeside` governance commit.
 
 This makes drift reviewable without vendoring the contract into every API or product repository.
+
+## Construct expertise on the operator surface
+
+Constructs join the continuity surface through a region-owned territory, not through free-text tags on every system component. The three planes stay mechanically separate:
+
+| Plane | Source | What it may claim |
+|---|---|---|
+| Orientation | `constructs info <slug> --json --rung local` → `orientation` | Prose that helps an operator frame the problem. Always `authoritative: false`. |
+| Mechanics | The same info payload → `mechanics`, plus `constructs capabilities --json` | Declared skills, commands, runtime requirements, verb mutation class, determinism, and provenance. It grants no authority. |
+| Authority | `grimoires/territory.yaml` projected through `constructs atlas --json`, then the graduated-trust ledger | The region's ceiling intersected with earned evidence. Unknown earned state collapses to `observe`. |
+
+Mechanical gates read the atlas's structured `ratification_status`; the neighboring `ratification` sentence is operator orientation and is never parsed as state.
+
+`loa-freeside` ratifies the first bounded territory at `grimoires/territory.yaml`. The joined operator receipt derives from the existing system component, the live Constructs producer, and that region-owned declaration:
+
+```bash
+# Deterministic fixture seam used by CI
+node tools/construct-operator.mjs render \
+  --snapshot tools/fixtures/construct-operator.snapshot.json
+
+# Live operator-local projection (producer path may also be set as CONSTRUCTS_CLI)
+CONSTRUCTS_DIR="$HOME/.loa/constructs/packs" \
+node tools/construct-operator.mjs render \
+  --constructs-cli /path/to/loa-constructs/packages/constructs-cli/bin/constructs.mjs
+```
+
+Live mode invokes only `capabilities`, `atlas`, and `info --rung local`. Mutation verbs are displayed in a separate set but are never executed. Until the territory is committed on the region's default branch and the producer can prove stationing, the overall receipt stays `partial` rather than presenting a working-tree declaration as ratified expertise.
--- a/docs/product/flow-moment-contract.md
+++ b/docs/product/flow-moment-contract.md
@@ -1,0 +1,62 @@
+# User-flow moment contract
+
+The user-flow moment is Freeside's stable join between product intent and implementation. It answers what progress a user is trying to make at one point in the journey, what we hypothesize will help, how the experience is exposed, and what evidence would support or falsify the hypothesis.
+
+It is deliberately not another decision or learning ledger. Each record points to existing issues, ADRs, Learning Memos, feature flags, components, and evidence. Agents render a disposable continuity receipt from those sources:
+
+```bash
+node tools/flow-moment.mjs validate product/flow-moments
+node tools/flow-moment.mjs render product/flow-moments/audit-community-composition.flow.json
+```
+
+System buildings and front-facing products consume those moments through a thinner manifest:
+
+```bash
+node tools/system-component.mjs validate product/system-components
+node tools/system-component.mjs render product/system-components/loa-freeside.system.json
+```
+
+The system manifest captures the engineering operator, component object, consumer question, stable responsibility, trust surface, actions, ownership boundary, and handoff. It links to a canonical `FM-*` record when the relationship is real. An unmapped component must say why instead of inventing a user-truth relationship.
+
+## Three independent axes
+
+| Axis | Answers | States |
+|---|---|---|
+| Hivemind learning status | Does the flow moment help the user progress? | `cant-make-a-conclusion`, `smol-evidence`, `directionally-correct`, `strongly-validated`, `hypothesis-failed` |
+| Component maturity | Can an agent safely reuse this implementation pattern? | `uncaptured`, `silver`, `gold`, `retired` |
+| Exposure | Who sees the experience, and why? | `dark`, `internal`, `partner`, `default-on`, `retiring`, `retired` |
+
+A Gold component can appear inside a falsified flow. A Silver prototype can produce strong user evidence. Component promotion never changes Hivemind learning status.
+
+## Promotion rules
+
+- A product thesis starts as an unproven hypothesis.
+- Dark research may be recorded before a flag is wired.
+- Exposed research requires a feature-flag reference and an evidence contract.
+- Default-on research requires a learning reference and an operator decision.
+- Gold requires Silver metadata, a taste owner, 14 production days, active use, no regressions, and evidence references.
+- CI validates claims and evidence pointers. It cannot decide whether the taste is good or the hypothesis is true.
+
+## Evidence privacy
+
+Product telemetry, website feedback, tickets, Discord, interviews, partner observations, and personal DMs may all orient learning. Raw private messages do not belong in the repository. Summarize them in operator voice, retain a provenance reference, and use direct quotations only when consented and redacted.
+
+## Adding a moment
+
+1. Copy an existing `product/flow-moments/*.flow.json` record.
+2. Give the moment a stable `FM-*` identifier.
+3. State the actor's entry state, desired progress, dream outcome, hypothesis, and falsification conditions.
+4. Attach the canonical Hivemind classification without adding an action field.
+5. Define behavioral and qualitative signals before exposure.
+6. Attach exemplars with both what to adopt and what to reject.
+7. Keep components `uncaptured` until their design intent is actually captured.
+8. Run validation and render the receipt before requesting review.
+
+## Cross-repository adoption
+
+`loa-freeside` owns the schemas and validators. Consumer repositories own only:
+
+- `product/system-components/<repo>.system.json`, containing their local responsibility and real flow links; and
+- a thin workflow caller pinned to an immutable `loa-freeside` governance commit.
+
+This makes drift reviewable without vendoring the contract into every API or product repository.
--- a/grimoires/territory.yaml
+++ b/grimoires/territory.yaml
@@ -34,6 +34,8 @@ scopes:
   - "tools/system-component.test.mjs"
   - "tools/construct-operator.mjs"
   - "tools/construct-operator.test.mjs"
+  - "tools/fixtures/construct-operator.snapshot.json"
+  - "tools/hivemind/hivemind-labels.v1.0.json"
 
 # Structural architecture and machine-readable projection are the bounded
 # expertise needed by this region. Both are observations at birth.
--- a/grimoires/territory.yaml
+++ b/grimoires/territory.yaml
@@ -1,0 +1,43 @@
+# Freeside product-governance territory.
+#
+# This file is owned by this region. loa-constructs indexes and validates it,
+# but does not author the loadout. A declared authority_tier is a ceiling, not
+# a grant; every construct below is observe-only until earned authority exists.
+
+schema_version: "1.0"
+region: loa-freeside
+maintainers:
+  - "@janitooor"
+  - "@zkSoju"
+
+outcomes:
+  - id: user-intent-continuity
+    description: "System components remain joined to explicit user-flow moments, evidence state, and ownership boundaries across context windows and repository seams."
+
+  - id: expertise-boundary-integrity
+    description: "Construct prose is visibly non-authoritative, mechanical claims resolve to producer-owned contracts, and declared expertise never silently becomes permission."
+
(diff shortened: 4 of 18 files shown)


zkSoju (commented):
## Summary

**Verdict: REQUEST_CHANGES.** The governance model has a strong separation of hypothesis maturity from component maturity, but its trust boundary currently validates internal shape more reliably than external truth. Canonical references, rendered prose, and CI triggers all need tighter closure before these receipts can safely represent authority.

A governance receipt is a miniature control plane: every identifier must resolve, every visual boundary must match the data boundary, and every input capable of changing validation must trigger validation.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "BF-001",
      "title": "System manifests can reference nonexistent canonical flow moments",
      "severity": "HIGH",
      "category": "validation",
      "file": "tools/system-component.mjs:48",
      "description": "Semantic validation only checks that canonical_ref equals flow:<flow_moment_id>. It never verifies that the referenced flow moment exists in the canonical governance records. A consumer manifest can therefore pass CI while claiming continuity with an invented or removed FM-* identifier.",
      "suggestion": "Load the canonical flow-moment records, require each linked flow_moment_id to resolve exactly once, and fail validation for missing or duplicate canonical records.",
      "confidence": 0.98,
      "dimension": "[Quality]",
      "faang_parallel": "Google's Bazel treats labels as resolvable graph edges, not merely well-formed strings; unresolved targets fail analysis.",
      "metaphor": "The address has valid handwriting, but nobody checks whether the building exists.",
      "teachable_moment": "Reference validation requires both syntax and referential integrity.",
      "connection": "This is the same integrity boundary enforced by foreign keys, linkers, and package resolvers.",
      "decision_trail": "Document the canonical record source, uniqueness rule, and behavior for retired IDs."
    },
    {
      "id": "BF-002",
      "title": "Raw construct prose can visually spoof authoritative receipt sections",
      "severity": "HIGH",
      "category": "output-integrity",
      "file": "tools/construct-operator.mjs:247",
      "description": "orientation.description and other producer-controlled strings are interpolated directly into Markdown. Although the JSON structure marks orientation as non-authoritative, embedded headings or formatted text can create counterfeit Mechanics or Authority sections in the rendered operator receipt.",
      "suggestion": "Render producer prose inside a clearly delimited block with Markdown control characters escaped or sanitized. Apply the same treatment to other producer-controlled receipt fields and test adversarial headings, links, and HTML.",
      "confidence": 0.93,
      "dimension": "[Security]",
      "faang_parallel": "GitHub's rendering pipeline sanitizes untrusted markup because presentation can impersonate trusted interface structure.",
      "metaphor": "A visitor can print an EMPLOYEE badge because the lobby printer accepts arbitrary formatting.",
      "teachable_moment": "Semantic trust labels fail if the presentation layer can counterfeit them.",
      "connection": "This is output injection: analogous to XSS even when the immediate medium is Markdown.",
      "decision_trail": "Specify allowed Markdown, escaping rules, HTML handling, and which fields are producer-controlled."
    },
    {
      "id": "BF-003",
      "title": "Territory loadouts may claim nonexistent outcomes",
      "severity": "MEDIUM",
      "category": "validation",
      "file": "tools/construct-operator.mjs:194",
      "description": "Unknown stationing outcome IDs are retained with description: null and do not affect integrity status. The resulting receipt can report that a construct answers for an outcome which the selected region never declares.",
      "suggestion": "Reject every stationing.outcomes entry that is absent from region.outcomes, and validate duplicate outcome IDs and duplicate construct stationings before rendering.",
      "confidence": 0.97,
      "dimension": "[Quality]",
      "faang_parallel": "Kubernetes rejects many invalid object relationships during admission rather than preserving ambiguous intent.",
      "metaphor": "A train manifest lists a station that is absent from the railway map.",
      "teachable_moment": "Null enrichment should not conceal a broken domain relationship.",
      "connection": "This is another graph-integrity gap, now inside the territory model.",
      "decision_trail": "Record whether unknown outcomes are invalid, forward-compatible, or warning-only—and why."
    },
    {
      "id": "BF-004",
      "title": "Dependency changes can bypass the governance workflow",
      "severity": "MEDIUM",
      "category": "ci",
      "file": ".github/workflows/flow-moment-governance.yml:4",
      "description": "The workflow installs dependencies from package.json and the pnpm lockfile, but neither file is included in the pull_request path filter. A dependency or lockfile change that breaks or alters these validators can merge without running this governance check.",
      "suggestion": "Add package.json, pnpm-lock.yaml, and any relevant workspace configuration files to the workflow path filter.",
      "confidence": 0.96,
      "dimension": "[Operational]",
      "faang_parallel": "Bazel derives rebuilds from the full dependency graph; CI path filters must model the same causal closure.",
      "metaphor": "The alarm watches the vault door but ignores changes to the alarm's wiring.",
      "teachable_moment": "A check must trigger on its implementation and dependency closure, not only its subject files.",
      "connection": "This is a supply-chain and change-detection boundary.",
      "decision_trail": "Document every file class capable of changing validator behavior."
    },
    {
      "id": "BF-005",
      "title": "Typed reference pattern accepts trailing whitespace and arbitrary suffixes",
      "severity": "LOW",
      "category": "schema",
      "file": "spec/product/flow-moment.schema.json:236",
      "description": "The reference pattern ^[a-z][a-z0-9+.-]*:\\S+ is not end-anchored, so values such as \"github:owner/repo trailing text\" satisfy the schema. The same pattern is repeated in the system-component schema.",
      "suggestion": "End-anchor the expression as ^[a-z][a-z0-9+.-]*:\\S+$ and add a rejection test for whitespace or trailing text.",
      "confidence": 0.99,
      "dimension": "[Quality]",
      "faang_parallel": "Protocol parsers generally require full-input consumption; accepting a valid prefix creates parser differentials.",
      "metaphor": "The ticket inspector checks only the first word and ignores everything appended.",
      "teachable_moment": "Validation regexes should describe the entire value.",
      "connection": "Full-match parsing prevents ambiguous downstream interpretation.",
      "decision_trail": "Centralize and document the typed-reference grammar to prevent schema drift."
    },
    {
      "id": "BF-006",
      "title": "Promotion rules are separated from component maturity",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/flow-moment.mjs:82",
      "description": "The validator independently enforces evidence confidence, exposure promotion, and Gold maturity, with explicit red-path tests. This prevents component reuse maturity from implicitly promoting an unproven product hypothesis.",
      "suggestion": "Preserve this separation as the schemas evolve and continue adding negative tests for every new promotion path.",
      "confidence": 0.99,
      "faang_parallel": "Google's SRE model separates software release state from service reliability evidence; one signal cannot silently promote another.",
      "metaphor": "A well-built ladder does not prove it is leaning against the right wall.",
      "teachable_moment": "Independent claims deserve independent evidence and promotion gates.",
      "connection": "This is strong separation of concerns across product truth, exposure, and implementation maturity.",
      "decision_trail": "Capture this invariant in the contract so future schema changes cannot collapse the axes."
    }
  ]
}
```


## Callouts

The strongest choice is the explicit separation of evidence confidence, exposure promotion, and Gold maturity. The negative-path tests make that architecture executable rather than aspirational.

The new artifacts also show disciplined intent: schemas, contracts, fixtures, validators, and workflow integration evolve together. Once referential integrity and presentation safety are closed, this structure can become a durable governance substrate.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. This PR shows a strong governance instinct: the fail-closed tests and construct boundary checks are the right shape for an authority-sensitive system. The main concern is that the reusable validator contract currently promises a thinner integration path than the implementation permits.

Architecturally, this is a classic boundary-design PR: it is defining what can be projected, validated, and promoted across contexts. The strongest parts treat authority as a capability that must be earned; the weakest part is where portability and local canonicality are still coupled.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Reusable system validator still requires local flow records",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "tools/system-component.mjs:88",
      "description": "The reusable workflow exposes validate-flow-moments as optional, but validateSystemComponent always requires each flow_moment_id to resolve to exactly one repo-local product/flow-moments/*.flow.json record. A consumer repository that only wants to validate a thin system-component manifest, as documented, will fail unless it vendors full flow records too.",
      "suggestion": "Add an option to validateSystemComponent to disable canonical flow-record resolution for reusable consumer validation, or make the reusable workflow require and document repo-local flow records whenever flow_moments is non-empty.",
      "confidence": 0.88,
      "faang_parallel": "Google's Borg config validation separated schema validity from cluster-local admission checks; portable config linting stayed useful only because it did not require every downstream environment to carry the full production universe.",
      "metaphor": "This is like issuing a passport checker that also requires every airport to keep a copy of the traveler's birth certificate. Sometimes that is the right gate, but it must be explicit.",
      "teachable_moment": "Reusable validation needs layers: shape validation, reference validation, and repository-canonical validation are different promises with different operational costs.",
      "connection": "This connects directly to contract design. If a workflow is reusable, its strictest repo-local assumptions should be opt-in or documented as mandatory inputs."
    },
    {
      "id": "F-002",
      "title": "Reference fields are rendered as raw Markdown",
      "severity": "LOW",
      "category": "security",
      "file": "tools/flow-moment.mjs:154",
      "description": "renderFlowMoment escapes prose fields but interpolates exemplar ref and decision_refs directly into Markdown. The schema only requires a no-whitespace URI-like string, so a malicious or malformed reference can still create raw links or inline HTML in generated receipts.",
      "suggestion": "Pass all reference values through escapeMarkdownText before rendering, including exemplar refs and decision_refs.",
      "confidence": 0.76,
      "faang_parallel": "GitHub's Markdown rendering pipeline treats even trusted-looking reference text as hostile input until it has passed through sanitization and link policy controls.",
      "metaphor": "Escaping only prose fields is like locking the front door while leaving a mail slot big enough to climb through.",
      "teachable_moment": "Anything rendered into Markdown is presentation input, not merely data. References often look harmless, which is why they are easy to miss in output-encoding reviews.",
      "connection": "This is part of the broader output-encoding pattern: escape at the render boundary, consistently, for every value that crosses into a user-readable format."
    },
    {
      "id": "F-003",
      "title": "Promotion gate tests cover the intended red paths",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tools/flow-moment.test.mjs:31",
      "description": "The tests explicitly prove important invalid states fail closed: Hivemind action injection, exposed research without a flag, default-on research without evidence, unsupported confidence claims, invalid dates, and premature Gold maturity.",
      "suggestion": "Keep these red-path tests as the contract evolves; they are the most useful regression protection in this PR.",
      "confidence": 0.95,
      "faang_parallel": "Meta's configuration review culture leans heavily on negative tests for policy gates because the expensive failures are usually invalid states being accepted, not valid states being rejected.",
      "metaphor": "These tests are guardrails on a mountain road: their value appears precisely when something drifts toward the edge.",
      "teachable_moment": "For governance code, red-path tests are often more important than happy paths. They encode what the system must refuse.",
      "connection": "This strengthens the authority boundary introduced by the PR and gives future contributors a clear regression contract."
    },
    {
      "id": "F-004",
      "title": "Construct projection separates prose, mechanics, and authority defensively",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/construct-operator.mjs:83",
      "description": "The construct operator surface rejects authoritative orientation prose, mechanics that claim authority, unknown authority promotion above observe, and unchecked ratification presenting as ok. This is a strong fail-closed boundary for the new governance surface.",
      "suggestion": "No change requested.",
      "confidence": 0.93,
      "faang_parallel": "Amazon's internal service ownership models distinguish documentation, operational mechanism, and launch authority; mixing those layers is how informal advice accidentally becomes production policy.",
      "metaphor": "This design uses separate keys for separate doors: a note, a procedure, and a ratification are not allowed to masquerade as one another.",
      "teachable_moment": "Authority should be typed. When prose, mechanism, and approval share the same channel, systems eventually promote claims by accident.",
      "connection": "This is the strongest architectural move in the PR: it makes governance auditable by preserving the distinction between expression, execution, and permission."
    }
  ]
}
```


## Callouts
[Test Coverage] The negative tests around promotion gates are especially strong because they protect against the highest-cost failure mode: invalid governance states becoming normal. That is the same lesson large infra teams learned from policy-as-code systems: refusals deserve first-class tests.

[Quality] The construct projection boundary is thoughtfully shaped. Separating orientation prose from mechanics and ratification keeps future agents and humans from confusing “someone said this” with “the system is allowed to do this.”

[Operational] The PR is already thinking in reusable contracts and workflows, which is the right direction. Tightening the documented portability boundary in F-001 will make that contract easier for downstream repositories to adopt without surprise vendoring requirements.

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#468. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "PRAISE-001",
      "title": "Promotion gates are backed by red-path tests",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tools/flow-moment.test.mjs:21",
      "description": "The new validator is exercised against invalid Hivemind extension, exposed research without a flag, unsupported default-on exposure, premature Gold maturity, invalid dates, and Markdown injection paths. That materially reduces the risk of the governance schema becoming documentation-only.",
      "suggestion": "Keep these red-path tests close to the validators as the schema evolves.",
      "confidence": 0.95
    },
    {
      "id": "PRAISE-002",
      "title": "Reusable workflow isolates governance dependencies from consumer workspaces",
      "severity": "PRAISE",
      "category": "ci",
      "file": ".github/workflows/reusable-flow-moment-governance.yml:39",
      "description": "The reusable workflow installs dependencies in the checked-out governance repository with --ignore-scripts and --ignore-workspace, which avoids a consumer repository's pnpm workspace influencing the validator install.",
      "suggestion": "Preserve this isolation when adding future reusable validation jobs.",
      "confidence": 0.9
    },
    {
      "id": "LOW-001",
      "title": "Date-sensitive Gold maturity checks can become nondeterministic in CI",
      "severity": "LOW",
      "category": "reliability",
      "file": "tools/flow-moment.mjs:132",
      "description": "The CLI defaults `today` to the current date, so a Gold component with a recent `production_since` value can fail when opened and later pass automatically after 14 days without any PR change. That makes governance outcomes time-dependent unless callers consistently pass --today.",
      "suggestion": "Consider requiring --today in CI or deriving the as-of date from an explicit record field so validation results are reproducible for a given commit.",
      "confidence": 0.78
    },
    {
      "id": "LOW-002",
      "title": "Unchecked construct fixture returns success despite partial status",
      "severity": "LOW",
      "category": "ci",
      "file": "tools/construct-operator.mjs:395",
      "description": "The construct operator exits nonzero only for `drift`, while an unchecked territory ratification renders `partial` and still exits 0. The workflow step therefore proves the receipt renders, but does not fail if the operator surface remains unratified or missing mechanics.",
      "suggestion": "If CI is intended to enforce a ratified operator surface after this branch lands, add a strict mode or workflow assertion that rejects `status: partial` for release-gating contexts.",
      "confidence": 0.72
    },
    {
      "id": "LOW-003",
      "title": "Unknown CLI flags are silently ignored",
      "severity": "LOW",
      "category": "usability",
      "file": "tools/flow-moment.mjs:277",
      "description": "The CLIs parse known flags with ad hoc scans and ignore unrecognized flags. A typo such as `--todays 2026-07-14` would be accepted and validation would run against the current date instead of failing fast.",
      "suggestion": "Reject unknown flags in `flow-moment.mjs`, `system-component.mjs`, and `construct-operator.mjs` before running validation or rendering.",
      "confidence": 0.7
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Verdict: COMMENT. This PR shows a strong governance architecture: schemas, portable validators, and rendered receipts are moving in the right direction. The main concerns are around time trust and ref immutability, both of which matter because governance systems only work when the verifier cannot be quietly moved underneath the record.

Architecture note: this change is building a contract layer, not just tooling. In systems like this, dates and refs are part of the security boundary: they decide what counts as mature, canonical, and reviewed.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Gold maturity can be validated against a future as_of date",
      "severity": "MEDIUM",
      "category": "validation",
      "file": "tools/flow-moment.mjs:129",
      "description": "validateFlowMoment defaults the Gold age check to document.as_of, but semanticErrors does not reject an as_of date in the future. A record can advance as_of arbitrarily and pass the 14-day production floor before that date has actually occurred.",
      "suggestion": "Reject records whose as_of is later than the validator's wall-clock date, or require CI to pass an explicit trusted --today and compare document.as_of <= today before using it for maturity checks.",
      "confidence": 0.86,
      "faang_parallel": "Google's Zanzibar paper is a useful parallel: authorization decisions are only meaningful when evaluated against a trusted consistency point. Here, as_of becomes a consistency point, so it must be constrained by trusted time.",
      "metaphor": "This is like letting someone postdate a driver's license and then accepting it today because it will eventually be valid.",
      "teachable_moment": "Any field used to prove maturity, age, freshness, or expiration is not ordinary metadata; it is verifier input and needs an independent trust anchor.",
      "connection": "This connects to the broader pattern of separating claimed document state from validator-observed state. Governance records may declare intent, but CI should decide whether the claim is presently valid.",
      "decision_trail": "Document whether as_of is authored truth, validator observation, or both. Future maintainers need to know why maturity uses this date and what prevents temporal self-certification."
    },
    {
      "id": "F-002",
      "title": "Reusable workflow accepts mutable governance refs",
      "severity": "MEDIUM",
      "category": "supply-chain",
      "file": ".github/workflows/reusable-flow-moment-governance.yml:6",
      "description": "The workflow input is described as an immutable loa-freeside commit, but the workflow never verifies that inputs.governance-ref is a full commit SHA. Consumers can pin a branch or tag, which makes governance validation change without a reviewed caller change.",
      "suggestion": "Validate governance-ref with an early shell step such as a full 40-character SHA regex, or document that mutable refs are accepted and rename the input/description accordingly.",
      "confidence": 0.82,
      "faang_parallel": "SLSA and Google's Binary Authorization practices both lean on immutable provenance. A mutable branch ref is the CI equivalent of rebuilding from a different source while keeping the same label.",
      "metaphor": "This is like approving a building inspection against blueprint version 'latest' instead of the stamped drawing set.",
      "teachable_moment": "Supply-chain guarantees depend less on where code comes from and more on whether the reference can move after review.",
      "connection": "This is the same trust-boundary issue as F-001 in another form: governance inputs must not be self-moving. Time and Git refs both need pinning when they affect validation outcomes.",
      "decision_trail": "Record whether consumers are required to pin immutable SHAs. If mutable refs are intentionally supported for development, name that mode explicitly and separate it from production governance."
    },
    {
      "id": "F-003",
      "title": "Flow observations can reference duplicate signal ids without being caught",
      "severity": "LOW",
      "category": "validation",
      "file": "tools/flow-moment.mjs:75",
      "description": "The observation signal_ids schema only requires strings, so syntactically valid duplicate ids are blocked by uniqueItems, but ids with the wrong format are only caught later when missing from the signal set. This produces weaker field-level validation and allows nonconforming ids to reach semantic logic.",
      "suggestion": "Reuse the signal id pattern for observation.signal_ids items in spec/product/flow-moment.schema.json so malformed references fail at the schema layer.",
      "confidence": 0.64,
      "faang_parallel": "Protocol Buffers culture at Google pushes shape constraints as close to the schema as practical, so downstream systems can reason about typed fields instead of defensive string soup.",
      "metaphor": "A schema should be the bouncer at the door, not a note asking the bartender to check IDs later.",
      "teachable_moment": "Put format validation at the earliest layer that has enough context. Semantic checks should answer 'does this valid reference exist?', not 'was this ever a valid reference shape?'",
      "connection": "This strengthens the contract between schema and semantic validator, keeping each layer crisp and easier to test.",
      "decision_trail": "Document the signal id grammar in one schema location and reference it wherever signal ids appear, so future additions do not drift."
    },
    {
      "id": "F-004",
      "title": "Receipts consistently escape untrusted prose before Markdown rendering",
      "severity": "PRAISE",
      "category": "security",
      "file": "tools/markdown-text.mjs:5",
      "description": "The shared Markdown escaping helper is applied across receipt renderers, and tests cover heading, list, link, and inline HTML injection attempts from untrusted record fields.",
      "suggestion": "Keep new receipt fields routed through the same helper by default.",
      "confidence": 0.95,
      "faang_parallel": "React's default escaping model is the industry lesson here: make the safe path the normal path, and require deliberate ceremony to render raw content.",
      "metaphor": "This is a clean kitchen rule: every ingredient from outside gets washed before it touches the plate.",
      "teachable_moment": "Security scales when escaping is centralized and boring. The best injection defense is the one future contributors use without needing to remember a policy.",
      "connection": "This is especially important because governance receipts may be copied into GitHub, docs, or chat surfaces where Markdown behavior differs.",
      "decision_trail": "Keep a short note near receipt rendering that all user-authored prose must pass through the shared Markdown escaping helper before display."
    },
    {
      "id": "F-005",
      "title": "Portable validation keeps consumer repositories from vendoring canonical flow records",
      "severity": "PRAISE",
      "category": "design",
      "file": "tools/system-component.mjs:119",
      "description": "The requireCanonicalFlowRecords option preserves local canonical checks for loa-freeside while letting consumers validate typed flow links without copying Freeside's flow record corpus.",
      "confidence": 0.93,
      "faang_parallel": "This resembles Kubernetes' API machinery split between core validation and admission policy: consumers can validate shape locally without importing the whole control-plane corpus.",
      "metaphor": "It gives downstream repos a passport checkpoint without making them carry the entire national archive.",
      "teachable_moment": "Portable validators should separate universal contract checks from repository-local authority checks. That makes adoption cheaper without weakening the source-of-truth repo.",
      "connection": "This is the strongest architectural move in the PR: it treats governance as federated infrastructure instead of a monolith.",
      "decision_trail": "Document which checks are portable and which are canonical-only, so consumers know what guarantees they get from each mode."
    }
  ]
}
```


## Callouts

[Security] The Markdown escaping helper is a strong foundation because it makes safe rendering the default behavior for receipts, which are exactly the kind of artifact that tends to travel across surfaces.

[Quality] The portable validation option is well-shaped. It respects the difference between Freeside as source of authority and consumer repositories as participants in the contract.

[Test Coverage] The receipt escaping tests cover practical Markdown injection shapes, not just happy paths. That gives future contributors executable examples of the security boundary.

zkSoju (commented):
## Summary
Verdict: COMMENT. This PR shows strong governance instincts: the validators avoid empty-success traps, generated receipts treat prose as hostile input, and reusable CI prefers immutable contracts over convenience. The main architectural theme is healthy: governance automation should be boring, explicit, and resistant to drift.

The throughline here is provenance. Like Spanner’s TrueTime or Borg’s declarative job model, the value is not only that the system works today; it is that future operators can tell what was checked, against which contract, and why that contract could not silently move.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FINDING-001",
      "title": "Validators fail closed when no records are found",
      "severity": "PRAISE",
      "category": "validation",
      "file": "tools/flow-moment.mjs:280",
      "description": "Both the flow-moment and system-component CLIs explicitly return a failure when their target directory contains no matching records, preventing CI from passing after validating an empty set.",
      "suggestion": "Keep this behavior; it is an important guard for path-scoped governance workflows.",
      "confidence": 0.95,
      "faang_parallel": "Google's Borg configuration systems treat empty target sets as a control-plane risk, because a successful no-op can hide a broken selector or misplaced ownership boundary.",
      "metaphor": "This is like a fire drill where nobody checks whether anyone was actually in the building; a clean report means little if the search area was empty.",
      "teachable_moment": "Validators should distinguish 'all records passed' from 'there were no records to validate.' Those are operationally different states and should produce different outcomes.",
      "connection": "This connects to fail-closed governance: scoped automation must prove it inspected the intended domain before it claims success."
    },
    {
      "id": "FINDING-002",
      "title": "Markdown receipts consistently escape untrusted prose",
      "severity": "PRAISE",
      "category": "security",
      "file": "tools/markdown-text.mjs:5",
      "description": "The shared Markdown escaping helper is used across generated receipts and is covered by injection tests for headings, links, and inline HTML.",
      "suggestion": "Keep generated receipt rendering routed through this helper for any future prose fields.",
      "confidence": 0.95,
      "faang_parallel": "React's default escaping model is the relevant industry pattern: make the safe path the normal path, and force unsafe rendering to be explicit.",
      "metaphor": "The helper is a customs checkpoint for text. Most prose can pass through, but it gets inspected before entering a rendered document.",
      "teachable_moment": "Receipt generation is a security boundary when user- or model-authored prose becomes Markdown. Centralizing escaping prevents each caller from re-solving the same problem inconsistently.",
      "connection": "This supports durable audit artifacts: generated governance documents remain readable without becoming a vehicle for HTML or Markdown injection."
    },
    {
      "id": "FINDING-003",
      "title": "Reusable workflow pins governance input to immutable commits",
      "severity": "PRAISE",
      "category": "ci",
      "file": ".github/workflows/reusable-flow-moment-governance.yml:21",
      "description": "The reusable workflow rejects branch and tag inputs before checking out the governance contract, which prevents consumers from silently moving validator behavior.",
      "suggestion": "Keep the full 40-character SHA requirement for cross-repository adoption.",
      "confidence": 0.9,
      "faang_parallel": "SLSA and Google's internal supply-chain practices both emphasize immutable, provenance-bearing inputs because mutable refs make rebuilds and audits ambiguous.",
      "metaphor": "A branch name is a street address that can point to a remodeled house tomorrow; a commit SHA is the exact brick in the wall.",
      "teachable_moment": "CI workflows that enforce policy should pin policy code immutably. Otherwise, consumers cannot distinguish intentional validator upgrades from accidental or compromised drift.",
      "connection": "This is the same design family as lockfiles and digest-pinned containers: reproducibility is a governance feature, not just a build convenience."
    }
  ]
}
```


## Callouts
[Security] The Markdown escaping path is especially strong because it is shared and tested. That combination matters: a helper without adoption is decoration, and adoption without tests is trust without a receipt.

[Test Coverage] The injection-oriented tests around generated prose are the right kind of paranoia. They encode the threat model directly, which gives future maintainers a concrete warning when expanding receipt fields.

[Operational] The immutable SHA requirement in the reusable workflow is excellent governance ergonomics. It asks consumers to make upgrades deliberate, which creates a natural decision trail for future agents and humans reviewing why behavior changed.

[Quality] The empty-record failure is a small implementation choice with large system value. It prevents the most dangerous kind of green check: the one produced by checking nothing.

zkSoju (commented):
## Summary

Verdict: COMMENT. This PR shows strong governance instincts: schemas, receipts, tests, and reusable workflow boundaries are being treated as product surfaces, not incidental tooling. The main concern is CLI argument validation: both validators fail open around extra positional arguments, which can make CI/operator evidence look stronger than it is.

Architecturally, the theme here is trust calibration. The Markdown and workflow choices reduce ambiguity at the boundaries; the CLI parsing should meet that same standard so “validated” always means “validated the thing the operator intended.”

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Flow validator accepts stray positional arguments",
      "severity": "LOW",
      "category": "cli-validation",
      "file": "tools/flow-moment.mjs:301",
      "description": "The validate command chooses only the first non-flag positional argument as the target and does not reject additional positional arguments. A command such as `node tools/flow-moment.mjs validate --today 2026-07-14 product/flow-moments` silently validates the default path instead of the supplied path, which can make operator or CI usage look correct while ignoring the intended target.",
      "suggestion": "Parse validate arguments into explicit positional and flag sets, reject more than one positional target, and support flags before or after the target.",
      "confidence": 0.83,
      "faang_parallel": "Google's Borg/SRE tooling culture strongly favors fail-closed command surfaces because automation logs become operational evidence. A command that appears successful while checking the wrong target weakens that evidence chain.",
      "metaphor": "This is like a security guard checking the first badge handed over and ignoring the second person walking in behind it.",
      "teachable_moment": "Validation tools are part of the contract. When a validator accepts ambiguous input, the ambiguity moves downstream into CI, release notes, and operator confidence.",
      "connection": "This connects directly to governance receipts: if the receipt is the artifact of trust, argument parsing is the front door to that trust.",
      "decision_trail": "Document the CLI grammar: accepted flags, whether flag order is free-form, and the exact failure mode for extra positional targets."
    },
    {
      "id": "F-002",
      "title": "System component validator accepts stray positional arguments",
      "severity": "LOW",
      "category": "cli-validation",
      "file": "tools/system-component.mjs:221",
      "description": "The validate command uses the first non-flag argument as the target and ignores any later positional arguments. Mistyped invocations can therefore pass while validating a different directory than the caller intended.",
      "suggestion": "Reject extra positional arguments after extracting the optional target, matching the existing unknown-flag fail-closed behavior.",
      "confidence": 0.8,
      "faang_parallel": "Meta's internal config validators commonly treat malformed invocation as a hard failure because large-scale infra depends on humans and bots reading validation status literally.",
      "metaphor": "It is a GPS accepting two destinations, driving to the first, and still saying the trip succeeded.",
      "teachable_moment": "Consistency is a quality feature. Since unknown flags already fail closed, extra positional arguments should follow the same contract.",
      "connection": "The flow-moment and system-component validators are sibling tools; their parse behavior should converge so operators build one mental model.",
      "decision_trail": "Add a short parser contract near the command implementation or docs so future tool additions preserve the same fail-closed rule."
    },
    {
      "id": "F-003",
      "title": "Markdown receipt escaping is centralized and tested against injection",
      "severity": "PRAISE",
      "category": "security",
      "file": "tools/markdown-text.mjs:5",
      "description": "The receipt renderers consistently route untrusted prose through `escapeMarkdownText`, and the tests cover forged headings, links, and inline HTML across flow, system component, and construct receipts.",
      "suggestion": "Keep this helper as the only path for rendering user-controlled text into derived Markdown receipts.",
      "confidence": 0.96,
      "faang_parallel": "This mirrors the React/Google Closure lesson: centralize escaping at rendering boundaries rather than relying on each caller to remember defensive formatting.",
      "metaphor": "This is a customs checkpoint for text. Every suitcase gets inspected before entering the rendered receipt.",
      "teachable_moment": "Escaping is strongest when it is boring and centralized. The best security helper is the one future code naturally reaches for.",
      "connection": "This supports the larger governance model: receipts can safely preserve user intent without allowing that intent to rewrite the receipt structure.",
      "decision_trail": "Record that receipt Markdown must pass through this helper, and add tests whenever a new renderer is introduced."
    },
    {
      "id": "F-004",
      "title": "Reusable workflow pins governance by immutable commit",
      "severity": "PRAISE",
      "category": "supply-chain",
      "file": ".github/workflows/reusable-flow-moment-governance.yml:19",
      "description": "The reusable workflow rejects non-SHA governance refs before checkout and installs the governance workspace with scripts and workspace linking disabled, reducing the chance that consumer repositories can accidentally or intentionally move the validator contract.",
      "suggestion": "Preserve the full-SHA gate and isolated install semantics when extending this reusable workflow.",
      "confidence": 0.93,
      "faang_parallel": "This is aligned with SLSA-style supply-chain practice: pin the producer, constrain install behavior, and make provenance less mutable.",
      "metaphor": "It is bolting the measuring ruler to the table before asking multiple teams to measure against it.",
      "teachable_moment": "Reusable workflows become policy distribution systems. Pinning by SHA turns policy from a moving suggestion into an auditable dependency.",
      "connection": "This is the strongest operational pattern in the PR: governance is treated as infrastructure, with immutability and isolation built in.",
      "decision_trail": "Document why full SHA refs are required and what process updates that pin, so future maintainers do not relax it for convenience."
    }
  ]
}
```


## Callouts

[Security] The centralized Markdown escaping is excellent because it protects the structure of governance receipts while still allowing rich operator prose to flow through the system.

[Operational] The reusable workflow’s immutable ref gate is a serious supply-chain move. It gives downstream repositories a stable governance contract instead of a branch-shaped promise.

[Test Coverage] The injection-oriented receipt tests are especially valuable: they test the failure mode people usually discover only after generated Markdown becomes an authority surface.

zkSoju (commented):
## Summary

Verdict: COMMENT. This PR shows unusually strong governance instincts: fail-closed validation, immutable workflow inputs, and explicit product contracts. The main concern is contract layering: CI, schema, and semantic validators currently enforce slightly different truths, so future consumers need clearer breadcrumbs about which layer is authoritative.

The architecture is strongest where it treats governance as executable policy, not prose. The remaining work is to make partial acceptance, schema-only validation, and semantic validation boundaries obvious enough that both humans and future agents cannot accidentally promote incomplete state.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "praise-001",
      "title": "Validators prove negative promotion paths",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tools/flow-moment.test.mjs:29",
      "description": "The test suite covers several fail-closed cases, including forbidden Hivemind actions, exposed research without flags, default-on research without evidence, premature Gold maturity, invalid dates, and Markdown injection in rendered receipts.",
      "suggestion": "Keep these red-path tests as the contract evolves; they are doing real governance work rather than only snapshotting happy paths.",
      "confidence": 0.95,
      "faang_parallel": "Google SRE practice treats failure-mode tests as part of the service contract, not as secondary coverage; this is the same instinct applied to governance artifacts.",
      "metaphor": "These tests are guardrails on a mountain road: their value appears when someone drifts, not when the road is empty.",
      "teachable_moment": "For policy code, negative tests often matter more than happy paths because they prove what cannot be promoted.",
      "connection": "This connects to the broader pattern of executable governance: the validator becomes the living boundary between observation, evidence, and authority."
    },
    {
      "id": "praise-002",
      "title": "Reusable workflow pins governance by immutable SHA",
      "severity": "PRAISE",
      "category": "supply-chain",
      "file": ".github/workflows/reusable-flow-moment-governance.yml:21",
      "description": "The reusable workflow rejects non-40-character lowercase commit refs before checking out the governance repository, which prevents consumers from silently tracking a mutable branch or tag.",
      "suggestion": "Preserve this requirement for cross-repository adoption; it makes validator upgrades reviewable in each consumer.",
      "confidence": 0.96,
      "faang_parallel": "Google's Binary Authorization and Meta's internal deployment controls both rely on immutable provenance because mutable references make audits retroactive guesses.",
      "metaphor": "A branch name is a street address; a commit SHA is a fingerprint.",
      "teachable_moment": "Supply-chain safety starts by making change explicit and reviewable at every trust boundary.",
      "connection": "This is the right pattern for governance reuse: consumers inherit policy intentionally instead of floating with upstream movement."
    },
    {
      "id": "low-001",
      "title": "Construct operator fixture is never required to reach ok status in CI",
      "severity": "LOW",
      "category": "validation",
      "file": ".github/workflows/flow-moment-governance.yml:55",
      "description": "The CI render step for the construct operator fixture does not pass `--require-ok`, while the checked-in fixture has `ratification_status: unchecked` and renders a `partial` surface. That means the governance workflow can stay green even when the operator expertise projection is intentionally incomplete, and only the separate unit test asserts that `--require-ok` would reject it.",
      "suggestion": "If this workflow is intended as a release gate for construct expertise, add a second explicit `node tools/construct-operator.mjs render --snapshot ... --require-ok` gate once ratification is expected. If partial is intentional for this PR, document that this CI step is smoke rendering rather than acceptance.",
      "confidence": 0.78,
      "faang_parallel": "Amazon commonly separates canary/smoke checks from production readiness gates; the danger is when the label does not make that distinction clear.",
      "metaphor": "This is like a fire alarm test that proves the bell rings, but not that the building is safe to occupy.",
      "teachable_moment": "CI jobs should name their authority level: render smoke test, semantic validation, or release gate.",
      "connection": "The broader pattern is promotion hygiene. Partial surfaces are useful, but only if the pipeline says whether partial is allowed or merely tolerated."
    },
    {
      "id": "low-002",
      "title": "Schema allows duplicate signal ids until semantic validation runs",
      "severity": "LOW",
      "category": "schema",
      "file": "spec/product/flow-moment.schema.json:115",
      "description": "`evidence_contract.signals` does not declare `uniqueItems: true`, so JSON Schema alone accepts duplicate signal objects and relies on the custom validator to reject duplicate ids. Consumers that use only the published schema could accept records the repository validator rejects.",
      "suggestion": "Add `uniqueItems: true` to `evidence_contract.signals` if whole-object duplicates should be schema-invalid, and consider documenting that id-level uniqueness requires the semantic validator.",
      "confidence": 0.73,
      "faang_parallel": "Protocol Buffers at Google separate wire shape from semantic invariants; teams document which invariants require service-side validation.",
      "metaphor": "The schema checks that every passport has the right fields; the semantic validator checks that two people are not using the same identity.",
      "teachable_moment": "Schemas are shape contracts, not complete truth contracts, unless the invariant can actually be expressed there.",
      "connection": "This should become explicit validator doctrine: schema consumers get structural safety, repository tooling gets semantic safety."
    },
    {
      "id": "low-003",
      "title": "System component flow_moments uniqueItems does not enforce duplicate flow ids",
      "severity": "LOW",
      "category": "schema",
      "file": "spec/product/system-component.schema.json:124",
      "description": "`flow_moments` uses `uniqueItems: true`, which only rejects identical objects. Two entries can still share the same `flow_moment_id` with different roles or contributions, and only `tools/system-component.mjs` catches that semantically.",
      "suggestion": "Keep the semantic duplicate-id check, and either document that validator as mandatory or add a schema-level constraint if the toolchain supports it.",
      "confidence": 0.76,
      "faang_parallel": "Kubernetes API validation has the same split: OpenAPI handles object shape, while admission controllers enforce cross-field and identity rules.",
      "metaphor": "The list can prevent two identical tickets, but it cannot tell that two different tickets reserve the same seat.",
      "teachable_moment": "Identity uniqueness is usually semantic, especially when duplicates can differ in surrounding metadata.",
      "connection": "This reinforces the need for a documented validation stack: JSON Schema for portable structure, custom tools for governance meaning."
    }
  ]
}
```


## Callouts

[Security] The Markdown injection coverage is especially strong because rendered receipts are trust surfaces. That test anticipates the exact place where governance text can become executable influence.

[Operational] The immutable SHA workflow is production-grade supply-chain thinking. It makes policy upgrades auditable instead of ambient.

[Test Coverage] The negative-path tests are the standout. They teach future contributors the real contract: not just how to create a valid flow moment, but what must never be smuggled through one.

zkSoju (commented):
## Summary

Verdict: `COMMENT`

This PR has a strong governance shape: schema validation is kept portable, semantic promotion rules live in code, and generated receipts treat prose as hostile input. The main concerns are contract completeness and reproducibility: references should resolve to inspectable component records, and rendering should not depend silently on today’s date.

The architectural theme here is traceability. Governance systems earn trust when every claim points to something reviewable, every generated artifact can be reproduced, and malformed producer input fails with a precise contract error.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "praise-001",
      "title": "Validators separate schema shape from semantic promotion rules",
      "severity": "PRAISE",
      "category": "validation",
      "file": "tools/flow-moment.mjs:38",
      "description": "The implementation uses JSON Schema for portable shape validation and a separate semantic validation pass for invariants like evidence confidence, exposure promotion, duplicate IDs, and Gold maturity timing.",
      "suggestion": "Keep this split; it makes the portable contract clear while preserving repository-specific acceptance checks.",
      "confidence": 0.95,
      "faang_parallel": "Google’s Protocol Buffers separate wire shape from service-level invariants for the same reason: portable structure should not pretend to encode every domain rule.",
      "metaphor": "The schema is the building code; semantic validation is the inspector who checks whether this specific building is safe to occupy.",
      "teachable_moment": "A schema says what can be parsed. A contract validator says what can be trusted.",
      "connection": "This matches the broader pattern of treating schemas as interfaces, not full governance authority."
    },
    {
      "id": "praise-002",
      "title": "Reusable workflow pins governance by immutable commit",
      "severity": "PRAISE",
      "category": "supply-chain",
      "file": ".github/workflows/reusable-flow-moment-governance.yml:18",
      "description": "The reusable workflow rejects non-SHA governance refs before checking out the validator source, preventing consumers from silently tracking a moving branch or tag.",
      "suggestion": "This is a strong default for cross-repository governance adoption.",
      "confidence": 0.95,
      "faang_parallel": "Google’s SLSA guidance pushes immutable provenance because mutable refs turn review into a moving target.",
      "metaphor": "A SHA is a sealed evidence bag; a branch name is a sticky note saying where the evidence used to be.",
      "teachable_moment": "Governance code is part of the supply chain. Pinning it is not bureaucracy; it is reproducibility.",
      "connection": "This supports cross-repo adoption without letting one repo’s future changes silently rewrite another repo’s policy."
    },
    {
      "id": "finding-001",
      "title": "Flow records can reference non-existent component signal IDs",
      "severity": "MEDIUM",
      "category": "validation",
      "file": "tools/flow-moment.mjs:118",
      "description": "The validator only checks duplicate component refs, but it does not verify that each flow component ref resolves to a known system-component manifest or a valid external component registry. This lets a flow moment claim component linkage that cannot be inspected or governed by the new system-component contract.",
      "suggestion": "For refs intended to point at local system components, resolve them against product/system-components and require exactly one matching manifest; otherwise make the external component namespace explicit and document that it is not locally verified.",
      "confidence": 0.72,
      "faang_parallel": "Kubernetes learned this boundary with object references: a syntactically valid ref is not the same as a resolvable object with enforceable ownership.",
      "metaphor": "This is like listing a fire extinguisher on the safety plan without checking whether it exists in the building.",
      "teachable_moment": "Governance references need referential integrity. Otherwise the graph looks connected while the review surface is hollow.",
      "connection": "The PR introduces system-component contracts, so flow moments should either bind to that contract or explicitly declare an external, unverifiable namespace."
    },
    {
      "id": "finding-002",
      "title": "CLI render uses wall-clock validation by default",
      "severity": "LOW",
      "category": "reproducibility",
      "file": "tools/flow-moment.mjs:333",
      "description": "The render command calls validateFlowMoment(document) without accepting a --today override. Validation therefore depends on the current UTC date, while validate has an explicit --today option. This makes historical review and fixture rendering harder to reproduce exactly.",
      "suggestion": "Allow render to accept the same --today YYYY-MM-DD option and pass it through to validateFlowMoment.",
      "confidence": 0.8,
      "faang_parallel": "Netflix-style delivery tooling treats time as an input in tests and replay paths because production debugging often means reconstructing yesterday exactly.",
      "metaphor": "A renderer that reads the wall clock is a camera whose lens changes color depending on the day.",
      "teachable_moment": "If validation can be time-sensitive, every command that validates should expose the same time control.",
      "connection": "This is a small CLI symmetry issue, but it matters for governance receipts because receipts become historical evidence."
    },
    {
      "id": "finding-003",
      "title": "Producer metadata is not validated before sorting by name",
      "severity": "LOW",
      "category": "robustness",
      "file": "tools/construct-operator.mjs:24",
      "description": "byName assumes both records have a string name. normalizeExecutionContract and normalizeConstructInfo sort producer-supplied verbs and commands without first validating required string fields, so malformed producer JSON can throw a generic TypeError instead of a field-specific contract error.",
      "suggestion": "Validate verb.name and command.name before sorting and throw a targeted error that names the producer payload path.",
      "confidence": 0.78,
      "faang_parallel": "Amazon service contracts commonly fail fast with field-path errors because generic parser exceptions slow down both producers and on-call responders.",
      "metaphor": "Sorting before validating is alphabetizing forms before checking whether the name field was filled in.",
      "teachable_moment": "Normalize after validation. Error messages are part of the API surface for every producer contract.",
      "connection": "This connects operational readiness with developer experience: precise failures make governance adoption easier to debug."
    },
    {
      "id": "praise-003",
      "title": "Markdown receipt rendering treats source prose as untrusted",
      "severity": "PRAISE",
      "category": "security",
      "file": "tools/markdown-text.mjs:5",
      "description": "The shared escaping helper and tests cover forged headings, links, inline HTML, and list injection in generated receipts, which is important because these receipts render untrusted product and producer text.",
      "suggestion": "Keep all future receipt renderers on this helper or an equivalent structured renderer.",
      "confidence": 0.9,
      "faang_parallel": "GitHub’s own Markdown rendering pipeline is defensive because user-authored text can reshape review context if treated as trusted markup.",
      "metaphor": "This helper is a turnstile: prose can enter the document, but it cannot grab the microphone and become structure.",
      "teachable_moment": "Escaping is not just XSS prevention. In governance documents, it also protects the integrity of headings, links, and evidence boundaries.",
      "connection": "This is the security counterpart to the schema work: untrusted inputs stay data, not instructions."
    }
  ]
}
```


## Callouts

[Security] The Markdown escaping work is especially strong because it recognizes document-structure injection, not only browser-script injection. That is the right threat model for generated governance receipts.

[Quality] The schema-plus-semantic-validator split is a mature design choice. It keeps the contract portable while allowing repository policy to evolve without overloading JSON Schema into a rule engine.

[Operational] The immutable SHA check in the reusable workflow is a high-leverage supply-chain guard. It makes downstream adoption safer because consumers can review the exact validator they are inheriting.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`

This PR is architecturally thoughtful: it separates product intent, system components, schemas, and operator-facing receipts in a way that can scale. The main blocker is cross-repo validation: the reusable workflow appears to validate consumer flow records against the governance checkout’s component registry, which can silently reject valid consumer records or accept the wrong universe of components.

The deeper pattern here is authority by location. In distributed governance systems, “where am I resolving truth from?” is as important as “is the schema valid?”

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Repo-local flow validation resolves components against governance checkout",
      "severity": "HIGH",
      "category": "correctness",
      "file": "tools/flow-moment.mjs:48",
      "description": "collectSystemComponentIds defaults to DEFAULT_SYSTEM_COMPONENTS_PATH derived from the validator repository root, and semanticErrors calls it without a repoRoot or target. In the reusable workflow, `node .flow-governance/tools/flow-moment.mjs validate product/flow-moments` runs from the consumer repository but the module ROOT is `.flow-governance`, so local `component:` refs in consumer-owned flow records are checked against `.flow-governance/product/system-components` instead of the consumer repo.",
      "suggestion": "Thread a repoRoot or systemComponentsPath through validateFlowMoment/semanticErrors and have the CLI pass process.cwd() for cross-repository validation.",
      "confidence": 0.92,
      "faang_parallel": "Google's monorepo tooling draws a hard line between tool location and package/workspace root; Blaze/Bazel actions make the execution root explicit because implicit path authority creates non-reproducible builds.",
      "metaphor": "This is like asking a visiting building inspector to check your house, but they accidentally compare every room against the blueprint for their own office.",
      "teachable_moment": "Reusable validators should treat the target repository as data authority and the validator checkout as implementation authority. Those are different roots and should be named separately in code.",
      "connection": "This connects to a broad governance pattern: schema validation proves shape, but semantic validation must resolve references against the correct trust domain."
    },
    {
      "id": "F002",
      "title": "Reusable workflow installs governance dependencies without Node cache protection",
      "severity": "LOW",
      "category": "ci",
      "file": ".github/workflows/reusable-flow-moment-governance.yml:38",
      "description": "The reusable workflow checks out the consumer first and the governance repo under `.flow-governance`, then installs inside `.flow-governance`. Unlike the local workflow, setup-node does not enable pnpm caching here. This is not a functional blocker, but repeated consumer runs will pay the full install cost.",
      "suggestion": "Either add setup-node `cache: pnpm` with `cache-dependency-path: .flow-governance/pnpm-lock.yaml`, or keep it uncached intentionally and document that choice.",
      "confidence": 0.68,
      "faang_parallel": "Meta's large-scale CI practice treats dependency cache keys as part of developer experience infrastructure; slow validation loops reduce how often teams run the checks that protect quality.",
      "metaphor": "The workflow works, but it is repacking the same suitcase for every short trip.",
      "teachable_moment": "CI performance choices are product choices for maintainers. If uncached installs are intentional for freshness or isolation, leave a breadcrumb so future agents do not “fix” it blindly.",
      "connection": "This is operational hygiene rather than correctness: governance checks should be cheap enough that downstream repos adopt them without resentment."
    },
    {
      "id": "F003",
      "title": "Markdown receipt rendering consistently escapes untrusted prose",
      "severity": "PRAISE",
      "category": "security",
      "file": "tools/markdown-text.mjs:5",
      "description": "The shared escapeMarkdownText helper neutralizes Markdown headings, links, lists, and inline HTML before rendering untrusted record and producer fields into operator receipts. The tests also cover forged headings and HTML/link injection attempts across flow, system, and construct renderers.",
      "suggestion": "Keep using this shared helper for any future receipt renderer instead of adding local escaping variants.",
      "confidence": 0.95,
      "faang_parallel": "GitHub's own Markdown rendering pipeline relies on centralized sanitization because scattered one-off escaping eventually misses an edge case.",
      "metaphor": "This is a single customs checkpoint for prose before it enters the operator interface.",
      "teachable_moment": "The safest rendering pattern is boring and centralized: untrusted text goes through one well-tested path before it can become presentation.",
      "connection": "This is especially important for governance receipts, where forged headings or links can make untrusted content look like system authority."
    },
    {
      "id": "F004",
      "title": "Authority projection fails closed on malformed producer claims",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "tools/construct-operator.mjs:169",
      "description": "normalizeAuthority checks ceiling, earned authority, and effective authority independently, rejects effective authority above the territory ceiling or earned evidence, and collapses unknown earned authority to observe. This directly protects the operator surface from promoting prose or incomplete stationing into gate authority.",
      "suggestion": "Maintain these checks as semantic validation rather than moving them into display-only code.",
      "confidence": 0.93,
      "faang_parallel": "AWS IAM policy evaluation is valuable here as a parallel: effective permission is computed from explicit constraints, and absence of valid authority does not become implicit grant.",
      "metaphor": "This behaves like a badge reader that defaults to the lobby when the credential is malformed, not the server room.",
      "teachable_moment": "Authority should be computed, bounded, and auditable before display. Presentation code should report authority, not decide it.",
      "connection": "This is the strongest design move in the PR: it treats governance as a semantic contract, not just a nicely formatted document."
    }
  ]
}
```


## Callouts
[Security] The Markdown escaping work is genuinely strong because it recognizes that operator-facing receipts are an authority surface. Preventing forged headings, links, and inline HTML protects the human reader, not just the parser.

[Quality] The authority normalization design deserves to stay close to validation. It gives future maintainers a clear decision trail: ceiling, earned authority, and effective authority are distinct concepts with distinct failure behavior.

[Test Coverage] The tests appear to cover hostile prose across multiple renderers, which is exactly the right pattern. Security helpers earn trust when they are tested at every consumer boundary, not only in isolation.

[Operational] The reusable workflow structure is a good foundation. Once the target repo root is made explicit, this becomes a portable governance check rather than a repo-local script wearing a reusable-workflow jacket.

zkSoju (commented):
## Summary

Verdict: `COMMENT`

This PR appears to be moving governance from implicit convention into explicit, testable contracts: schemas define portable shape, semantic validators enforce promotion rules, and workflows guard supply-chain drift. The analytical pass found no blocking issues; the strongest signal is architectural discipline around treating governance artifacts as executable contracts rather than prose-only documentation.

The larger pattern here is healthy: schema, validation, receipt rendering, and workflow enforcement each own a different layer of trust. That separation is the same principle behind mature platform systems: make the easy path auditable, make dangerous ambiguity hard to introduce silently.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "praise-001",
      "title": "Validators separate schema shape from semantic promotion gates",
      "severity": "PRAISE",
      "category": "validation",
      "file": "tools/flow-moment.mjs:55",
      "description": "The implementation uses JSON Schema for portable shape validation, then applies semantic checks for duplicate IDs, evidence requirements, exposure promotion, local component resolution, and Gold maturity. This keeps the contract strict without overloading the schema layer.",
      "suggestion": "Keep this split as the contract evolves; add semantic tests whenever a new promotion invariant is introduced.",
      "confidence": 0.95,
      "faang_parallel": "Google's Protocol Buffers ecosystem follows this same split: schemas describe wire shape, while higher-level services enforce business invariants and rollout policy outside the IDL.",
      "metaphor": "The schema is the passport format; the semantic validator is the border officer checking whether the trip itself is allowed.",
      "teachable_moment": "A schema should answer 'is this shaped correctly?' not 'is this organizationally wise to promote?' Keeping those questions separate makes both layers easier to test and evolve.",
      "connection": "This connects to the broader 'schema is not the contract' pattern: durable systems pair structural validation with explicit behavioral invariants."
    },
    {
      "id": "praise-002",
      "title": "Receipt rendering treats source prose as untrusted content",
      "severity": "PRAISE",
      "category": "security",
      "file": "tools/markdown-text.mjs:5",
      "description": "Derived Markdown receipts escape headings, links, inline HTML, and list syntax from record fields before rendering. The tests cover forged Markdown sections and link injection attempts across flow, system-component, and construct receipts.",
      "suggestion": "Continue routing all record-controlled prose and references through this helper before rendering Markdown.",
      "confidence": 0.95,
      "faang_parallel": "React's default HTML escaping embodies this practice at web scale: user-controlled text is inert unless a caller explicitly opts into dangerous rendering.",
      "metaphor": "This is like putting visitor notes behind glass in a museum display: people can read them, but they cannot rearrange the exhibit.",
      "teachable_moment": "Generated Markdown is still an output surface. If records can become headings, links, or HTML, then data has quietly become control flow.",
      "connection": "This is a strong trust-boundary pattern for governance receipts, changelogs, bot comments, and any artifact later consumed by humans or agents."
    },
    {
      "id": "praise-003",
      "title": "Reusable workflow pins governance by immutable commit",
      "severity": "PRAISE",
      "category": "supply-chain",
      "file": ".github/workflows/reusable-flow-moment-governance.yml:23",
      "description": "The reusable governance workflow rejects non-SHA refs before checking out the validator repository, which prevents consumers from silently moving their governance contract via a branch or tag.",
      "suggestion": "Preserve the full-SHA requirement for all cross-repository governance callers.",
      "confidence": 0.9,
      "faang_parallel": "Google's SLSA guidance and internal build provenance practices emphasize immutable, verifiable inputs for exactly this reason: moving refs weaken reproducibility.",
      "metaphor": "A branch name is a street address that can be reassigned; a commit SHA is a fingerprint.",
      "teachable_moment": "Governance code is part of the trusted computing base. If it can move without review, the policy being enforced can change without the product changing.",
      "connection": "This aligns CI governance with supply-chain hardening: pin the thing that decides trust, then update it through an intentional review path."
    }
  ]
}
```


## Callouts

[Quality] The validator layering is excellent because it leaves room for the contract to grow. New maturity rules, evidence gates, or construct-resolution checks can be added as semantic tests without making the JSON Schema unreadable or non-portable.

[Security] The Markdown escaping helper is the kind of small shared primitive that prevents an entire class of downstream problems. It recognizes that receipts are not just documentation; they are rendered interfaces and future agent inputs.

[Operational] The reusable workflow’s SHA pinning is especially strong governance hygiene. It gives future maintainers a clean decision trail: policy updates become visible dependency changes instead of invisible branch movement.

The architectural throughline is trust minimization: every layer receives input, narrows it, and emits something more constrained. That is the right instinct for governance tooling.

Issue #472, opened 2026-07-16
[coord] collection-report Collection recognition and asynchronous report fulfillment — see 0xHoneyJar/freeside-dashboard


zkSoju:
## Cross-repo coordination — Collection recognition and asynchronous report fulfillment

**Coordinator**: 0xHoneyJar/freeside-dashboard (cycle `collection-report`)
**This repo's role**: Owns shared protocol, Ordering, Shadow Audit, authorization, trust envelopes, persistence, generation, projections, rollout, privacy, and operational gates.
**Master plan content hash (at issue-open time)**: `sha256:05a8539467ca1b61aa2916445d29ecf5141f5ba4484294c03459f8173b793e80`

### What this issue is

This is a **coordination tracker** for Collection recognition and asynchronous report fulfillment work that lives in this repo. It is not itself a piece of work — it is an index pointing at the master plan and tracking aggregate progress.

The actual work is broken down into per-task PRs. Each PR description should include:

```
Closes 0xHoneyJar/freeside-dashboard#<coord-issue-num> (master task A.B)
```

so the coordinator's `bin/coord-sync.sh` can correlate upstream merges back to master beads.

### Tasks owned by this repo

- [ ] **Master Task ACCEPT-LOA** (sprint loa-acceptance) — Loa platform owners accept shared protocol, Ordering, Shadow Audit, privacy, deployment, capacity, rollback, and operations boundaries before CR work starts
- [ ] **Master Task CR-000** (sprint loa-foundation) — Discord restricted-tier viability go/no-go
- [ ] **Master Task CR-001** (sprint loa-foundation) — Cross-VM collection identity contract
- [ ] **Master Task CR-002** (sprint loa-foundation) — Ratify Gate Leak readiness and work-key scopes
- [ ] **Master Task CR-005** (sprint loa-foundation) — Cross-repository contract harness
- [ ] **Master Task CR-006** (sprint loa-foundation) — Durable resolution and stale-selection contract
- [ ] **Master Task CR-007A** (sprint loa-foundation) — Public resolution, order, and demand authorization
- [ ] **Master Task CR-007B** (sprint loa-foundation) — Restricted artifact access, retention, and deletion policy
- [ ] **Master Task CR-009** (sprint loa-foundation) — Signed trust-envelope protocol
- [ ] **Master Task CR-010** (sprint loa-foundation) — Hierarchical artifact-manifest contract
[...]

PR #475, opened 2026-07-16
feat: define cross-VM collection identity protocol (CR-001)

zkSoju:
## Summary

- adds the versioned cross-VM collection identity protocol for EVM and Solana
- defines stable normalized deployment and logical collection digests
- binds finality policy into candidate, work, evidence, cache, and report inputs
- adds capability-registry epoch/sequence semantics and strict Effect boundary schemas
- publishes positive, malformed, digest, Dashboard-shaped, and Sonar-shaped fixtures

## Verification

- `pnpm typecheck`
- `pnpm test` — 22 tests
- `pnpm clean && pnpm build`
- independent Fable review: PASS
- independent SHA-256/JCS vector reproduction

## Coordination

- Task: CR-001
- Coordinator epic: `collection-report-coordinator-f09`
- Aggregate issue: #472

This PR is intentionally draft. It must not be merged until the coordinated collection-report gates permit it.


zkSoju:
CI baseline note: the two red gates are reproduced on pre-feature main. Main CI run 29392769035 fails the same cross-system conservation assertion (`CLAIMS_SCHEMA` vs `OVERSPEND`), and main Agent Subsystem run 28840198320 fails the exact same `agent-gateway.test.ts` pair (per-user limit and SSE `UND_ERR_SOCKET`). The collection protocol package does not touch those paths. I have requested one failed-job rerun to distinguish transient recovery from the known baseline.

zkSoju:
Rerun result: both failed jobs reproduced. Integration attempt 2 still includes the baseline `CLAIMS_SCHEMA`/`OVERSPEND` failure (plus other existing full-suite failures); agent-ci attempt 2 again failed the exact per-user assertion and SSE `UND_ERR_SOCKET`. This is now classified as an external baseline gate, not a feature regression. No unrelated CI code will be folded into this branch.

zkSoju:
Tracked the reproduced required-check baseline separately as #477. This PR remains scoped to CR-001.

zkSoju (commented):
## Summary
Verdict: COMMENT. This PR has the right architectural instinct: define identity as a canonical protocol surface, then defend it with fixtures and consumer tests. The main concerns are narrow but important: exported helpers should protect runtime callers with the same rigor as schema boundaries, and digest integrity checks should stay whole-object by habit.

The larger pattern here is protocol durability. Cross-VM identity will outlive today’s TypeScript types, so every public helper is part of the wire contract, not just convenience code.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Public digest helper can emit structurally invalid VersionedDigest values",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/protocol/collection/src/canonical.ts:234",
      "description": "digestVersioned accepts domain and majorVersion as already-valid typed values, but it does not runtime-decode them before returning a VersionedDigest. JavaScript callers or TypeScript callers using unchecked values can produce digests whose domain or major_version fail the exported VersionedDigest schema.",
      "suggestion": "Decode the inputs with DigestDomain and an integer >= 1 schema before canonicalizing, or return a typed validation error when domain/majorVersion are invalid.",
      "faang_parallel": "Google’s Protocol Buffers practice treats generated types as helpful, but still validates at service boundaries because wire data and runtime callers can bypass compile-time guarantees.",
      "metaphor": "This is like printing an official passport without checking the date of birth field first; the seal looks real, but the document may not pass the next border.",
      "teachable_moment": "In protocol packages, exported helpers are boundary APIs. If they mint branded values, they should either validate or make invalid construction impossible.",
      "connection": "This connects to the broader schema-vs-runtime gap: TypeScript narrows authoring mistakes, but published packages still serve JavaScript, tests, migrations, and future agents.",
      "decision_trail": "Document whether public helpers are trusted-internal constructors or boundary-safe constructors. Future maintainers need to know where validation is guaranteed.",
      "confidence": 0.86
    },
    {
      "id": "F002",
      "title": "Registry version advancement can throw outside Effect for malformed inputs",
      "severity": "LOW",
      "category": "error-handling",
      "dimension": "[Operational]",
      "file": "packages/protocol/collection/src/registry-version.ts:146",
      "description": "advanceCapabilityRegistryVersion calls BigInt directly on current.registry_sequence and candidate.registry_sequence. If this exported function is called from JavaScript or with unchecked boundary data instead of decoded CapabilityRegistryVersion values, malformed sequences throw synchronously instead of failing through the declared Effect error channel.",
      "suggestion": "Decode current and candidate with decodeCapabilityRegistryVersion inside the function, or wrap BigInt conversion in Effect.try and include parse failures in the function's typed error surface.",
      "faang_parallel": "Netflix’s Hystrix-era resilience work popularized the rule that failures should travel through expected channels; surprise synchronous throws bypass bulkheads and observability.",
      "metaphor": "The Effect error channel is the marked return lane, but BigInt can open a trapdoor beside it.",
      "teachable_moment": "Once an API advertises typed effects, every fallible operation inside it should honor that contract, including primitive parsing.",
      "connection": "This is an operational consistency issue more than a correctness flaw: consumers can recover from declared failures, but uncaught throws often become crashes or noisy incident paths.",
      "decision_trail": "Record whether these functions require pre-decoded inputs. If yes, name them accordingly; if no, make decoding part of the implementation.",
      "confidence": 0.73
    },
    {
      "id": "F003",
      "title": "Registry baseline integrity check should compare the full digest object",
      "severity": "LOW",
      "category": "correctness",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/registry-version.ts:111",
      "description": "decodeCapabilityRegistryBaseline validates the expected baseline by comparing only expected.digest to baseline.baseline_digest.digest. The schema currently filters domain and major_version and VersionedDigest fixes the algorithm, so this is not exploitable through the present decoder, but it is weaker than the digest integrity checks used elsewhere and easy to regress if the schema changes.",
      "suggestion": "Use the same full digest comparison pattern as contracts.ts, comparing algorithm, domain, major_version, and digest.",
      "faang_parallel": "Amazon’s internal service contracts commonly compare full identity tuples rather than a single field, because partial equality becomes fragile when schemas evolve.",
      "metaphor": "Matching only the digest string is like checking a house key’s teeth while ignoring which building it belongs to.",
      "teachable_moment": "Integrity checks should compare the whole identity surface, even when today’s schema makes the extra fields redundant.",
      "connection": "This keeps the registry protocol future-proof as domains, algorithms, or version semantics expand.",
      "decision_trail": "Add a short comment or helper documenting the invariant: digest equality means algorithm, domain, major_version, and digest all match.",
      "confidence": 0.64
    },
    {
      "id": "F004",
      "title": "Strict boundary decoders and digest integrity checks are well covered",
      "severity": "PRAISE",
      "category": "test-coverage",
      "file": "packages/protocol/collection/src/__tests__/protocol.test.ts:30",
      "description": "The tests exercise strict fixture decoding, malformed boundary inputs, digest-vector stability, canonicalization edge cases, registry sequencing, and finality-policy propagation. This gives the new wire protocol meaningful regression coverage at the contract boundary.",
      "suggestion": "Keep these fixture tests as required publication checks for future protocol revisions.",
      "faang_parallel": "Meta’s Thrift and Google’s protobuf ecosystems both rely on golden fixtures and compatibility tests to keep independently deployed consumers from drifting.",
      "metaphor": "These fixtures are survey stakes in the ground; future builders can extend the property line without guessing where it started.",
      "teachable_moment": "Protocol tests are strongest when they encode invalid examples, not only happy paths.",
      "connection": "This is the right foundation for a cross-VM contract because it tests the serialized boundary where real consumers will meet the system.",
      "decision_trail": "Document that fixture compatibility is part of the release gate for CR-001 revisions.",
      "confidence": 0.95
    }
  ]
}
```


## Callouts
[Test Coverage] The malformed fixture suite is unusually strong for a first protocol slice. It teaches future contributors what must be rejected, which is often more valuable than showing only valid examples.

[Quality] The separation across `canonical`, `contracts`, `registry-version`, and `scalars` gives the protocol a readable shape. That modularity will matter when other VMs or registry versions arrive.

[Operational] Consumer tests for dashboard and sonar are a good signal: this PR is not just defining an abstract schema, it is checking that real downstream shapes can depend on it.

zkSoju (commented):
## Summary

Verdict: `REQUEST_CHANGES`. The protocol shape is strong: digest integrity checks and decimal-string registry sequences show disciplined boundary design. The two medium issues are both hash-boundary problems, and they deserve tightening before this becomes a cross-VM identity contract.

Canonical protocols live or die at the boundary between “typed in memory” and “valid on the wire.” The broad lesson here is the same one behind systems like Bigtable, Borg, and Tink: canonical forms must be narrower than application objects, and digest domains must be explicit enough that a valid hash cannot mean the wrong thing.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Canonical digest encodes undeclared object properties",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/protocol/collection/src/contracts.ts:634",
      "description": "The digest helper functions accept already-typed values and pass them directly to digestVersioned without re-decoding with strict excess-property handling. If a caller constructs a value with extra enumerable properties after decoding or through a type assertion, those properties become part of the canonical digest even though they are outside the declared wire contract.",
      "suggestion": "Have each digestCollection* helper decode the material through its strict schema before hashing, or provide only decode-and-digest APIs that reject excess properties at the hashing boundary.",
      "faang_parallel": "Google's Protocol Buffers guidance treats canonical wire data as distinct from arbitrary in-memory objects for exactly this reason.",
      "metaphor": "This is like sealing an envelope after letting someone slip extra pages behind the signed form.",
      "teachable_moment": "Hashing is a trust boundary; normalize and reject before digesting, even for already-typed inputs.",
      "connection": "This connects to content-addressed storage and Merkle systems: the hash must represent the declared contract, not incidental object shape.",
      "decision_trail": "Document whether digest helpers accept raw application objects or only schema-clean protocol material."
    },
    {
      "id": "F002",
      "title": "Collection evidence digest accepts any digest domain",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/contracts.ts:427",
      "description": "CollectionEvidenceReference validates deployment_ids and finality_policies but leaves evidence_digest as a bare VersionedDigest. This means decodeCollectionEvidenceReference accepts unrelated domains such as collection.cache-key or fixture domains, while CollectionReportInput has a sorted evidence_digests set with the same loose domain behavior.",
      "suggestion": "Filter evidence_digest and evidence_digests to the intended evidence/provenance digest domains, or split the fields into explicitly named digest types if both domains are valid.",
      "faang_parallel": "Google Tink separates key types and primitives so a valid byte string in one domain cannot be mistaken for another.",
      "metaphor": "A passport number is valid only in the passport office; it should not also unlock a bank vault.",
      "teachable_moment": "A digest without a constrained domain is structurally valid but semantically ambiguous.",
      "connection": "Cross-VM identity needs anti-confusion properties; domain-specific digest types prevent accidental equivalence between unrelated artifacts.",
      "decision_trail": "Record the allowed evidence digest domains and why each belongs in report inputs."
    },
    {
      "id": "F003",
      "title": "Canonicalizer rejects valid JSON objects with non-plain prototypes",
      "severity": "LOW",
      "category": "compatibility",
      "dimension": "[Operational]",
      "file": "packages/protocol/collection/src/canonical.ts:40",
      "description": "canonicalizeValue only treats objects with Object.prototype or null prototype as records. Values decoded from other JSON-compatible sources, or class instances with enumerable JSON data, fail even though JSON.stringify would serialize them as JSON objects.",
      "suggestion": "If the boundary is intended to canonicalize JSON-compatible data, accept any non-null non-array object and enumerate own string keys, or document and test the plain-object-only restriction as part of the protocol.",
      "faang_parallel": "Bigtable-style APIs tend to publish exact accepted encodings because client ecosystems produce equivalent data through many runtimes.",
      "metaphor": "The form is filled out correctly, but rejected because it came from the wrong printer.",
      "teachable_moment": "Canonicalization rules are interoperability contracts; strictness is fine, but it must be intentional and tested.",
      "connection": "This is a compatibility edge, not a security flaw, but protocol consumers will feel it as unexplained rejection.",
      "decision_trail": "Document whether CR-001 requires plain JSON-decoded objects or broader JSON-compatible objects."
    },
    {
      "id": "F004",
      "title": "Digest integrity validation is consistently enforced on decoded identities",
      "severity": "PRAISE",
      "category": "correctness",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/contracts.ts:548",
      "description": "decodeCollectionDeploymentRef, decodeCollectionIdentity, and decodeCollectionCandidate recompute canonical digests and fail when provided identifiers do not match the canonical material. This prevents callers from trusting well-shaped but forged identifiers.",
      "suggestion": "Keep this validation pattern at every boundary that accepts persisted protocol identifiers.",
      "faang_parallel": "Certificate Transparency and Merkle-log systems use the same discipline: structure alone is not trust; recomputation is trust.",
      "metaphor": "This checks the wax seal against the letter, not just whether the seal looks official.",
      "teachable_moment": "Decoders for persisted identifiers should verify semantic integrity, not merely parse syntax.",
      "connection": "This is the core safety rail for cross-VM identity claims."
    },
    {
      "id": "F005",
      "title": "Registry sequence is safely represented as a bounded decimal string",
      "severity": "PRAISE",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/protocol/collection/src/scalars.ts:83",
      "description": "RegistrySequence rejects leading-zero, numeric, and uint64-overflow representations while comparison uses BigInt. This avoids lexical ordering bugs and unsafe JavaScript number precision loss.",
      "suggestion": "Maintain the decimal-string wire form for sequence values and keep malformed edge fixtures committed.",
      "faang_parallel": "Spanner-style distributed systems are careful about ordered identifiers because tiny representation mistakes become consistency bugs.",
      "metaphor": "The ruler is marked in exact ticks instead of stretchy rubber.",
      "teachable_moment": "When JavaScript precision is insufficient, make the wire format explicit and validate the edge cases.",
      "connection": "This gives registry ordering a portable contract across runtimes."
    }
  ]
}
```


## Callouts

[Security] The recompute-on-decode pattern is excellent because it makes forged-but-well-shaped identifiers fail at ingress, where the mental model is simplest.

[Quality] The registry sequence design is especially strong: decimal string on the wire, `BigInt` for comparison, and malformed fixtures for leading zeros and overflow. That is the kind of boring precision protocol code should aspire to.

[Test Coverage] The malformed fixture suite is doing real work here. It turns edge cases into executable doctrine, which is how future maintainers learn the contract without guessing.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. The protocol shape is strong: strict boundary decoding and digest recomputation show real care for cross-VM identity. The main concern is correctness at the contract edge: one test appears to assert Solana case behavior with an invalid base58 key, and packaging expectations for fixtures should be made explicit before this becomes a downstream dependency.

The architectural through-line here is identity determinism. Cross-VM protocols succeed when every consumer can independently arrive at the same bytes, the same digest, and the same rejection reasons.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Solana case-sensitivity test uses an invalid base58 key",
      "severity": "MEDIUM",
      "category": "test-correctness",
      "file": "packages/protocol/collection/src/__tests__/protocol.test.ts:137",
      "description": "The test expects makeCollectionDeploymentRef to accept the lower-cased Solana key `so111...`, but lowercase `o` is not in the Solana/base58 alphabet enforced by SolanaPublicKey. This makes the test fail before it can assert case-sensitive identity behavior.",
      "suggestion": "Use two distinct valid base58-encoded 32-byte Solana public keys whose difference is not produced by lowercasing an excluded character, or change the assertion to expect failure for the invalid lowercased key.",
      "confidence": 0.95,
      "faang_parallel": "Google's protobuf conformance suites separate lexical validity tests from semantic equivalence tests for this reason: one failing layer must not mask the behavior the test claims to prove.",
      "metaphor": "This is like testing whether two passports belong to different people by first using a passport number the border scanner cannot read.",
      "teachable_moment": "For protocol tests, keep validation fixtures and identity-behavior fixtures distinct. Otherwise the test suite can pass or fail for the wrong reason.",
      "connection": "This sits directly on the protocol's normalization boundary. If the suite cannot distinguish invalid encoding from valid-but-distinct identity, future agents may weaken the Solana validator to satisfy the wrong test.",
      "decision_trail": "Document whether Solana public keys are treated as opaque base58 identities with no case normalization. Add a short test name or comment making that rule explicit."
    },
    {
      "id": "F002",
      "title": "Published package omits fixture files from build output",
      "severity": "MEDIUM",
      "category": "packaging",
      "dimension": "[Operational]",
      "file": "packages/protocol/collection/package.json:14",
      "description": "`files` includes `fixtures`, but fixtures live at the package root while `tsconfig.json` only emits `src` into `dist`. Consumers importing the package get runtime code, but any tests or downstream fixture contract checks that rely on package-installed fixtures need the root fixture tree to be intentionally included and path-stable.",
      "suggestion": "Confirm the root `fixtures` directory is included in the packed artifact with `pnpm pack --dry-run`, and document or export fixture paths if consumers are expected to use them.",
      "confidence": 0.6,
      "faang_parallel": "Meta's Buck/Bazel-style package discipline treats golden files as API when consumers depend on them; they are not incidental repo debris once published.",
      "metaphor": "The fixtures are the measuring blocks shipped with the instrument. If they stay in the workshop, every buyer calibrates differently.",
      "teachable_moment": "Packaging is part of the protocol surface. A fixture promised by `files` should be verified the same way exported TypeScript declarations are verified.",
      "connection": "Because this package defines cross-consumer identity rules, fixture availability affects dashboard, sonar, and future VM adapters as much as code availability.",
      "decision_trail": "Decide and document whether fixtures are public contract artifacts. If yes, add a pack dry-run check or exported fixture path convention; if no, remove them from the package surface."
    },
    {
      "id": "F003",
      "title": "Protocol decoders validate digest integrity at trust boundaries",
      "severity": "PRAISE",
      "category": "correctness",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/contracts.ts:503",
      "description": "Deployment, identity, and candidate decoders recompute canonical digests instead of only checking digest shape. This prevents well-formed but stale or forged identifiers from passing boundary validation.",
      "suggestion": "Keep this behavior as the shared consumer path and avoid adding alternate decoders that skip integrity checks.",
      "confidence": 0.95,
      "faang_parallel": "Amazon's Dynamo paper leans on content-derived identifiers and validation at boundaries; shape alone is never enough when replicas and clients can drift.",
      "metaphor": "This is checking both the envelope address and the wax seal. A forged-looking seal does not pass just because it is round.",
      "teachable_moment": "For identity protocols, recomputation beats trust. If a digest can be derived, the boundary should derive it and compare.",
      "connection": "This is the strongest architectural move in the PR because it makes canonicalization executable, not merely documented.",
      "decision_trail": "Record that exported decoders are integrity-enforcing boundary APIs. Future relaxed decoders, if any, should be named as unsafe/internal."
    },
    {
      "id": "F004",
      "title": "Strict excess-property decoding is consistently applied",
      "severity": "PRAISE",
      "category": "security",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/contracts.ts:466",
      "description": "The public decode helpers use `onExcessProperty: \"error\"`, and the digest helpers decode before hashing. This reduces schema drift and prevents hidden fields from being silently accepted at protocol boundaries.",
      "suggestion": "Preserve strict decode options for future exported boundary helpers.",
      "confidence": 0.9,
      "faang_parallel": "Google's API Improvement Proposals emphasize explicit unknown-field behavior because silent acceptance creates long-lived compatibility ambiguity.",
      "metaphor": "This is a customs checkpoint that rejects undeclared luggage instead of waving it through because the suitcase shape looks familiar.",
      "teachable_moment": "Strict decoding turns schema design into an enforceable contract. That is especially valuable before multiple clients learn different interpretations.",
      "connection": "Together with digest recomputation, this closes the two common protocol gaps: hidden data and stale derived data.",
      "decision_trail": "Document strict excess-property rejection as part of CR-001 compatibility, so later consumers do not treat it as an implementation detail."
    }
  ]
}
```


## Callouts
[Security] The digest validation path is notably strong because it makes forgery resistance the default consumer experience, not an optional helper.

[Quality] The malformed fixture set is a good investment. It gives future maintainers executable examples of what the protocol refuses, which is often more valuable than only documenting valid shapes.

[Operational] The dashboard and sonar consumer tests are a healthy signal: this protocol is being tested from the perspective of downstream readers, not only its authors.

zkSoju:
Current-head Bridgebuilder adjudication (`8c3af5ad7476e3f319477270d9e790e5b1bbb0f2`): the two requested changes are false positives, so no code change is warranted.

- The Solana comparison uses `So...` versus `so...`. Lowercase `o` is valid Base58; only `0`, uppercase `O`, uppercase `I`, and lowercase `l` are excluded. Both values pass the protocol decoder, and the test correctly proves case-sensitive identity.
- `packages/protocol/collection/package.json` explicitly publishes `fixtures`, and `pnpm pack --dry-run --json` at this exact head reports 15 `fixtures/` entries. Fixtures are package-root runtime assets; they are not expected under `dist`.

The digest-boundary and excess-property praise remains applicable. This comment adjudicates the review findings only; it does not authorize merge or release.

Issue #477, opened 2026-07-16
CI: repair deterministic integration and agent-gateway baseline failures


zkSoju:
## Problem

Required checks on `main` and unrelated feature branches are deterministically red, preventing trustworthy regression attribution.

### Reproduced on main

- CI run https://github.com/0xHoneyJar/loa-freeside/actions/runs/29392769035: `cross-system-conservation.test.ts` expects `OVERSPEND` but receives `CLAIMS_SCHEMA`, alongside additional full-suite baseline failures.
- Agent Subsystem run https://github.com/0xHoneyJar/loa-freeside/actions/runs/28840198320: `agent-gateway.test.ts` fails the per-user free-tier assertion and SSE drop recovery with `UND_ERR_SOCKET`.

### Independent reproduction

Draft PRs #475 and #476 touch separate new protocol packages, yet both produced the same failures on initial runs and on failed-job reruns (attempt 2). Neither branch touches the failing paths.

## Impact

- Required checks cannot distinguish feature regressions from baseline defects.
- Collection-report protocol PRs #475 and #476 remain red despite focused package tests and all other checks passing.

## Acceptance

- Fix or correctly quarantine the `CLAIMS_SCHEMA` vs `OVERSPEND` fixture/contract mismatch.
- Make per-user rate-limit tests isolated and deterministic.
- Make SSE drop recovery assert the intended reconciliation behavior without leaking the expected socket termination as an uncaught test failure.
- Run the affected workflows twice on `main` with both runs green.
- Preserve fail-closed production behavior; do not weaken runtime controls merely to satisfy tests.

## Scope boundary

This issue tracks baseline CI health only. It must not be folded into the collection protocol feature branches.

zkSoju:
### Triage provenance

**Record type:** Classification provenance  
**Effect:** Audit trail only. No work was dispatched, approved, closed, or resolved by this comment.

| Classification | Applied value |
| --- | --- |
| Workstream | `workstream:tech-debt` |
| Artifact type | `artifact-type:bug-report` |
| Priority | `priority:high` |
| Route | `triage:bug-queued` |

| Grounding | Value |
| --- | --- |
| Triaged at | `2026-07-16T16:06:28Z` |
| Repository revision | [`3782fd4`](https://github.com/0xHoneyJar/loa-freeside/commit/3782fd47e8a20cdaf6325621962bd0443e6781b8) |

**Evidence basis**

[...]

PR #478, opened 2026-07-16
feat: ratify Gate Leak protocol

zkSoju:
## What changed

- adds the Gate Leak readiness, capability, evidence, classification, mapping, migration, and work-key protocol package
- derives readiness from six decoded evidence capabilities and a server-recomputed compute-input digest
- enforces a single active immutable gate mapping with cryptographic configuration and reveal-evidence integrity
- rejects empty or duplicate deployment/role sets and canonicalizes equivalent permutations
- binds community-scoped consent, invalidation, retention, authorization, and tombstone watermarks into restricted evidence
- binds Discord snapshot work to the exact ratified mapping version and configuration

## Why

Gate Leak admission and computation need a shared fail-closed contract before Ordering, Identity, Shadow Audit, and report workers can exchange evidence safely. This package makes stale, incomplete, replayed, ambiguously scoped, or structurally valid-but-unproven inputs explicit typed failures.

## Validation

- Gate Leak typecheck and build
- Gate Leak tests: 124/124
- CR-001 collection protocol regression: 22/22 and typecheck
- deterministic fixture regeneration and golden assertions
- independent adversarial review

## External evidence gate

The three recorded-shape community fixtures are synthetic and do **not** satisfy the three-live-community G1 evidence requirement. The implementation can be reviewed, but CR-002 and G1 remain open until operator-supplied exports from three live communities are replayed through these contracts.

## Coordination

- Master task: CR-002
- Aggregate issue: #472
- Stacked on CR-001 branch `coord/collection-report-coordinator-f09.11`

Draft only. Do not merge.


zkSoju:
CI note: `agent-ci` is red on the same pre-existing `tests/integration/agent-gateway.test.ts` failures tracked in #477 (per-user rate-limit expectation and SSE drop recovery `UND_ERR_SOCKET`). The CR-002 Gate Leak package is independently green (124/124; CR-001 22/22), and all other PR checks pass. No unrelated baseline CI fix is included in this draft.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. The protocol work shows strong integrity instincts and unusually good contract-test coverage, but F001 leaves readiness as the one path that does not apply the same aggregate-envelope discipline as transition code. For a ratified protocol, the bigger theme is consistency: every consumer of signed or denormalized state should pay the same verification tax.

Readiness is the user-facing verdict surface. That makes it less like a helper and more like a checkpoint: if it can be reached with malformed aggregate history, future operators may trust a “ready” answer that bypassed the protocol’s own audit envelope.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Readiness does not verify the mapping aggregate envelope",
      "severity": "MEDIUM",
      "category": "integrity",
      "file": "packages/protocol/gate-leak/src/readiness.ts:178",
      "description": "evaluateGateLeakReadiness consumes context.mapping_aggregate directly, then calls mappingSatisfiesReadiness and verifies only the selected active version. Unlike ratify/revoke and decodeGateMappingAggregate, this path does not strict-decode the aggregate envelope or verify every stored version/audit record, so in-process malformed aggregate state with invalid envelope fields, excess fields, or tampered inactive history can still reach a ready verdict if the active version verifies.",
      "suggestion": "Run the mapping aggregate through the same strict aggregate verification used by decodeGateMappingAggregate or transition paths before readiness evaluation, and return gate_mapping_malformed or mapping_integrity_violation on failure.",
      "confidence": 0.82,
      "dimension": "[Security]",
      "faang_parallel": "Google's BeyondCorp model treats every access decision as a fresh verification point, not a trust handoff from an earlier layer.",
      "metaphor": "This is checking the passport photo but not whether the passport book was forged.",
      "teachable_moment": "Integrity checks should sit at every trust boundary, especially at read-time verdict surfaces.",
      "connection": "This connects directly to F005: the transition path has the right pattern, but readiness has not inherited it."
    },
    {
      "id": "F002",
      "title": "Ordering of refusal reason codes is nondeterministic by insertion side effects",
      "severity": "LOW",
      "category": "testability",
      "file": "packages/protocol/gate-leak/src/readiness.ts:484",
      "description": "Readiness accumulates reason codes in a Set and returns [...reasons]. The order depends on evaluation order across a long function, so adding or moving checks changes the public array order even when the semantic refusal set is identical. Several tests use containment, but one path asserts exact equality for empty_deployment_selection only because it returns early.",
      "suggestion": "Return reason codes sorted by a fixed registry order or define the API as an unordered set and keep tests/assertions aligned with that contract.",
      "confidence": 0.66,
      "dimension": "[Quality]",
      "faang_parallel": "Meta's GraphQL APIs commonly stabilize response shapes because clients accidentally depend on ordering even when the schema does not promise it.",
      "metaphor": "The reasons are a checklist; shuffling the checklist should not change what the inspector found.",
      "teachable_moment": "Public arrays are contracts unless explicitly documented otherwise.",
      "connection": "This is small now, but protocol consumers often turn incidental ordering into brittle downstream logic."
    },
    {
      "id": "F003",
      "title": "Readiness returns partial coverage disclosure after proving finality over partial coverage",
      "severity": "SPECULATION",
      "category": "semantics",
      "file": "packages/protocol/gate-leak/src/readiness.ts:236",
      "description": "missingOwnership treats coverage entries with completeness: partial as missing, but evaluateOwnershipFinalityProof still accepts the same partial entry if its source position and timestamp match the attestation. With partial_policy disclose, readiness can return ready with a disclosed gap while finality proof has also treated the partial coverage entry as source-bound proof.",
      "suggestion": "Clarify whether partial coverage should participate in finality proof at all. If not, pass only complete coverage into finality proof or make evaluateOwnershipFinalityProof reject partial entries for selected deployments.",
      "confidence": 0.42,
      "dimension": "[Quality]",
      "faang_parallel": "Amazon's two-pizza service reviews force teams to document ambiguous state semantics before they become API behavior.",
      "metaphor": "This is stamping a form final while one box says incomplete.",
      "teachable_moment": "When two validators interpret the same field differently, document the intended lattice before patching code.",
      "connection": "This may be valid design, but the decision trail needs to say whether partial evidence is proof, disclosure, or both."
    },
    {
      "id": "F004",
      "title": "Strong regression coverage for canonical set and digest invariants",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/protocol/gate-leak/src/__tests__/canonical-sets.test.ts:26",
      "description": "The new tests cover permutation-invariant digests, sorted-unique decode, duplicate refusal, wrong-domain refusal, and in-process bypasses for deployment sets, role sets, disclosure keys, and authorization watermarks. This directly protects the work-key and privacy-ledger fragmentation risks.",
      "suggestion": "Keep these tests as contract tests for any future key-schema evolution.",
      "confidence": 0.93,
      "faang_parallel": "Spanner-style invariants survive because canonicalization and identity rules are tested as contracts, not examples.",
      "metaphor": "These tests are calibration weights for the scale.",
      "teachable_moment": "Digest code deserves adversarial fixture tests because tiny ordering drift becomes identity drift.",
      "connection": "This coverage is the strongest argument that the protocol is being built as infrastructure, not just application code."
    },
    {
      "id": "F005",
      "title": "Good integrity-first transition design",
      "severity": "PRAISE",
      "category": "integrity",
      "file": "packages/protocol/gate-leak/src/mapping.ts:601",
      "description": "ratifyGateMapping and revokeGateMappingVersion verify the stored aggregate before idempotency lookup or transition branching, which prevents replay or CAS behavior from trusting tampered denormalized digest fields.",
      "suggestion": "Apply the same aggregate verification discipline to readiness evaluation for consistency.",
      "confidence": 0.9,
      "dimension": "[Security]",
      "faang_parallel": "Borg and Kubernetes controllers both re-read and reconcile observed state before acting, because cached intent is not authority.",
      "metaphor": "This is measuring the bridge before driving the truck across it.",
      "teachable_moment": "State transition code should verify before branching; this PR does that well.",
      "connection": "Use this as the house style for all Gate Leak verdict and transition paths."
    }
  ]
}
```


## Callouts
[Security] The ratify/revoke path is thoughtfully defensive: verifying aggregate state before idempotency or CAS logic blocks a whole class of replay-by-denormalization bugs.

[Test Coverage] The canonical-set tests are excellent because they defend invariants, not implementation trivia. That is the right shape for protocol code: future refactors can move freely while identity, sorting, uniqueness, and domain boundaries stay pinned.

[Quality] The decision documentation investment is visible. The remaining ask is to make the ambiguous partial-coverage semantics equally explicit so future agents can preserve intent instead of reverse-engineering it.

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#478. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "GL-001",
      "title": "Gateway sequence pin is not bound to Discord evidence",
      "severity": "HIGH",
      "category": "integrity",
      "file": "packages/protocol/gate-leak/src/readiness.ts:316",
      "description": "Readiness compares `discord.attestation.gateway_epoch` with `context.pins.gateway_epoch`, but never compares `discord.attestation.gateway_resume_sequence` with `context.pins.gateway_sequence`. Evidence from the same gateway epoch but a different resume sequence can satisfy readiness while the compute input remains bound to the old sequence pin.",
      "suggestion": "Require `String(discord.attestation.gateway_resume_sequence) === context.pins.gateway_sequence` during readiness and add a regression test that drifts only the sequence.",
      "confidence": 0.93
    },
    {
      "id": "GL-002",
      "title": "Deployment scope comparison is order-sensitive",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/protocol/gate-leak/src/readiness.ts:151",
      "description": "`deploymentRefDigestsEqual` compares `collection_identity.deployments` and `selected_deployments` by array index. The selected deployment set is otherwise canonicalized for digesting, so the same logical set in a different order can be rejected with `evidence_scope_mismatch`.",
      "suggestion": "Canonicalize both deployment ref arrays by `deployment_id.digest` before comparing, or compare by digest-to-network maps instead of positional arrays.",
      "confidence": 0.86
    },
    {
      "id": "GL-003",
      "title": "Duplicate wallet deployment entries make classification order-dependent",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/protocol/gate-leak/src/compute.ts:134",
      "description": "`LinkedWalletEvidence.deployments` only requires a non-empty array, while classification converts it to a `Map` keyed by deployment digest. Duplicate entries for the same deployment silently use the last value, so input order can change a subject from eligible to proven-ineligible or indeterminate.",
      "suggestion": "Reject duplicate deployment IDs per wallet, ideally with the same sorted-unique set discipline used for selected deployment sets.",
      "confidence": 0.84
    },
    {
      "id": "GL-004",
      "title": "Duplicate subjects can inflate cohort and disclosure bands",
      "severity": "MEDIUM",
      "category": "privacy_correctness",
      "file": "packages/protocol/gate-leak/src/compute.ts:267",
      "description": "`classifyGateLeakSubjects` maps every input subject into a row and `computeGateLeakMeasure` counts rows directly. There is no uniqueness check for `(discord_user_id, role_id)`, so duplicated subject inputs can inflate the denominator, leak count, and disclosure bands.",
      "suggestion": "Validate the subject cohort as sorted-unique by subject identity before classification, or fail closed on duplicates.",
      "confidence": 0.78
    },
    {
      "id": "GL-005",
      "title": "Measure computation can return schema-invalid bot counts",
      "severity": "LOW",
      "category": "validation",
      "file": "packages/protocol/gate-leak/src/compute.ts:360",
      "description": "`computeGateLeakMeasure` accepts `excludedBotCount: number` and returns it directly as `excluded_bot_count` without validating that it is a non-negative integer. Callers can receive an object typed as `GateLeakMeasure` that would fail the exported schema decoder.",
      "suggestion": "Validate `excludedBotCount` against `NonNegativeInt` inside `computeGateLeakMeasure` or return an Effect failure for invalid inputs.",
      "confidence": 0.7
    },
    {
      "id": "GL-P01",
      "title": "Mapping transitions verify integrity before branch logic",
      "severity": "PRAISE",
      "category": "integrity",
      "file": "packages/protocol/gate-leak/src/mapping.ts:472",
      "description": "`verifyStoredMappingAggregate` strict-decodes the aggregate and verifies every version before replay, CAS, ratify, or revoke logic. The tests also cover in-process grafts, which gives this transition boundary strong regression coverage.",
      "suggestion": "Keep this integrity-first pattern for future protocol transition entrypoints.",
      "confidence": 0.95
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. The protocol work shows strong integrity-oriented design and unusually good adversarial testing, but two public proof/ratification paths currently allow self-consistent forged material to pass. The build lifecycle concern is lower risk but important because install-time network mutation makes the repository less reproducible.

Architecturally, the common thread is “closed evidence must stay closed.” Once a protocol exposes digest-backed attestations or ratified command material, every public verifier has to re-bind the bytes, permissions, and semantic claims at the trust boundary.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "gate-leak-finality-proof-skips-digest-verification",
      "title": "Ownership finality proof can accept tampered in-process attestations",
      "severity": "HIGH",
      "category": "integrity",
      "file": "packages/protocol/gate-leak/src/evidence.ts:1138",
      "description": "evaluateOwnershipFinalityProof walks finality_attestations and checks policy, VM, adapter, status, and coverage linkage, but it never recomputes or validates attestation_digest. A caller that bypasses decodeOwnershipIndexEvidence can pass an OwnershipIndexEvidence object with a forged digest and otherwise matching fields and still receive { proven: true }, which contradicts the closed attestation contract exposed by this public proof function.",
      "suggestion": "Make evaluateOwnershipFinalityProof verify every attestation with verifyOwnershipFinalityAttestationIntegrity, or change it to require a verified evidence wrapper produced only by decodeOwnershipIndexEvidence. Add an in-process tampered attestation_digest readiness/proof regression.",
      "confidence": 0.86,
      "faang_parallel": "Google Certificate Transparency’s Merkle proof model works because verifiers bind the presented claim back to the signed tree head; checking only neighboring metadata would collapse the guarantee.",
      "metaphor": "This is like accepting a notarized document after reading the title and date, without checking the notary seal.",
      "teachable_moment": "Integrity helpers should sit on every public trust boundary, not only on the decode path.",
      "connection": "This is the same broader pattern as JWT libraries that decode payloads without verifying signatures: the object shape looks valid, but the evidence is not bound."
    },
    {
      "id": "gate-leak-mapping-material-permission-not-bound",
      "title": "Stored command material does not prove the ratifier held the ratify permission",
      "severity": "HIGH",
      "category": "authorization",
      "file": "packages/protocol/gate-leak/src/mapping.ts:367",
      "description": "commandMaterialMatchesVersion compares scope, rule, provenance, subject, timing, idempotency key, and reveal basis, but it does not compare or validate command_material.ratifier_permissions against the version's ratifier_permission. A self-consistent persisted version can be minted with command_material lacking gate-config:ratify, recomputed command_digest and mapping_version_id, and still decode as a ratified mapping with ratifier_permission set to gate-config:ratify.",
      "suggestion": "Add a stored-version invariant that command_material.ratifier_permissions includes GATE_CONFIG_RATIFY_PERMISSION and matches the denormalized ratifier_permission semantics. Cover a fixture where command_material permissions are removed and all dependent digests are recomputed.",
      "confidence": 0.78,
      "faang_parallel": "AWS IAM-style authorization systems treat action, principal, and resource as inseparable evaluation inputs; persisting the decision without the action grant would be an audit break.",
      "metaphor": "The receipt says someone bought the house, but the file never proves they had signing authority.",
      "teachable_moment": "A digest proves consistency, not legitimacy, unless every authorization-relevant field is inside the material being checked.",
      "connection": "This is an authorization binding issue: denormalized summaries must be re-derived from canonical command material or explicitly proven equivalent."
    },
    {
      "id": "root-prepare-clones-network-dependency",
      "title": "Root prepare lifecycle now clones a GitHub repository during install",
      "severity": "MEDIUM",
      "category": "build",
      "file": "package.json:7",
      "description": "Adding prepare: ./scripts/prepare-effect.sh makes ordinary installs run a network git clone into .repos/effect. This makes install and packaging depend on GitHub availability, credentials/network policy, and filesystem mutation outside the package graph.",
      "suggestion": "Move this to an explicit opt-in script such as prepare:effect or bootstrap:effect, or model it as a pinned submodule/workspace dependency. Keep lifecycle scripts deterministic and offline-capable.",
      "confidence": 0.91,
      "faang_parallel": "Bazel and Google’s hermetic build practices avoid undeclared network fetches during ordinary builds because reproducibility depends on making inputs explicit.",
      "metaphor": "Install becomes a recipe that quietly drives to another store mid-cook; it works until the store is closed.",
      "teachable_moment": "Lifecycle scripts should be boring: deterministic, offline-capable, and limited to declared package inputs.",
      "connection": "This connects to supply-chain hygiene and CI reliability; hidden network mutation turns dependency setup into operational state."
    },
    {
      "id": "gate-leak-adversarial-tests-strong",
      "title": "Adversarial tests cover persisted and in-process graft paths",
      "severity": "PRAISE",
      "category": "test-coverage",
      "file": "packages/protocol/gate-leak/src/__tests__/mapping.test.ts:362",
      "description": "The mapping suite exercises exact replay, same-key changed-command conflicts, persisted tamper fixtures, and in-process grafts before ratify/revoke branches. That is valuable coverage for the protocol's strict-decode-plus-integrity model.",
      "suggestion": "Keep these probes as contract tests when the mapping schema evolves.",
      "confidence": 0.95,
      "faang_parallel": "Netflix’s failure-testing culture is the right analogy here: the suite does not only test the happy path, it tests the system’s immune response.",
      "metaphor": "These tests are controlled break-in drills for the protocol doors.",
      "teachable_moment": "Protocol tests earn their keep when they preserve negative guarantees, not just successful serialization.",
      "connection": "This is the testing counterpart to the two high-severity issues: once the missing bindings are fixed, similar adversarial fixtures should lock them down."
    }
  ]
}
```


## Callouts
[Test Coverage] The mapping tests are a real strength because they model attacker behavior across persisted tamper and in-process graft paths. That is exactly the discipline this protocol needs.

[Quality] The decision trail in the added protocol docs is directionally strong: future maintainers are being given context, not just code. Extend that habit into the two binding fixes by documenting which fields are canonical evidence and which are derived summaries.

[Operational] The package boundary appears intentionally carved out, with fixtures and verification scripts colocated. Moving the network clone out of `prepare` would preserve that structure while keeping installs predictable.

zkSoju (commented):
## Summary
Verdict: **REQUEST_CHANGES**. This PR shows strong protocol discipline around digest recomputation and evidence-derived readiness, but two high-confidence boundary issues mean invalid or unreachable states can still enter through public-looking paths. The theme is clear: the protocol is strongest where it recomputes truth, and weakest where it trusts already-shaped objects.

Architecturally, Gate Leak is aiming for the same posture as large distributed systems: never trust the caller’s declared state when the system can derive it. The remaining fixes are about making that philosophy uniform at every transition boundary.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "GL-001",
      "title": "Ratify and revoke transitions do not strict-decode incoming commands",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/protocol/gate-leak/src/mapping.ts:706",
      "faang_parallel": "Google's protocol-buffer guidance treats decode/validation boundaries as security and correctness boundaries; internal callers are not assumed trustworthy just because they are in-process.",
      "metaphor": "This is like checking IDs at the building entrance, then letting someone use a side door straight into the vault.",
      "teachable_moment": "Typed objects are not validated objects. Public transition functions should re-establish invariants before constructing canonical state.",
      "connection": "This is the same integrity pattern GL-005 gets right for stored versions; apply it symmetrically to incoming commands.",
      "specific_recommendation": "Strict-decode command at the start of ratifyGateMapping and revokeGateMappingVersion, then use only the decoded value for checks and emitted state.",
      "decision_trail": "Document whether command decoders are the sole canonical boundary or whether transition functions independently enforce wire invariants."
    },
    {
      "id": "GL-002",
      "title": "Partial ownership policy is unreachable through the public decoder when finality is missing",
      "severity": "HIGH",
      "category": "protocol",
      "file": "packages/protocol/gate-leak/src/evidence.ts:535",
      "faang_parallel": "Amazon's service APIs commonly distinguish representability from policy: schemas accept the domain shape, then business logic decides reject/allow/degrade.",
      "metaphor": "The road sign offers three exits, but the guardrail blocks two of them before any driver can choose.",
      "teachable_moment": "If policy has multiple branches, fixtures should prove each branch is reachable through the same public API consumers will use.",
      "connection": "This is a schema-vs-policy split: evidence decoding should describe possible evidence, while readiness should decide acceptable evidence.",
      "specific_recommendation": "Allow empty finality_attestations in the evidence schema and let readiness enforce reject/disclose/allow, or remove disclose/allow for missing finality.",
      "decision_trail": "Record whether missing finality is a valid V1 wire state or intentionally unrepresentable."
    },
    {
      "id": "GL-003",
      "title": "Aggregate audit and version counters are shape-checked but not semantically verified",
      "severity": "MEDIUM",
      "category": "integrity",
      "file": "packages/protocol/gate-leak/src/mapping.ts:274",
      "faang_parallel": "Spanner-style systems do not just store timestamps and logs; they verify ordering relationships that make history meaningful.",
      "metaphor": "A receipt with valid-looking line items still needs the total, dates, and merchant to agree.",
      "teachable_moment": "Append-only metadata is only useful when the append relationship is verified, not merely shaped.",
      "connection": "This extends the digest-integrity story from individual mapping versions to aggregate history.",
      "specific_recommendation": "Verify aggregate_version, audit event count/order, event version monotonicity, mapping IDs, and expected history counts in verifyStoredMappingAggregate.",
      "decision_trail": "Document the aggregate invariants future migrations must preserve."
    },
    {
      "id": "GL-004",
      "title": "Package toolchain can install on unsupported Node 22 minors",
      "severity": "LOW",
      "category": "packaging",
      "file": "packages/protocol/gate-leak/package.json:38",
      "faang_parallel": "Meta's monorepo tooling pins runtime floors because 'works on most minors' becomes CI entropy at scale.",
      "metaphor": "The key says it opens every Node 22 door, but the lock was only cut for 22.12 and later.",
      "teachable_moment": "Toolchain engines are part of the contract, especially for newly introduced package boundaries.",
      "connection": "Operational readiness includes developer-machine reproducibility, not just runtime behavior.",
      "specific_recommendation": "Declare \"engines\": { \"node\": \">=22.12.0\" } for this package or lower dependencies to match the repo floor.",
      "decision_trail": "Record whether gate-leak intentionally raises the package-local Node floor."
    },
    {
      "id": "GL-005",
      "title": "Version integrity verification recomputes all committed mapping identities",
      "severity": "PRAISE",
      "category": "integrity",
      "file": "packages/protocol/gate-leak/src/mapping.ts:529",
      "faang_parallel": "This resembles Merkle-style validation in large storage systems: persisted identity is accepted only after recomputation.",
      "metaphor": "Instead of trusting the label on the jar, the code re-weighs and re-tests the contents.",
      "teachable_moment": "Canonical systems should derive identity from contents whenever possible.",
      "connection": "This is the north star for GL-001 and GL-003.",
      "specific_recommendation": "Keep this integrity-first pattern on every public transition boundary.",
      "decision_trail": "Document this as a protocol invariant: stored identities are claims until recomputed."
    },
    {
      "id": "GL-006",
      "title": "Readiness recomputes the compute-input digest server-side",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/protocol/gate-leak/src/readiness.ts:534",
      "faang_parallel": "Netflix-style cache-key discipline derives keys from canonical inputs rather than caller-supplied result identities.",
      "metaphor": "The cashier totals the cart instead of accepting a sticky note that says '$42'.",
      "teachable_moment": "Result identity belongs to the verifier, not the presenter.",
      "connection": "This creates a clean trust boundary for downstream compute and cache consumers.",
      "specific_recommendation": "Keep server-derived digest behavior as the contract boundary.",
      "decision_trail": "Capture that readiness verdict identity is evidence-derived, not caller-declared."
    }
  ]
}
```


## Callouts
[Quality] GL-005 is excellent because it recomputes rule, config, command, and version digests before accepting stored state. That makes tampering expensive and visible.

[Operational] GL-006 is also strong: deriving the compute-input digest from evidence protects downstream caches from caller-shaped identity drift.

The PR’s best pattern is “derive before trust.” The requested changes are mostly about applying that pattern consistently.

zkSoju (commented):
## Summary

**Verdict: REQUEST_CHANGES.** The protocol boundary and digest-focused tests are strong, but the install lifecycle change introduces a supply-chain and CI reliability risk that should be fixed before merge. The main architectural theme is good protocol discipline paired with one operational escape hatch that deserves to be closed.

A ratified protocol package should behave like infrastructure: deterministic, inspectable, and boring to install. The PR mostly moves in that direction, especially through explicit exports and canonical vectors; the prepare-time remote clone is the outlier because it makes install semantics depend on mutable external state.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Install lifecycle clones a remote repository",
      "severity": "HIGH",
      "category": "supply-chain",
      "file": "scripts/prepare-effect.sh:6",
      "description": "The new root prepare script runs during install lifecycle and clones https://github.com/Effect-TS/effect-smol when .repos/effect is absent. This makes ordinary installs depend on network availability and mutable remote state, and it can execute in consumer/CI contexts that do not expect repository mutation.",
      "suggestion": "Remove the clone from the prepare lifecycle. If this checkout is required for development, make it an explicit opt-in script with a pinned commit or use a submodule/locked dependency.",
      "confidence": 0.95,
      "faang_parallel": "Google's SLSA guidance treats uncontrolled dependency acquisition during builds as a provenance break: builds should resolve pinned inputs, not discover mutable ones mid-flight.",
      "metaphor": "This is like a locked recipe that quietly sends the cook to a street market halfway through dinner; the dish now depends on whoever is selling that day.",
      "teachable_moment": "Install hooks are part of the supply chain. Anything they fetch, mutate, or execute must be as intentional as a production dependency.",
      "connection": "This contrasts with the rest of the PR's protocol determinism: digest vectors and fixtures aim for reproducibility, while prepare-time cloning reintroduces ambient state.",
      "decision_trail": "Document whether effect-smol is a dev convenience, vendored dependency, or protocol input. That choice determines whether it belongs in an opt-in script, submodule, or lockfile."
    },
    {
      "id": "F002",
      "title": "Root engine range is looser than the new package/tooling requirement",
      "severity": "MEDIUM",
      "category": "compatibility",
      "file": "package.json:4",
      "description": "The root package still allows Node >=22, while the new gate-leak package and Vite/Rolldown stack require >=22.12.0. A workspace install or CI job running Node 22.0 through 22.11 can satisfy the root engine but fail inside the new package/tooling.",
      "suggestion": "Raise the root engines.node constraint to >=22.12.0, or otherwise ensure workspace commands for this package cannot run on unsupported Node versions.",
      "confidence": 0.82,
      "faang_parallel": "Meta-style monorepos commonly centralize toolchain versions because one loose root constraint lets leaf packages fail nondeterministically across developer machines and CI shards.",
      "metaphor": "The lobby sign says anyone over 22 can enter, but one room requires 22.12 shoes; people only discover the mismatch after they are already inside.",
      "teachable_moment": "In workspaces, the root engine is the contract humans and automation see first. It should be at least as strict as the strictest routinely-invoked package.",
      "connection": "This is an operational readiness issue, not just package metadata: CI reproducibility starts at the first version gate.",
      "decision_trail": "Record why Node 22.12.0 is the minimum, likely tied to Vite/Rolldown behavior, so future upgrades know whether the constraint is technical or incidental."
    },
    {
      "id": "F003",
      "title": "Gate Leak package is well-scoped as a protocol boundary",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/protocol/gate-leak/src/index.ts:13",
      "description": "The public index exports a narrow, explicit protocol surface and separates scalars, reason codes, gate rules, capabilities, work keys, mapping, evidence, compute, readiness, and migration. That makes downstream imports clearer and reduces accidental reach into private implementation modules.",
      "suggestion": "Keep this export boundary intentional as the package grows; avoid adding wildcard exports from internal modules.",
      "confidence": 0.9,
      "faang_parallel": "Google's API design discipline favors small, explicit surfaces because every exported symbol becomes a compatibility promise.",
      "metaphor": "This is a museum with labeled public galleries and locked staff doors; visitors can understand the exhibit without wandering into storage.",
      "teachable_moment": "A protocol package is healthiest when its public API tells consumers what concepts are stable and what remains implementation detail.",
      "connection": "The explicit boundary supports the PR's ratification goal: downstream systems can depend on named protocol concepts instead of file layout accidents.",
      "decision_trail": "As new modules appear, document why each export is public. That breadcrumb prevents convenience exports from becoming permanent protocol surface by accident."
    },
    {
      "id": "F004",
      "title": "Canonical set tests cover important digest-fragmentation risks",
      "severity": "PRAISE",
      "category": "test-coverage",
      "file": "packages/protocol/gate-leak/src/__tests__/canonical-sets.test.ts:32",
      "description": "The tests explicitly verify permutation-invariant deployment set digests, duplicate rejection, sorted unique role sets, disclosure key canonicalization, and authorization watermark equality. These are high-value regression tests for cache/work-key correctness.",
      "suggestion": "Keep these vectors stable and require fixture regeneration checks in CI so digest semantics cannot drift silently.",
      "confidence": 0.92,
      "faang_parallel": "Bazel's remote cache model depends on stable content-addressed keys; tiny canonicalization drift can fragment cache identity across otherwise equivalent inputs.",
      "metaphor": "These tests make sure every copy of the same key opens the same door, even if the teeth were described in a different order.",
      "teachable_moment": "Canonicalization tests are not cosmetic. They protect identity, cacheability, and auditability in any protocol that turns structured data into digests.",
      "connection": "This is the strongest counterweight to protocol entropy in the PR: it makes semantic equivalence executable rather than merely documented.",
      "decision_trail": "Document the fixture regeneration rule and CI expectation so future changes distinguish intentional digest migrations from accidental drift."
    }
  ]
}
```


## Callouts

[Quality] The module decomposition is unusually clear for a new protocol surface. Separating scalars, mapping, evidence, readiness, and migration gives future reviewers natural fault lines to inspect.

[Test Coverage] The canonical-set tests are doing real architectural work. They encode invariants that humans often miss in review, especially ordering, duplicate rejection, and digest equality across equivalent inputs.

[Operational] The remaining fix is narrow: remove or pin the install-time clone path, then align the root Node engine with the package/tooling floor. After that, the PR’s deterministic protocol posture will be much more coherent.

zkSoju (commented):
## Summary
Verdict: REQUEST_CHANGES. The Gate Leak protocol work shows strong integrity instincts, especially around replay safety, but the root `prepare` hook introduces a supply-chain risk that should not ship as-is. The second issue is narrower, but finality APIs deserve explicit non-vacuous guards.

Architecturally, both concerns are about reproducibility: installs must reproduce the same inputs, and proofs must prove over a real domain. Protocol code earns trust by making the impossible states unrepresentable, then checking again at the boundary.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FINDING-001",
      "title": "Root prepare hook performs an unpinned network clone",
      "severity": "HIGH",
      "category": "supply_chain",
      "dimension": "[Security]",
      "file": "scripts/prepare-effect.sh:12",
      "description": "The new root prepare lifecycle runs ./scripts/prepare-effect.sh, which clones https://github.com/Effect-TS/effect-smol at the default branch with no commit pin or integrity check. Because prepare runs during install/pack flows, installs become network-dependent and can pick up changed upstream code over time.",
      "suggestion": "Remove this from the root prepare lifecycle, or replace it with an explicit opt-in setup command that pins a commit/tag and verifies it. If this repository is required for builds, make it a declared submodule or vendored artifact with reproducible provenance.",
      "confidence": 0.94,
      "faang_parallel": "Google's SLSA work exists largely because unpinned build inputs make provenance unverifiable; the build graph must name exact materials, not moving branches.",
      "metaphor": "This is like printing a recipe that says 'buy whatever flour is on the shelf tomorrow' and expecting the same bread every time.",
      "teachable_moment": "Lifecycle hooks are privileged execution paths. Anything they fetch should be pinned, verified, and intentionally invoked.",
      "connection": "Gate Leak is a protocol ratification package; a non-reproducible install path weakens the same integrity story the protocol is trying to encode.",
      "decision_trail": "Document whether effect-smol is a dev convenience, build prerequisite, or vendored dependency. That decision determines whether it belongs in an opt-in setup script, submodule, or published artifact."
    },
    {
      "id": "FINDING-002",
      "title": "Finality proof can succeed vacuously for an empty selected deployment map",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/protocol/gate-leak/src/evidence.ts:1215",
      "description": "evaluateOwnershipFinalityProof derives selectedDeploymentIds from selectedDeploymentNetworks and never rejects an empty map. If called directly with an empty selectedDeploymentNetworks map and no attestations, the selected-deployment loop is skipped and the function returns { proven: true }, even though an empty deployment selection is invalid elsewhere.",
      "suggestion": "Add an explicit selectedDeploymentNetworks.size === 0 failure, returning partial_deployment_coverage or empty_deployment_selection before any proof work. Keep the readiness-level empty-selection check as defense in depth.",
      "confidence": 0.86,
      "faang_parallel": "Amazon's internal service reviews often call this out as a 'vacuous success' class: a checker that passes because it never evaluated the thing it was meant to check.",
      "metaphor": "It is like a security guard saying every guest had a badge because nobody walked through the door.",
      "teachable_moment": "Proof and validation APIs should reject empty universes unless emptiness is explicitly meaningful in the domain model.",
      "connection": "Readiness-level validation helps, but public or reusable proof functions should carry their own precondition checks so future callers cannot bypass the intended contract.",
      "decision_trail": "Record the invariant: ownership finality requires at least one selected deployment network. Future agents need that breadcrumb at the proof boundary, not only in the caller."
    },
    {
      "id": "FINDING-003",
      "title": "Transition boundary verifies stored mapping integrity before replay",
      "severity": "PRAISE",
      "category": "integrity",
      "dimension": "[Quality]",
      "file": "packages/protocol/gate-leak/src/mapping.ts:785",
      "description": "The ratification path strict-decodes the stored aggregate and recomputes version integrity before idempotency replay or conflict handling. That prevents a grafted command_digest or mapping_version_id from turning an unverified in-memory object into a successful replay.",
      "suggestion": "Keep this integrity-first ordering as the public transition pattern for future protocol aggregates.",
      "confidence": 0.9,
      "faang_parallel": "Bigtable-style systems rely on validating persisted state before interpreting it; storage is never treated as inherently trusted just because it came from the system.",
      "metaphor": "This is checking the seal on an evidence bag before reading the label.",
      "teachable_moment": "Replay safety is not just idempotency. It is idempotency over authenticated, revalidated state.",
      "connection": "This pattern directly supports the protocol's broader ratification posture: decode, verify, then decide.",
      "decision_trail": "Document this ordering as a transition invariant so later aggregate handlers copy the discipline instead of only the surface API."
    }
  ]
}
```


## Callouts
[Quality] The integrity-first ratification path is the strongest part of this PR. Verifying stored aggregates before replay gives future maintainers a pattern that scales beyond this package.

[Test Coverage] The fixture-heavy shape is a good sign for protocol work: golden vectors and malformed cases make the contract inspectable, not just executable.

[Operational] The dedicated packed-consumer verification script points in the right direction. Once the root prepare hook is made reproducible, that consumer-path thinking will materially improve release confidence.

PR #479, opened 2026-07-16
feat: add durable collection resolution

zkSoju:
## Summary

- add the shared collection-resolution protocol for create, confirm, refresh, stale selection, and order admission
- add an Ordering-owned deterministic reference store/service with CAS and bounded exact-command idempotency
- bind immutable original request, authorization scope, candidate snapshot, selected equivalence subset, capability safety, and fixed server expiry
- keep Sonar stateless behind a probe port and expose deterministic fixtures for Dashboard/Sonar consumers

## Verification

- collection-resolution typecheck, build, and protocol tests
- CR-001 collection regressions
- focused Ordering resolution/store and adjacent Ordering regressions
- fixture determinism and changed-path lint
- independent mutation, replay, concurrency, stale-selection, scope, and composition review

## External production blockers

- no production Postgres `ResolutionStore` adapter or migration is claimed
- authenticated HTTP endpoint/ACL wiring remains CR-007A
- Dashboard and Sonar production adoption remain downstream
- existing unrelated `metadata_snapshot` Ordering drift remains separately tracked

Tracks CR-006 in #472.

Draft only. Do not merge.


zkSoju:
CI note: `agent-ci` is red on the same pre-existing `tests/integration/agent-gateway.test.ts` failures tracked in #477 (per-user rate-limit expectation and SSE drop recovery `UND_ERR_SOCKET`). The CR-006 resolution suites are independently green (23 protocol, 29 Ordering resolution, 22 CR-001), and all other PR checks pass. No unrelated baseline CI fix is included in this draft.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. The protocol shape is strong: strict decoding, canonical fingerprints, and fixture-backed contracts show real care. The main risk is that idempotent refresh semantics are not yet durable enough under replay/concurrency, which is exactly where this feature needs to be most boring.

Durable resolution systems live or die on one invariant: the second observation of an operation must mean the same thing as the first. The findings below all orbit that invariant.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Refresh stale-path can throw stale after idempotent replay",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/resolution-service.ts:511",
      "description": "The stale refresh branch calls mapRefreshResult(result) only for side effects, then always throws SelectionStaleError. If the store returns kind:\"replay\" from refreshCas, mapRefreshResult returns the retained replay projection, but that value is discarded and the caller receives a fresh stale error instead of the idempotent historical response. This can occur with a production adapter under concurrent duplicate refresh attempts between the pre-probe lookup and atomic write.",
      "suggestion": "Handle refreshCas replay explicitly before throwing stale, returning the replay projection. Only throw SelectionStaleError when the CAS result is the newly accepted stale refresh.",
      "confidence": 0.78,
      "faang_parallel": "Amazon's Builders' Library guidance on idempotent APIs makes this exact point: retries must return the original operation result, not a newly interpreted result.",
      "metaphor": "This is like a coat-check ticket that sometimes gives you your coat and sometimes says the coat was never checked, depending on which clerk handles the retry.",
      "teachable_moment": "Idempotency is not just duplicate suppression; it is outcome preservation.",
      "connection": "The service already has the right CAS/replay vocabulary. The stale branch needs to honor that vocabulary before mapping fresh stale failures.",
      "decision_trail": "Document whether stale refresh is modeled as a replayable response, a replayable typed error, or a fresh validation failure. Future adapters need that contract."
    },
    {
      "id": "F-002",
      "title": "Stale refresh replays lose selection_stale signal",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/resolution-store.ts:407",
      "description": "refreshCas stores only response_snapshot for idempotent replay. For a stale refresh, the service persists a cleared-selection record and throws SelectionStaleError, but the retained replay response has no selection_stale flag because that flag is not part of ConfirmedResolutionRecord and sealProjection is called without stale context. A duplicate request can therefore replay as an ordinary unconfirmed projection rather than preserving the original stale outcome.",
      "suggestion": "Persist the accepted operation outcome in the idempotency record, or include enough metadata to replay stale refreshes as the same typed error/selection_stale projection originally returned.",
      "confidence": 0.7,
      "faang_parallel": "Dynamo-style conditional writes separate state mutation from client-visible outcome; both must be recorded when retries are expected.",
      "metaphor": "The ledger records that money moved, but forgets whether the ATM showed success or failure to the customer.",
      "teachable_moment": "A durable idempotency record should capture the externally observed result, not only the internal post-state.",
      "connection": "This pairs with F-001: one bug discards replay, the other cannot faithfully construct replay.",
      "decision_trail": "Add a short contract note for idempotency records: persisted projection fields, typed error fields, and which outcomes are replayable."
    },
    {
      "id": "F-003",
      "title": "In-memory store trusts caller subject for CAS idempotency",
      "severity": "LOW",
      "category": "defense-in-depth",
      "file": "packages/services/ordering/src/resolution-store.ts:253",
      "description": "confirmCas and refreshCas key idempotency by input.subject_id but do not verify that input.subject_id matches the stored record's requester_subject before mutating. The service currently checks scope before calling the store, but the store port is exported and documented as the durability boundary, so alternate callers or future adapters could accidentally persist a cross-subject mutation.",
      "suggestion": "Have the store CAS methods compare input.subject_id to current.requester_subject and return not_found/conflict on mismatch, so the persistence boundary enforces the same invariant as the service layer.",
      "confidence": 0.62,
      "faang_parallel": "Google BeyondCorp's lesson applies inside services too: trust should be continuously re-established at each boundary, not inherited from an upstream check.",
      "metaphor": "The lobby checked your badge, but the vault door still needs to check it again.",
      "teachable_moment": "Ports that claim to be durability boundaries should enforce ownership invariants themselves.",
      "connection": "This becomes more important when the in-memory store is replaced by Postgres or another adapter with independent callers.",
      "decision_trail": "Document whether ResolutionStore is an internal helper or an authority boundary. The exported API currently reads like the latter."
    },
    {
      "id": "F-004",
      "title": "Good strict boundary decoding and canonical idempotency",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/protocol/collection-resolution/src/contracts.ts:259",
      "description": "The protocol exposes strict decoders with excess-property rejection and the service builds idempotency fingerprints from decoded canonical material rather than raw JSON object order. This directly reduces replay ambiguity and input smuggling risk across create, confirm, refresh, and admission.",
      "suggestion": "Keep this pattern for the future HTTP and Postgres adapter surfaces.",
      "confidence": 0.96,
      "faang_parallel": "Protocol Buffer and Google API design both lean on canonical shapes at boundaries so clients cannot smuggle alternate meanings through incidental encoding details.",
      "metaphor": "This is a customs checkpoint that checks the manifest, not the suitcase brand.",
      "teachable_moment": "Canonicalization before fingerprinting is the difference between durable identity and byte-level coincidence.",
      "connection": "This is the architectural foundation that makes the replay fixes above straightforward instead of ad hoc.",
      "decision_trail": "Write this as an explicit adapter rule: decode strictly, canonicalize, then fingerprint. That breadcrumb will prevent drift."
    }
  ]
}
```


## Callouts
[Security] The excess-property rejection is excellent because it closes a common input-smuggling path before business logic sees the object.

[Quality] The separation between protocol contracts, selection freshness, digests, admission, and store/service layers gives this package a clean evolutionary path.

[Test Coverage] The fixture suite is unusually valuable here: durable protocols need executable examples, and these fixtures make future adapter behavior easier to verify.

[Operational] The design is already thinking in replay, freshness, and sealed projections. Once stale outcomes are persisted as first-class operation results, this becomes a much stronger production boundary.

zkSoju (commented):
## Summary
REQUEST_CHANGES. The protocol shape is strong, especially around sealed replay boundaries, but the PR’s “durable” claim currently outruns the implementation because persistence is in-memory only. The remaining concerns are mostly about binding/canonicalization edges where future operators need explicit breadcrumbs.

Durable resolution is a contract with time: retries, restarts, and later audits all need the same answer trail. The strongest parts of this PR understand that; the weakest parts still leave too much durability and equivalence semantics implicit.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "finding-1",
      "title": "Production feature ships with only in-memory persistence",
      "severity": "HIGH",
      "category": "durability",
      "dimension": "[Operational]",
      "file": "packages/services/ordering/src/resolution-store.ts:5",
      "description": "The PR title and new service surface describe durable collection resolution, but the implementation exports only InMemoryResolutionStore and explicitly notes that no production Postgres adapter or migration is included. Any process restart loses resolution sessions and idempotency history, which breaks the durable-session and replay guarantees outside tests.",
      "suggestion": "Add the production ResolutionStore adapter and migration in this PR, or keep the new service behind a non-production/internal flag and avoid presenting it as durable until persistence is implemented.",
      "faang_parallel": "Google’s Chubby and Spanner designs treat persistence as part of the API contract, not an implementation detail; clients build retry logic around that promise.",
      "metaphor": "This is a bank ledger written on a whiteboard: accurate until the lights go out.",
      "teachable_moment": "If a feature name promises durability, storage, migration, and replay survival are acceptance criteria.",
      "connection": "This connects directly to operational readiness: process restarts are normal production events, not edge cases.",
      "decision_trail": "Document whether this PR is protocol-only, test-only, or production-ready; future agents need to know whether durable storage is intentionally deferred.",
      "confidence": 0.96
    },
    {
      "id": "finding-2",
      "title": "Selection freshness ignores diagnostics changes",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/protocol/collection-resolution/src/freshness.ts:32",
      "description": "selectionRelevantMaterial projects only snapshot.candidates and drops diagnostics entirely. If a refresh changes searched/timed_out/unavailable networks while returning the same candidates, compareCandidateFreshness treats the selection as still relevantly equal and can preserve confirmation even though the probe coverage changed.",
      "suggestion": "Include selection-relevant diagnostic coverage in the freshness material, or explicitly document and test that diagnostics are intentionally display-only and cannot affect resolution validity.",
      "faang_parallel": "Amazon’s distributed systems guidance often distinguishes data equality from observation equality; two results can match while the confidence envelope changes.",
      "metaphor": "It is like confirming a flight because the arrival city is unchanged, while ignoring that half the radar network went offline.",
      "teachable_moment": "Freshness comparisons should include the evidence needed to trust the answer, not only the answer itself.",
      "connection": "This is a broader protocol-design issue: equivalence must define both payload and proof surface.",
      "decision_trail": "Record whether diagnostics are validity material or display-only material; either choice is defensible, but silence makes later behavior accidental.",
      "confidence": 0.78
    },
    {
      "id": "finding-3",
      "title": "Confirmed resolution fixture has stale hard-coded digests",
      "severity": "MEDIUM",
      "category": "test",
      "file": "packages/protocol/collection-resolution/fixtures/confirmed-resolution.valid.json:19",
      "description": "The fixture hard-codes request_digest and candidate_snapshot_digest values, while tests recompute only the candidate snapshot digest. There is no assertion that request_digest matches original_request plus authorization_scope, so the fixture can publish inconsistent request-binding material without failing the protocol suite.",
      "suggestion": "Add a fixture test that recomputes digestResolutionRequest from confirmed-resolution.original_request/authorization_scope and asserts it equals request_digest, or generate these fixture digests from a checked script.",
      "faang_parallel": "Meta-style schema fixture suites commonly regenerate or verify golden hashes because stale goldens silently train consumers on invalid contracts.",
      "metaphor": "The wax seal looks official, but nobody checks whether it belongs to this envelope.",
      "teachable_moment": "Golden fixtures are only trustworthy when their derived fields are independently recomputed in tests.",
      "connection": "This touches security-adjacent integrity: request binding is what prevents valid-looking artifacts from drifting away from their source.",
      "decision_trail": "Document the canonical digest inputs and add a fixture-generation or verification path so future edits do not rely on memory.",
      "confidence": 0.72
    },
    {
      "id": "finding-4",
      "title": "Admission compatibility digest is order-sensitive for supported_standards",
      "severity": "LOW",
      "category": "canonicalization",
      "dimension": "[Quality]",
      "file": "packages/protocol/collection-resolution/src/admission.ts:257",
      "description": "compatibilityDigest includes view.supported_standards as provided by the local capability view. If the same standards arrive in a different order, admission still succeeds because includes() is order-insensitive, but the compatibility_digest changes for semantically equivalent input.",
      "suggestion": "Canonicalize supported_standards before digesting, for example by sorting and de-duplicating each view's standards array.",
      "faang_parallel": "Merkle-tree systems at Google and elsewhere canonicalize child ordering because semantically equal sets must hash identically.",
      "metaphor": "The guest list is the same people, but the door code changes because their names were printed in a different order.",
      "teachable_moment": "Hash inputs must match semantic equality rules; otherwise digests become noise instead of evidence.",
      "connection": "This is the same family as content-addressed storage and reproducible builds: canonical form is the contract.",
      "decision_trail": "State whether supported_standards is a set or sequence; if it is a set, sort and dedupe before hashing.",
      "confidence": 0.74
    },
    {
      "id": "finding-5",
      "title": "Good boundary design around idempotent historical replay",
      "severity": "PRAISE",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/services/ordering/src/resolution-store.ts:40",
      "description": "The store records sealed response snapshots for idempotency replay instead of returning current mutable store truth. The service tests cover create replay after confirmation and confirm replay after expiry, which protects a subtle but important retry contract.",
      "suggestion": "Keep this behavior covered when adding the production store adapter; adapter conformance tests should reuse these replay cases.",
      "faang_parallel": "Stripe popularized this operational pattern for idempotency keys: replay the original outcome, not whatever current state happens to say.",
      "metaphor": "This is keeping the original receipt instead of re-ringing the register after prices changed.",
      "teachable_moment": "Idempotency is historical consistency, not just duplicate detection.",
      "connection": "This is the architectural center of the PR and should become the conformance standard for every store implementation.",
      "decision_trail": "Promote sealed replay behavior into store-interface documentation so production adapters cannot accidentally weaken it.",
      "confidence": 0.94
    }
  ]
}
```


## Callouts
[Quality] The sealed snapshot replay design is genuinely strong because it protects callers from time-dependent retry behavior. That is the kind of boundary that makes distributed systems boring in the best way.

[Test Coverage] The replay-after-confirmation and replay-after-expiry cases show good instincts: they test the contract users will actually depend on, not just the happy path.

[Operational] The remaining work is clear: once persistence and adapter conformance land, this can become a durable protocol rather than a durable-shaped in-memory prototype.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. The architecture is pointed in the right direction: server-side digest recomputation, strict decoding, and immutability boundaries are strong choices. The blocker is install-time supply-chain behavior; the correctness risks are about canonicalization, which matters especially because this feature is trying to make collection resolution durable.

Durable protocols live or die on boring determinism. If two equivalent observations can hash differently, operators eventually start treating noise as truth; if installs fetch mutable remote code, the build graph stops being auditable.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Root prepare script clones a remote repo on every install",
      "severity": "HIGH",
      "category": "supply-chain",
      "file": "package.json:8",
      "description": "Adding a root `prepare` script means package installation now runs `./scripts/prepare-effect.sh`, which performs a network `git clone` of `https://github.com/Effect-TS/effect-smol` into `.repos/effect`. This makes normal installs depend on external network availability and executes unpinned remote source acquisition as an install lifecycle side effect.",
      "suggestion": "Remove the root `prepare` hook. If the Effect repo is only needed for optional local development, expose it as an explicit script such as `pnpm run prepare:effect` and pin the expected commit or tag.",
      "confidence": 0.95,
      "faang_parallel": "Google's SLSA work treats build inputs as provenance-bearing artifacts; implicit network fetches during install break that chain.",
      "metaphor": "This is like letting a delivery driver swap ingredients while the kitchen is closed.",
      "teachable_moment": "Install hooks should be deterministic, pinned, and boring; optional developer bootstrap belongs behind explicit commands.",
      "connection": "Durable resolution needs reproducible tooling as much as reproducible protocol state."
    },
    {
      "id": "F002",
      "title": "Candidate order changes can invalidate stable selections",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/protocol/collection-resolution/src/freshness.ts:31",
      "description": "`selectionRelevantMaterial` preserves `snapshot.candidates` array order when computing the selection-relevant digest. If Sonar returns the same candidate set in a different order, `selection_relevant_equal` becomes false even though the flattened stale classifier sees no semantic difference and falls back to `deployment_changed`. This can force unnecessary reconfirmation for ordering-only probe noise.",
      "suggestion": "Canonicalize the selection-relevant candidate collection as a sorted set keyed by stable candidate identity or deployment digests before hashing, matching the set semantics used elsewhere for selections and capability views.",
      "confidence": 0.8,
      "faang_parallel": "Bigtable and Spanner APIs lean on stable key ordering so equivalent state has one durable representation.",
      "metaphor": "A guest list should not change because the names were read aloud in a different order.",
      "teachable_moment": "If the domain meaning is a set, hash the set canonically, not the transport array.",
      "connection": "Freshness classifiers become trustworthy only when noise dimensions are removed before comparison."
    },
    {
      "id": "F003",
      "title": "Full snapshot digest also depends on non-canonical candidate order",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/protocol/collection-resolution/src/digests.ts:31",
      "description": "`digestCandidateSnapshot` hashes the raw `CandidateSnapshot` object, including the input order of `candidates`, `diagnostics.searched`, `timed_out`, and `unavailable`. Because the schemas accept arrays without canonical set sorting, semantically identical probe results can produce different server digests and reject a client echo as stale or forged after harmless ordering changes.",
      "suggestion": "Decode candidate snapshots into a canonical order before digesting, or define these arrays as canonical sorted sets with duplicate rejection.",
      "confidence": 0.72,
      "faang_parallel": "Merkle systems at Meta and Amazon canonicalize leaves before comparison to avoid false divergence.",
      "metaphor": "The fingerprint should identify the person, not which pocket they put their keys in.",
      "teachable_moment": "Digest boundaries are contracts; define duplicate handling, sorting, and identity explicitly.",
      "connection": "This is the same family as F002, but at the broader snapshot-integrity layer."
    },
    {
      "id": "F004",
      "title": "In-memory store is exported beside durable service despite production blocker",
      "severity": "LOW",
      "category": "operational-risk",
      "file": "packages/services/ordering/src/index.ts:153",
      "description": "`InMemoryResolutionStore` is exported from the service package public index even though the implementation comments state there is no production Postgres adapter and durable deployment requires one. This makes it easy for downstream wiring to accidentally use the test/reference backend in non-test code.",
      "suggestion": "Keep `ResolutionStore` exported, but move `InMemoryResolutionStore` behind a test-only export path or clearly named development module, and add production wiring that refuses to start without a durable adapter.",
      "confidence": 0.7,
      "faang_parallel": "Netflix popularized fail-fast service config because silent fallback dependencies cause production drift.",
      "metaphor": "A cardboard fire door is useful in a demo, dangerous if it fits the real frame.",
      "teachable_moment": "Public exports are product surface; make unsafe defaults hard to import accidentally.",
      "connection": "Durability is not just an interface promise; packaging must steer consumers toward durable implementations."
    },
    {
      "id": "F005",
      "title": "Strict boundary decoding and server-side digest recomputation are well covered",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/services/ordering/src/resolution-service.ts:173",
      "description": "The service strict-decodes Sonar output before persistence and recomputes candidate snapshot digests server-side before confirm/admit decisions. The tests exercise excess properties, client digest forgery, immutable request binding, and stale selection cases.",
      "suggestion": "Keep this boundary pattern when adding the production persistence adapter and HTTP handlers.",
      "confidence": 0.95,
      "faang_parallel": "Chrome's security boundaries follow the same rule: distrust cross-boundary data and revalidate inside the trusted process.",
      "metaphor": "The clerk does not accept a stamped form from outside; they restamp it at the counter.",
      "teachable_moment": "The right trust boundary both rejects malformed input and recomputes security-relevant facts.",
      "connection": "This is the strongest pattern in the PR and should become the template for persistence and transport layers."
    }
  ]
}
```


## Callouts
[Security] The strict decode plus server-side digest recomputation is the kind of boundary discipline durable protocols need. It resists both accidental schema drift and intentional client-side digest forgery.

[Test Coverage] The tests cover malformed fixtures, immutable bindings, stale selections, and consumer package behavior. That breadth is valuable because this PR is not just adding code; it is establishing a protocol surface others will build against.

[Quality] The separation between protocol contracts, admission, freshness, selection, and store/service layers is promising. Once canonicalization and production-store guardrails are tightened, this structure should be easier to reason about than a single orchestration module.

zkSoju (commented):
## Summary
Verdict: REQUEST_CHANGES. The protocol work shows strong boundary discipline, especially around strict decoding and server-side digest recomputation, but the root install hook introduces a high-risk supply-chain regression. The main architectural theme is durable identity: every automatic action must bind to stable, intentional inputs.

Durability is only valuable when replay is deterministic. Borg-style infrastructure, Bazel-style hermeticity, and Spanner-style decision logs all converge on the same lesson: record exactly what was decided, from exactly which inputs, and avoid ambient mutable state.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Root prepare script performs unpinned network clone during install",
      "severity": "HIGH",
      "category": "supply_chain",
      "dimension": "[Security]",
      "file": "package.json:8",
      "description": "Adding a root `prepare` script makes normal installs run `scripts/prepare-effect.sh`, which clones `https://github.com/Effect-TS/effect-smol` without a pinned commit or offline fallback. This makes dependency installation non-deterministic, network-dependent, and able to execute against a moving upstream repository.",
      "suggestion": "Remove the root prepare hook, or replace it with an explicit developer-only command that pins a commit/tag and is not run automatically during install.",
      "confidence": 0.98,
      "faang_parallel": "Google's Bazel culture treats network access during builds as a hermeticity violation for exactly this reason.",
      "metaphor": "This is like changing the recipe by letting a stranger restock the pantry mid-cook.",
      "teachable_moment": "Install-time hooks are privileged execution paths; keep them deterministic, pinned, and explicit.",
      "connection": "Durable collection resolution depends on reproducible protocol artifacts; nondeterministic installs weaken that foundation.",
      "decision_trail": "Document whether effect-smol is a vendored build dependency, a dev bootstrap step, or a temporary workaround."
    },
    {
      "id": "F-002",
      "title": "Refresh idempotency ignores suppliedRequest identity",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/services/ordering/src/resolution-service.ts:450",
      "description": "The refresh command digest includes the persisted `effectiveRequest` but not the caller-provided `suppliedRequest`. A first refresh with no supplied request and a retry with a matching supplied request use the same idempotency key and digest, so the second call replays instead of fingerprinting the exact request shape. The tests assert exact command fingerprinting for refresh, but this path is not covered.",
      "suggestion": "Include the strict-decoded supplied request, or an explicit null marker when absent, in the refresh command digest so idempotency reflects every caller-provided refresh input.",
      "confidence": 0.86,
      "faang_parallel": "Stripe's idempotency guidance binds replay semantics to the full request shape, not just the eventual effect.",
      "metaphor": "Two envelopes may contain the same letter, but the postmark still matters.",
      "teachable_moment": "Idempotency keys protect retries only when the fingerprint includes every caller-controlled semantic input.",
      "connection": "This mirrors the broader protocol goal: replay should prove the same command happened, not merely a compatible command.",
      "decision_trail": "Document whether refresh idempotency is command-shape exact or effect-equivalent."
    },
    {
      "id": "F-003",
      "title": "Admission compatibility digest includes all local views, not only selected views",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/protocol/collection-resolution/src/admission.ts:263",
      "description": "The compatibility digest material filters selected prepare views, but `decodeLocalCapabilitySnapshot` returns all views sorted by canonical key. If the local snapshot contains unrelated capability views, they are included in the filter input ordering only indirectly through the returned array and can affect admission processing cost, while the digest omits them. This makes the decision digest fail to bind the full local capability snapshot that was accepted.",
      "suggestion": "Either document and enforce that compatibility digests bind only selected prepare views, or include the complete decoded local capability snapshot metadata used for admission in the digest material.",
      "confidence": 0.62,
      "faang_parallel": "Spanner-style commit records distinguish chosen write sets from ambient read context to preserve auditability.",
      "metaphor": "The receipt lists only purchased items, but the cashier also used unseen items to decide the price.",
      "teachable_moment": "A digest should bind either the whole accepted input or a clearly documented projection.",
      "connection": "This is a decision-boundary question, not just a hashing question.",
      "decision_trail": "Record why non-selected capability views are excluded or make exclusion impossible."
    },
    {
      "id": "F-004",
      "title": "Display-only freshness exception allows collection name changes without reconfirmation",
      "severity": "LOW",
      "category": "product_risk",
      "dimension": "[Operational]",
      "file": "packages/protocol/collection-resolution/src/freshness.ts:15",
      "description": "`identity.name`, `identity.symbol`, and `identity.image` are excluded from selection-relevant freshness. If the selected collection's displayed identity changes materially while deployment identity remains stable, refresh preserves confirmation and admission can continue without user reconfirmation.",
      "suggestion": "Confirm that display-name drift is intentionally non-binding for this workflow, or narrow the allowlist to fields that cannot affect user selection confidence.",
      "confidence": 0.55,
      "faang_parallel": "Apple review flows separate cryptographic app identity from user-visible brand identity because both can affect trust.",
      "metaphor": "The passport number is unchanged, but the badge shown at the door now says something else.",
      "teachable_moment": "User-facing labels can be security-adjacent when humans make irreversible selections from them.",
      "connection": "Freshness policy should encode which facts are protocol identity and which are user-confidence inputs.",
      "decision_trail": "Document why display drift does or does not require reconfirmation."
    },
    {
      "id": "F-005",
      "title": "Strict boundary decoding and server-side digest recomputation are well covered",
      "severity": "PRAISE",
      "category": "security",
      "dimension": "[Security]",
      "file": "packages/services/ordering/src/resolution-service.ts:213",
      "description": "Sonar probe output is cloned, strict-decoded, and then digested server-side before persistence. The same pattern is used for refresh output, which reduces boundary trust and client digest forgery risk.",
      "suggestion": "Keep this boundary pattern when adding the production persistence adapter and HTTP wiring.",
      "confidence": 0.96,
      "faang_parallel": "This resembles Chrome's process-boundary validation: deserialize, validate, then trust the normalized form.",
      "metaphor": "It is a customs desk that opens the suitcase itself instead of trusting the traveler's inventory list.",
      "teachable_moment": "The safest digest is computed from canonical server-validated material.",
      "connection": "This is the strongest pattern in the PR and should become the template for future adapters.",
      "decision_trail": "Capture this as a protocol invariant so future integrations preserve it."
    },
    {
      "id": "F-006",
      "title": "Historical idempotency replay avoids rolling current state back",
      "severity": "PRAISE",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/services/ordering/src/resolution-store.ts:48",
      "description": "Idempotency records store a sealed response snapshot and replay that historical response without mutating current store truth. The tests cover create-after-confirm and confirm-after-expiry replay behavior.",
      "suggestion": "Preserve this exact semantic in the future durable store implementation.",
      "confidence": 0.95,
      "faang_parallel": "Dynamo-style systems separate historical request replay from current authoritative state to avoid time-travel bugs.",
      "metaphor": "A receipt reprint should not put yesterday's groceries back on the shelf.",
      "teachable_moment": "Replay is an observational act, not a mutation.",
      "connection": "This cleanly supports durability without letting retries corrupt live resolution state.",
      "decision_trail": "Write this replay semantic into the storage contract before swapping persistence backends."
    }
  ]
}
```


## Callouts
[Security] The strict decode plus server-side digest recomputation is genuinely strong. It rejects the tempting shortcut of trusting client-provided hashes, which is where many protocol boundaries quietly become porous.

[Quality] The sealed historical idempotency response is excellent because it separates replay from rollback. That distinction is subtle, and it is exactly the sort of invariant durable systems need before adding real persistence.

[Test Coverage] The tests already cover important temporal behavior like create-after-confirm and confirm-after-expiry replay. Add the refresh suppliedRequest/null-marker case and the suite becomes much more convincing.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. The protocol work shows strong boundary thinking, especially around digest grafting and public projection minimization, but the PR currently wires durable resolution into Ordering without implementation and adds a root prepare hook that appears non-functional.

The architectural theme here is ownership: durable protocols only become durable when the service boundary that persists, refreshes, confirms, and projects them is real. Right now the protocol layer is much stronger than the integration layer.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Ordering integration is only empty stubs",
      "severity": "HIGH",
      "category": "implementation-gap",
      "file": "packages/services/ordering/src/resolution-service.ts:1",
      "description": "The PR adds durable collection-resolution protocol packages and wires Ordering dependencies, but the new Ordering service, store, and related tests are empty files. That leaves no persistence, confirmation, refresh, public projection, or order-binding enforcement path in the service that is supposed to own durable resolution sessions.",
      "suggestion": "Implement the Ordering resolution store/service and non-empty tests before merging, or remove the service dependency wiring until the integration exists.",
      "confidence": 0.94,
      "faang_parallel": "Amazon’s service ownership model treats an API contract without an operational owner as unfinished work; the owning service must enforce the lifecycle, not just import the schema.",
      "metaphor": "This is like installing a vault door in a bank lobby without building the vault room behind it.",
      "teachable_moment": "A durable protocol needs a durable executor. Persistence, idempotency, confirmation, refresh, and projection are part of the feature, not follow-up polish.",
      "connection": "This is the core release blocker because the new protocol package creates confidence while the service boundary still cannot uphold the contract."
    },
    {
      "id": "F-002",
      "title": "Prepare hook points at an empty new script",
      "severity": "MEDIUM",
      "category": "build",
      "file": "package.json:7",
      "description": "The root package now runs `./scripts/prepare-effect.sh` from `prepare`, but the added script is empty and the `prepare` key is duplicated in the same scripts object. Depending on file mode this can either fail during install or silently no-op while appearing to prepare the Effect checkout.",
      "suggestion": "Remove the duplicate `prepare` key and either implement `scripts/prepare-effect.sh` with a shebang and executable mode or remove the hook.",
      "confidence": 0.9,
      "faang_parallel": "Google’s build discipline favors hermetic, explicit setup steps because hidden install-time hooks are a common source of non-reproducible developer environments.",
      "metaphor": "It is a sign on a door saying emergency exit, but the door opens onto a painted wall.",
      "teachable_moment": "Build hooks are operational promises. If they exist, they should be executable, deterministic, and tested in a clean checkout.",
      "connection": "This weakens contributor trust in the repo setup path and can mask missing Effect preparation until much later."
    },
    {
      "id": "F-003",
      "title": "Capture completeness can accept invalid in-process windows",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/protocol/gate-leak/src/compute.ts:718",
      "description": "`evaluateCaptureCompleteness` computes `windowSeconds` with `Date.parse` and only checks whether it exceeds the max duration. If callers pass an in-process attestation with invalid timestamps or an end before start, `NaN` or a negative duration will not exceed the limit and the capture can still qualify.",
      "suggestion": "Validate `DiscordCaptureAttestation` at the evaluator boundary or explicitly reject non-finite and negative capture durations before returning `qualifying: true`.",
      "confidence": 0.78,
      "faang_parallel": "Netflix’s resilience patterns emphasize validating at trust boundaries because downstream assumptions compound quickly in distributed workflows.",
      "metaphor": "A stopwatch that reads `NaN` or runs backward should never be accepted as proof that a race was completed.",
      "teachable_moment": "When computing eligibility from time windows, reject invalid, non-finite, and inverted intervals before applying policy thresholds.",
      "connection": "This is a small predicate bug with larger trust implications because capture completeness feeds readiness evidence."
    },
    {
      "id": "F-004",
      "title": "Gate-leak package lacks a project reference to collection",
      "severity": "MEDIUM",
      "category": "build",
      "file": "packages/protocol/gate-leak/tsconfig.json:23",
      "description": "The gate-leak source imports `@freeside/collection-protocol`, but its `tsconfig.json` has no `references` entry for `../collection`, unlike the collection-resolution package. A clean filtered `tsc -b` can depend on prebuilt collection `dist` artifacts instead of building the referenced project graph.",
      "suggestion": "Add `\"references\": [{ \"path\": \"../collection\" }]` to the gate-leak tsconfig so clean builds follow the protocol dependency graph.",
      "confidence": 0.62,
      "faang_parallel": "Meta-style monorepo builds rely on explicit dependency graphs so incremental and clean builds converge on the same result.",
      "metaphor": "The package is borrowing a ladder from the neighbor’s garage instead of declaring it as part of the job plan.",
      "teachable_moment": "TypeScript project references are not decoration; they encode build order and prevent stale artifacts from becoming accidental inputs.",
      "connection": "This sits in the same family as F-002: the repo should make dependency preparation explicit rather than ambient."
    },
    {
      "id": "F-005",
      "title": "Boundary tests cover digest grafting and public projection minimization",
      "severity": "PRAISE",
      "category": "test-coverage",
      "file": "packages/protocol/collection-resolution/src/__tests__/protocol.test.ts:83",
      "description": "The collection-resolution tests explicitly reject forged nested candidate identity digests at both persisted-record and public-projection boundaries, and verify public projections omit requester, authorization, original request, request digest, and full candidate snapshot fields.",
      "suggestion": "Keep these tests as regression coverage for the durable-resolution trust boundary.",
      "confidence": 0.95,
      "faang_parallel": "Chrome’s security regression culture treats boundary tests as long-lived guardrails, especially where serialization and public projection meet.",
      "metaphor": "These tests are a customs checkpoint: they verify both the passport and what is allowed to leave the country.",
      "teachable_moment": "The best protocol tests assert what must be rejected and what must not be disclosed, not only the happy path.",
      "connection": "This is the strongest part of the PR and should become the standard for the Ordering integration tests requested in F-001."
    }
  ]
}
```


## Callouts
[Security] The digest-forgery and projection-minimization tests are genuinely strong because they protect both integrity and confidentiality at the protocol edge.

[Test Coverage] The malformed fixture set is a good pattern: it makes invalid states named, reviewable artifacts instead of burying them inside test bodies.

[Quality] The protocol/package separation is directionally sound. Once Ordering owns the lifecycle with real store/service behavior, this can become a clean contract boundary rather than a schema-only island.

PR #480, opened 2026-07-16
feat: add collection protocol compatibility harness

zkSoju:
## Summary

- add a deterministic build, pack, manifest, and verification harness for `@freeside/collection-protocol`
- bind package/schema/source identity and the complete compatibility fixture set to the artifact
- prove isolated clean-build byte reproducibility and exclude stale generated output
- provide one consumer-neutral verification entrypoint plus temporary CR-003/CR-105 legacy pin support

## Verification

- frozen install, typecheck, and package tests
- two isolated clean packs with byte-identical artifact and manifest
- seeded stale-output pack matches clean output
- artifact, manifest identity, fixture, version, and source-commit negative probes
- dependency-direction/cycle check
- legacy CR-003/CR-105 checksum verification
- independent reproducibility review

## Consumer follow-up

Sonar and Dashboard must pin the produced tarball plus manifest and run the shared verify-only entrypoint before install/tests. They must consume the package fixtures rather than fork schemas.

Tracks CR-005 in #472.

Draft only. Do not merge.


zkSoju:
The collection protocol compatibility gate is green after the portable GNU/PAX precedence test amendment.

The remaining `agent-ci` failure is the existing repository baseline tracked in #477:

- per-user rate-limit behavior
- SSE drop recovery (`UND_ERR_SOCKET`)

Those failures reproduce outside this change and are unrelated to the collection protocol harness. No merge is requested; this remains a draft PR.

zkSoju (commented):
## Summary

Verdict: REQUEST_CHANGES. This PR shows strong supply-chain instincts: isolated harness builds and archive preflight checks are exactly the right architectural direction. The remaining concerns are narrow but important: failure cleanup in CI and tar checksum validation are both boundary-integrity issues.

Compatibility harnesses are trust boundaries. The core question is not just “does this artifact pass?” but “can every actor in the path agree on what the artifact is?”

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "ci-compat can leave poisoned dist behind on failure",
      "severity": "MEDIUM",
      "file": "packages/protocol/collection/scripts/ci-compat.sh:98",
      "description": "The script appends a throwing export to dist/harness/pack.js and restores it only after pnpm run pack:artifact succeeds. With set -e, any failure in that command exits before the mv restore runs, leaving the checkout dist modified for the rest of the job or local run.",
      "suggestion": "Wrap the poison probe in a local trap or restore immediately via an ERR-safe cleanup function, similar to the try/finally pattern used in the Vitest version.",
      "confidence": 0.95,
      "faang_parallel": "Google's TAP-style test infrastructure treats cleanup as part of the test contract, not as best effort, because poisoned shared state causes misleading follow-on failures.",
      "metaphor": "This is like testing a smoke alarm by filling the kitchen with smoke, then only opening the windows if the test passes.",
      "teachable_moment": "Any test that mutates the checkout needs shell-level finally semantics. In bash, that usually means trap-based restoration before the risky command runs.",
      "connection": "This connects directly to hermetic testing: compatibility suites should fail closed, but they should not contaminate the next diagnostic step."
    },
    {
      "id": "F-002",
      "title": "Tar parser does not validate header checksums",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/protocol/collection/src/harness/archive.ts:243",
      "description": "listTarGzipMembers parses 512-byte tar headers but never validates the tar header checksum before trusting name, typeflag, size, PAX, and GNU metadata. A corrupted or deliberately malformed archive can be interpreted by this parser before extraction tools reject or interpret it differently, weakening the pre-extract validation boundary.",
      "suggestion": "Compute and verify the tar checksum for each non-zero header block, accepting standard signed/unsigned variants as needed, and fail with ArtifactArchiveMismatch before reading member payload metadata.",
      "confidence": 0.82,
      "faang_parallel": "Android's APK verifier history is the relevant lesson: validators and consumers must agree byte-for-byte on container interpretation, or malformed containers become bypass material.",
      "metaphor": "The checksum is the tamper seal on the envelope. Reading the address before checking the seal lets a forged envelope steer the process.",
      "teachable_moment": "Pre-extract validation is only as strong as its parser equivalence with the extractor. Validate container integrity before trusting metadata.",
      "connection": "This is the same class of issue as archive-slip defenses: path checks are necessary, but parser disagreement can reopen the door."
    },
    {
      "id": "F-003",
      "title": "Global PAX path override is rejected for all normal archives",
      "severity": "LOW",
      "category": "compatibility",
      "file": "packages/protocol/collection/src/harness/archive.ts:338",
      "description": "Global PAX path/linkpath overrides are merged into every subsequent member, so a valid archive using a global PAX record with unrelated keys is fine, but any global path/linkpath makes every member share an effective path and fail. That is stricter than tar semantics and may reject archives produced by tools that emit global path metadata.",
      "suggestion": "If this strictness is intentional, document that global PAX path/linkpath records are unsupported. Otherwise, handle global records according to the target extractor behavior or explicitly reject them at the global header rather than applying them repeatedly.",
      "confidence": 0.66,
      "faang_parallel": "Bazel's remote cache compatibility work leans hard on documenting stricter-than-spec behavior so producers know which artifacts are portable.",
      "metaphor": "This is like a hotel giving every future guest the same room number because one lobby note mentioned a room.",
      "teachable_moment": "Strictness is fine when it is explicit. Undocumented strictness becomes accidental incompatibility.",
      "connection": "The broader pattern is protocol design: unsupported features should be rejected at the boundary where they appear, not after transformation makes the error look like something else."
    },
    {
      "id": "F-004",
      "title": "Isolated build avoids trusting checkout dist",
      "severity": "PRAISE",
      "category": "supply-chain",
      "file": "packages/protocol/collection/scripts/lib/isolated-harness.mjs:54",
      "description": "The pack and verify CLIs compile the harness from a fresh staging copy when TypeScript is available, and the tests include a poisoned checkout dist probe. This is a strong control for preventing stale or tampered generated output from influencing artifact creation.",
      "suggestion": "Keep this invariant covered by both the unit test and CI compatibility script.",
      "confidence": 0.98,
      "faang_parallel": "Chrome's release infrastructure similarly separates source truth from generated artifacts to reduce stale-build and supply-chain risk.",
      "metaphor": "You are baking from the recipe in a clean kitchen, not trusting a cake someone left on the counter.",
      "teachable_moment": "Generated output is a cache, not authority. Rebuilding from source inside the harness is the right default for provenance-sensitive tooling.",
      "connection": "This supports the PR's central compatibility goal: consumers can trust the harness behavior without trusting incidental local build state."
    },
    {
      "id": "F-005",
      "title": "Archive preflight covers high-risk tar member classes",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/protocol/collection/src/harness/archive.ts:538",
      "description": "The preflight rejects non-file members, link targets, path escapes, duplicate effective names, case collisions, and NFC collisions before extraction. The negative fixtures exercise PAX and GNU path override behavior, which is exactly where archive verification commonly drifts.",
      "suggestion": "Add checksum validation to complete this boundary, but the member-set model itself is sound.",
      "confidence": 0.94,
      "faang_parallel": "Docker and Kubernetes image handling both learned that archive member policy must be explicit before unpacking, especially around links and path normalization.",
      "metaphor": "This is a bouncer checking IDs, duplicate names, side doors, and fake badges before anyone enters the venue.",
      "teachable_moment": "Archive security is mostly about refusing ambiguity before extraction. This implementation already covers many of the dangerous ambiguity classes.",
      "connection": "Once checksum validation lands, this becomes a coherent layered boundary: container integrity, member identity, then extraction."
    }
  ]
}
```


## Callouts

[Security] The archive preflight model is notably disciplined. Rejecting duplicate effective names, Unicode normalization collisions, link targets, and path escapes before extraction shows mature threat modeling.

[Quality] The isolated harness design is the strongest part of the PR. It encodes a valuable principle: compatibility artifacts should be produced from declared source, not whatever generated files happen to exist in the checkout.

[Test Coverage] The poisoned-dist probe is excellent test intent. After the cleanup trap fix, it will be a high-signal regression guard rather than a possible source of state leakage.

zkSoju (commented):
## Summary
Verdict: `COMMENT`. The harness is thoughtfully shaped around strictness and artifact integrity, with especially strong archive metadata handling; the main concern is resource bounding around untrusted compressed inputs. Compatibility infrastructure becomes supply-chain infrastructure quickly, so size limits and documented trust boundaries deserve first-class treatment.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "finding-1",
      "title": "Artifact verifier decompresses untrusted tarballs fully into memory",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/protocol/collection/src/harness/archive.ts:236",
      "description": "listTarGzipMembers reads the whole tarball and passes it through gunzipSync before any size or member validation. Since verifyArtifact calls this on pinned artifact input before extraction, a malformed or highly compressed tarball can cause excessive memory use in CI or consumer verification.",
      "suggestion": "Parse gzip/tar as a bounded stream or enforce explicit compressed and decompressed size limits before gunzipSync. Fail with ArtifactArchiveMismatch when limits are exceeded.",
      "confidence": 0.86,
      "faang_parallel": "Google SRE treats memory as an explicit production quota, not an implementation detail; Borg-style resource isolation exists because one hostile or malformed workload must not starve the fleet.",
      "metaphor": "This is like inspecting a suitcase by first expanding a vacuum-packed warehouse inside the lobby.",
      "teachable_moment": "Verification code is security code. It must reject bad inputs before paying attacker-controlled decompression costs.",
      "connection": "This is the same class as zip bombs and decompression bombs: structural validation is good, but resource ceilings are the other half of trust."
    },
    {
      "id": "finding-2",
      "title": "Packer can extract unbounded package archives during manifest generation",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/protocol/collection/src/harness/pack.ts:154",
      "description": "buildManifestFromTarball validates member shape, but then invokes system tar extraction on the generated archive without enforcing total size or file count limits. If packCommand is overridden or pnpm pack output is compromised, this can write unexpectedly large content into the temporary directory.",
      "suggestion": "Use the preflight member list to enforce a maximum file count and total unpacked size before extraction, and consider extracting with a controlled tar reader rather than shelling out to tar.",
      "confidence": 0.7,
      "faang_parallel": "Container registries and build farms commonly preflight layer counts and byte sizes before unpacking; Docker image handling learned this lesson through years of hostile archive inputs.",
      "metaphor": "The current check verifies every box has a legal label, but not whether the truck contains ten boxes or ten million.",
      "teachable_moment": "Archive validation should cover names, paths, count, and bytes. Shape safety without volume safety is incomplete.",
      "connection": "Because this harness may become a consumer-facing compatibility contract, bounded extraction should be part of the protocol expectation, not just CI hygiene."
    },
    {
      "id": "finding-3",
      "title": "Verify-only script argument construction is fragile",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/protocol/collection/scripts/ci-compat.sh:29",
      "description": "Optional CLI flags are built with parameter expansion containing embedded quotes. This is brittle for shell word splitting and can mis-handle values containing whitespace or shell-special characters if future expectations are not simple numeric/semver values.",
      "suggestion": "Build an args array, append optional flags with separate array elements, then call node with \"${args[@]}\".",
      "confidence": 0.62,
      "faang_parallel": "Bazel-style command construction keeps argv as structured data as long as possible; shell strings are treated as a lossy serialization boundary.",
      "metaphor": "Quoting inside a string is like writing seating assignments on napkins, then asking the wind to preserve the table plan.",
      "teachable_moment": "In shell, arrays are the type-safe path for argv. Build commands as arrays, not interpolated strings.",
      "connection": "Low severity today because values are simple, but compatibility scripts tend to accrete flags; fixing the pattern now prevents future footguns."
    },
    {
      "id": "finding-4",
      "title": "Strict JSON parser intentionally preserves reserved keys without prototype pollution",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/protocol/collection/src/harness/json-strict.ts:213",
      "description": "Object.defineProperty is used when assigning parsed object fields, which lets __proto__ and constructor remain inert own properties instead of mutating the object prototype. The added test covers this behavior directly.",
      "suggestion": "Keep this parser path as the shared manifest and identity JSON entrypoint.",
      "confidence": 0.95,
      "faang_parallel": "Chrome and Node security reviews routinely insist that parsed attacker-controlled objects do not inherit ambient authority through prototypes.",
      "metaphor": "This puts suspicious keys in evidence bags instead of letting them touch the building wiring.",
      "teachable_moment": "A strict parser is not only about syntax; it also preserves data as data, without triggering language-level behavior.",
      "connection": "This is exactly the right posture for protocol artifacts, where consumers need deterministic bytes and semantics across runtimes."
    },
    {
      "id": "finding-5",
      "title": "Archive preflight accounts for effective PAX and GNU paths before extraction",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/protocol/collection/src/harness/archive.ts:228",
      "description": "The tar member parser resolves PAX path/linkpath and GNU longname/longlink metadata before validating the archive inventory, and rejects ambiguous mixed metadata instead of relying on platform tar precedence.",
      "suggestion": "Maintain these negative fixtures because they protect the highest-risk part of the artifact verifier.",
      "confidence": 0.92,
      "faang_parallel": "Supply-chain systems such as container image scanners normalize archive metadata before policy checks so the checker and extractor cannot disagree.",
      "metaphor": "This checks the real address on the envelope, not just the nickname scribbled on top.",
      "teachable_moment": "Validate the effective path the extractor will honor, not merely the raw header field you first notice.",
      "connection": "This pairs well with strict JSON parsing: both reduce ambiguity before trust decisions are made."
    }
  ]
}
```


## Callouts
[Security] The strict JSON work is excellent because it handles hostile-but-valid keys without collapsing into prototype mutation. That is mature protocol engineering.

[Security] The tar metadata normalization is also strong: resolving PAX/GNU effective names before validation avoids the classic checker/extractor disagreement. Keep those negative fixtures close; they are regression armor.

[Quality] The PR shows a clear compatibility-contract mindset. The next step is making resource budgets as explicit as schema rules, so future consumers inherit both correctness and operational safety.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. The harness work shows strong security instincts around archive verification, but the root install hook creates a high-risk operational dependency on external network state during ordinary installs.

The architectural theme here is supply-chain boundaries: verifier code is doing the right thing by distrusting artifacts, while package/install plumbing needs the same discipline. Google’s SRE practice treats builds and installs as reproducible systems, not conveniences; this PR is close, but the install path currently has too much ambient authority.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Root prepare script introduces network cloning during install",
      "severity": "HIGH",
      "category": "reliability",
      "dimension": "[Operational]",
      "file": "package.json:9",
      "description": "Adding `prepare` at the repository root runs `./scripts/prepare-effect.sh`, which performs `git clone https://github.com/Effect-TS/effect-smol` during install. This makes ordinary dependency installation depend on network access and an external repository, and can break offline CI, clean local installs, or any environment without GitHub access.",
      "suggestion": "Remove the root `prepare` hook, or make the Effect checkout an explicit developer command that CI invokes only where needed.",
      "confidence": 0.95,
      "faang_parallel": "Google’s hermetic build discipline exists precisely to avoid this class of failure: a build should not silently depend on the current availability or contents of a remote service.",
      "metaphor": "This is like making the office front door call a locksmith every time someone enters. It may work most days, but entry now depends on someone outside the building answering.",
      "teachable_moment": "Install hooks should be boring and deterministic. Network fetches belong in explicit bootstrap or CI steps with caching, pinning, retries, and ownership.",
      "connection": "The harness is about compatibility and reproducibility; the root prepare hook weakens that same story at the repository boundary.",
      "decision_trail": "Document why Effect checkout is needed, who owns updating it, whether it must be pinned, and which command or CI job is allowed to perform the fetch."
    },
    {
      "id": "F-002",
      "title": "Verify CLI can report success with invalid numeric expectation flags",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/protocol/collection/scripts/verify-artifact.mjs:137",
      "description": "`--expect-major` and `--expect-minor` are converted with `Number(...)` without validating that the result is a finite integer. Values like `--expect-major abc` become `NaN`; comparisons against `NaN` are always true for mismatch, but the error output reports `expected` as `null` when stringified, making operator errors hard to diagnose. Values like `1.5` are accepted as expectations even though contract majors/minors are integers.",
      "suggestion": "Parse these flags with explicit integer validation and fail with `UsageError` before invoking `verifyArtifact` when the value is not a non-negative integer.",
      "confidence": 0.82,
      "faang_parallel": "Amazon’s internal CLI guidance emphasizes fail-fast argument validation because ambiguous operator input turns simple mistakes into incident-debugging sessions.",
      "metaphor": "This is a turnstile that accepts a blank ticket, then complains later that the destination was unclear.",
      "teachable_moment": "CLI boundaries are API boundaries. Parse into domain types first, then call business logic only with values that already satisfy the contract.",
      "connection": "The harness validates artifact structure rigorously; expectation flags deserve the same strictness because they steer the verifier’s verdict.",
      "decision_trail": "Record the numeric contract for major/minor values in CLI help and tests: finite, integer, non-negative, and rejected before verification starts."
    },
    {
      "id": "F-003",
      "title": "Package file allowlist publishes internal CI harness scripts",
      "severity": "LOW",
      "category": "packaging",
      "dimension": "[Operational]",
      "file": "packages/protocol/collection/package.json:25",
      "description": "The package `files` allowlist includes the entire `scripts` directory. That publishes `ci-compat.sh` and helper scripts intended for repository CI alongside the package binaries. Consumers only need the pack/verify binaries and their loader; publishing the full compatibility script increases package surface area and exposes producer-specific workflow assumptions.",
      "suggestion": "Narrow the published files to the exact CLI runtime files needed by the `bin` entries, such as `scripts/pack-artifact.mjs`, `scripts/verify-artifact.mjs`, and required `scripts/lib/*.mjs`.",
      "confidence": 0.78,
      "faang_parallel": "Meta’s Buck/Bazel packaging patterns separate internal build orchestration from shipped runtime artifacts to keep consumer contracts small and stable.",
      "metaphor": "Shipping the whole scripts folder is like mailing the restaurant’s kitchen checklist with the takeout order.",
      "teachable_moment": "Published package contents are part of the public interface. Smaller artifacts are easier to secure, document, cache, and support.",
      "connection": "This is a boundary hygiene issue: CI machinery and consumer CLI runtime have different audiences and stability promises.",
      "decision_trail": "Document which scripts are public CLI surface, which are repo-only automation, and how future scripts should be classified before publication."
    },
    {
      "id": "F-004",
      "title": "Archive preflight rejects unsafe tar metadata before extraction",
      "severity": "PRAISE",
      "category": "security",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/harness/archive.ts:255",
      "description": "The harness validates effective tar member paths, PAX/GNU overrides, link types, duplicates, case/NFC collisions, and manifest membership before extraction. That is a strong defensive choice for verifier code that consumes pinned artifacts.",
      "suggestion": "Keep this preflight path as the single verifier entrypoint for both CLI and library use.",
      "confidence": 0.96,
      "faang_parallel": "Chrome’s update and Safe Browsing pipelines use layered validation before trusting downloaded payloads; unsafe archive metadata is treated as hostile input, not an edge case.",
      "metaphor": "This is airport security before boarding, not after the plane is in the air.",
      "teachable_moment": "The safest extraction code is code that proves the archive shape before extraction begins.",
      "connection": "This is the strongest architectural move in the PR: it centralizes distrust at the verifier boundary instead of scattering checks after files exist on disk.",
      "decision_trail": "Document that all artifact consumers must enter through this preflight path, so future CLI/library additions do not bypass the security invariant."
    }
  ]
}
```


## Callouts
[Security] The archive preflight deserves real credit. Rejecting path tricks, link types, duplicate identities, Unicode/case collisions, and manifest drift before extraction shows a mature threat model for artifact verification.

[Test Coverage] The added compatibility and harness test files suggest the PR is treating the harness as a contract, not just a helper script. That is the right instinct for protocol tooling: tests become executable documentation for future consumers.

[Quality] The split into archive, digests, identity, manifest, pack, strict, and verify modules is a promising shape. It gives future maintainers named places to put policy instead of letting verifier logic become one large procedural tunnel.

zkSoju (commented):
## Summary

Verdict: `REQUEST_CHANGES`. The harness shows serious care around artifact verification, especially archive preflight and strict JSON handling, but the provenance path currently allows artifacts to claim a different source than the bytes being packed. The install-time network clone is also too operationally risky for a root lifecycle hook.

This PR is strongest where it treats compatibility artifacts as hostile inputs. It needs the same discipline on provenance and install behavior: identity metadata is part of the trust boundary, not decoration.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F001",
      "title": "Pack provenance can name an ancestor while packing the current tree",
      "severity": "HIGH",
      "category": "provenance",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/harness/source.ts:69",
      "description": "assertReachableSourceCommit accepts any commit that is an ancestor of HEAD, while packArtifact copies and builds the current working tree. A caller can pass an older reachable commit and produce an artifact whose protocol-identity.json source_commit points at that older commit even though the packed bytes came from current HEAD.",
      "suggestion": "Require sourceCommit to equal git rev-parse HEAD for this pack flow, or actually materialize/build the requested commit in the isolated staging tree. If non-HEAD trees remain allowed, surface that state in the manifest and fail consumer verification unless explicitly expected.",
      "confidence": 0.86,
      "faang_parallel": "Google's SLSA provenance model is built around this exact invariant: the claimed source revision must correspond to the built artifact, or the attestation loses evidentiary value.",
      "metaphor": "This is a shipping label from last week's box placed on today's package.",
      "teachable_moment": "Reachability proves ancestry, not identity. For supply-chain claims, ancestor-of is almost never strong enough.",
      "connection": "This connects directly to reproducible builds and artifact attestations: metadata must bind to bytes, not merely to a plausible repository history."
    },
    {
      "id": "F002",
      "title": "Invalid source-commit CLI input silently falls back to HEAD",
      "severity": "MEDIUM",
      "category": "cli",
      "dimension": "[Quality]",
      "file": "packages/protocol/collection/scripts/pack-artifact.mjs:45",
      "description": "resolveSourceCommit only honors an explicit value when it is a full sha40. Any non-empty typo or shorthand passed via --source-commit or SOURCE_COMMIT is ignored and the command packs HEAD instead.",
      "suggestion": "Distinguish omitted input from invalid input. If an explicit source commit is provided and does not match the accepted format, fail with a UsageError instead of falling back.",
      "confidence": 0.9,
      "faang_parallel": "Kubernetes CLIs generally fail fast on malformed explicit flags because silent fallback turns operator intent into ambiguous state.",
      "metaphor": "If someone writes an address on an envelope, the post office should not quietly deliver it to the sender's house because the ZIP code looked odd.",
      "teachable_moment": "Defaults are safe only when input is absent. Invalid explicit input should be treated as information, not ignored.",
      "connection": "This is the CLI version of strict parsing: preserve user intent by separating omission, shorthand, and malformed values."
    },
    {
      "id": "F003",
      "title": "Root prepare hook adds an unpinned network clone to install",
      "severity": "HIGH",
      "category": "ci",
      "dimension": "[Operational]",
      "file": "package.json:8",
      "description": "The new root prepare script runs ./scripts/prepare-effect.sh during installs, and that script clones https://github.com/Effect-TS/effect-smol without a pinned commit. This makes normal installs depend on network availability and mutable remote state.",
      "suggestion": "Remove this from the prepare lifecycle. Make it an explicit developer-only script, or pin and verify a commit if the checkout is required for a controlled workflow.",
      "confidence": 0.88,
      "faang_parallel": "Meta's build systems and Google's hermetic build guidance both push dependency resolution out of ambient install hooks and into pinned, cacheable inputs.",
      "metaphor": "This turns `npm install` into a door that sometimes opens into the internet and sometimes into a wall.",
      "teachable_moment": "Lifecycle hooks run in more places than authors expect: CI, consumers, packagers, and security scanners.",
      "connection": "Hermeticity is operational reliability. Mutable remote state in install paths creates flake, supply-chain exposure, and hard-to-reproduce failures."
    },
    {
      "id": "F004",
      "title": "Strict JSON test uses an invalid escape sequence",
      "severity": "MEDIUM",
      "category": "test",
      "file": "packages/protocol/collection/src/__tests__/harness.test.ts:89",
      "description": "The test case intended to accept escaped control characters passes JSON text containing \\^, which is not a valid JSON escape. parseJsonStrict correctly rejects unknown escapes, so this assertion should fail.",
      "suggestion": "Use valid JSON escapes such as \\u0000, \\n, and \\t in the accepted-form assertion, or update the expected behavior if \\^ was meant to be rejected.",
      "confidence": 0.82,
      "faang_parallel": "Chromium's parser tests distinguish accepted grammar from rejected grammar with very small fixtures because one wrong fixture can invert the meaning of a conformance suite.",
      "metaphor": "The test is asking the bouncer to admit someone with a fake ticket, then calling the bouncer broken for refusing.",
      "teachable_moment": "Parser tests are specifications. A malformed positive fixture teaches future maintainers the wrong language.",
      "connection": "This matters more because the strict JSON parser is a trust boundary; its tests should be canonical examples."
    },
    {
      "id": "F005",
      "title": "Archive verification validates effective tar paths before extraction",
      "severity": "PRAISE",
      "category": "security",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/harness/archive.ts:271",
      "description": "The harness resolves PAX and GNU extended path/link metadata before extraction and rejects symlinks, hardlinks, duplicate members, path escapes, and case/NFC collisions. That is a strong guardrail for artifact verification.",
      "suggestion": "Keep this preflight path as the single archive gate for both CLI and library verification.",
      "confidence": 0.94,
      "faang_parallel": "Android's package tooling has long treated archive paths as hostile because extraction bugs become arbitrary file writes.",
      "metaphor": "You inspect every suitcase before it enters the building, including hidden name tags and false bottoms.",
      "teachable_moment": "Validate the effective path, not the visible path. Archive formats have multiple layers of naming.",
      "connection": "This is exactly the right shape for compatibility artifacts: reject ambiguity before bytes touch disk."
    },
    {
      "id": "F006",
      "title": "Manifest parsing rejects duplicate JSON keys before schema decode",
      "severity": "PRAISE",
      "category": "validation",
      "dimension": "[Security]",
      "file": "packages/protocol/collection/src/harness/json-strict.ts:218",
      "description": "The strict JSON parser detects duplicate keys before handing data to Effect schemas, avoiding JSON.parse last-writer-wins behavior for manifests and identity files.",
      "suggestion": "Continue routing CLI manifest verification through the raw JSON-text decoder so this protection is not bypassed.",
      "confidence": 0.93,
      "faang_parallel": "AWS and Google API signing systems avoid ambiguous canonicalization for the same reason: two readers must not see two different documents.",
      "metaphor": "A contract cannot have two prices on the same line and let each party choose the convenient one.",
      "teachable_moment": "Schema validation starts too late if the parser has already discarded evidence.",
      "connection": "This complements the archive gate: both remove ambiguity before higher-level validation begins."
    }
  ]
}
```


## Callouts

The archive verifier is notably strong: resolving extended tar metadata before extraction is the difference between checking the map and checking the terrain. The duplicate-key rejection is also excellent because it prevents `JSON.parse` from laundering ambiguity before Effect schemas see the data.

The broader architecture is promising: compatibility harnesses become durable only when they are stricter than normal application code. Fix the provenance binding and install lifecycle issues, and the harness will better match the rigor already visible in its artifact verification core.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`.

This PR is doing serious protocol work: compatibility harnesses, strict fixtures, packed-consumer checks, and supply-chain-conscious CI are all the right architectural instincts. The blocking concern is identity binding: idempotency that is not scoped to the resolution can make a correct-looking replay answer the wrong question.

The broader pattern here is that protocol harnesses live or die by their boundaries: every digest, cache key, build edge, and CI invocation needs to say exactly what world it belongs to.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Confirm and refresh idempotency keys are not scoped to the resolution",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/resolution-store.ts:175",
      "description": "The idempotency map key is only operation + subject + idempotency_key, while confirm/refresh commands do not carry resolution_id. Reusing the same idempotency key and same command body across two resolutions for the same subject can replay the first resolution's stored response instead of applying to the second resolution.",
      "suggestion": "Scope confirm/refresh idempotency by resolution_id, or include resolution_id in the command digest and validate it on replay. Add a regression test that uses the same confirm/refresh idempotency key on two different resolution IDs.",
      "confidence": 0.88,
      "dimension": "[Quality]",
      "faang_parallel": "Google Cloud-style idempotency guidance treats request identity as the full operation target, not just the retry token; the token is a replay guard, not a global command alias.",
      "metaphor": "This is like giving two hotel rooms the same spare key because the guest name matches. The key works, but it opens the wrong door.",
      "teachable_moment": "Idempotency keys must bind to the resource whose state may change. Otherwise retries become cross-resource cache hits.",
      "connection": "This cuts against the PR's otherwise strong digest-domain work: the same domain-separation discipline should apply to operational replay keys.",
      "specific_recommendation": "Include resolution_id in the confirm/refresh command shape or derive the idempotency-store key from operation + subject + resolution_id + idempotency_key.",
      "decision_trail": "Document why idempotency is scoped at the resolution boundary so future service changes do not collapse it back to subject-level identity."
    },
    {
      "id": "F-002",
      "title": "Root package.json contains duplicate prepare script keys",
      "severity": "MEDIUM",
      "category": "build",
      "file": "package.json:7",
      "description": "The diff adds the same \"prepare\" key three times in the root scripts object. JSON parsers typically keep only the last value, but duplicate keys are ambiguous and can fail stricter tooling or hide future script edits.",
      "suggestion": "Keep a single \"prepare\": \"./scripts/prepare-effect.sh\" entry in the scripts object.",
      "confidence": 0.96,
      "dimension": "[Operational]",
      "faang_parallel": "Google's Bazel ecosystem is strict about one declared value for one build edge because ambiguous configuration makes builds non-reproducible.",
      "metaphor": "A duplicate JSON key is a signpost with three arrows nailed on top of each other; most drivers follow the last one, but nobody should rely on that.",
      "teachable_moment": "Build metadata should be boring and singular. Ambiguity in root config spreads quickly because every package inherits the uncertainty.",
      "connection": "A compatibility harness is meant to increase confidence across consumers; duplicate root scripts make the harness's own entrypoint less trustworthy.",
      "specific_recommendation": "Collapse the root scripts object to exactly one prepare key.",
      "decision_trail": "No ADR needed, but the commit should make clear that prepare is the single shared bootstrap hook."
    },
    {
      "id": "F-003",
      "title": "Gate-leak build lacks a project reference to collection-protocol",
      "severity": "MEDIUM",
      "category": "build",
      "file": "packages/protocol/gate-leak/tsconfig.json:24",
      "description": "The gate-leak package imports @freeside/collection-protocol but its tsconfig has no reference to ../collection, unlike collection-resolution. A clean `tsc -b` can depend on prebuilt collection dist/types instead of building the local protocol dependency in order.",
      "suggestion": "Add a TypeScript project reference to ../collection and ensure package build/test commands run through `tsc -b` with that reference.",
      "confidence": 0.74,
      "dimension": "[Operational]",
      "faang_parallel": "Bazel's core lesson from Google's monorepo is that dependency edges must be explicit or incremental builds eventually lie.",
      "metaphor": "This is a bridge with the road painted on the map but no actual support beam underneath.",
      "teachable_moment": "Package imports and TypeScript build graph references should tell the same story, especially for protocol packages with generated artifacts.",
      "connection": "The harness is testing compatibility across packages; the build graph should encode that compatibility path directly.",
      "specific_recommendation": "Add `{ \"path\": \"../collection\" }` to gate-leak's references and validate with a clean `tsc -b` from the workspace root.",
      "decision_trail": "Document whether protocol packages are built as a TS project-reference graph or as independently prebuilt npm-style packages."
    },
    {
      "id": "F-004",
      "title": "Compatibility workflow may depend on executable bit for shell script",
      "severity": "LOW",
      "category": "ci",
      "file": ".github/workflows/collection-protocol-compat.yml:44",
      "description": "The workflow executes `./scripts/ci-compat.sh` directly. If the newly added script is committed without executable mode, CI will fail with permission denied.",
      "suggestion": "Either ensure the script is committed with executable permissions or invoke it as `bash ./scripts/ci-compat.sh`.",
      "confidence": 0.52,
      "dimension": "[Operational]",
      "faang_parallel": "Hermetic build systems such as Bazel avoid ambient filesystem assumptions; the interpreter is part of the action, not folklore.",
      "metaphor": "Relying on the executable bit is like assuming every door is already unlocked before the inspection starts.",
      "teachable_moment": "CI should remove incidental platform state wherever possible.",
      "connection": "This is small, but compatibility jobs should fail only on compatibility, not on file-mode drift.",
      "specific_recommendation": "Prefer `bash ./scripts/ci-compat.sh` in the workflow unless executable mode is intentionally enforced in repo policy.",
      "decision_trail": "If executable mode is required, document that convention near the scripts or in CONTRIBUTING."
    },
    {
      "id": "F-005",
      "title": "GitHub Actions are pinned to immutable SHAs",
      "severity": "PRAISE",
      "category": "supply-chain",
      "file": ".github/workflows/collection-protocol-compat.yml:28",
      "description": "The workflow pins third-party actions to full commit SHAs and uses read-only contents permission, which reduces CI supply-chain risk for the compatibility harness.",
      "suggestion": "Keep this pattern for the new protocol workflows and any downstream consumer harness jobs.",
      "confidence": 0.95,
      "dimension": "[Security]",
      "faang_parallel": "Google's SLSA work pushes the same principle: provenance and immutable references are the foundation of trustworthy builds.",
      "metaphor": "This pins the recipe to a numbered page instead of trusting that tomorrow's bookmark points to the same instructions.",
      "teachable_moment": "Supply-chain hardening starts with making CI inputs stable and least-privileged.",
      "connection": "Protocol compatibility is only meaningful if the harness itself is not mutable under your feet.",
      "specific_recommendation": "Preserve SHA pinning and read-only permissions as the default for protocol CI.",
      "decision_trail": "Record this as a workflow convention so future jobs do not regress to tag-based action refs."
    }
  ]
}
```


## Callouts
[Security] The CI hardening is genuinely strong: pinned actions plus read-only permissions show the harness is being treated as part of the protocol trust boundary, not just as convenience automation.

[Test Coverage] The volume and shape of fixture coverage is promising. The malformed cases around grafted digests, excess properties, and unsupported versions suggest the author is testing the negative space of the protocol, which is where compatibility contracts usually fail first.

[Quality] The split between collection, collection-resolution, and gate-leak packages points in the right direction. This keeps protocol concepts from becoming one large undifferentiated utility layer, and that separation will pay off as consumer harnesses multiply.

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#480. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Duplicate prepare script keys in root package.json",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "package.json:6",
      "description": "The root scripts object adds the same \"prepare\" key three times. JSON parsers typically keep only the last value, but duplicate keys make the manifest ambiguous and can cause inconsistent behavior across tooling that validates or rewrites package.json.",
      "suggestion": "Keep a single \"prepare\": \"./scripts/prepare-effect.sh\" entry in the scripts object.",
      "confidence": 0.98
    },
    {
      "id": "F-002",
      "title": "Strict JSON escape test uses an invalid escape",
      "severity": "MEDIUM",
      "category": "test",
      "file": "packages/protocol/collection/src/__tests__/harness.test.ts:51",
      "description": "The test case described as accepting escaped forms passes JSON containing \\^, which is not a valid JSON escape sequence. A strict JSON parser should reject it, so this test either fails or weakens strictness if made to pass.",
      "suggestion": "Replace the fixture with valid JSON escapes, for example \"{\\\"value\\\":\\\"\\\\u0000\\\\n\\\\t\\\"}\", and assert the parsed null/newline/tab characters.",
      "confidence": 0.86
    },
    {
      "id": "F-003",
      "title": "Compat workflow ignores harness implementation changes",
      "severity": "MEDIUM",
      "category": "ci",
      "file": ".github/workflows/collection-protocol-compat.yml:13",
      "description": "The workflow only runs on changes under packages/protocol/collection/** and the workflow file. This PR adds compatibility consumers and protocol packages under packages/protocol/collection-resolution/** and packages/protocol/gate-leak/**, but changes there will not trigger the compatibility harness even though they depend on and exercise the new contracts.",
      "suggestion": "Add the dependent protocol package paths, or a shared protocol glob such as packages/protocol/**, to the pull_request and push path filters if those changes must be covered by this CI entrypoint.",
      "confidence": 0.72
    },
    {
      "id": "F-004",
      "title": "Harness artifacts are built with git HEAD in tests",
      "severity": "LOW",
      "category": "test",
      "file": "packages/protocol/collection/src/__tests__/harness.test.ts:27",
      "description": "The harness tests derive source_commit from `git rev-parse HEAD`. In PR and local dirty-tree contexts, the files under test can differ from HEAD, while packArtifact is still asked to stamp the artifact with HEAD. That can let tests pass with a source_commit that does not actually identify the packed source state.",
      "suggestion": "Use an explicit test fixture repository for source-commit assertions, or have packArtifact/assertReachableSourceCommit reject dirty source trees when isolatedBuild is intended to publish a source-bound artifact.",
      "confidence": 0.62
    },
    {
      "id": "F-005",
      "title": "Protocol export surface is cleanly centralized",
      "severity": "PRAISE",
      "file": "packages/protocol/gate-leak/src/index.ts:1",
      "description": "The new gate-leak protocol exposes a single explicit barrel for version constants, schemas, decoders, digest helpers, policy declarations, and typed errors. That makes consumer imports auditable and keeps boundary APIs discoverable.",
      "suggestion": "Keep this explicit export discipline as the protocol grows.",
      "confidence": 0.9
    },
    {
      "id": "F-006",
      "title": "In-memory store protects persisted truth from aliasing",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/services/ordering/src/resolution-store.ts:179",
      "description": "The reference store seals stored records and replay snapshots with deepCloneFreeze before retaining or returning them. That is a strong guard against caller-side object mutation bypassing the CAS/idempotency boundary.",
      "suggestion": "Preserve the same clone/freeze or serialization boundary in the production Postgres adapter.",
      "confidence": 0.88
    }
  ]
}
```


## Callouts

PR #481, opened 2026-07-16
test(agent): repair baseline gateway contracts

zkSoju:
## Summary

- isolate the free-tier user sliding-window assertion from the stricter burst bucket without resetting user state
- exercise real AgentGateway stream-drop behavior and prove reconciliation scheduling after partial forwarding
- replace direct BullMQ facade coupling with the narrow typed queue capability the gateway actually uses

## Why

The current `agent-ci` baseline fails these same two tests on `main`, blocking the collection-report PR stack even though none of those PRs touches the agent subsystem. This repairs the shared gate without suppressing the job or changing runtime behavior.

## Verification

- `npx vitest run --config vitest.agent-ci-integration.config.ts` — 90 passed, 10 skipped
- agent unit suite — 448 passed, 1 skipped
- `packages/adapters: npm run typecheck` — pass
- targeted gateway integration test — 22/22, repeated five times
- `git diff --check` — pass

## Operational posture

Draft until current-head Fable Bridgebuilder review and GitHub checks are green.


zkSoju (commented):
## Summary

This PR does the quiet, high-leverage work of making a test suite tell the truth: the drop-recovery test now drives a real `AgentGateway` through a narrow `ReconciliationQueue` port instead of asserting on stub output, and the gateway's public API sheds its BullMQ dependency in the process. Nothing here blocks, but one decision deserves to be made deliberately rather than by accident — the new test pins undici's `TypeError('terminated')` as the error callers of `stream()` receive. **Verdict: COMMENT.**

The two PRAISE findings are one idea seen from two sides. The port (`agent-gateway.ts`) is what made the honest test (`agent-gateway.test.ts`) cheap to write — testability is a property of the architecture, not the test suite. The MEDIUM finding is Hyrum's Law in miniature: the moment CI asserts on a vendor's error message, that message *is* your contract, and a routine Node upgrade becomes a product incident. **Decision trail:** add a one-line note (PR description or follow-up issue) stating whether raw transport errors escaping `stream()` is intended baseline or acknowledged debt — future readers will otherwise treat the assertion as gospel. The three findings below 0.8 confidence come from the analytical pass without full runtime context; each names its own two-minute verification (does `BudgetManager` expose reservation lookup? are key-builders already exportable?) — worth confirming before filing follow-ups.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "drop-recovery-test-exercises-real-gateway",
      "title": "Drop-recovery test now exercises the gateway instead of the stub",
      "severity": "PRAISE",
      "category": "testing",
      "file": "tests/integration/agent-gateway.test.ts",
      "description": "The previous test fetched the stub endpoint directly and asserted on raw SSE text, so the gateway's drop-recovery path (event forwarding, error propagation, reconciliation scheduling) was never executed despite the test name claiming 'reconciliation fires'. The rewrite wires a real AgentGateway with real BudgetManager, AgentRateLimiter, and LoaFinnClient against the stub, consumes gateway.stream(), and asserts the forwarded events plus the exact reconciliation job payload and retry options. The typed capture of queue calls via Parameters<ReconciliationQueue['add']>[n] keeps the fake in sync with the interface.",
      "suggestion": "No change needed. This is the right shape for the remaining SSE proxy tests that still assert on raw stub output.",
      "confidence": 0.95,
      "faang_parallel": "Google's Testing on the Toilet series calls the old shape 'testing the mock' — a test that asserts on stub output verifies the stub, not the system. The rewrite is consumer-driven contract testing in the Pact tradition, with the fake's shape derived from the real interface so drift becomes a compile error.",
      "metaphor": "The old test checked the smoke detector by reading the box it shipped in; this one fills the room with smoke and listens for the alarm.",
      "teachable_moment": "A test's value equals the production code it forces to execute. Test names are claims and wiring is the proof — 'reconciliation fires' only means something when a real gateway is the thing firing it.",
      "connection": "This test only became cheap to write because of the ReconciliationQueue port (see reconciliation-queue-port-decouples-bullmq) — testability is an architecture property, purchased at design time."
    },
    {
      "id": "reconciliation-queue-port-decouples-bullmq",
      "title": "ReconciliationQueue structural interface cleanly replaces the BullMQ Queue dependency",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/adapters/agent/agent-gateway.ts",
      "description": "Replacing Queue<StreamReconciliationJob> with a narrow structural interface removes the bullmq type from the gateway's public API while remaining backward compatible: a real BullMQ queue still satisfies it (its add accepts string ⊇ 'stream-reconcile', the options object is a valid JobsOptions, and Promise<Job> is assignable to Promise<unknown>). Tests can now provide a dependency-free fake, and re-exporting the type from index.ts makes that available to consumers.",
      "suggestion": "No change needed. If production wiring passes a BullMQ Queue, a one-line type-level assignment test (const q: ReconciliationQueue = realQueue) would lock in the structural compatibility.",
      "confidence": 0.9,
      "faang_parallel": "Go's 'accept interfaces, return structs' discipline across Google's codebase: the consumer declares the narrowest interface it needs, and structural typing lets the vendor type satisfy it with no adapter class.",
      "metaphor": "Hardwiring BullMQ into the signature was wiring the lamp straight into the wall; the structural interface installs an outlet — same electricity, but now anything with the right plug works, including a test fixture.",
      "teachable_moment": "Define ports at the consumer, sized to what the consumer actually uses. The narrower the port, the more implementations satisfy it for free — and the less a vendor upgrade can break your public API.",
      "connection": "The suggested const q: ReconciliationQueue = realQueue assignment is TypeScript's equivalent of Go's var _ Interface = (*Impl)(nil) compile-time conformance check — one line that makes the structural compatibility a build guarantee."
    },
    {
      "id": "error-contract-pins-undici-internals",
      "title": "Baselined error contract pins undici's TypeError('terminated') as the gateway's public behavior",
      "severity": "MEDIUM",
      "category": "error-handling",
      "file": "tests/integration/agent-gateway.test.ts",
      "description": "Lines ~448-449 assert the error escaping gateway.stream() is a TypeError with message 'terminated'. That is undici's internal error for an abruptly terminated response body, not a gateway-owned error. This has two costs: (1) the assertion is brittle across Node/undici upgrades, which can change the error class or message and break CI without any product change; (2) the test now codifies that callers of gateway.stream() receive raw transport errors rather than a stable domain error, making downstream error handling depend on fetch-implementation details.",
      "suggestion": "In the gateway's stream error path (after scheduling reconciliation), wrap the transport failure in AgentGatewayError with a stable code (e.g. STREAM_INTERRUPTED) and the original error as cause, then assert on the domain error here. If this PR is intentionally baselining current behavior only, loosen the assertion to expect(streamError).toBeInstanceOf(Error) and file a follow-up for the wrapping.",
      "confidence": 0.7,
      "faang_parallel": "Hyrum's Law, coined at Google: with enough users, every observable behavior becomes a contract. Asserting a fetch-implementation's error message promotes it into the gateway's API — and undici error surfaces have shifted across Node releases before, breaking suites with zero product change.",
      "metaphor": "It's a contract that names the courier's van model instead of the delivery guarantee — accurate today, void the day the courier buys new vans.",
      "teachable_moment": "Errors are API surface. Wrap transport failures at the boundary into a domain error with a stable code and the original as cause — callers get a contract, debuggers keep the truth.",
      "connection": "The same anti-corruption instinct as the ReconciliationQueue port, applied to the error channel: this PR gave the happy path a clean seam, while the failure path still leaks the vendor through."
    },
    {
      "id": "tests-hardcode-redis-key-schemas",
      "title": "Tests duplicate internal Redis key formats as string literals",
      "severity": "LOW",
      "file": "tests/integration/agent-gateway.test.ts",
      "description": "The rate-limit test deletes `agent:rl:burst:${params.userId}` (~line 213) and the drop test seeds `agent:budget:limit:${communityId}` (~line 384). Both hardcode key schemas owned by AgentRateLimiter and BudgetManager. If either class renames its keys, the burst deletion becomes a no-op and the test fails at iteration 3 with a misleading 'allowed=false' assertion rather than an error pointing at the key coupling. The comment explaining why the burst bucket is cleared (isolating the user sliding-window dimension from free-tier burst capacity) is a good touch and should be kept either way.",
      "suggestion": "Export key-builder functions (or a small test helper) from the rate limiter and budget manager modules and use them in tests, so key-schema changes update both sides together.",
      "confidence": 0.6,
      "faang_parallel": "Meta's 'Scaling Memcache at Facebook' era centralized key construction for exactly this reason: hand-built key strings scattered across callers turn every rename into a silent-staleness hunt instead of a compile-time change.",
      "metaphor": "Two departments each keep their own copy of the building's door codes; when facilities rotates one, nobody's badge fails until the fire drill.",
      "teachable_moment": "When a test must touch internals, reach through the same function production uses. The coupling isn't the hazard — the duplication is: a renamed key should fail loudly at the key builder, not as a misleading assertion three steps downstream.",
      "connection": "Same single-source-of-truth instinct as the Parameters<ReconciliationQueue['add']> capture praised above — derive, don't transcribe."
    },
    {
      "id": "drop-test-missing-reservation-state-assertion",
      "title": "Drop-recovery contract does not assert the budget reservation is left pending",
      "severity": "LOW",
      "category": "testing",
      "file": "tests/integration/agent-gateway.test.ts",
      "description": "The test verifies the reconciliation job is enqueued with the expected payload, but not the Redis-side state the reconcile job depends on: that the budget reservation made before streaming remains held (neither settled nor refunded) after the drop. If a future change settles or releases the reservation on stream error, this test would still pass while the reconciliation job becomes a no-op or double-counts.",
      "suggestion": "After the stream error, assert the reservation is still pending — via the BudgetManager API if it exposes reservation lookup, or by checking the reservation key for the idempotencyKey in Redis.",
      "confidence": 0.5,
      "faang_parallel": "Payments reconciliation at Stripe rests on an invariant pair: the reconcile job AND the held state it settles against. Testing only the enqueue verifies one half of a saga's compensation step.",
      "metaphor": "Scheduling the tow truck without confirming the car is still parked where you said — the dispatch succeeds, the recovery does nothing.",
      "teachable_moment": "Async handoffs have two-sided contracts: the message sent, and the state its handler assumes. Assert both, or a refactor can keep the test green while hollowing out the recovery path.",
      "connection": "Closes the last gap that drop-recovery-test-exercises-real-gateway opened up — the queue side of the contract is now real; the state side still rides on faith."
    }
  ]
}
```


## Callouts

- **[Test Coverage]** The `Parameters<ReconciliationQueue['add']>[n]` typed capture is the standout small move: the fake derives its shape from the interface, so interface drift becomes a compile error instead of a silently stale test double. This is what "mock correctness" looks like when the type system does the enforcement.
- **[Quality]** The burst-bucket comment explains *why* the key is cleared (isolating the sliding-window dimension from free-tier burst capacity) — a constraint the code can't show on its own. That's precisely the kind of comment that earns its keep; preserve it through any key-builder refactor.
- **[Quality]** Re-exporting `ReconciliationQueue` from `index.ts` finishes the port properly: consumers and tests type against the seam, not the vendor. Many teams stop one step short and leave the interface buried in the implementation file.

zkSoju (commented):
## Summary
Verdict: `REQUEST_CHANGES`. The PR is moving the gateway toward cleaner ports and less Redis key drift, but the stream error path deserves one more pass: recovery code should not be able to obscure the failure that operators most need to see.

Gateway code is a contract with production incidents. The design question here is not only “does the test pass,” but “when the stream breaks at 03:00, which failure does the system preserve as truth?”

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Stream failures are now swallowed until cleanup succeeds",
      "severity": "MEDIUM",
      "category": "error-handling",
      "dimension": "[Operational]",
      "file": "packages/adapters/agent/agent-gateway.ts:570",
      "description": "The catch block stores the stream error and defers rethrowing until after the finally block. If lifecycle failure handling or reconciliation scheduling throws, the original upstream or transport failure is masked, changing the observable error and making diagnosis harder.",
      "suggestion": "Preserve the original stream error even if cleanup fails, for example by wrapping cleanup failures as suppressed/logged errors or by ensuring cleanup paths cannot replace the primary stream failure.",
      "faang_parallel": "Google SRE practice treats the first user-visible failure as the primary signal; cleanup failures are recorded as secondary context because replacing the root cause slows incident response.",
      "metaphor": "This is like an ambulance report that records the broken clipboard instead of the patient’s injury because the paperwork failed afterward.",
      "teachable_moment": "Cleanup is important, but it should not outrank the operation that failed. Preserve the primary error, then attach cleanup failures as suppressed errors, structured logs, or diagnostic metadata.",
      "connection": "This sits in the broader reliability pattern of causal error chains: recovery paths must be observable without rewriting history.",
      "decision_trail": "Document whether stream errors or reconciliation errors are intended to be primary when both occur, and encode that precedence in a focused test.",
      "confidence": 0.78
    },
    {
      "id": "F2",
      "title": "Reconciliation is scheduled for non-transport upstream errors",
      "severity": "LOW",
      "category": "behavioral-risk",
      "dimension": "[Quality]",
      "file": "packages/adapters/agent/agent-gateway.ts:580",
      "description": "The finally block schedules reconciliation whenever the stream was not finalized, regardless of whether the failure was a dropped stream or an explicit LoaFinnError such as HTTP 503. The new test verifies the LoaFinnError is preserved, but it does not assert that reconciliation is not queued for ordinary upstream failures.",
      "suggestion": "Gate reconciliation on the same failure class intended for recovery, or add tests that document why reconciliation should also run for explicit upstream HTTP failures.",
      "faang_parallel": "Netflix-style resilience systems distinguish retryable transport failures from explicit upstream refusals; mixing them can create avoidable retry pressure during partial outages.",
      "metaphor": "It is the difference between redialing after a dropped call and redialing after someone clearly said they are unavailable.",
      "teachable_moment": "Recovery queues should encode intent, not just absence of completion. A failure taxonomy keeps compensating work from becoming accidental traffic amplification.",
      "connection": "This connects to idempotency and backpressure: reconciliation is powerful, but only when scheduled for failures it is meant to heal.",
      "decision_trail": "Document which failure classes are reconciliation-worthy and add one assertion for explicit upstream errors so future agents do not infer policy from absence.",
      "confidence": 0.66
    },
    {
      "id": "F3",
      "title": "Redis key builders remove duplicated key literals",
      "severity": "PRAISE",
      "dimension": "[Quality]",
      "file": "packages/adapters/agent/budget-manager.ts:81",
      "description": "Exporting shared Redis key builders and using them from BudgetManager, AgentGateway, and tests reduces drift between production behavior and assertions.",
      "suggestion": "Keep future budget-related Redis access routed through these helpers.",
      "faang_parallel": "Meta’s large PHP/Hack codebases leaned heavily on shared typed helpers for storage keys because duplicated string protocols become migration hazards at scale.",
      "metaphor": "This turns several handwritten mailing labels into one address book entry.",
      "teachable_moment": "When tests and production code share a protocol helper, assertions track behavior instead of reimplementing fragile string folklore.",
      "connection": "This is the right direction for adapter boundaries: key shape becomes a small contract rather than ambient convention.",
      "decision_trail": "Keep the key format decision near these helpers, including the namespace and compatibility expectations for existing Redis data.",
      "confidence": 0.95
    },
    {
      "id": "F4",
      "title": "Narrow queue port improves testability",
      "severity": "PRAISE",
      "category": "design",
      "dimension": "[Quality]",
      "file": "packages/adapters/agent/agent-gateway.ts:52",
      "description": "Replacing the direct BullMQ Queue dependency with a small ReconciliationQueue interface makes AgentGateway easier to unit test and reduces coupling to the queue implementation.",
      "suggestion": "Keep the port limited to the operations AgentGateway actually needs.",
      "faang_parallel": "The Borg and Kubernetes control-plane tradition favors narrow interfaces around schedulers and queues so core orchestration logic can be tested without the whole substrate.",
      "metaphor": "The gateway now uses a wall switch instead of reaching into the breaker panel.",
      "teachable_moment": "A small port is not abstraction for its own sake; it is a seam that lets failure policy be tested without dragging infrastructure into every test.",
      "connection": "This supports the same architectural direction as the Redis helper work: make contracts explicit at the adapter edge.",
      "decision_trail": "Document that this interface is intentionally minimal and should grow only from gateway use cases, not from BullMQ’s full API surface.",
      "confidence": 0.92
    }
  ]
}
```


## Callouts
[Quality] The Redis key-builder extraction is strong because it removes a quiet class of future test lies: production and tests now speak through the same naming contract.

[Quality] The `ReconciliationQueue` port is a clean design move. It keeps the gateway focused on policy while letting the queue implementation remain replaceable infrastructure.

[Test Coverage] The added integration coverage around preserving `LoaFinnError` is valuable. The remaining gap is not volume, but policy precision: which failures should reconcile, and which should simply surface.

zkSoju (commented):
## Summary

Verdict: COMMENT. This PR strengthens the agent gateway’s executable contract: the strongest signal is that tests now drive the real facade where budget, streaming, error preservation, and reconciliation meet. The architectural direction is also cleaner because Redis budget key construction is moving behind named builders instead of scattered string formats.

The broader pattern here is contract gravity: gateways and budget ledgers become safer when the test boundary matches the production boundary, and when persistence keys are treated as API surface rather than incidental strings.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "praise-001",
      "title": "Gateway streaming tests now exercise the real facade contract",
      "severity": "PRAISE",
      "category": "test-coverage",
      "file": "tests/integration/agent-gateway.test.ts:355",
      "description": "The new streaming tests drive AgentGateway directly, assert partial event forwarding, reconciliation enqueue payloads, pending reservations, HTTP-domain error preservation, and enqueue-failure behavior. This is a stronger contract than only checking the stub HTTP response.",
      "suggestion": "Keep this style for gateway behavior tests where budget, rate limiting, upstream errors, and reconciliation interact.",
      "confidence": 0.95,
      "faang_parallel": "Google SRE practice treats service-level behavior as the contract, not just handler return values; these tests similarly verify the gateway boundary where multiple subsystems converge.",
      "metaphor": "This is testing the airport control tower during real takeoff coordination, not just checking whether one runway light turns on.",
      "teachable_moment": "The most valuable integration tests usually sit at coordination points: places where independently-correct components can still fail together.",
      "connection": "This reinforces the facade as the stable behavioral boundary for streaming, budget reservation, error semantics, and reconciliation side effects.",
      "decision_trail": "Document which gateway behaviors are contractual: partial event forwarding, pending budget reservation shape, domain error preservation, and enqueue-failure semantics. Future maintainers need to know these are not incidental assertions."
    },
    {
      "id": "praise-002",
      "title": "Redis budget key builders reduce cross-file key drift",
      "severity": "PRAISE",
      "file": "packages/adapters/agent/budget-manager.ts:77",
      "description": "Centralizing budget Redis key construction and reusing it from AgentGateway and tests removes duplicated string formats for committed, reserved, limit, reservation, expiry, and finalized keys.",
      "suggestion": "Continue routing new budget-related Redis access through these builders instead of inline string templates.",
      "confidence": 0.95,
      "faang_parallel": "Amazon’s large service fleets commonly treat storage key formats as versioned interface contracts; central builders give the same protection at repo scale.",
      "metaphor": "A Redis key format is a street address. If every team writes addresses by hand, mail eventually lands at the wrong door.",
      "teachable_moment": "Persistence naming is not plumbing trivia; once multiple files depend on it, it becomes shared API surface and deserves a single owner.",
      "connection": "This reduces hidden coupling between gateway logic, budget accounting, and tests while making future key migration or namespacing changes mechanically safer.",
      "decision_trail": "Add a short note near the builders describing the key namespace and compatibility expectations, especially whether existing keys are migration-sensitive."
    }
  ]
}
```


## Callouts

[Test Coverage] The real win is not just “more tests”; it is better placement. By testing `AgentGateway` directly, the suite now watches the contract where streaming, budgets, upstream HTTP errors, and reconciliation interact.

[Quality] The Redis key builders are a strong maintainability move because they turn string convention into named code. That makes drift visible in review instead of discovering it later through missing budget records.

[Operational] The enqueue-failure coverage is especially valuable. Systems fail at handoff points, and this PR gives that handoff explicit behavioral weight.

zkSoju:
## CI adjudication (Codex wave)

Head `450c9047cdc4b1bbad7db8b97ff2923bcee66c06` remains BEHIND with red **agent-ci / floor / Security Scan / Integration Tests / PR Summary**.

Prior wave evidence (runbook): Agent CI contract repairs in this PR are green locally (90 agent-ci tests); remaining reds reproduce shared baseline debt tracked in #477 / #483 / #484 — not regressions introduced by the agent-gateway contract fix.

Still required: exact-current human APPROVE + `loa-signoff` + operator update-from-main. External collection-report descendants must not treat these reds as CR-specific blockers without re-checking `#477` baseline.

PR #497, opened 2026-07-17
feat(protocol): CR-009 signed trust-envelope protocol

zkSoju:
## Summary
- Adds `@freeside/trust-envelope-protocol` (`packages/protocol/trust-envelope`) ratifying CR-009 Ordering trust envelopes distinct from events-pillar `acvp-l1-v2`.
- Headers bind producer, signing key, contract, event ID, stream ID, epoch/sequence, issued/expiry, tenant scope digest, capability, and body digest; strict decoders fail closed on unknown major/required-field violations with explicit mixed-minor rules.
- Shared producer/consumer fixtures cover rotation, revocation, replay, expiry, gap repair, epoch baseline reset, and disaster-recovery refusal paths; Sonar producer and Ordering consumer tests consume the same committed vectors.

## Test plan
- [x] `cd packages/protocol/trust-envelope && pnpm test` (10/10)
- [x] `cd packages/protocol/trust-envelope && pnpm typecheck`
- [ ] CR-011A Sonar producer adoption (downstream)
- [ ] CR-012A Ordering dependency ledger wiring (downstream)
- [ ] CR-013 production signing-key custody (downstream)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/protocol/trust-envelope/README.md
+++ b/packages/protocol/trust-envelope/README.md
@@ -1,0 +1,28 @@
+# `@freeside/trust-envelope-protocol`
+
+CR-009 signed trust-envelope wire contract for collection-report Ordering
+dependency intake. Distinct from the events-pillar `acvp-l1-v2` NATS envelope.
+
+The package owns:
+
+- strict header/body envelope schemas and decoders;
+- Ed25519 signing bytes (JCS + SHA-256 digest, empty-string signature placeholder);
+- fixture service-key registry binding (producer, capability, tenant scope);
+- producer signing helpers and consumer inbox/stream verification;
+- shared rotation, revocation, replay, expiry, gap-repair, retention, and
+  disaster-recovery fixtures consumed by planned producers and Ordering replay.
+
+Canonical rules:
+
+- header fields bind producer, signing key, contract, event ID, stream ID,
+  epoch/sequence, issued/expiry, tenant/privacy scope digest, capability, and
+  body digest;
+- `event_id` is the idempotency key; contiguous sequence is enforced only when
+  `trust_stream` is true;
+- epoch reset requires a signed complete baseline before a new epoch is accepted;
+- unknown contract major or required-field violation fails closed; mixed-minor
+  acceptance is explicit in `versioning.ts`;
+- Ordering database `accepted_at` evaluates issued/expiry (30s future skew).
+
+Production signing-key custody remains CR-013. This package publishes wire
+semantics and non-production fixture keys only.
--- a/packages/protocol/trust-envelope/fixtures/scenarios/producer-consumer.shared.json
+++ b/packages/protocol/trust-envelope/fixtures/scenarios/producer-consumer.shared.json
@@ -1,0 +1,315 @@
+{
+  "schema_version": 1,
+  "accepted_at_ms": 1784322001000,
+  "registry": {
+    "schema_version": 1,
+    "keys": [
+      {
+        "signing_key_id": "sonar-fixture-primary",
+        "public_key_hex": "3b6a27bcceb6a42d62a3a8d02a6f0d73653215771de243a63ac048a18b59da29",
+        "producer": "sonar-api",
+        "capabilities": [
+          "collection-report.capability-evidence.v1"
+        ],
+        "tenant_scope_digests": [
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+        ],
+        "activated_at": "2026-07-01T00:00:00.000Z"
+      },
+      {
+        "signing_key_id": "sonar-fixture-rotated",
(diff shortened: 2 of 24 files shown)


zkSoju (commented):
# Bridgebuilder review — CR-009 trust-envelope protocol

**HEAD:** `0804dfb1907ced2aec1e619c26656ffdab3357b9`
**Scope:** `packages/protocol/trust-envelope` only (greenfield package)
**Verdict: merge-BLOCKED** — consumer gap-repair state machine is incorrect; do not land until fixed + tested.

## Security

**Ed25519 / binding — OK**
- Algorithm locked to `Ed25519` via Effect schema + `@noble/curves`.
- TBS = SHA-256(JCS(envelope with empty signature)); body bound via JCS digest in header; `signTrustEnvelope` rewrites `signing_key_id` + `body_digest` before signing so producer cannot desync.
- Verify path checks body digest, key registry (active/revoked/compromise), producer/capability/tenant bindings, issued/expiry skew, then signature — fail-closed on invalid crypto.

**Fail-closed unknown major — OK (with taxonomy note)**
- `assertSupportedSchemaMajor` rejects `schema_version !== 1` before signature verify (covered by protocol test).
- Strict decode uses `onExcessProperty: "error"`; mixed-minor rules document higher-minor fail-closed.
- Note (non-blocking): unknown major/minor throw bare `Error`, not `TrustEnvelopeRejectedError` (`unknown_schema_major` / `schema_minor_unsupported` exist but unused). Wire still fail-closed; typed reject surface is inconsistent.

**Replay / idempotency — PARTIAL (blocking interaction with gaps)**
- Duplicate `event_id` rejects with `event_id_replay` — good for simple replay.
- **Blocking:** `ingestTrustEnvelope` adds `event_id` to `seenEventIds` *before* sequence checks. On `sequence_gap`, it returns mutated state that already recorded the premature envelope's `event_id`. Gap-fill redelivery of that same envelope then fails as `event_id_replay`, so the advertised repair path cannot accept the out-of-order envelope after missing sequences arrive.
- `replayEnvelopeIdempotently` does not implement idempotent accept (`replay: true` is never set); it only event-id-matches then calls `ingestTrustEnvelope` (which rejects seen IDs). Dead / misleading API for CR-012A consumers.

## Quality — BLOCKING

In `consumer.ts`, `gaps` is populated with *missing* sequences when `sequence > expected`, and `requestGapRepairRange` correctly exposes that range. But after accepting the next contiguous sequence, this loop treats gap membership as “already filled”:

```ts
while (state.gaps.has(state.highestContiguousSequence + 1)) {
  state.highestContiguousSequence += 1;
  state.gaps.delete(state.highestContiguousSequence);
}
```

Example: after seq 1–2 accepted, seq 6 arrives → gaps `{3,4,5}`, reject. Later accept seq 3 → loop advances contiguous through 4 and 5 without ever verifying those envelopes. That breaks trust-stream contiguity and makes “gap repair” unsafe.

**Required fix direction**
1. Do not commit `event_id` / retention stamps on sequence reject (or buffer OO envelopes separately from `seenEventIds`).
2. Either remove the while-advance-over-gaps loop, or flip the model to an OO *buffer* of received sequences (advance only when next buffered seq is present).
3. Add an end-to-end test: gap reject → repair range → ingest missing seqs → re-ingest the original higher seq → accept with contiguous highest.

Minor (non-blocking): producer mismatch maps to `capability_not_bound`; fixture seeds exported from package index are clearly non-prod (CR-013).

## Tests

- Package fixtures + Sonar/Ordering contract tests cover accept/verify reject (revoke, expiry, tamper), ingest rejects (gap, epoch-without-baseline), replay of already-accepted id, and epoch baseline install + resume-forbidden.
- **Gap:** no test exercises post-gap *acceptance* of repaired sequences — so the while-loop bug and premature `event_id` commit are uncaught.
- CI: **Unit Tests pass** for this PR. Integration Tests fail on pre-existing billing `initDatabase` / webhook path; agent-ci fails on unrelated agent-gateway SSE assertion. Those failures are out of scope for CR-009 and would not alone block merge.

## Explicit out-of-scope (OK for this PR)

- **CR-011A** Sonar producer adoption — downstream
- **CR-012A** Ordering dependency ledger wiring — downstream (but must not inherit the broken consumer helper as-is)
- **CR-013** production signing-key custody — downstream; fixture seeds appropriately labeled

## Merge decision

**Do not merge** until gap-repair state + `event_id` commit ordering are fixed and covered by an end-to-end ingest test. Crypto/verify/decode core of CR-009 looks sound; the shared consumer helper is not yet safe for Ordering adoption.

zkSoju (commented):
## Re-review after gap-repair fix (`0791a090`)

**Verdict: merge OK**

Blocking findings from prior review addressed in `0791a090`:
1. Trust-stream ingest accepts **only** `sequence === highestContiguous + 1` — no while-loop skip across gaps.
2. `sequence_gap` records missing sequences **without** committing `event_id` — redelivery is not poisoned as `event_id_replay`.
3. Regression assertions added in producer-consumer suite.

Unit package tests 10/10. Unrelated Integration/agent-ci flakes remain out of CR-009 scope.

PR #501, opened 2026-07-17
feat(ordering): CR-201A public shared preparation persistence

zkSoju:
## Summary

- Adds expand-only migration `007_shared_preparation_work.sql` for `shared_preparation_work`, `preparation_work_items`, and `report_work_links` with public-only capability CHECK constraints (`collection_identity.v1`, `ownership_index.v1`) and partial unique index on active work keys.
- Implements in-memory + Postgres store ports with transactional join/create, lease reclaim fencing (`lease_epoch`), retry CAS (`retry_wait` → `queued`), subscriber fan-in (100 joiners → one active row), zero-subscriber abandonment, and multi-deployment child evidence gating before parent `ready`.
- Exports shared-preparation types/service from `@freeside/ordering-service` for downstream CR-201C/CR-204A wiring.

## Test plan

- [x] `pnpm exec vitest run src/__tests__/shared-preparation-migration.test.ts src/__tests__/shared-preparation-store.test.ts` (9 tests)
- [x] Migration expand-only / rollback-safety static check
- [x] 100 concurrent equivalent joiners → one active work row + 100 links
- [x] Lease reclaim increments epoch; stale fenced child write rejected
- [x] Retry scheduling + stale wake rejection
- [x] Subscriber detachment → deterministic abandonment
- [x] Supersession before successor generation
- [x] Parent ready only after all child deployment evidence

## Gaps (honest)

- Not yet wired into collection-report intake / orchestrator (orders still idle in `producing` without shared-work links).
- No live Postgres integration test harness in CI (Postgres adapter is mechanical behind the same port).
- CR-201C admission capacity reservation and CR-204A Sonar dispatch remain downstream.
- 100–1000 joiner Postgres latency proof and mixed-version migration tests deferred to CR-201C/CR-209A.
- No customer Gate Leak artifact / no restricted Discord or identity evidence (by design for T1).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/services/ordering/migrations/007_shared_preparation_work.sql
+++ b/packages/services/ordering/migrations/007_shared_preparation_work.sql
@@ -45,7 +45,8 @@ CREATE TABLE IF NOT EXISTS shared_preparation_work (
   CONSTRAINT shared_preparation_work_unique_generation UNIQUE (work_key_digest, generation)
 );
 
-CREATE INDEX IF NOT EXISTS shared_preparation_work_key_active_idx
+-- Active is exactly queued/preparing/retry_wait; one active row per work key.
+CREATE UNIQUE INDEX IF NOT EXISTS shared_preparation_work_key_active_idx
   ON shared_preparation_work (work_key_digest)
   WHERE state IN ('queued', 'preparing', 'retry_wait');
 
--- a/packages/services/ordering/migrations/007_shared_preparation_work.sql
+++ b/packages/services/ordering/migrations/007_shared_preparation_work.sql
@@ -1,0 +1,100 @@
+-- CR-201A expand: public shared preparation persistence (T1 fixtures only).
+-- Expand-only: no destructive DDL. Restricted Discord/identity evidence is CR-201B.
+
+CREATE TABLE IF NOT EXISTS shared_preparation_work (
+  work_id                   TEXT PRIMARY KEY,
+  work_key_digest           TEXT NOT NULL,
+  deployment_set_digest     TEXT NOT NULL,
+  capability                TEXT NOT NULL,
+  capability_version        TEXT NOT NULL,
+  scope_class               TEXT NOT NULL DEFAULT 'deployment',
+  scope_digest              TEXT NOT NULL,
+  privacy_class             TEXT NOT NULL DEFAULT 'public_chain',
+  source_identity           JSONB NOT NULL,
+  readiness_policy_version  TEXT NOT NULL,
+  evidence_boundary_kind    TEXT NOT NULL,
+  evidence_boundary_digest  TEXT,
+  adapter_version           TEXT NOT NULL,
+  finality_policy_version   TEXT NOT NULL,
+  state                     TEXT NOT NULL,
+  generation                INTEGER NOT NULL,
+  readiness_evidence        JSONB,
+  attempt                   INTEGER NOT NULL DEFAULT 0,
+  next_attempt_at           TIMESTAMPTZ,
+  retry_deadline            TIMESTAMPTZ,
+  lease_until               TIMESTAMPTZ,
+  lease_epoch               BIGINT NOT NULL DEFAULT 0,
+  failure_reason            JSONB,
+  sharing_scope_kind        TEXT NOT NULL DEFAULT 'public',
+  work_tenant_scope_digest  TEXT,
+  created_at                TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  updated_at                TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  CONSTRAINT shared_preparation_work_state_check CHECK (
+    state IN ('queued', 'preparing', 'retry_wait', 'ready', 'failed', 'abandoned', 'superseded')
+  ),
+  CONSTRAINT shared_preparation_work_privacy_public_only CHECK (
+    privacy_class = 'public_chain'
+  ),
+  CONSTRAINT shared_preparation_work_sharing_scope_public CHECK (
+    sharing_scope_kind = 'public' AND work_tenant_scope_digest IS NULL
+  ),
+  CONSTRAINT shared_preparation_work_capability_public CHECK (
+    capability IN ('collection_identity.v1', 'ownership_index.v1')
+  ),
+  CONSTRAINT shared_preparation_work_generation_positive CHECK (generation >= 1),
+  CONSTRAINT shared_preparation_work_unique_generation UNIQUE (work_key_digest, generation)
+);
+
+CREATE INDEX IF NOT EXISTS shared_preparation_work_key_active_idx
+  ON shared_preparation_work (work_key_digest)
+  WHERE state IN ('queued', 'preparing', 'retry_wait');
+
+CREATE INDEX IF NOT EXISTS shared_preparation_work_ready_idx
+  ON shared_preparation_work (work_key_digest, generation DESC)
+  WHERE state = 'ready';
+
+CREATE TABLE IF NOT EXISTS preparation_work_items (
+  work_item_id              TEXT PRIMARY KEY,
+  work_id                   TEXT NOT NULL REFERENCES shared_preparation_work (work_id),
+  deployment_id             JSONB NOT NULL,
+  capability                TEXT NOT NULL,
+  adapter_version           TEXT NOT NULL,
+  external_job_ref          TEXT,
+  state                     TEXT NOT NULL,
+  attempt                   INTEGER NOT NULL DEFAULT 0,
+  lease_epoch               BIGINT NOT NULL DEFAULT 0,
+  evidence_envelope         JSONB,
+  failure_reason            JSONB,
+  created_at                TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  updated_at                TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  CONSTRAINT preparation_work_items_state_check CHECK (
+    state IN ('queued', 'preparing', 'retry_wait', 'ready', 'failed', 'abandoned', 'superseded')
+  ),
+  CONSTRAINT preparation_work_items_unique_deployment UNIQUE (work_id, deployment_id)
+);
+
+CREATE INDEX IF NOT EXISTS preparation_work_items_work_idx
+  ON preparation_work_items (work_id);
+
+CREATE TABLE IF NOT EXISTS report_work_links (
+  order_id                  TEXT NOT NULL REFERENCES orders (order_id),
+  work_id                   TEXT NOT NULL REFERENCES shared_preparation_work (work_id),
+  joined_at                 TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  detached_at               TIMESTAMPTZ,
+  generation                INTEGER NOT NULL,
+  order_tenant_scope_digest TEXT NOT NULL,
+  sharing_scope_kind        TEXT NOT NULL DEFAULT 'public',
+  work_tenant_scope_digest  TEXT,
+  PRIMARY KEY (order_id, work_id),
+  CONSTRAINT report_work_links_public_scope CHECK (
+    sharing_scope_kind = 'public' AND work_tenant_scope_digest IS NULL
+  )
+);
+
+CREATE INDEX IF NOT EXISTS report_work_links_work_active_idx
+  ON report_work_links (work_id)
+  WHERE detached_at IS NULL;
+
+CREATE INDEX IF NOT EXISTS report_work_links_order_active_idx
+  ON report_work_links (order_id)
+  WHERE detached_at IS NULL;
--- a/packages/services/ordering/src/__tests__/shared-preparation-fixtures.ts
+++ b/packages/services/ordering/src/__tests__/shared-preparation-fixtures.ts
@@ -1,0 +1,87 @@
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { createHash } from "node:crypto";
+import { decodeCollectionCandidate } from "@freeside/collection-protocol";
+import { Effect } from "effect";
+import {
+  buildPublicWorkKeyMaterial,
+  digestPublicWorkKey,
+} from "../shared-preparation-work-key.js";
+import type {
+  PublicPreparationWorkKeyMaterial,
+  ReadinessEvidenceEnvelope,
+  VersionedDigest,
+} from "../shared-preparation-types.js";
+
+const fixtureDir = join(
+  dirname(fileURLToPath(import.meta.url)),
+  "../../../../protocol/collection/fixtures",
+);
+
+function runEffect<A>(effect: Effect.Effect<A, unknown>): A {
+  return Effect.runSync(effect);
+}
+
+export function loadEvmFixtureCandidate() {
+  const raw = JSON.parse(readFileSync(join(fixtureDir, "evm-candidate.valid.json"), "utf8"));
+  return runEffect(decodeCollectionCandidate(raw));
+}
+
+export function fixturePublicWorkKey(input?: {
+  capability?: "collection_identity.v1" | "ownership_index.v1";
+  communityScopeDigest?: string;
+}): PublicPreparationWorkKeyMaterial {
+  const candidate = loadEvmFixtureCandidate();
+  const deploymentIds = candidate.identity.deployments.map((d) => d.deployment_id);
+  return buildPublicWorkKeyMaterial({
+    capability: input?.capability ?? "ownership_index.v1",
+    capability_version: "v1",
+    collection_id: candidate.identity.collection_id,
+    deployment_ids: deploymentIds,
+    finality_policies: candidate.finality_policies,
+    source_identity: {
+      schema_version: 1,
+      producer: "sonar-api.fixture",
+      upstream_evidence_source: "sonar.public-capability.v1",
+    },
+    readiness_policy_version: "gate-leak-public-prep.v1",
+    adapter_version: "sonar-kitchen.v1",
+  });
+}
+
+export function fixtureCommunityScopeDigest(communityRef: string): string {
+  return createHash("sha256").update(`community:${communityRef}`).digest("hex");
+}
+
+export function fixtureReadinessEvidence(deploymentIds: readonly VersionedDigest[]): ReadinessEvidenceEnvelope {
+  const digest = (domain: string, seed: string): VersionedDigest => ({
+    algorithm: "sha-256",
+    domain,
+    major_version: 1,
+    digest: createHash("sha256").update(seed).digest("hex"),
+  });
+  return {
+    schema_version: 1,
+    producer: "sonar-api.fixture",
+    schema: "collection.public-evidence.v1",
+    adapter: "sonar-kitchen.v1",
+    readiness_policy_version: "gate-leak-public-prep.v1",
+    privacy_scope: "public_chain",
+    deployment_coverage: [...deploymentIds],
+    observation_window: {
+      observed_at: "2026-07-17T21:00:00.000Z",
+      as_of: "2026-07-17T21:00:00.000Z",
+    },
+    freshness: {
+      qualified: true,
+      max_age_ms: 60_000,
+    },
+    source_digest: digest("collection.evidence-source", "fixture-source"),
+    evidence_digest: digest("collection.evidence", "fixture-evidence"),
+  };
+}
+
+export function fixtureWorkKeyDigest(key: PublicPreparationWorkKeyMaterial): string {
+  return digestPublicWorkKey(key);
+}
--- a/packages/services/ordering/src/__tests__/shared-preparation-migration.test.ts
+++ b/packages/services/ordering/src/__tests__/shared-preparation-migration.test.ts
@@ -12,7 +12,12 @@ describe("CR-201A migration rollback safety", () => {
     expect(sql).toMatch(/CREATE TABLE IF NOT EXISTS shared_preparation_work/);
     expect(sql).toMatch(/CREATE TABLE IF NOT EXISTS preparation_work_items/);
     expect(sql).toMatch(/CREATE TABLE IF NOT EXISTS report_work_links/);
-    expect(sql).toMatch(/shared_preparation_work_key_active_idx/);
+    expect(sql).toMatch(
+      /CREATE UNIQUE INDEX IF NOT EXISTS shared_preparation_work_key_active_idx/,
+    );
+    expect(sql).toMatch(
+      /WHERE state IN \('queued', 'preparing', 'retry_wait'\)/,
+    );
     expect(sql).toMatch(/capability IN \('collection_identity\.v1', 'ownership_index\.v1'\)/);
     expect(sql).toMatch(/privacy_class = 'public_chain'/);
   });
--- a/packages/services/ordering/src/__tests__/shared-preparation-migration.test.ts
+++ b/packages/services/ordering/src/__tests__/shared-preparation-migration.test.ts
@@ -1,0 +1,19 @@
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+
+describe("CR-201A migration rollback safety", () => {
+  it("007_shared_preparation_work.sql is expand-only", () => {
+    const dir = dirname(fileURLToPath(import.meta.url));
+    const sql = readFileSync(join(dir, "../../migrations/007_shared_preparation_work.sql"), "utf8");
+    expect(sql).not.toMatch(/\bDROP\b/i);
+    expect(sql).not.toMatch(/\bTRUNCATE\b/i);
+    expect(sql).toMatch(/CREATE TABLE IF NOT EXISTS shared_preparation_work/);
+    expect(sql).toMatch(/CREATE TABLE IF NOT EXISTS preparation_work_items/);
+    expect(sql).toMatch(/CREATE TABLE IF NOT EXISTS report_work_links/);
+    expect(sql).toMatch(/shared_preparation_work_key_active_idx/);
+    expect(sql).toMatch(/capability IN \('collection_identity\.v1', 'ownership_index\.v1'\)/);
+    expect(sql).toMatch(/privacy_class = 'public_chain'/);
+  });
+});
--- a/packages/services/ordering/src/__tests__/shared-preparation-postgres.test.ts
+++ b/packages/services/ordering/src/__tests__/shared-preparation-postgres.test.ts
@@ -1,0 +1,243 @@
+/**
+ * CR-201A Postgres behavioral suite (F4).
+ *
+ * Gated on ORDERING_PG_TEST_URL (or DATABASE_URL when ORDERING_PG_TEST=1).
+ * Skips cleanly when unset — no Docker required for unit CI.
+ *
+ *   ORDERING_PG_TEST_URL=postgres://… pnpm exec vitest run \
+ *     src/__tests__/shared-preparation-postgres.test.ts
+ */
+
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import pg from "pg";
+import { PostgresSharedPreparationStore } from "../shared-preparation-store-postgres.js";
+import { digestPublicWorkKey } from "../shared-preparation-work-key.js";
+import {
+  fixtureCommunityScopeDigest,
+  fixturePublicWorkKey,
+  fixtureReadinessEvidence,
+} from "./shared-preparation-fixtures.js";
+
+const pgUrl =
+  process.env.ORDERING_PG_TEST_URL ??
+  (process.env.ORDERING_PG_TEST === "1" ? process.env.DATABASE_URL : undefined);
+
+const describePg = describe.skipIf(!pgUrl);
+
+async function seedOrder(pool: pg.Pool, orderId: string): Promise<void> {
+  const now = Math.floor(Date.now() / 1000);
+  await pool.query(
+    `INSERT INTO orders (
+      order_id, product, placed_by, inputs, inputs_digest, state,
+      placed_at_unix, created_at_unix, updated_at_unix
+    ) VALUES ($1, 'collection-report', 'subject-fixture', '{}'::jsonb, 'digest', 'placed', $2, $2, $2)
+    ON CONFLICT (order_id) DO NOTHING`,
+    [orderId, now],
+  );
+}
+
+describePg("CR-201A Postgres shared preparation store", () => {
+  let store: PostgresSharedPreparationStore;
+  let pool: pg.Pool;
+
+  beforeAll(async () => {
+    store = await PostgresSharedPreparationStore.connect(pgUrl!, { migrate: true });
+    pool = store.getPool();
+    // Ensure unique index statement is present in applied migration text.
+    const migration = readFileSync(
+      join(dirname(fileURLToPath(import.meta.url)), "../../migrations/007_shared_preparation_work.sql"),
+      "utf8",
+    );
+    expect(migration).toMatch(/CREATE UNIQUE INDEX IF NOT EXISTS shared_preparation_work_key_active_idx/);
+  }, 60_000);
+
+  afterAll(async () => {
+    await store.close();
+  });
+
+  it("joins, leases, retries, abandons, and reaches ready under fencing", async () => {
+    const workKey = fixturePublicWorkKey({ capability: "collection_identity.v1" });
+    const orderA = `ord_pg_a_${Date.now()}`;
+    const orderB = `ord_pg_b_${Date.now()}`;
+    await seedOrder(pool, orderA);
+    await seedOrder(pool, orderB);
+
+    const joined = await store.joinPublicWork({
+      order_id: orderA,
+      order_tenant_scope_digest: fixtureCommunityScopeDigest("community-alpha"),
+      work_key: workKey,
+      now_ms: Date.now(),
+    });
+    expect(joined.kind).toBe("joined");
+    if (joined.kind !== "joined") return;
+
+    const fanIn = await store.joinPublicWork({
+      order_id: orderB,
+      order_tenant_scope_digest: fixtureCommunityScopeDigest("community-alpha"),
+      work_key: workKey,
+      now_ms: Date.now(),
+    });
+    expect(fanIn.kind).toBe("joined");
+    if (fanIn.kind !== "joined") return;
+    expect(fanIn.work.work_id).toBe(joined.work.work_id);
+    expect(fanIn.created).toBe(false);
+
+    const lease = await store.acquireLease({
+      work_id: joined.work.work_id,
+      worker_id: "pg-worker",
+      lease_duration_ms: 30_000,
+      now_ms: Date.now(),
+    });
+    expect(lease.kind).toBe("acquired");
+    if (lease.kind !== "acquired") return;
+
+    await store.transitionToPreparing({
+      work_id: joined.work.work_id,
+      expected_lease_epoch: lease.work.lease_epoch,
+      now_ms: Date.now(),
+    });
+
+    const failed = await store.recordRetryableFailure({
+      work_id: joined.work.work_id,
+      expected_lease_epoch: lease.work.lease_epoch,
+      next_attempt_at_unix_ms: Date.now() - 1,
+      retry_deadline_unix_ms: Date.now() + 86_400_000,
+      failure: { code: "dependency_unavailable", reason: "pg fixture" },
+      now_ms: Date.now(),
+    });
+    expect(failed.state).toBe("retry_wait");
+
+    const woke = await store.wakeRetryWait({
+      work_id: joined.work.work_id,
+      expected_attempt: failed.attempt,
+      now_ms: Date.now(),
+    });
+    expect(woke.kind).toBe("woke");
+
+    const lease2 = await store.acquireLease({
+      work_id: joined.work.work_id,
+      worker_id: "pg-worker-2",
+      lease_duration_ms: 30_000,
+      now_ms: Date.now(),
+    });
+    expect(lease2.kind === "acquired" || lease2.kind === "reclaimed").toBe(true);
+    if (lease2.kind !== "acquired" && lease2.kind !== "reclaimed") return;
+
+    await store.transitionToPreparing({
+      work_id: joined.work.work_id,
+      expected_lease_epoch: lease2.work.lease_epoch,
+      now_ms: Date.now(),
+    });
+
+    const evidence = fixtureReadinessEvidence(workKey.deployment_ids);
+    for (const item of await store.listWorkItems(joined.work.work_id)) {
+      await store.publishChildEvidence({
+        work_item_id: item.work_item_id,
+        expected_lease_epoch: lease2.work.lease_epoch,
+        evidence,
+        now_ms: Date.now(),
+      });
+    }
+
+    const ready = await store.finalizeReadyIfQualified({
+      work_id: joined.work.work_id,
+      expected_lease_epoch: lease2.work.lease_epoch,
+      readiness_evidence: evidence,
+      now_ms: Date.now(),
+    });
+    expect(ready.kind).toBe("ready");
+
+    await store.detachSubscriber({
+      order_id: orderA,
+      work_id: joined.work.work_id,
+      now_ms: Date.now(),
+    });
+    const stillReady = await store.getWork(joined.work.work_id);
+    expect(stillReady?.state).toBe("ready");
+  });
+
+  it("enforces unique active work key via partial unique index", async () => {
+    const workKey = fixturePublicWorkKey({ capability: "ownership_index.v1" });
+    const workKeyDigest = digestPublicWorkKey(workKey);
+    const orderId = `ord_pg_uniq_${Date.now()}`;
+    await seedOrder(pool, orderId);
+
+    const joined = await store.joinPublicWork({
+      order_id: orderId,
+      order_tenant_scope_digest: fixtureCommunityScopeDigest("community-alpha"),
+      work_key: workKey,
+      now_ms: Date.now(),
+    });
+    expect(joined.kind).toBe("joined");
+
+    await expect(
+      pool.query(
+        `INSERT INTO shared_preparation_work (
+          work_id, work_key_digest, deployment_set_digest, capability, capability_version,
+          scope_digest, source_identity, readiness_policy_version, evidence_boundary_kind,
+          adapter_version, finality_policy_version, state, generation, attempt, lease_epoch
+        ) VALUES (
+          $1, $2, 'depset', 'ownership_index.v1', 'v1',
+          'scope', '{}'::jsonb, 'gate-leak-public-prep.v1', 'continuous_latest',
+          'sonar-kitchen.v1', 'finality', 'queued', 99, 0, 0
+        )`,
+        [`spw_conflict_${Date.now()}`, workKeyDigest],
+      ),
+    ).rejects.toMatchObject({ code: "23505" });
+  });
+
+  it("does not abandon when a concurrent join reattaches before commit", async () => {
+    const workKey = fixturePublicWorkKey({ capability: "ownership_index.v1" });
+    const orderDetach = `ord_pg_det_${Date.now()}`;
+    const orderJoin = `ord_pg_join_${Date.now()}`;
+    await seedOrder(pool, orderDetach);
+    await seedOrder(pool, orderJoin);
+
+    const joined = await store.joinPublicWork({
+      order_id: orderDetach,
+      order_tenant_scope_digest: fixtureCommunityScopeDigest("community-alpha"),
+      work_key: workKey,
+      now_ms: Date.now(),
+    });
+    if (joined.kind !== "joined") throw new Error("join failed");
+
+    const [detachResult, joinResult] = await Promise.all([
+      store.detachSubscriber({
+        order_id: orderDetach,
+        work_id: joined.work.work_id,
+        now_ms: Date.now(),
+      }),
+      store.joinPublicWork({
+        order_id: orderJoin,
+        order_tenant_scope_digest: fixtureCommunityScopeDigest("community-alpha"),
+        work_key: workKey,
+        now_ms: Date.now(),
+      }),
+    ]);
+
+    expect(detachResult.kind === "detached" || detachResult.kind === "not_linked").toBe(true);
+    expect(joinResult.kind === "joined" || joinResult.kind === "serialization_retry").toBe(true);
+
+    if (joinResult.kind === "joined") {
+      const work = await store.getWork(joinResult.work.work_id);
+      const links = await store.listActiveLinks(joinResult.work.work_id);
+      if (work && ["queued", "preparing", "retry_wait"].includes(work.state)) {
+        expect(links.length).toBeGreaterThan(0);
+        expect(work.state).not.toBe("abandoned");
+      }
+    }
+  });
+});
+
+describe("CR-201A Postgres suite gate", () => {
+  it("documents ORDERING_PG_TEST_URL skip behavior", () => {
+    if (!pgUrl) {
+      expect(pgUrl).toBeFalsy();
+    } else {
+      expect(pgUrl).toMatch(/^postgres/);
+    }
+  });
+});
--- a/packages/services/ordering/src/__tests__/shared-preparation-postgres.test.ts
+++ b/packages/services/ordering/src/__tests__/shared-preparation-postgres.test.ts
@@ -218,7 +218,11 @@ describePg("CR-201A Postgres shared preparation store", () => {
       }),
     ]);
 
-    expect(detachResult.kind === "detached" || detachResult.kind === "not_linked").toBe(true);
+    expect(
+      detachResult.kind === "detached" ||
+        detachResult.kind === "not_linked" ||
+        detachResult.kind === "serialization_retry",
+    ).toBe(true);
     expect(joinResult.kind === "joined" || joinResult.kind === "serialization_retry").toBe(true);
 
     if (joinResult.kind === "joined") {
--- a/packages/services/ordering/src/__tests__/shared-preparation-store.test.ts
+++ b/packages/services/ordering/src/__tests__/shared-preparation-store.test.ts
@@ -1,0 +1,345 @@
+import { describe, expect, it } from "vitest";
+import {
+  countActiveRows,
+  InMemorySharedPreparationStore,
+  SharedPreparationFencingError,
+} from "../shared-preparation-store.js";
+import { buildPublicWorkKeyMaterial } from "../shared-preparation-work-key.js";
+import { PUBLIC_PREP_CAPABILITIES } from "../shared-preparation-types.js";
+import {
+  fixtureCommunityScopeDigest,
+  fixturePublicWorkKey,
+  fixtureReadinessEvidence,
+  fixtureWorkKeyDigest,
+  loadEvmFixtureCandidate,
+} from "./shared-preparation-fixtures.js";
+
+describe("CR-201A public shared preparation store", () => {
+  it("enforces one active work row per work key under 100 equivalent joiners", async () => {
+    const store = new InMemorySharedPreparationStore();
+    const workKey = fixturePublicWorkKey();
+    const digest = fixtureWorkKeyDigest(workKey);
+    const now = Date.parse("2026-07-17T22:00:00.000Z");
+
+    const results = await Promise.all(
+      Array.from({ length: 100 }, (_, index) =>
+        store.joinPublicWork({
+          order_id: `ord_join_${index}`,
+          order_tenant_scope_digest: fixtureCommunityScopeDigest("community-alpha"),
+          work_key: workKey,
+          now_ms: now + index,
+        }),
+      ),
+    );
+
+    for (const result of results) {
+      expect(result.kind).toBe("joined");
+      if (result.kind !== "joined") continue;
+      expect(result.work.work_key_digest).toBe(digest);
+    }
+
+    const workIds = new Set(
+      results
+        .filter((result): result is Extract<typeof result, { kind: "joined" }> => result.kind === "joined")
+        .map((result) => result.work.work_id),
+    );
+    expect(workIds.size).toBe(1);
+    expect(countActiveRows(store, digest)).toBe(1);
+
+    const links = await store.listActiveLinks([...workIds][0]!);
+    expect(links).toHaveLength(100);
+  });
+
+  it("does not share community-scoped keys across different scope digests", async () => {
+    const store = new InMemorySharedPreparationStore();
+    const candidate = loadEvmFixtureCandidate();
+    const deploymentIds = candidate.identity.deployments.map((d) => d.deployment_id);
+
+    const alphaKey = buildPublicWorkKeyMaterial({
+      capability: "ownership_index.v1",
+      capability_version: "v1",
+      collection_id: candidate.identity.collection_id,
+      deployment_ids: deploymentIds,
+      finality_policies: candidate.finality_policies,
+      source_identity: {
+        schema_version: 1,
+        producer: "sonar-api.fixture",
+        upstream_evidence_source: "sonar.public-capability.v1",
+      },
+      readiness_policy_version: "gate-leak-public-prep.v1",
+      adapter_version: "sonar-kitchen.v1",
+    });
+
+    const betaKey = buildPublicWorkKeyMaterial({
+      ...{
+        capability: "ownership_index.v1" as const,
+        capability_version: "v1",
+        collection_id: candidate.identity.collection_id,
+        deployment_ids: deploymentIds,
+        finality_policies: candidate.finality_policies,
+        source_identity: {
+          schema_version: 1,
+          producer: "sonar-api.fixture",
+          upstream_evidence_source: "sonar.public-capability.v1",
+        },
+        readiness_policy_version: "gate-leak-community-beta.v1",
+        adapter_version: "sonar-kitchen.v1",
+      },
+    });
+
+    expect(fixtureWorkKeyDigest(alphaKey)).not.toEqual(fixtureWorkKeyDigest(betaKey));
+
+    const first = await store.joinPublicWork({
+      order_id: "ord_alpha",
+      order_tenant_scope_digest: fixtureCommunityScopeDigest("community-alpha"),
+      work_key: alphaKey,
+      now_ms: 1,
+    });
+    const second = await store.joinPublicWork({
+      order_id: "ord_beta",
+      order_tenant_scope_digest: fixtureCommunityScopeDigest("community-beta"),
+      work_key: betaKey,
+      now_ms: 2,
+    });
+
+    expect(first.kind).toBe("joined");
+    expect(second.kind).toBe("joined");
+    if (first.kind !== "joined" || second.kind !== "joined") return;
+    expect(first.work.work_id).not.toEqual(second.work.work_id);
+  });
+
+  it("increments lease epoch on reclaim and rejects stale fenced writes", async () => {
+    const store = new InMemorySharedPreparationStore();
+    const workKey = fixturePublicWorkKey();
+    const joined = await store.joinPublicWork({
+      order_id: "ord_lease",
+      order_tenant_scope_digest: fixtureCommunityScopeDigest("community-alpha"),
+      work_key: workKey,
+      now_ms: 1_000,
+    });
+    expect(joined.kind).toBe("joined");
+    if (joined.kind !== "joined") return;
+
+    const acquired = await store.acquireLease({
+      work_id: joined.work.work_id,
+      worker_id: "worker-a",
+      lease_duration_ms: 30_000,
+      now_ms: 2_000,
+    });
+    expect(acquired.kind).toBe("acquired");
+    if (acquired.kind !== "acquired") return;
+    expect(acquired.work.lease_epoch).toBe(1);
+
+    const preparing = await store.transitionToPreparing({
+      work_id: joined.work.work_id,
+      expected_lease_epoch: 1,
+      now_ms: 2_500,
+    });
+    expect(preparing.state).toBe("preparing");
+
+    const reclaimed = await store.acquireLease({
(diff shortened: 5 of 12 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#501. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Active work-key index is not UNIQUE",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql:48",
      "description": "`shared_preparation_work_key_active_idx` is a non-unique partial index on `work_key_digest` for active states. The Postgres adapter treats violations of this index name as `serialization_retry` (`err.message.includes(\"shared_preparation_work_key_active_idx\")`), but a non-unique index never raises a unique violation. The one-active-row-per-work-key invariant is therefore only enforced by advisory locks in application code, with no DB fallback if locking is skipped or fails.",
      "suggestion": "Change to `CREATE UNIQUE INDEX IF NOT EXISTS shared_preparation_work_key_active_idx ON shared_preparation_work (work_key_digest) WHERE state IN ('queued', 'preparing', 'retry_wait');` and keep the conflict→`serialization_retry` path.",
      "confidence": 0.95
    },
    {
      "id": "F2",
      "title": "detachSubscriber can abandon work after a concurrent join",
      "severity": "HIGH",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:520",
      "description": "After detaching a link, the adapter counts remaining active links, then separately updates the work row to `abandoned` if count is 0. A concurrent `joinPublicWork` can attach a new link between the count and the abandon update (join only advisory-locks the work key; detach does not). The new subscriber can remain linked to abandoned work.",
      "suggestion": "Hold `FOR UPDATE` on the work row (or the same work-key advisory lock) for the detach+count+abandon sequence, and re-check active link count in the abandon `UPDATE ... WHERE` (e.g. only abandon when no active links exist via a subquery or CTE).",
      "confidence": 0.88
    },
    {
      "id": "F3",
      "title": "finalizeReadyIfQualified is non-atomic and unscoped by state",
      "severity": "HIGH",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:470",
      "description": "Pending-child count and parent `ready` update run as separate pool queries with no transaction and no `state = 'preparing'` predicate. Concurrent child updates or lease reclaim can race the check; a matching `lease_epoch` alone can mark non-preparing work ready.",
      "suggestion": "Run count + parent update in one transaction with `SELECT ... FOR UPDATE` on the work row; require `state = 'preparing'` (and optionally re-verify all children ready) in the same statement.",
      "confidence": 0.9
    },
    {
      "id": "F4",
      "title": "Postgres adapter has no behavioral tests",
      "severity": "HIGH",
      "category": "testing",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts:1",
      "description": "All concurrency, fencing, retry CAS, abandon, and ready-finalization coverage exercises only `InMemorySharedPreparationStore`. The production Postgres path (advisory locks, partial index assumptions, timestamptz CAS, link upsert, abandon) is untested beyond a static SQL expand-only string check.",
      "suggestion": "Add a Postgres-backed suite (testcontainers or existing ordering DB harness) that mirrors the in-memory join/lease/retry/abandon/ready cases, including unique-active conflict and detach/join races.",
      "confidence": 0.93
    },
    {
      "id": "F5",
      "title": "Work-key digest is order-sensitive for arrays",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-work-key.ts:79",
      "description": "`digestPublicWorkKey` hashes `JSON.stringify(canonical)` including `deployment_ids` and `finality_policies` in caller order. `deploymentSetDigest` / `finalityPolicyVersion` sort for stored columns, but the share key does not. Equivalent deployment sets in different orders produce different `work_key_digest` values and will not join the same shared work.",
      "suggestion": "Canonicalize by sorting `deployment_ids` (and finality policies) by a stable key before hashing, matching `deploymentSetDigest` / `finalityPolicyVersion`.",
      "confidence": 0.85
    },
    {
      "id": "F6",
      "title": "In-memory vs Postgres error semantics diverge on fenced writes",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:390",
      "description": "`recordRetryableFailure` (and similarly `publishChildEvidence` / parent ready) maps every zero-row update to `SharedPreparationFencingError`. The in-memory store distinguishes missing work / wrong state (`SharedPreparationStateError`) from stale epoch (`SharedPreparationFencingError`). Callers cannot reliably branch on error type across backends.",
      "suggestion": "On zero rows, load current row and throw `FencingError` only when epoch mismatches; otherwise throw `StateError` (and align publish/finalize the same way).",
      "confidence": 0.9
    },
    {
      "id": "F7",
      "title": "Parent/child state updates are not transactional in Postgres",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:350",
      "description": "`transitionToPreparing` updates the parent with `RETURNING`, then issues a separate child `UPDATE` on the pool. A crash or pool failure between them leaves parent `preparing` and children still `queued`/`retry_wait`. `acquireLease` similarly commits parent epoch then relies on a post-commit `getWork` reload.",
      "suggestion": "Perform parent and child updates on the same client inside one transaction; return the in-transaction `RETURNING` row instead of a post-commit re-read where possible.",
      "confidence": 0.82
    },
    {
      "id": "F8",
      "title": "supersedeActiveGeneration return type lies about serialization_retry",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:560",
      "description": "`supersedeActiveGeneration` casts `withWorkKeyLock(...)` to `Promise<SharedPreparationWorkRecord | undefined>`, but `withWorkKeyLock` can return `{ kind: \"serialization_retry\" }` after exhausting retries. Callers typed to expect work|undefined may mis-handle that object.",
      "suggestion": "Have supersede loop/retry internally until success or return a typed result; do not cast away `serialization_retry`.",
      "confidence": 0.8
    },
    {
      "id": "F9",
      "title": "wakeRetryWait uses exact timestamptz equality",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:410",
      "description": "CAS wake requires `next_attempt_at = $2` where `$2` is `new Date(ms).toISOString()`. Round-trip through `TIMESTAMPTZ` can change sub-ms representation relative to the value stored at `recordRetryableFailure`, causing persistent `stale_wake` for otherwise valid wakes. In-memory compares numeric ms and avoids this class of bug.",
      "suggestion": "Compare with a tolerance, store/compare unix ms as `BIGINT`, or wake with `work_id + attempt` (or generation) as the CAS token instead of timestamp equality.",
      "confidence": 0.7
    },
    {
      "id": "F10",
      "title": "Community-scope test does not exercise community isolation",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts:60",
      "description": "The test titled for community-scoped key isolation builds two keys that differ by `readiness_policy_version`, not by community/tenant scope. Public work keys omit tenant scope by design (`work_tenant_scope_digest` must be NULL); `order_tenant_scope_digest` only lives on links. The test does not prove cross-community non-sharing of public preparation.",
      "suggestion": "Rename the test to match readiness-policy key separation, and add an explicit case that two communities joining the same public work key share one `work_id` while retaining distinct link tenant digests.",
      "confidence": 0.85
    },
    {
      "id": "F11",
      "title": "publishChildEvidence ignores item state",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:440",
      "description": "Child evidence publish only fences on `lease_epoch`, not on an expected state (e.g. `preparing`). A fenced writer can mark items ready from `queued`/`failed`/etc., weakening the intended state machine.",
      "suggestion": "Add `AND state IN ('preparing', ...)` (matching the in-memory intended lifecycle) to the update predicate.",
      "confidence": 0.75
    },
    {
      "id": "F12",
      "title": "Lease fencing and expand-only public constraints",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql:1",
      "description": "Expand-only migration with public-only CHECKs on capability/privacy/sharing scope, plus lease-epoch fencing across prepare/retry/evidence/ready paths and a 100-joiner single-active-row in-memory test, are strong foundations for CR-201A.",
      "suggestion": "Keep these invariants; close the UNIQUE active-index and Postgres transactional gaps so the DB matches the in-memory proofs.",
      "confidence": 0.95
    },
    {
      "id": "F13",
      "title": "Advisory lock plus generation uniqueness is a solid join shape",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:160",
      "description": "`pg_advisory_xact_lock` around ready-reuse / active-join / create, combined with `UNIQUE (work_key_digest, generation)`, is a clear approach for serialized join/create without remote calls under lock.",
      "suggestion": "Pair it with a UNIQUE partial active index so conflict handling is real defense-in-depth.",
      "confidence": 0.9
    },
    {
      "id": "F14",
      "title": "In-memory acquireLease lacks mutual exclusion",
      "severity": "SPECULATION",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:320",
      "description": "Unlike `joinPublicWork`, `acquireLease` on the in-memory store mutates `lease_epoch` / `lease_until` with no per-work lock. Concurrent acquires could both observe an expired lease and both succeed. Postgres uses `FOR UPDATE`, so this may only affect the reference backend and tests.",
      "suggestion": "If the in-memory store is a concurrency oracle, serialize acquire/transition/finalize per `work_id` the same way work-key joins are locked.",
      "confidence": 0.55
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#501. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "mem-lease-no-mutex",
      "title": "In-memory acquireLease has no mutual exclusion",
      "severity": "HIGH",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:360",
      "description": "InMemorySharedPreparationStore.acquireLease (and transitionToPreparing / publishChildEvidence / finalizeReadyIfQualified / recordRetryableFailure / wakeRetryWait) mutate shared maps with no work-key or per-work lock. Two concurrent acquireLease calls can both observe an expired lease, both increment lease_epoch from the same base, and both return acquired/reclaimed with the same epoch—breaking the fencing model the module claims to prove. join/detach use withWorkKeyLock; lease/state paths do not. Unit tests exercise fencing only sequentially, so this gap is invisible in CI.",
      "suggestion": "Serialize lease and fenced state transitions on a per-work_id (or work_key) lock in the in-memory store, matching Postgres FOR UPDATE semantics, and add a concurrent double-acquire test that expects only one winner.",
      "confidence": 0.95
    },
    {
      "id": "retry-schedule-bypass",
      "title": "transitionToPreparing accepts retry_wait and skips wake gate",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:398",
      "description": "Both backends allow state IN ('queued','retry_wait') for transitionToPreparing. wakeRetryWait is the only path that enforces next_attempt_at <= now and attempt CAS, but callers can skip it and move directly from retry_wait to preparing before the scheduled time. That defeats retry backoff/deadline scheduling encoded on the row.",
      "suggestion": "Restrict transitionToPreparing to state = 'queued' only (after wake), or require the same next_attempt_at/attempt predicates when leaving retry_wait.",
      "confidence": 0.9
    },
    {
      "id": "wake-ignores-deadline",
      "title": "wakeRetryWait does not enforce retry_deadline",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:468",
      "description": "recordRetryableFailure persists retry_deadline, but wakeRetryWait only CAS-matches attempt and next_attempt_at <= now. A wake after the deadline still returns woke and re-enters queued, with no transition to failed. The deadline field is therefore advisory storage only.",
      "suggestion": "Reject or fail terminal when now_ms > retry_deadline (or next_attempt_at past deadline), and cover with a unit test.",
      "confidence": 0.85
    },
    {
      "id": "finalize-empty-children-parity",
      "title": "Empty child set: Postgres finalizes ready, in-memory stays pending",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:148",
      "description": "In-memory allChildrenReady requires items.length > 0. Postgres finalizeReadyIfQualified treats COUNT(state <> 'ready') = 0 as success, so zero child rows yields ready. Normal joins create children, but any orphan/partial write path diverges between backends.",
      "suggestion": "Align both sides: either require at least one ready child before parent ready, or document and test empty-set behavior identically.",
      "confidence": 0.88
    },
    {
      "id": "child-evidence-ignores-parent-state",
      "title": "publishChildEvidence fences lease_epoch only, not parent state",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:495",
      "description": "Child evidence updates only check work_item lease_epoch and item state = preparing. After parent abandon/supersede/retry_wait, children can still be preparing with a matching epoch and accept evidence, leaving ready children under a non-preparing parent.",
      "suggestion": "Join parent row and require parent.state = 'preparing' and matching lease_epoch in the same conditional update (transaction in Postgres).",
      "confidence": 0.8
    },
    {
      "id": "finalize-skips-qualified-check",
      "title": "finalizeReadyIfQualified does not require freshness.qualified",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:530",
      "description": "Parent can be marked ready with readiness_evidence.freshness.qualified = false. joinPublicWork reuse requires qualified = true, so unqualified ready rows are skipped and a new active generation can be created beside them—surprising dual rows for the same key.",
      "suggestion": "Reject finalize unless evidence.freshness.qualified === true (SQL check and in-memory guard), matching join reuse predicates.",
      "confidence": 0.82
    },
    {
      "id": "work-key-json-stringify-hash",
      "title": "digestPublicWorkKey depends on JSON.stringify key order",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-work-key.ts:118",
      "description": "Canonicalization sorts deployments and finality policies but hashes via JSON.stringify of nested objects (collection_id, source_identity, digests). Property insertion order on those objects affects the digest. Callers reconstructing material with different key orders get different work_key_digest values for semantically identical keys.",
      "suggestion": "Hash a explicitly ordered canonical encoding (sorted keys recursively, or field-by-field length-prefixed), and add a test with shuffled object key insertion orders.",
      "confidence": 0.75
    },
    {
      "id": "misleading-community-isolation-test",
      "title": "Community isolation test does not exercise community scope in the work key",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts:70",
      "description": "The test titled for different scope digests builds two keys that differ by readiness_policy_version, while order_tenant_scope_digest is only stored on report_work_links and is not part of digestPublicWorkKey. It does not prove community isolation of shared preparation; public_chain work keys are shareable across tenant link digests by design.",
      "suggestion": "Rename the test to match readiness_policy_version separation, and add an explicit test that identical work keys with different order_tenant_scope_digest fan into the same work_id.",
      "confidence": 0.92
    },
    {
      "id": "link-reattach-skips-tenant-digest",
      "title": "insertLink ON CONFLICT does not refresh order_tenant_scope_digest",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:318",
      "description": "Reattach after detach updates detached_at, generation, and joined_at but leaves the previous order_tenant_scope_digest. A rejoin with a new tenant scope digest keeps the stale value.",
      "suggestion": "Include order_tenant_scope_digest = EXCLUDED.order_tenant_scope_digest in the ON CONFLICT DO UPDATE set list.",
      "confidence": 0.7
    },
    {
      "id": "advisory-lock-hashtext-collision",
      "title": "Advisory locks use 32-bit hashtext",
      "severity": "LOW",
      "category": "performance",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:176",
      "description": "pg_advisory_xact_lock(hashtext(work_key_digest)) can collide across distinct digests, causing unrelated keys to serialize on one lock. Correctness still rests on the partial unique index, but contention and latency can spike under hash collisions.",
      "suggestion": "Use a 64-bit digest-derived lock key (e.g. two int4 from sha256 bytes) to reduce collision probability.",
      "confidence": 0.65
    },
    {
      "id": "worker-id-ignored",
      "title": "acquireLease accepts worker_id but never records it",
      "severity": "LOW",
      "category": "api",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:330",
      "description": "worker_id is part of the store port and both implementations ignore it. Operators cannot attribute the lease holder or debug reclaim races by worker.",
      "suggestion": "Persist lease_owner/worker_id on the work row on acquire/reclaim, or remove the parameter until ownership is modeled.",
      "confidence": 0.9
    },
    {
      "id": "detach-serialization-retry-mapped-not-linked",
      "title": "In-memory detach maps serialization_retry to not_linked",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:560",
      "description": "If withWorkKeyLock returns { kind: 'serialization_retry' }, detachSubscriber reports not_linked. That conflates lock exhaustion with absence of a link and can cause callers to skip retry.",
      "suggestion": "Propagate serialization_retry (or throw) from detach/supersede instead of coercing to not_linked/undefined.",
      "confidence": 0.6
    },
    {
      "id": "pg-suite-gated-off-default-ci",
      "title": "Postgres behavioral suite skipped unless ORDERING_PG_TEST_URL set",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-postgres.test.ts:36",
      "description": "F4 Postgres tests correctly skip without a URL, so default unit CI only validates the in-memory backend. Postgres-specific behaviors (partial unique index, advisory-lock detach/join race, FOR UPDATE leasing) are easy to regress unnoticed.",
      "suggestion": "Run the Postgres suite in at least one CI job with a service container, or document a required merge gate for CR-201A.",
      "confidence": 0.85
    },
    {
      "id": "praise-partial-unique-active-index",
      "title": "Partial unique index enforces one active row per work key",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql:48",
      "description": "shared_preparation_work_key_active_idx on work_key_digest WHERE state IN (queued, preparing, retry_wait), plus capability/privacy CHECKs and expand-only DDL, give a strong DB-level invariant for public shared prep without destructive migrations.",
      "suggestion": "Keep join paths treating 23505 as serialization_retry and cover the index in the gated Postgres suite (already present).",
      "confidence": 1.0
    },
    {
      "id": "praise-abandon-not-exists",
      "title": "Abandonment re-checks zero subscribers in the updating statement",
      "severity": "PRAISE",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:620",
      "description": "detachSubscriber updates the link then abandons only when NOT EXISTS active links in the same UPDATE, under work-key advisory locking. That closes the join-reattach vs abandon race the Postgres test targets.",
      "suggestion": "Retain this pattern for any future restricted-scope detach paths.",
      "confidence": 0.95
    },
    {
      "id": "praise-retry-cas-wake",
      "title": "Retry wake uses attempt CAS and inequality on next_attempt_at",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:468",
      "description": "wakeRetryWait avoids timestamptz equality pitfalls by using attempt as a monotonic CAS token and next_attempt_at <= now, matching the in-memory reference and the dedicated unit test.",
      "suggestion": "Extend the same style when enforcing retry_deadline.",
      "confidence": 0.95
    },
    {
      "id": "speculation-child-state-on-parent-retry",
      "title": "Parent retry_wait leaves children in preparing/ready",
      "severity": "SPECULATION",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:430",
      "description": "recordRetryableFailure updates only the parent row. Children may remain preparing or ready across retries. A later finalize can succeed immediately from stale child readiness without re-running child work. This may be intentional partial-progress reuse, but it is unspecified and untested.",
      "suggestion": "Document reuse-vs-reset policy; if reset is required, move non-terminal children to retry_wait/queued and clear evidence on parent failure.",
      "confidence": 0.45
    },
    {
      "id": "reframe-public-cross-tenant-share",
      "title": "order_tenant_scope_digest does not isolate public work keys",
      "severity": "REFRAME",
      "category": "design",
      "file": "packages/services/ordering/src/shared-preparation-work-key.ts:99",
      "description": "Work-key digests cover collection/capability/finality/readiness/adapter identity for privacy_class public_chain. Tenant/community digests live only on report_work_links. Cross-tenant fan-in to one shared preparation row is the intended CR-201A public sharing model, not a missing ACL on the work key.",
      "suggestion": "Keep link-level tenant digests for subscriber accounting; reserve work_tenant_scope_digest / non-public sharing for CR-201B. Avoid tests that imply community digests partition public work keys.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#501. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Retryable failure leaves child items ready; finalize can skip re-work",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "recordRetryableFailure only moves the parent to retry_wait and does not reset preparation_work_items. After wakeRetryWait → queued → transitionToPreparing, only children still in queued become preparing; children already ready stay ready. finalizeReadyIfQualified then sees zero non-ready children and can mark the parent ready without re-running failed preparation. The in-memory store has the same behavior.",
      "suggestion": "On recordRetryableFailure (or on wake), reset non-terminal child items to queued (and clear evidence_envelope if retry invalidates prior evidence), or reject finalize until attempt-scoped child evidence matches the current parent attempt.",
      "confidence": 0.82
    },
    {
      "id": "F2",
      "title": "finalizeReadyIfQualified does not require freshness.qualified",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "joinPublicWork only reuses ready rows when readiness_evidence.freshness.qualified is true, but finalizeReadyIfQualified writes state=ready for any caller-supplied envelope. An unqualified envelope yields a ready row that later joins will not reuse, leaving a dead-end ready generation.",
      "suggestion": "Reject finalize (or no-op) unless readiness_evidence.freshness.qualified === true, matching the join reuse predicate in both Postgres and in-memory backends.",
      "confidence": 0.88
    },
    {
      "id": "F3",
      "title": "Date.parse used on node-pg timestamptz Date objects",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "rowToWork/rowToItem/rowToLink call Date.parse(row.* as string) for timestamptz columns. node-pg returns JavaScript Date instances by default, so this relies on Date→string→parse coercion and can yield NaN under non-V8 or unusual toString output, corrupting lease_until_unix_ms and related fields used for fencing.",
      "suggestion": "Normalize with a helper: if value instanceof Date use getTime(); else if string use Date.parse; else reject. Apply in all three row mappers.",
      "confidence": 0.78
    },
    {
      "id": "F4",
      "title": "Empty child-set ready check diverges between backends",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "InMemory allChildrenReady requires items.length > 0. Postgres finalizeReadyIfQualified treats COUNT(state <> 'ready') = 0 as complete, which is true for zero child rows. buildPublicWorkKeyMaterial forbids empty deployment_ids today, but the store contract already diverges and would silently disagree if a row were inserted without items.",
      "suggestion": "Align both backends: either require at least one child before ready, or allow empty-set ready explicitly in both, and add a paired unit/Postgres assertion.",
      "confidence": 0.7
    },
    {
      "id": "F5",
      "title": "Reattach link does not refresh order_tenant_scope_digest",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "insertLink ON CONFLICT (order_id, work_id) clears detached_at and updates generation/joined_at but leaves the prior order_tenant_scope_digest. A re-join with a new digest keeps the stale tenant scope on the link row.",
      "suggestion": "Include order_tenant_scope_digest = EXCLUDED.order_tenant_scope_digest in the DO UPDATE SET clause (and mirror that in InMemory attachLink).",
      "confidence": 0.85
    },
    {
      "id": "F6",
      "title": "Lease reclaim while preparing leaves no legal transitionToPreparing path",
      "severity": "MEDIUM",
      "category": "api-contract",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "After lease expiry in preparing, acquireLease returns reclaimed with a new lease_epoch but state remains preparing. transitionToPreparing only allows queued→preparing, so a reclaiming worker that always calls transitionToPreparing will get SharedPreparationStateError even though publishChildEvidence/finalizeReadyIfQualified with the new epoch are valid. Tests cover stale-epoch rejection, not the reclaim-continue path.",
      "suggestion": "Document that reclaim during preparing must skip transition, or make transitionToPreparing idempotent when state is already preparing and lease_epoch matches.",
      "confidence": 0.8
    },
    {
      "id": "F7",
      "title": "pg_advisory_xact_lock(hashtext(...)) is 32-bit collision-prone",
      "severity": "LOW",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "Advisory locks key only on hashtext(work_key_digest). Distinct digests can share a lock or, worse, interleave with unrelated keys. The partial unique index still prevents two actives, but collision causes false contention and pushes more join paths into serialization_retry.",
      "suggestion": "Prefer pg_advisory_xact_lock on two int4 hashes derived from the full digest (or hashtextextended where available) to reduce collision rate.",
      "confidence": 0.65
    },
    {
      "id": "F8",
      "title": "In-memory attachLink ignores generation/joined_at on live re-join",
      "severity": "LOW",
      "category": "consistency",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "When an active link already exists, InMemory attachLink returns the existing row unchanged. Postgres insertLink ON CONFLICT updates generation and joined_at. Callers observing links after re-join see different metadata depending on backend.",
      "suggestion": "Update generation and joined_at_unix_ms on the in-memory path when the link is already active, matching Postgres.",
      "confidence": 0.9
    },
    {
      "id": "F9",
      "title": "worker_id is accepted but never recorded or checked",
      "severity": "LOW",
      "category": "api-contract",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "acquireLease requires worker_id in both the port and Postgres implementation, but neither persists nor validates it. Fencing is solely lease_epoch-based; the parameter suggests holder identity that does not exist.",
      "suggestion": "Persist lease_owner/worker_id on acquire and enforce it on fenced writes, or remove worker_id from the port until ownership is implemented.",
      "confidence": 0.95
    },
    {
      "id": "F10",
      "title": "capability.replace('.v1','') is a brittle collection-protocol bridge",
      "severity": "LOW",
      "file": "packages/services/ordering/src/shared-preparation-work-key.ts",
      "description": "buildPublicWorkKeyMaterial strips '.v1' via String.replace before digestCollectionWorkKey. Only the first match is removed, and newer capability version suffixes will silently produce wrong collection digests / scope_digest values.",
      "suggestion": "Map PublicPrepCapability to collection-protocol capability IDs with an explicit table, or strip a trailing /\\.v\\d+$/ with validation.",
      "confidence": 0.72
    },
    {
      "id": "F11",
      "title": "Partial unique index + lease epoch fencing + expand-only DDL",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql",
      "description": "One-active-row invariant is enforced in the schema (shared_preparation_work_key_active_idx), public-only CHECKs gate privacy/capability, fenced updates key off lease_epoch, wakeRetryWait CASes on attempt with next_attempt_at inequality, and detach abandons via NOT EXISTS in the same transaction under an advisory lock. In-memory and gated Postgres suites exercise join fan-in, reclaim races, and detach/join races.",
      "suggestion": "Keep the Postgres race suite in CI behind ORDERING_PG_TEST_URL so the unique-index and advisory-lock paths stay covered.",
      "confidence": 1.0
    },
    {
      "id": "F12",
      "title": "Community-scope test actually varies readiness_policy_version",
      "severity": "LOW",
      "category": "tests",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts",
      "description": "The test titled 'does not share community-scoped keys across different scope digests' builds alpha/beta keys that differ by readiness_policy_version, while fixturePublicWorkKey's communityScopeDigest option is unused. order_tenant_scope_digest is not part of digestPublicWorkKey, so the test does not prove tenant-scope isolation.",
      "suggestion": "Rename the test to match readiness_policy_version isolation, or add a real case that different order_tenant_scope_digest values still share the same public work_id when the work key matches.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#501. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "detach-serialization-misreport",
      "title": "In-memory detach maps lock exhaustion to not_linked",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:640",
      "description": "detachSubscriber returns { kind: \"not_linked\" } when withWorkKeyLock yields serialization_retry. That conflates \"link was never active / already detached\" with \"detach did not run because the work-key lock retry budget was exhausted\". Callers can skip retry or treat the subscriber as gone while the link remains attached, leaving work unreclaimable or preventing abandonment.",
      "suggestion": "Propagate a distinct failure (throw SharedPreparationStateError, or return kind: \"serialization_retry\") instead of not_linked when the lock path is exhausted.",
      "confidence": 0.9
    },
    {
      "id": "finalize-skips-qualified-check",
      "title": "finalizeReadyIfQualified does not require freshness.qualified",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:520",
      "description": "joinPublicWork only reuses ready rows when readiness_evidence.freshness.qualified === true, but finalizeReadyIfQualified writes state=ready with whatever envelope the caller supplies. An unqualified ready row is terminal for that generation (not in the active partial index), so joiners skip reuse and open a new active generation while a misleading ready row remains. Same gap exists in the in-memory finalize path.",
      "suggestion": "Reject finalize unless readiness_evidence.freshness.qualified === true (and ideally matches readiness_policy_version), mirroring the join reuse predicate.",
      "confidence": 0.85
    },
    {
      "id": "empty-children-parity",
      "title": "Empty child set: Postgres finalizes ready, in-memory stays pending",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:545",
      "description": "In-memory allChildrenReady requires items.length > 0. Postgres pending count treats zero children as zero pending and promotes parent to ready. Join normally creates children, so this is mostly a corruption/parity hazard, but the two backends disagree on the same store contract.",
      "suggestion": "Align Postgres with in-memory: require COUNT(*) > 0 AND no non-ready children before promoting to ready.",
      "confidence": 0.8
    },
    {
      "id": "insert-link-stale-tenant-digest",
      "title": "Reattach link does not refresh order_tenant_scope_digest",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:318",
      "description": "report_work_links upsert on (order_id, work_id) clears detached_at and updates generation/joined_at but leaves order_tenant_scope_digest unchanged. A later join from the same order with a different tenant scope digest keeps the original digest on the active link.",
      "suggestion": "Add order_tenant_scope_digest = EXCLUDED.order_tenant_scope_digest to the ON CONFLICT DO UPDATE set list.",
      "confidence": 0.85
    },
    {
      "id": "lease-during-retry-wait",
      "title": "acquireLease treats retry_wait as leasable",
      "severity": "MEDIUM",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:365",
      "description": "isActivePublicWorkState includes retry_wait, so workers can acquire/reclaim a lease while the parent is waiting on next_attempt_at. transitionToPreparing still requires queued, so the lease is held without a legal preparing transition until wakeRetryWait runs. Contended reclaim during the backoff window is allowed and adds noise/races without progressing work.",
      "suggestion": "Restrict acquireLease to queued (and optionally preparing), or clear/deny leases while state is retry_wait.",
      "confidence": 0.75
    },
    {
      "id": "classify-zero-row-toctou",
      "title": "Zero-row error classification uses a post-ROLLBACK read",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:175",
      "description": "classifyZeroRowWorkUpdate rolls back, then getWork() on a different pool connection. Concurrent mutators can change state/lease_epoch between the failed CAS and the diagnostic read, so callers may see SharedPreparationStateError instead of SharedPreparationFencingError (or the reverse).",
      "suggestion": "Classify from the locked row inside the same transaction before ROLLBACK, or return a generic conflict without a second read.",
      "confidence": 0.7
    },
    {
      "id": "pg-date-parse-coercion",
      "title": "Timestamptz fields coerced via Date.parse",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:42",
      "description": "rowToWork/rowToItem/rowToLink use Date.parse(row.* as string). node-pg typically returns Date objects for timestamptz, so this relies on Date→string→parse round-tripping. That is brittle across drivers/type parsers and can yield NaN if a parser returns an unexpected shape.",
      "suggestion": "Normalize with a helper: if value instanceof Date use getTime(); else Date.parse(String(value)); reject NaN.",
      "confidence": 0.7
    },
    {
      "id": "service-no-serialization-retry",
      "title": "Service facade does not handle serialization_retry",
      "severity": "MEDIUM",
      "category": "api",
      "file": "packages/services/ordering/src/shared-preparation-service.ts:16",
      "description": "Postgres joinPublicWork returns kind serialization_retry on unique-active conflicts; the in-memory store mostly retries under its lock. createSharedPreparationService only injects now_ms and re-exports store, so HTTP/workers using the service must know to retry. Easy to miss at the first integration call site.",
      "suggestion": "Retry joinPublicWork inside the service (bounded backoff) until joined or budget exhausted, matching in-memory behavior.",
      "confidence": 0.65
    },
    {
      "id": "jsonb-deployment-unique",
      "title": "Child uniqueness depends on JSONB binary equality",
      "severity": "LOW",
      "category": "data-model",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql:74",
      "description": "UNIQUE (work_id, deployment_id) on JSONB means equivalent VersionedDigest values with different key order or whitespace normalize to different values. Inserts use JSON.stringify from object key order, which is usually stable in V8 but is a fragile uniqueness boundary.",
      "suggestion": "Store a canonical deployment digest text column for uniqueness, or normalize JSONB with jsonb_strip_nulls and sorted keys before insert.",
      "confidence": 0.55
    },
    {
      "id": "hashtext-advisory-collisions",
      "title": "Advisory locks use 32-bit hashtext",
      "severity": "LOW",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:168",
      "description": "pg_advisory_xact_lock(hashtext($1)) can collide across distinct work_key_digest values, serializing unrelated keys. The partial unique index still prevents two actives, but collision increases serialization_retry traffic.",
      "suggestion": "Use a 64-bit lock key derived from the digest (e.g. two int4s from sha256 prefix) to reduce collisions.",
      "confidence": 0.6
    },
    {
      "id": "capability-replace-v1",
      "title": "Collection work-key capability uses string replace(\".v1\", \"\")",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-work-key.ts:68",
      "description": "input.capability.replace(\".v1\", \"\") strips only the first \".v1\" substring. Safe for the two allowlisted names today, but a future capability containing \".v1\" elsewhere would hash under the wrong collection capability id.",
      "suggestion": "Map via an explicit table (e.g. ownership_index.v1 → ownership_index) or strip only a trailing \".v1\" suffix.",
      "confidence": 0.5
    },
    {
      "id": "community-scope-test-mislabel",
      "title": "Community-scope test actually varies readiness_policy_version",
      "severity": "LOW",
      "category": "tests",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts:66",
      "description": "The test title claims different community scope digests do not share keys, but alphaKey/betaKey differ by readiness_policy_version while order_tenant_scope_digest is not part of digestPublicWorkKey. Public sharing across tenants is schema-intentional; the test does not prove tenant isolation.",
      "suggestion": "Rename the test to assert readiness_policy_version isolation, and add an explicit test that identical work keys with different order_tenant_scope_digest share one work row if that is the product rule.",
      "confidence": 0.9
    },
    {
      "id": "retry-deadline-unenforced",
      "title": "wakeRetryWait ignores retry_deadline",
      "severity": "SPECULATION",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:470",
      "description": "retry_deadline is persisted on failure but wake only CAS-matches attempt and next_attempt_at <= now. If deadline enforcement lives only in a future worker, overdue work can be woken forever.",
      "suggestion": "If CR-201A owns deadline semantics, fail or refuse wake when now_ms > retry_deadline_unix_ms; otherwise document that callers must enforce it.",
      "confidence": 0.4
    },
    {
      "id": "praise-partial-unique-and-fencing",
      "title": "Partial unique index plus lease-epoch fencing",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql:48",
      "description": "One-active-row invariant is enforced in the database (partial unique index), join uses transactional advisory locking with serialization_retry, and mutators CAS on lease_epoch. Contested reclaim and stale child publish are covered in tests for the in-memory backend, with a gated Postgres behavioral suite.",
      "suggestion": "Keep Postgres race coverage (join/detach, unique active) in CI where ORDERING_PG_TEST_URL is available so the index remains the source of truth.",
      "confidence": 1.0
    },
    {
      "id": "praise-retry-invalidates-children",
      "title": "Retry clears child evidence before re-prepare",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:445",
      "description": "recordRetryableFailure resets ready/preparing children and nulls evidence_envelope so finalize cannot assemble parent ready from pre-retry child evidence. The matching unit test locks that regression.",
      "suggestion": "Retain this invariant if terminal failure paths are added later.",
      "confidence": 1.0
    },
    {
      "id": "praise-expand-only-public-constraints",
      "title": "Expand-only migration with public-only CHECKs",
      "severity": "PRAISE",
      "category": "safety",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql:1",
      "description": "007 is IF NOT EXISTS / expand-only, pins privacy_class and sharing_scope to public, allowlists capabilities, and keeps work_tenant_scope_digest NULL for public rows—aligned with CR-201A vs CR-201B split.",
      "suggestion": "None.",
      "confidence": 1.0
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#501. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "pg-timestamptz-date-parse",
      "title": "rowTo* uses Date.parse on pg timestamptz values that are already Date objects",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "node-pg parses TIMESTAMPTZ columns into JavaScript Date by default. rowToWork/rowToItem/rowToLink still do Date.parse(row.* as string) for created_at, updated_at, lease_until, next_attempt_at, retry_deadline, joined_at, and detached_at. Date.parse(Date) stringifies via Date#toString() (locale/offset-dependent) rather than using getTime(), so lease expiry, retry wake thresholds, and link timestamps can be wrong or NaN depending on runtime locale and pg type parsers.",
      "suggestion": "Normalize timestamps with a helper: if value instanceof Date use getTime(); else if string use Date.parse/ISO; reject unexpected types. Apply it in all three rowTo* mappers and cover with a Postgres test that asserts round-tripped unix_ms equals the written instant.",
      "confidence": 0.88
    },
    {
      "id": "lease-acquire-during-retry-wait",
      "title": "acquireLease treats retry_wait as leasable active state",
      "severity": "HIGH",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "isActivePublicWorkState includes retry_wait, and both in-memory and Postgres acquireLease only require an active state plus expired/missing lease_until. A worker can therefore take/bump lease_epoch while the parent is still in retry_wait, before wakeRetryWait moves it to queued. That blocks other workers (busy) during backoff, lets the premature holder call transitionToPreparing (which correctly fails), and desynchronizes lease ownership from the intended wake→queued→acquire sequence covered by the tests.",
      "suggestion": "Restrict acquireLease to state === 'queued' (and optionally preparing only for same-epoch renew if that is a product need). Keep retry_wait in the partial unique index / active-row definition for join fan-in, but not for leasing. Add a regression test that acquireLease during retry_wait returns not_active or busy-without-epoch-bump.",
      "confidence": 0.86
    },
    {
      "id": "child-evidence-ignores-parent-state",
      "title": "publishChildEvidence does not verify parent work is still preparing",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "Postgres and in-memory publishChildEvidence only CAS the child row (lease_epoch + state=preparing). detachSubscriber/supersedeActiveGeneration can move the parent to abandoned/superseded without resetting child rows. Children can remain preparing and accept evidence after the parent is no longer an active preparation, leaving orphan ready children and weakening the preparing-only write boundary the suite asserts for finalize.",
      "suggestion": "Under the same work-key/work row lock (or a single SQL UPDATE ... FROM shared_preparation_work WHERE parent.state = 'preparing'), require parent preparing (and matching lease_epoch) before accepting child evidence. On abandon/supersede, optionally reset non-terminal children in the same transaction.",
      "confidence": 0.84
    },
    {
      "id": "ready-reuse-skips-freshness-age",
      "title": "Ready-row reuse only checks freshness.qualified, not max_age / as_of",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "joinPublicWork reuses the latest ready row when readiness_policy_version matches and freshness.qualified is true (SQL and highestReadyRow). It never evaluates freshness.max_age_ms against observation_window/as_of or now_ms. Qualified-but-stale evidence can be fan-in reused indefinitely after finalize, which conflicts with storing max_age_ms on the envelope and with the unqualified-poison test’s intent to avoid bad ready reuse.",
      "suggestion": "At reuse time, apply the same readiness policy gate used at finalize (qualified + privacy_scope + age/as_of vs now_ms). If stale, skip reuse and create/join an active generation (or supersede). Add a test where qualified evidence older than max_age_ms is not reused_ready.",
      "confidence": 0.78
    },
    {
      "id": "child-ready-without-evidence-gate",
      "title": "Child ready transition does not assert evidence qualification or coverage",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "finalizeReadyIfQualified runs assertReadinessEvidenceQualified on the parent envelope and only requires children state === 'ready'. publishChildEvidence stores arbitrary ReadinessEvidenceEnvelope with no qualified/privacy/deployment_coverage checks. Unqualified or mismatched child envelopes still satisfy pending_children, so parent qualification alone can mark shared work ready.",
      "suggestion": "Apply assertReadinessEvidenceQualified (and deployment_id ⊆ coverage / capability checks) inside publishChildEvidence, or require child evidence_envelope.freshness.qualified when counting children ready in finalize.",
      "confidence": 0.8
    },
    {
      "id": "reattach-skips-tenant-scope-digest",
      "title": "insertLink ON CONFLICT does not refresh order_tenant_scope_digest",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "Re-join after detach updates detached_at, generation, and joined_at but leaves the previous order_tenant_scope_digest in place. An order that reattaches with a new tenant scope digest keeps the stale digest on the link row, which is the field intended to record subscriber tenancy for public shared work.",
      "suggestion": "Include order_tenant_scope_digest = EXCLUDED.order_tenant_scope_digest in the ON CONFLICT DO UPDATE set list, and mirror that in InMemorySharedPreparationStore.attachLink when reactivating a detached link.",
      "confidence": 0.83
    },
    {
      "id": "supersede-swallows-serialization-retry",
      "title": "In-memory supersedeActiveGeneration maps serialization_retry to undefined",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "When withWorkKeyLock exhausts retries, supersedeActiveGeneration returns undefined—the same signal as “no active row.” Callers cannot distinguish lock exhaustion from nothing to supersede, and may proceed to join/create under the belief the prior generation was cleared when it was not.",
      "suggestion": "Propagate serialization_retry (change return type) or throw SharedPreparationStateError on lock exhaustion, consistent with unwrapLocked used by lease/finalize paths.",
      "confidence": 0.9
    },
    {
      "id": "postgres-detach-no-serialization-retry",
      "title": "Postgres detachSubscriber never returns serialization_retry",
      "severity": "MEDIUM",
      "category": "api-consistency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "SharedPreparationStore.detachSubscriber documents serialization_retry; in-memory can return it and tests assert that distinction. The Postgres adapter only returns detached/not_linked or throws. Callers written against the port cannot rely on a uniform retry signal across backends for contested detach/join.",
      "suggestion": "Map deadlock/serialization failures (40P01/40001) and unique-active races to { kind: 'serialization_retry' }, or remove serialization_retry from the port if Postgres will always throw and callers must catch.",
      "confidence": 0.82
    },
    {
      "id": "standalone-migrate-skips-intermediate-migrations",
      "title": "PostgresSharedPreparationStore.runMigrations applies only 001 and 007",
      "severity": "MEDIUM",
      "category": "ops",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "connect({ migrate: true }) runs 001_orders.sql then 007_shared_preparation_work.sql, skipping 002–006 that PostgresOrderStore applies. A store brought up via SharedPreparationStore.connect alone can miss earlier ordering schema while still creating CR-201A tables, diverging from the production migration chain now extended in store-postgres.ts.",
      "suggestion": "Reuse the same ordered migration list as PostgresOrderStore (or call a shared migrate helper) instead of a hard-coded 001+007 subset.",
      "confidence": 0.75
    },
    {
      "id": "misleading-community-scope-test",
      "title": "Community isolation test keys off readiness_policy_version; unused fixture param",
      "severity": "LOW",
      "category": "tests",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts",
      "description": "The test titled “does not share community-scoped keys across different scope digests” builds distinct work keys by changing readiness_policy_version, not community/tenant scope inside the work key. fixturePublicWorkKey accepts communityScopeDigest but never uses it. order_tenant_scope_digest differences alone would still correctly fan into one public work row—so the test name overclaims tenancy isolation that CR-201A intentionally does not put in the work key.",
      "suggestion": "Rename the test to policy-version / work-key divergence, or add an assertion that identical work keys with different order_tenant_scope_digest share one work_id. Remove or wire communityScopeDigest in the fixture.",
      "confidence": 0.92
    },
    {
      "id": "classify-zero-row-toctou",
      "title": "classifyZeroRowWorkUpdate re-reads work on a different connection after ROLLBACK",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts",
      "description": "After a fenced UPDATE returns zero rows, the adapter ROLLBACKs then calls getWork via the pool. Concurrent lease reclaim can change lease_epoch/state between the failed UPDATE and the classification read, so callers may see SharedPreparationStateError instead of SharedPreparationFencingError (or the reverse).",
      "suggestion": "Classify using a SELECT in the same transaction before ROLLBACK (or FOR UPDATE), or return a generic conflict and let callers refresh—avoid a post-rollback pool read for fencing decisions.",
      "confidence": 0.7
    },
    {
      "id": "worker-id-ignored",
      "title": "acquireLease accepts worker_id but never persists holder identity",
      "severity": "LOW",
      "category": "observability",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "Both backends take worker_id on acquireLease but only bump lease_epoch/lease_until. Contested reclaim tests cannot assert which worker won, and production debugging cannot see the current holder from the store.",
      "suggestion": "Add lease_owner / leased_by column (expand-only) and set it on acquire/reclaim, or drop worker_id from the API until owned.",
      "confidence": 0.95
    },
    {
      "id": "praise-partial-unique-active-key",
      "title": "Partial unique index plus advisory lock correctly enforces one active generation",
      "severity": "PRAISE",
      "category": "concurrency",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql",
      "description": "shared_preparation_work_key_active_idx on work_key_digest WHERE state IN (queued, preparing, retry_wait), combined with join’s advisory xact lock and unique-violation → serialization_retry, is a solid expand-only way to prove single active public prep work under fan-in. Privacy/capability CHECKs keep CR-201A scoped to public_chain fixtures.",
      "suggestion": "Keep the Postgres suite’s 23505 assertion in CI behind ORDERING_PG_TEST_URL so the index remains the source of truth, not only the in-memory lock.",
      "confidence": 1.0
    },
    {
      "id": "praise-lease-epoch-and-retry-cas",
      "title": "Lease epoch fencing and attempt-CAS wake are coherently tested",
      "severity": "PRAISE",
      "category": "concurrency",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts",
      "description": "Stale epoch rejection after reclaim, contested single-winner reclaim, retry_wait blocking direct preparing transitions, attempt-mismatched/early wake → stale_wake, and child evidence reset on retryable failure form a tight behavioral spec that matches the in-memory and Postgres control flow.",
      "suggestion": "Extend the same cases onto the gated Postgres suite (contested reclaim, stale epoch publish) so fencing cannot regress only in the adapter.",
      "confidence": 1.0
    },
    {
      "id": "praise-ready-reuse-qualification-gate",
      "title": "Ready reuse and finalize both require freshness.qualified",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts",
      "description": "assertReadinessEvidenceQualified on finalize, SQL qualified check on ready reuse, child invalidation before retry, and the poisoned-ready join test (reused_ready false + new generation) show deliberate defense against promoting unqualified evidence as shared ready state.",
      "suggestion": "Keep the poison-reuse test when a Postgres backend path for ready reuse is asserted end-to-end.",
      "confidence": 1.0
    },
    {
      "id": "reframe-public-work-key-excludes-tenant",
      "title": "Public work identity is deployment/capability scoped; tenancy lives on links",
      "severity": "REFRAME",
      "category": "design",
      "file": "packages/services/ordering/src/shared-preparation-types.ts",
      "description": "sharing_scope_kind=public with work_tenant_scope_digest IS NULL, plus order_tenant_scope_digest only on report_work_links, means cross-tenant fan-in of identical public work keys is intentional for CR-201A. Review pressure to “fix community cross-talk” at the work-key layer would fight the data model; isolation belongs in capability allowlists, evidence privacy_class, and link-level tenant digests—not a second active row per community.",
      "suggestion": "Document in the store module header that order_tenant_scope_digest is subscriber metadata only and must not be mixed into digestPublicWorkKey for public capabilities.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-freeside#501. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "All unique violations mapped to serialization_retry",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:34",
      "description": "isUniqueActiveKeyViolation returns true for any PostgreSQL 23505, not only shared_preparation_work_key_active_idx. joinPublicWork therefore converts unrelated unique conflicts (work_key_digest+generation, preparation_work_items (work_id, deployment_id), or other constraints) into { kind: \"serialization_retry\" }, hiding hard failures and prompting callers to retry operations that cannot succeed.",
      "suggestion": "Match constraint/index name explicitly (e.g. constraint === \"shared_preparation_work_key_active_idx\" or message/detail includes that index). Treat other 23505s as thrown errors.",
      "confidence": 0.92
    },
    {
      "id": "F2",
      "title": "retry_deadline persisted but never enforced",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:448",
      "description": "recordRetryableFailure stores retry_deadline_unix_ms, and the schema includes a failed state, but wakeRetryWait (Postgres and in-memory) only CAS-matches attempt and next_attempt_at. Work can be woken and leased indefinitely after the deadline; nothing transitions work to failed.",
      "suggestion": "Reject wakes when now_ms > retry_deadline (or CAS-update to failed), and add tests that deadline expiry is terminal.",
      "confidence": 0.88
    },
    {
      "id": "F3",
      "title": "Ready-row reuse ignores freshness age",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:210",
      "description": "joinPublicWork reuses ready rows when readiness_evidence.freshness.qualified is true and readiness_policy_version matches. It does not evaluate max_age_ms, observation_window, or as_of against now_ms, so permanently stale qualified evidence remains shareable across new orders.",
      "suggestion": "Apply the same readiness policy clock checks used at finalize (at least max_age_ms vs observed_at/as_of and now_ms) before reused_ready, or supersede ready rows when stale.",
      "confidence": 0.85
    },
    {
      "id": "F4",
      "title": "Child evidence content not validated",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:560",
      "description": "publishChildEvidence only checks parent/child state and lease_epoch. It does not require evidence.freshness.qualified, privacy_scope public_chain, or that deployment_coverage matches the child deployment_id. finalizeReadyIfQualified only gates on child state === ready and the parent envelope, so unqualified or mismatched child envelopes can qualify the parent.",
      "suggestion": "Validate child envelopes (qualified, privacy_scope, deployment coverage ⊇ item.deployment_id) before marking ready, mirroring assertReadinessEvidenceQualified.",
      "confidence": 0.8
    },
    {
      "id": "F5",
      "title": "Work-key digest uses non-canonical JSON.stringify",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-work-key.ts:99",
      "description": "digestPublicWorkKey sorts deployment_ids and finality_policies but hashes JSON.stringify of nested objects (collection_id, source_identity, VersionedDigest fields). Property insertion order on those objects can change the digest for semantically identical keys, splitting what should be one shared work row.",
      "suggestion": "Canonicalize nested digests with a fixed field order (or reuse digestCollectionWorkKey / an explicit codec) before hashing.",
      "confidence": 0.75
    },
    {
      "id": "F6",
      "title": "Post-ROLLBACK error classification is racy",
      "severity": "MEDIUM",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:175",
      "description": "classifyZeroRowWorkUpdate runs after ROLLBACK and re-reads via this.getWork on the pool. Concurrent lease reclaim or state transitions can change lease_epoch/state between the failed UPDATE and the diagnostic read, so callers may get SharedPreparationStateError instead of SharedPreparationFencingError (or the reverse).",
      "suggestion": "Classify from the locked row before release, or SELECT ... FOR UPDATE in the same transaction before deciding which error to throw.",
      "confidence": 0.78
    },
    {
      "id": "F7",
      "title": "supersedeActiveGeneration conflates lock failure with no-op",
      "severity": "MEDIUM",
      "category": "api-design",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:700",
      "description": "In-memory supersedeActiveGeneration returns undefined when withWorkKeyLock yields serialization_retry, the same signal as \"no active row\". Callers cannot tell lock exhaustion from a successful no-op. Postgres never returns serialization_retry for detach either, despite the shared interface advertising that kind.",
      "suggestion": "Propagate serialization_retry (or throw) from supersede; align Postgres detach/join retry signaling with the SharedPreparationStore contract.",
      "confidence": 0.7
    },
    {
      "id": "F8",
      "title": "Reattach link does not refresh tenant scope digest",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:318",
      "description": "insertLink ON CONFLICT DO UPDATE resets detached_at, generation, and joined_at but leaves order_tenant_scope_digest unchanged. A rejoin with a different digest keeps the stale value on the link.",
      "suggestion": "SET order_tenant_scope_digest = EXCLUDED.order_tenant_scope_digest in the ON CONFLICT update.",
      "confidence": 0.72
    },
    {
      "id": "F9",
      "title": "Community-scope test does not exercise community isolation",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts:64",
      "description": "The test titled as not sharing across scope digests actually builds different work keys via distinct readiness_policy_version values. order_tenant_scope_digest is not part of the work key; fixturePublicWorkKey also accepts communityScopeDigest and ignores it. The test does not prove tenant/community isolation behavior.",
      "suggestion": "Rename the test to match readiness_policy_version separation, or add an explicit case that identical public work keys with different order_tenant_scope_digest intentionally fan into one work_id.",
      "confidence": 0.9
    },
    {
      "id": "F10",
      "title": "worker_id accepted but never recorded or checked",
      "severity": "LOW",
      "category": "api-design",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:390",
      "description": "acquireLease takes worker_id in both backends but neither persists nor validates it; fencing is lease_epoch-only. Misattributed ownership will be hard to debug and any future owner checks cannot be added without a schema change.",
      "suggestion": "Persist lease_owner/worker_id on acquire/reclaim, or remove the parameter until ownership is modeled.",
      "confidence": 0.65
    },
    {
      "id": "F11",
      "title": "capability.v1 stripping is brittle",
      "severity": "LOW",
      "file": "packages/services/ordering/src/shared-preparation-work-key.ts:78",
      "description": "buildPublicWorkKeyMaterial maps protocol capability via capability.replace(\".v1\", \"\"), which only handles a single suffix pattern and can silently mis-map future capability names.",
      "suggestion": "Use an explicit allowlisted map from PublicPrepCapability to CollectionWorkKeyMaterial.capability.",
      "confidence": 0.6
    },
    {
      "id": "F12",
      "title": "hashtext advisory locks can collide across keys",
      "severity": "SPECULATION",
      "category": "concurrency",
      "file": "packages/services/ordering/src/shared-preparation-store-postgres.ts:188",
      "description": "pg_advisory_xact_lock(hashtext($1)) collapses work_key_digest into a 32-bit space. Collisions only over-serialize unrelated keys (latency), and the partial unique index still protects one-active-row integrity, but hot key pairs could contend unexpectedly.",
      "suggestion": "Prefer two-int advisory locks from a 64-bit hash of the digest if lock contention becomes measurable.",
      "confidence": 0.45
    },
    {
      "id": "F13",
      "title": "Lease epoch fencing and retry child invalidation",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/services/ordering/src/shared-preparation-store.ts:480",
      "description": "Stale lease_epoch is rejected on preparing/retry/evidence/finalize paths; contested reclaim tests enforce a single winner; recordRetryableFailure clears child ready evidence so finalize cannot skip republish after retry.",
      "suggestion": "Keep Postgres and in-memory suites paired whenever fencing rules change.",
      "confidence": 1.0
    },
    {
      "id": "F14",
      "title": "Expand-only schema with one-active partial unique index",
      "severity": "PRAISE",
      "category": "data-model",
      "file": "packages/services/ordering/migrations/007_shared_preparation_work.sql:48",
      "description": "Migration is expand-only with public-only CHECKs, capability allowlist, and a partial unique index on work_key_digest for queued/preparing/retry_wait—strong DB-level enforcement backing the join/create race story.",
      "suggestion": "Retain the migration text assertion in CI as the guard against accidental destructive DDL.",
      "confidence": 1.0
    },
    {
      "id": "F15",
      "title": "Concurrency coverage for join/detach and reclaim",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/services/ordering/src/__tests__/shared-preparation-store.test.ts:1",
      "description": "In-memory tests cover 100-way join fan-in, 50-way lease reclaim, retry CAS wakes, detach serialization_retry vs not_linked, and concurrent detach/join; Postgres suite adds unique-index and detach/join race checks behind an explicit env gate.",
      "suggestion": "Run ORDERING_PG_TEST_URL in at least one CI job so the gated suite cannot bit-rot.",
      "confidence": 0.95
    }
  ]
}
```


## Callouts

